Issue #918, opened 2012-08-02
Make font selection global and persistent


vext01:
If you change the font to a monospaced font, the next time you load the pad, this decision is forgotten. Would it not make sense to have this selection global and persistent?

Cheers

akien-mga:
Hi there. I know this has been closed as there is a plugin that addresses this issue, but so far I couldn't find a single instance (https://github.com/ether/etherpad-lite/wiki/Sites-that-run-Etherpad-Lite) that proposes this feature, and it's quite annoying when a monospace font is required. See e.g. this pad which shows with a non-monospace font by default, so each user has to change their configured font manually: https://public.etherpad-mozilla.org/p/godot-classref-status

Most etherpad instances that are listed above seem to have a "Global View" category, but I never found any instance which actually had configurable settings under this category. I understand that it's a matter of configuration of each etherpad hoster, but I think it would help a lot if some of those features were toggled on by default in the main trunk.

PR #1415, opened 2013-01-27
Global, persistent, per-pad View-Settings

luto:
This adds a set of settings which are synchronized for all users, persistent in the database and per-pad.

It looks like this: (updated)
img

Kinda related to https://github.com/ether/etherpad-lite/issues/918
Ready to be **tested** for bugs and usability.


t2d:
Sorry for digging this up. Has any similar functionality been committed to the main code base? Or is there a plugin to handle this functionality?

PR #1638, opened 2013-03-17
A nicer way to handle the API versions

disy-mk:
- Instead of manually copy/pasting the API version functions every time we version bump, we should reuse the content of a previous API version.
- Also a possibility for querying a list of all available API versions is available under /API/versions


Gared:
@nelsonsilva Any progress on that?

JohnMcLear:
Closing due to neglect.

Issue #1763, opened 2013-04-27
Slow down with lots of different contributions (~ @ 700 lines)


JohnMcLear:
https://bugzilla.wikimedia.org/show_bug.cgi?id=46637

Please attempt to replicate and profile.

JohnMcLear:
@dlongley Hey man just wondering if you had time to look into this further yet? :)

dlongley:
I'm still time strapped at the moment. However, I'm thinking the right approach to solving this, probably a number of other bugs, and getting maintenance into a better state would be:

1. Rewrite the core editor using a new version of [Ace](https://github.com/ajaxorg/ace). I think etherpad-lite should have more of a dependency-based relationship with Ace (vs. what appears to be highly-modified check-ins of very old version (2009)). This would let us pull in fixes from Aces independently of etherpad-lite and off-load a ton of maintenance-related bugs, reduce etherpad-lite's code base size, keep concerns limited to etherpad-lite's value-add and therefore keep things simpler for devs, etc.
2. Etherpad-lite's core design should be around tracking, storing, and transmitting changes to an Ace document (from multiple users), nothing more. On top of the core should be the same sort of existing feature set that exists now (hopefully changing core wouldn't require us to make too many changes elsewhere). In short, after a rewrite of core, we should still have all the features and a similar feel to what we had before, but core's scope would be appropriately limited, the code would be easier to understand, and we could focus on etherpad-lite features instead of worrying about the underlying editor itself.
3. As I'm always short on time, I've only briefly looked at Ace's APIs, but it seems like the first cut at a redesign would involve writing an [EditSession](http://ace.c9.io/#nav=api&api=edit_session) that tracked changes, communicated them to the server, and received changes from other users from the server. Each user would  just get their own EditSession on a document. Changes would be serialized to/from JSON and, at least it seems, could be applied as a change set via a simple call to `applyDeltas` on the shared [Document](http://ace.c9.io/#nav=api&api=document). I haven't looked into all of what goes into an Ace change object or what etherpad-lite is currently serializing for changes; we may/may not need to update the format or write something to convert legacy pads. We'd still need a way to leave meta tags on elements that were changed when applying each delta (to colorize for different users) -- but, maybe something easy/clever can be done here as well. I didn't look into it; maybe there's even a way to abuse a custom syntax highlighter to accomplish it.

[...]

PR #1764, opened 2013-04-29
Slow responsiveness on larger pads

JohnMcLear:
Test for https://github.com/ether/etherpad-lite/issues/1763


JohnMcLear:
Note that this test doesn't actually cover the bug, it would need to create edits as multiple authors.

PR #2295, opened 2014-11-04
Socket io v1 upgrade (Socket IO v1 has a different API from what we currently use)

JohnMcLear:
Things still to address:
- Minify SocketIO Javascript before delivering

Why this work is required (or why an upgrade to Socket v1 is painful):
- Heavily used API deprecation
- Pointless changes to Debug/Logging API
- Dropping of minified JS
- SocketIO API has moved on, we should too, this may fix quite a few issues with connectivity and ultimately it's worth staying up to date with our upstream deps

Things to test before Merge:
- [ ] Authenticated access (Cookie based)
- [ ] Works behind proxy
- [ ] Page load speeds aren't decreased

### DIFF
--- a/src/node/handler/PadMessageHandler.js
+++ b/src/node/handler/PadMessageHandler.js
@@ -117,8 +117,6 @@ exports.handleDisconnect = function(client)
   if(session && session.author)
   {
 
-console.log(remoteAddress);
-
     // Get the IP address from our persistant object
     var ip = remoteAddress[client.id];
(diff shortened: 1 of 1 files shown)


flotho:
Hi @JohnMcLear ,

You wrote that it should 'Works behind proxy'
I'm struggling with API issue and httpS proxy #2754 , if you think that this PR is related I can help to test this part.

regards

Issue #1241, opened 2012-12-02
Add a sanitizer step for custom PadIDs


itguy51:
Based on the code used to input PadIDs, there does not seem to be a sanitization step, for example, if I type anything with the slash character '/' or '\', I will be given the "Cannot GET /p/..." error, which could be confusing for users.

Gared:
Fixed in #2302

Issue #2287, opened 2014-10-29
Special characters in pad name when using API to create pads


ja67:
I am using Etherpad API to make some interaction between my website and Etherpad. I am trying to implement a feature that let users create a pad with a custom name using createPad and createGroupPad( we also use grouppad as private pad for users).

 I found that if user inputs a pad name with some special characters there are some strange behaviours. For example, if the name of the pad for createPad is called "foo?bar" then the actual name of the pad is "foo". Moreover, we can also use createPad to create a pad called "foo?bar2" and these two pads are both listed when we call the api function listPads, but these too pads actually point to the same pad "foo". Another example is we can create a pad whose name contains a pattern like "../../" and then this pad can no longer be accessed directly.

We have managed to check the input name and replace special characters before sending to API but I was wondering if those API functions themselves could check and refuse invalid name input. The documentation of Etherpad didn't tell us what characters are allow in datatypes in padID and authorID. Thanks!

Gared:
Fixed this in #2302

PR #2302, opened 2014-11-08
Add check for special url characters to createPad API function

Gared:
This will add a check to forbid special url characters (like /, ?, & and #) on the createPad API function.
This is only a small improvement, because it doesn't fix the whole bug reported in #2287
Characters like ':' are sanitized by the function padManager.sanitizePadId in APIHandler.js line 414. We should move this function inside the createPad method so we could give back an error if the sanitizedPadId is not equal with the original padId. Is there any reason why every API call with a padId uses this function?


Gared:
@JohnMcLear Is there anything I should do here?

JohnMcLear:
I'm happy with it if you are

On 14 Mar 2015 22:48, Stefan <notifications@github.com> wrote:

@JohnMcLear<https://github.com/JohnMcLear> Is there anything I should do here?

Gared:
@marcelklehr What do you think? Are you happy with this kind of validation?

JohnMcLear:
Bump @marcelklehr

JohnMcLear:
@Gared if you want to merge this in just for for it as long as you are happy with maintaining it as it grows =+)

Gared:
Fix #1241

PR #2329, opened 2014-11-19
Restore pad to new location at a given revision

### LINKED ISSUE #1791: Add option to restore revisions
It would be a very important addition to the revision system to add the ability to restore previous revisions. It was mentioned in issue #229, but the ticket was closed.

[Explanatory note: if a restoration takes place it should happen as a document change and not as a reset back to rev x.]

timrchavez:
This script gives an admin with shell access the ability to restore a
pad at a given revision by essentially rebuilding it at a new location
with data associated with the original pad.  The upsides to creating a
new pad vs. changing the original are: 1) avoiding service disruptions
(no deletes, no moving targets - builds from previous revision); and 2)
preservation of data (no deletes, no overwriting of the source pad).
The most obvious downside is the pad has a new ID which could require
folks to update their links, bookmarks, etc. to point at the new
location.

Our requirements for restoring pads at given revisions may be slightly
different than other folks.  We are not looking to enable a new feature
for users, but really another tool an admin can use to effectively fix
broken pads.

I would not consider this script to be complete.  I've only been looking
at etherpad-lite code for a few days now.  I was hoping to get some 
feedback / comments on places I can improve it and in the general
strategy.

Fixes #1791 
(so it shows up in that conversation)

### DIFF
--- a/bin/rebuildPad.js
+++ b/bin/rebuildPad.js
@@ -3,8 +3,8 @@
   known "good" revision.
 */
 
-if(process.argv.length != 4) {
-  console.error("Use: node bin/repairPad.js $PADID $REV");
+if(process.argv.length != 4 && process.argv.length != 5) {
+  console.error("Use: node bin/repairPad.js $PADID $REV [$NEWPADID]");
   process.exit(1);
 }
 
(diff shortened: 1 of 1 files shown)


dcht00:
For anyone just looking to simply restore contents of a pad - see "ep_cristo_restore_revision" plugin. Install via /admin.

When installed, you get a button in Timeslider, or you can also restore with a link, for example:
http://sometherpad.org/restore/[pad_name]/[revision_no]

Issue #1604, opened 2013-03-11
Indenting problems when exporting to HTML


jco-odoo:
If you use bullets points and make an empty line between 2 bullets of a same group, the HTML export decreases the level of bullet points

e.g: bullet in the user interface: 
- dhgfhdgfdf
- jshfjdhfdf
  - dfkhjdfgh
  - dfvdfv
  - fdvgdjfgkg
  - dfkjlhgdhfggf

Will give in the HTML export: 
- dhgfhdgfdf
- jshfjdhfdf
  - dfkhjdfgh
  - dfvdfv
- fdvgdjfgkg
- dfkjlhgdhfggf

PR #2403, opened 2014-12-27
Test coverage and backend support for setHTML.

JohnMcLear:
don't merge yet, I still need to add full test coverage

1) Reports correct failure for setHTML
2) Provides setHTML test coverage
3) Ultimately needs to provide functionality to create li intented stuff 

Attempting to add test coverage for https://github.com/ether/etherpad-lite/issues/1604

### DIFF
--- a/src/node/db/API.js
+++ b/src/node/db/API.js
@@ -408,12 +408,18 @@ exports.setHTML = function(padID, html, callback)
   getPadSafe(padID, true, function(err, pad)
   {
     if(ERR(err, callback)) return;
-
     // add a new changeset with the new html to the pad
-    importHtml.setPadHTML(pad, cleanText(html), callback);
-
-    //update the clients on the pad
-    padMessageHandler.updatePadClients(pad, callback);
+    importHtml.setPadHTML(pad, cleanText(html), function(e){
+      if(e){
+        callback(new customError("HTML is malformed","apierror"));
+        return;
+      }else{
+        //update the clients on the pad
+        padMessageHandler.updatePadClients(pad, callback);
+        callback();
+        return;
+      }
+    });
 
   });
 }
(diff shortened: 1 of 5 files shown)


JohnMcLear:
@webzwo0i note below..  Do you have a test case for this in your coverage?  It appears we miss the case where a user might have empty nested ul's as below..

```
  // will fail due to https://github.com/ether/etherpad-lite/issues/1604
  // reminder to self this is how the HTML looks
  // <ul>
  //   <li>one</li>
  //   <li>2</li>
  // </ul>
  // <br>
  // <ul class="bullet">
  //   <ul class="bullet">
  //     <li>UL2</li>
  //   </ul>
  // </ul>
  // It will look right in the browser but the export will get it horriby wrong

  // This is what the export puts out
  // <ul class="bullet">
  //  <li>one</li>
  //  <li>2</li>
  // </ul>
  // <br>
  // <ul class="bullet">
  // NOTE THIS IS WHAT'S MISSING
  //   <li>UL2</li>
  // </ul>
  // <br>
```

webzwo0i:
:-( did you set the html via API or manually in the browser?
I tried to replicate but if I put your html in importexport.js (one time with "br" - this will close the first ul, then newline, then open two uls; the other time without "br" it won't close the first ul and just open one other ul) everything seems to be okay.
So I wonder if there is some difference between setting via API and setting directly in the editor/doing an import.

webzwo0i:
r u sure this is still a problem? I see that you have this in the backend tests ("ULhtml") and I think it works (currently only the "import crappy html"-test fails).
I will add a case for this to the frontend-tests but as I said above it also seems to work there (with and without the newline)

JohnMcLear:
Yea should be okay now, I updated the tests (see above)

Issue #2375, opened 2014-12-16
Server crash "Failed assertion: line assembler not finished"


Gared:
My server crashes nearly every hour caused by this exception:
```
[2014-12-16 09:43:42.298] [ERROR] console - Error: Failed assertion: line assembler not finished
    at Object.error (/eplite/src/static/js/Changeset.js:39:11)
    at Object.assert (/eplite/src/static/js/Changeset.js:53:13)
    at Object.exports.mutateAttributionLines (/eplite/src/static/js/Changeset.js:1219:11)
    at async.series.callback.forwardsChangesets (/eplite/src/node/handler/PadMessageHandler.js:1399:19)
    at async.series.results (/eplite/src/node_modules/async/lib/async.js:486:21)
    at _asyncMap (/eplite/src/node_modules/async/lib/async.js:185:13)
    at async.forEachSeries.iterate (/eplite/src/node_modules/async/lib/async.js:108:13)
    at async.forEachSeries.iterate (/eplite/src/node_modules/async/lib/async.js:119:25)
    at _asyncMap (/eplite/src/node_modules/async/lib/async.js:187:17)
    at async.series.results (/eplite/src/node_modules/async/lib/async.js:491:34)
```

Gared:
The last time the error occured was a few days ago. So if I pull develop I can't be really sure that this fixes the problem. I will reopen this issue if I get further errors.

JohnMcLear:
Same :-) thanks
----- Reply message -----

PR #2372, opened 2014-12-14
Deny changesets with wrong linenumber

webzwo0i:
https://github.com/ether/etherpad-lite/issues/2215

PR #2405, opened 2014-12-27
Fix lineassembler not finished

webzwo0i:
taken from https://github.com/ether/etherpad-lite/pull/2372#issuecomment-68175271
I have another script at https://github.com/webzwo0i/ep-debug-helper/blob/master/checkrep-line-assembler-not-finished.rb
If applied to a new default pad, it sends a changeset "Z:6c<1|3=g=1o-1$" which is wrong, because within the first "g"(16 in base 36) letter there are less than 3 newlines. All clients will disconnect with the error "line assembler not finished" and if you reconnect and go to the timeslider the whole node instance will crash with the same error. Might be that this is the same as #2375
With the commits in this pull request, the changeset is not applied.

applied to develop frontend-tests are okay

### DIFF
--- a/src/static/js/Changeset.js
+++ b/src/static/js/Changeset.js
@@ -929,7 +929,7 @@ exports.applyToText = function (cs, str) {
       removedLines += op.lines;
       newlines = strIter.newlines()
       strIter.skip(op.chars);
-      if(!(newlines - strIter.newlines() == op.lines)){
+      if(!(newlines - strIter.newlines() == 0) && (newlines - strIter.newlines() != op.lines)){
         newlinefail = true
       }
       break;


dmitryuv:
@JohnMcLear you merged this? wow... it changes the API and have a very strange way to throw an error message in the outer function rather than in the inner where error happened.

webzwo0i:
Actually I did a grep in all the public plugins on npmjs for applyToAText and applyToText and not a single plugin used this functions directly. Also, they are not exposed in any (http) API endpoint(maybe indirectly somewhere?). So for me this is not a "real" API but just a function that is exported - like so many others in the current design of epl.

YMMV regarding dynamically constructed return values(array is a normal object type in EcmaScript), but think about this: CoffeeScript contains "coffee" - nice! Javascript contains..."java" - duuh! :)

Regarding the strange throwing: You are right (and I just realized that @marcelklehr asked the same question some time ago and I did not answer) that this is not good style. I will fix it in this other pull request. I remember asking myself the same question and was sure at that time that it is the better way to throw. mhh, thx for the hint

dmitryuv:
If its exported, it's a public API. Also you had to update tests and include strange [0] for this.
Yes, you can return arrays, but that usually means array of values, not a combination of results.
So here I suggest throw error in place and revert to the original API.

PR #2411, opened 2014-12-28
Getattrselectionlogic

JohnMcLear:


JohnMcLear:
This is still slightly broken by the looks of things.. Steps to replicate..

Check out this branch
Install ep_subscript_and_superscript
Type, because the caret is on the end of the author span the getAttributeOnSelection is still returning true for all attributes.  This is not the desired behaviour.

Make sense?  If not just bump me I will explain further :)

JohnMcLear:
oh im an idiot.

PR #2418, opened 2014-12-29
Full Pad portability (Export/Import)

JohnMcLear:
- [x] Requires export of Author data
- [x] Make it work cross server (requires All data, weirdly read only pad id, author data etc. is not enough but not getting an error telling me what's missing) -- help here appreciated :)
- [x] Requires a way to prevent import of "mature" pads
- [ ] Requires testing -- help here appreciated :)
- [x] Requires a way of updating the users UI without a refresh (currently we nuke all Changesets so a user might think they are further into the pad) -- help here appreciated :)

Do not merge, do contribute =)
- [x] Do throw kittens

### DIFF
--- a/src/node/handler/ImportHandler.js
+++ b/src/node/handler/ImportHandler.js
@@ -247,9 +247,16 @@ exports.doImport = function(req, res, padId)
       padManager.getPad(padId, function(err, _pad){
         var pad = _pad;
         padManager.unloadPad(padId);
-        padMessageHandler.updatePadClients(pad, function(){
+
+        // direct Database Access means a pad user should perform a switchToPad
+        // and not attempt to recieve updated pad data..
+        if(!directDatabaseAccess){
+          padMessageHandler.updatePadClients(pad, function(){
+            callback();
+          });
+        }else{
           callback();
-        });
+        }
       });
 
     },
@@ -282,7 +289,7 @@ exports.doImport = function(req, res, padId)
     }
 
     ERR(err);
-  
+
     //close the connection
     res.send(
       "<head> \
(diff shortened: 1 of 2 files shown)


src/node/handler/ImportHandler.js
```diff
@@ -116,9 +119,33 @@ exports.doImport = function(req, res, padId)
...
     },
+    function(callback) {
+      var fileEnding = path.extname(srcFile).toLowerCase()
+      var fileIsEtherpad = (fileEnding === ".etherpad");
+
+      if(fileIsEtherpad){
+        // we do this here so we can see if the pad has quit ea few edits
+        padManager.getPad(padId, function(err, _pad){
+          var headCount = _pad.head;
+          if(headCount >= 10){
```
marcelklehr: Why 10?

JohnMcLear: Seemed like a sensible constant.

src/node/utils/ExportEtherpad.js
```diff
@@ -0,0 +1,68 @@
...
+        // Get the Pad Authors
+        if(entry.pool && entry.pool.numToAttrib){
+          var authors = entry.pool.numToAttrib;
+          async.forEachSeries(Object.keys(authors), function(k, c){
+            if(authors[k][0] === "author"){
+              var authorId = authors[k][1];
+
+              // Get the author info
+              db.get("globalAuthor:"+authorId, function(e, authorEntry){
+                if(!e) data["globalAuthor:"+authorId] = authorEntry;
```
marcelklehr: Possible security hole (reported here: https://groups.google.com/forum/#!topic/etherpad-lite-dev/X31IhThTtIM)

marcelklehr: Might need a filter akin to this one: https://github.com/ether/etherpad-lite/blob/develop/src/node/handler/PadMessageHandler.js#L1079

JohnMcLear:
@marcelklehr I'd take that filter if you can pop it in please :)

PR #2423, opened 2014-12-31
Re-implement ace_getAttributeOnSelection

marcelklehr:
This is an implementation of the behavior proposed in #2411


JohnMcLear:
Doh if selStart = selEnd false is always returned..

JohnMcLear:
a selection can be the caret location so really we should be mindful of the previous character at this point..

JohnMcLear:
The idea behind the above statement is that you know what attributes will be applied to the next character typed.

marcelklehr:
Ahh, so that's why you need this. I'd create a separate method for that, though. Something like ace_getAttribsOnCaret

JohnMcLear:
Thanks! :)

PR #2442, opened 2015-01-05
Finally fix AttribManager#removeAttributeOnLine

### LINKED ISSUE #2394: Outdent to 0 destroys all line attributes
Create new line
Set it as h1 in ep_headings (applies to all plugins that set line attributes)
Indent
Outdent
Line Attributes will be lost.

marcelklehr:
return safely if the attribute wasn't found.
fixes #2394


JohnMcLear:
Thanks! :)

PR #2447, opened 2015-01-08
Removed useless use of cat

badboy:


JohnMcLear:
What does this fix?
----- Reply message -----

badboy:
Sorry about the "typo" thing. `ggrep` is indeed the right command on Solaris.

The other does not really "fix" anything but just cleans up the command used to extract the version.

JohnMcLear:
tnx

PR #2459, opened 2015-01-16
check for io.js

nemetz:
Added a check for io.js

### DIFF
--- a/bin/installDeps.sh
+++ b/bin/installDeps.sh
@@ -23,6 +23,7 @@ hash curl > /dev/null 2>&1 || {
 }
 
 #Is node installed?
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Tnx

Gared:
I think this check is very bad...
The bin/run.sh script also throws a exception that iojs command could not be found.
We should handle the node and iojs executable the same way.

JohnMcLear:
+! @Gared 

@nemetz please put forward a better fix that doesn't throw up an error...

JohnMcLear:
@nemetz are you going to get this done?

PR #2464, opened 2015-01-18
Plugin page improvements

Gared:
- Prevents a page reload when pressing enter on the admin plugin page
- Prevents dublicates in search results on plugin page


JohnMcLear:
n1 I have been meaning to fix this for ages..

I also suggest we drop infinite scrolling, it causes more problems than it solves and it's really easy to remove :)  Fancy doing that while you are in these files?  

The other thing is maybe moving the pluginfw code into the right location..  At the moment client/server code is in static/js and this is confusing..

PR #2465, opened 2015-01-18
Change domain for plugins.json request

Gared:
The file on http://static.etherpad.org/plugins.json will be updated every our


JohnMcLear:
I suggest we put an SSL Cert on this request as the next step.

PR #2467, opened 2015-01-19
Show users and Chat at the same time

JohnMcLear:
Adds new functionality that should be in core.

To test
1. Go into a pad
2. Click Settings
3. Click "Show Chat and Users"

It will need some polish but anyone can do that, the basic functionality is in..

Maybe cc @0ip for the polish


JohnMcLear:
After polish it needs tests..

Issue #2025, opened 2013-12-07
A setting to specify the url prefix of the installation


marcelklehr:
- No more confusing relative paths
- plugins can use the setting to build the correct urls, too

(_Asked for over here: https://groups.google.com/forum/#!msg/etherpad-lite-dev/l6tSr1Y75ZI/GSMzMIctPOkJ_)

JohnMcLear:
Is this actually a problem?

gabrielliwerant:
Not sure if this is the same issue, but etherpad seems to require being in a top level directory. If etherpad is installed in a subdirectory, such as `applications/libs/etherpad`, the server cannot locate the `settings.json` file. Moving the settings file does not work because this confuses the database directory...

This is less than ideal. I'm working on another project in git and I'd like to share a modified settings file with other coders. Under this arrangement, I cannot do this in my existing project. It should be possible to treat etherpad as a part of another application for source control and development purposes, but etherpad does not seem to be set up this way.

gabrielliwerant:
Possible mitigation for this issue: https://github.com/ether/etherpad-lite/pull/2477

llaurent:
If this feature request is about a new path option to make it work on http://host:port/path/ instead of http://host:port/ I'm all for it ! :-)
That would make reverse far more simpler when you don't want to use a dedicated vhost...

flotho:
@llaurent, I'm also struggling with making it work under subdirectory

JohnMcLear:
Is this still an issue?  Afaik it is not...

PR #2477, opened 2015-01-20
Add subdirectory installation instructions

gabrielliwerant:
Out of the box, etherpad-lite does not work correctly if cloned/installed into a subdirectory within an existing project. To do so, a few minor tweaks to the installation process are necessary which are documented here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,12 @@ Now, run `start.bat` and open <http://localhost:9001> in your browser.
 
 Update to the latest version with `git pull origin`, then run `bin\installOnWindows.bat`, again.
 
+If cloning to a subdirectory within another project, you may need to do the following:
+
+1. Start the server manually (e.g. `node/node_modules/ep_etherpad-lite/node/server.js]`)
+2. Edit the db `filename` in `settings.json` to the relative directory with the file (e.g. `application/lib/etherpad-lite/var/dirty.db`)
+3. Add auto-generated files to the main project `.gitignore`
+
 [Next steps](#next-steps).
 
 ## GNU/Linux and other UNIX-like systems


JohnMcLear:
Please provide more info, see: http://mclear.co.uk/2015/01/15/how-to-write-a-bug-report/

+1. Start the server manually (e.g. `node/node_modules/ep_etherpad-lite/node/server.js]`) <-- is wrong, you should use bin/run.sh

Path is relative anyway...  

I must be missing something as this PR doesn't seem to solve an actual problem ;\

JohnMcLear:
Ah this PR is for https://github.com/ether/etherpad-lite/issues/2025

JohnMcLear:
Anyway not using bin/run.sh will present a set of problems so you should use the settings CLI option IE

bin/run.sh --settings foo/var/la.json

And set the database path absolute in the settings file..  Peep https://github.com/ether/etherpad-lite#tweak-the-settings

gabrielliwerant:
Hey, thanks for the feedback.

I may not have been clear about a few things, so I will attempt to restate some of my findings.

The installation instructions for windows indicate, "Now, run `start.bat`..." to finish the process. The first problem is that `start.bat` fails when the etherpad-lite is installed in a non-root directory. E.g. my_project/application/lib/etherpad-lite.

All `start.bat` does is run the node server: https://github.com/ether/etherpad-lite/blob/develop/start.bat. So my instructions are to run it from the CLI instead of using `start.bat` so the developer can input the proper path. Since `start.bat` loads the node server via a `node` command, I suggest doing the same via CLI. The instructions to run `bin/run.sh` are given for running multiple instances of etherpad with multiple settings files. This is not what I want. All I want to do is bypass the single command in `start.bat`. It's easy enough, but the point here is to modify the instructions to make them more idiot-proof. If the instructions fail when the path is changed, at the very least, this should be indicated to prevent confusion and save developer time.

I give instructions for changing the path to the database file. Absolute paths might cause issues when migrating to a production environment. At the very least, it will make deployment more complicated. Is there any reason to prefer it over relative here when the relative works just fine?

Also, the Travis CI build seems to have failed due to "Missing required arguments USERNAME, API_KEY." Not sure how to handle that since I have only modified text in the README.md.

gabrielliwerant:
Ok, so I think I'm gaining a better understanding. Is it the case that cloning etherpad-lite to a subdirectory of an existing project is not supported? I understand that I can track a settings file separately by passing an argument to the settings parameter of `run.sh` but does this still require that the library itself live on a root directory? Basically, a lot of things seem to break if the project is cloned to and installed from a subdirectory of an existing project, whether or not the settings file is loaded. If needed, I can be more specific when I have the time.

Either way, I think the documentation could make the case more clear, and I can work on that if I know which situation is correct.

Issue #2486, opened 2015-01-24
getLineHTMLForExport produces multiple copies of a line


JohnMcLear:
If you have ep_headings2 and ep_align installed multiple copies of that line will be exported in the HTML.

To replicate install both plugins, set the attribute on both lines.  Hit up the /export/html endpoint.

What happens:  You get 2 lines of the same contents
What should happen: You get 1 line with both stylings applied.

Here is some example HTML

```
<body>
*<span style='font-size:16px'>Welcome to Etherpad!</span>  
<h2>Welcome to Etherpad!</h2>  
<div style="width:100%;margin:0 auto;list-style-position:inside;text-align:right">*<fs16>Welcome to Etherpad!</fs16></div>
</body>
</html>
```

https://github.com/ether/etherpad-lite/blob/develop/src/node/utils/ExportHtml.js#L414 is the offending hook.

JohnMcLear:
I created a new hook to solve this problem.

https://github.com/ether/etherpad-lite/pull/2487

The only caveat here is that the old hook set a variable and didn't append a ``<br>`` if it was used.  This new hook doesn't do that so I'm not sure what problem this solves..

rugk:
So this issue seems to happen again...

rugk:
It also happens with [`ep_mediawiki`](https://github.com/johnmclear/ep_mediawiki) instead of `ep_align`. In https://github.com/ether/etherpad-lite/issues/2981 I had installed both however...

ilmartyrk:
Had a look inside the code and found why this issue is accuring. The fix is quite simple, but most of the plugins will need a little update also.
in https://github.com/ether/etherpad-lite/blob/develop/src/node/utils/ExportHtml.js#L454

as `getLineHTMLForExport ` stacks all the return values we get a new line from each plugin that returns a value from it and it is handeled as if something is returned from getLineHTMLForExport then add it instead of the original lineContent. But as each of the plugins gets in the context param as a original value and it's not altered then we get these multiple lines each differing from another in some way.
To really make it work in every ep_ plugin it should set `context.lineContent = 'whatever you change on the line'` and return true if anything was changed  and in  the ExportHtml.js use 
`if (lineContentFromHook)
      {
        pieces.push(context.lineContent, '');
      }
      else
      {
        pieces.push(lineContent, '<br>');
      }`  instead og
`
if (lineContentFromHook)
      {
        pieces.push(lineContentFromHook, '');
      }
      else
      {
[...]

PR #2487, opened 2015-01-24
Better HTML Export rewrite hook allows functions to be passed instead of strings means state is saved between plugins

JohnMcLear:
Fix for https://github.com/ether/etherpad-lite/issues/2486

### DIFF
--- a/doc/api/hooks_server-side.md
+++ b/doc/api/hooks_server-side.md
@@ -245,7 +245,7 @@ Things in context:
 2. attribLine - line attributes
 3. text - line text
 
-This hook will allow a plug-in developer to re-write each line when exporting to HTML.
+This hook will allow a plug-in developer to re-write each line when exporting to HTML.  Note that you problably don't want to use this plugin and will probably get better results from `asyncLineHTMLForExport`
 
 Example:
 ```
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Use like this
```
exports.asyncLineHTMLForExport = function (hook, context, cb) {
  var align = _analyzeLine(context.attribLine, context.apool);
  if (align) {
    cb('<div style="width:100%;margin:0 auto;list-style-position:inside;text-align:' + align + '">' + context.lineContent + '</div>');
  }
}
```

JohnMcLear:
@Gared I want to get this in for 1.5.1

Gared:
@JohnMcLear Okay, I'm aware of this PR :-)
Maybe we should mark "getLineHTMLForExport" as deprecated?

JohnMcLear:
I tested this quite a lot and it's working well :)

Gared:
I tried to use this new hook in my plugin: https://github.com/Gared/ep_export_authors
but I still can't use asynchronous calls (like db calls) in this hook, because I have to _return_ the new lineContent in this function. 
It would be better (and more the nodejs-way) if the lineContent would be returned as the callback parameter like this:
```
exports.asyncLineHTMLForExport = function (hook, context, cb) {
  ...
  db.getSub("globalAuthor:" + author, ["colorId"], function(err, result) {
    ...
    cb("newLineContent");
  });
}
```

JohnMcLear:
Okay I will rewrite now

lpagliari:
Hey, guys, any news on this? We might need it to fix ep_font_color export to HTML...

PR #2496, opened 2015-01-25
Merge branch master back to develop

Gared:


Gared:
I hope this is the right way to do this...

Issue #1603, opened 2013-03-11
pad url is leaked whenever a link is clicked


elijh:
**The problem:** if a user clicks on a link in the main body or in the chat, the referrer is sent by the browser to the new website showing the full url of the pad, thus revealing the pad's content and history. Typically, this referrer information information is logged and stored by the target website.

**Why this is bad:** most users probably do not realized that clicking on a link in a pad automatically shares the pad with the website they clicked on (!).

Solutions:

**(A)** HTML5 has a `noreferrer` option for links, but is is not widely supported. Move along.

**(B)** One option would be to disable the auto-links in etherpad. As far as I can tell, these can be disabled in ace2_inner.js `handleClick()` and pad_utils.js `escapeHtmlWithClickableLinks()`. 

**(C)** Another option would be to change the way pad names are encoded in the URL to use the location hash instead of the path. For example, instead of `https://pad.riseup.net/p/0Jr0e` etherpad could use `https://pad.riseup.net/#0Jr0e`. The location hash (anchor) is not sent as part of the referrer.

**(D)** A final option is to create an interstitial page as part of etherpad. For example, every link would really go to `/redirect/`, where it would then get auto redirected on to the target website. The final target website could be sent to the request to `/redirect` either as a POST or in the location hash. For example, `/redirect#https%3A%2F%2Friseup.net%2Fwhat-is-my-referrer' would redirect to https://riseup.net/what-is-my-referrer.

Option D seems to me that it is probably the best for the moment: it would be easy to implement and it retains the usability of clickable links. On the downside, interstitial pages are slightly annoying because they make it a bit slower to get to the site you want.

Notes:

https -- If the pad is loaded via https, then the referrer is only sent to other https websites. This is small comfort, but important to know if you are trying this at home. 

chrome -- Chrome is not able to figure out what the referrer should be of the inner ace iframe. I suspect this is because it is dynamically generated. The result is that if you click on a link in a pad body using chrome, no referrer is sent. Firefox has no such confusion, and links from the chat always send referrer in all browsers.

testing -- You can use https://riseup.net/what-is-my-referrer as a link to test if the referrer is being sent by the browser. Make sure to specify `https` and not `http` because the plain `http` link will redirect and erase the referrer.

joelpurra:
Here's a test page, showing leaking pad URL to external links through HTTP referer.
https://etherpad.mozilla.org/eFJXwtxPJu

JohnMcLear:
[...]

PR #2498, opened 2015-01-27
Fix pad URL leaks through HTTP referer header in HTML5-compatible browsers

joelpurra:
See #1603.

Added `rel="noreferrer"` to automatically generated links in the main pad
window as well as the chat window. Also applied the fix to exported HTML documents.

[`rel="noreferrer"` is part of the HTML5 standard.](http://www.w3.org/TR/html5/links.html#link-type-noreferrer) While browser support
isn't 100%, it's better than nothing. Future alternative solutions with
wider browser support, such as intermediary redirect pages, are unaffected
by this change.

### DIFF
--- a/src/node/utils/ExportHtml.js
+++ b/src/node/utils/ExportHtml.js
@@ -284,7 +284,10 @@ function getHTMLFromAtext(pad, atext, authorColors)
         var url = urlData[1];
         var urlLength = url.length;
         processNextChars(startIndex - idx);
-        assem.append('<a href="' + Security.escapeHTMLAttribute(url) + '">');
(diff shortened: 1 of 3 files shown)


JohnMcLear:
This should be a plugin

slel:
Isn't that what this plugin is for: https://github.com/ether/ep_hide_referrer

muxator:
This PR was incorporated in #3636 and will be part of **1.8.0**.

#3044 (merged in **1.6.1**) already started addressing this issue, and mentioned:

> Makes https://github.com/ether/ep_hide_referrer obsolete.

PR #2509, opened 2015-02-09
Adding api call for appending a chat message.

derosm2:


JohnMcLear:
Looks good!  Please add this API call to the backend tests! :)
https://github.com/ether/etherpad-lite/blob/develop/tests/backend/specs/api/pad.js

You could also add other tests for chat now that the backend tests can make chat messages!  That's awesome, let me know if you need help!

derosm2:
Haven't had a chance to write up the tests yet, but will do so when I get some free time!

JohnMcLear:
Bump @derosm2

JohnMcLear:
Moving here: https://github.com/ether/etherpad-lite/pull/2573

PR #2510, opened 2015-02-09
better css

JohnMcLear:


JohnMcLear:
do not merge, will do a few tweeks first.

PR #2518, opened 2015-02-15
Deny changesets if newline count is wrong

webzwo0i:
This is similar to https://github.com/ether/etherpad-lite/pull/2405 
It adds the same sanity checks for + and - ops - just as the aforementioned pull request did for = op.
It prevents epl from accepting changesets that lie about newline counts. If such a cs is accepted it can result in disconnecting all clients and a crash of the whole epl instance can be triggered later on. ("lineassembler not finished")

Note that it does not prevent epl from crashing in case such a changeset was accepted in the past, but at least it prints the bad changeset now. You have to manually remove it from the database and repair/reimport - although this can be non-trivial because you may not precisely understand what the changeset was meant to do.

frontend&backend&easysync tests pass

Note also, that I actually made two mistakes in this other pull request and fixed them here. First I was off by 1 when counting newlines (I tried to check for wrong line count in - op although at that time I was not sure if this could trigger the same bug) and second I had an always-false branch condition. No harm was caused due to this, but also no use.

### DIFF
--- a/src/node/easysync_tests.js
+++ b/src/node/easysync_tests.js
@@ -149,7 +149,7 @@ function runTests() {
 
     var correctText = correct.join('');
     //print(literal(cs));
-    var outText = Changeset.applyToText(cs, inText);
+    var outText = Changeset.applyToText(cs, inText)[0];
     assertEqualStrings(correctText, outText);
   }
 
@@ -585,9 +585,9 @@ function runTests() {
     var change123a = Changeset.checkRep(Changeset.compose(change1, change23, p));
     assertEqualStrings(change123, change123a);
 
-    assertEqualStrings(text2, Changeset.applyToText(change12, startText));
-    assertEqualStrings(text3, Changeset.applyToText(change23, text1));
-    assertEqualStrings(text3, Changeset.applyToText(change123, startText));
+    assertEqualStrings(text2, Changeset.applyToText(change12, startText)[0]);
+    assertEqualStrings(text3, Changeset.applyToText(change23, text1)[0]);
+    assertEqualStrings(text3, Changeset.applyToText(change123, startText)[0]);
   }
 
   for (var i = 0; i < 30; i++) testCompose(i);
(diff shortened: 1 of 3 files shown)


JohnMcLear:
@marcelklehr @dmitryuv @Pita you guys wanna have any opinion on this?

dmitryuv:
Intentions are good but the code is not. Also commented on #2405. 
Code does not follow basic guidelines at least (missing ";"),  and I don't really like using split function to count lines, matching regex is better performance wise IMO. 

btw in my change sets i do even more checks including check for actually removed chars ;-)

webzwo0i:
could you tell my where I can find you changeset-checks?

webzwo0i:
Regarding the changeset checks, do you mean you check if removing x character really decreases the string size by x? I think I like this idea, but also I do like the idea how in the timeslider inverse-changesets are computed. If cs a applies to atext1 (results in atext2) and inverse-cs applies to atext2 and results in atext1 this is a good guess that the changeset was okay. However this would have more performance implications and maybe no benefit at all.
Still I cannot find the checks you are talking about :-(

split vs regex match, https://github.com/webzwo0i/ep-debug-helper/blob/master/testregex.js
in v8:
match with regex 4s 244.26525ms
split with non-regex seperator 2s 508.15675ms
split with regex seperator 6s 864.429ms
in my browser (firefox):
"match with regex " 10500.412867000006
"split with non-regex seperator" 169.00710699998308
"split with regex seperator 20821s ms"

so either I made a mistake or for single-character splits/regexes, split is faster. Note that I change the string in every iteration to circumvent optimization

webzwo0i:
in firefox the second value seems to be very low, strange I will look into this. Still with v8 the values seem to make sense

webzwo0i:
in chromium:
match with regex  19071.179000195116
split with non-regex seperator 10244.592000031844
split with regex seperator 40643.17000005394
so I still dont know what optimization happens in firefox but at least for v8-based engines split with a single character seems to be faster

webzwo0i:
Need to check if every call to applyToAText is wrapped in try-catch at some level

dmitryuv:
> could you tell my where I can find your changeset-checks?
it's here https://github.com/dmitryuv/atext-changeset
but the latest version is here https://github.com/dmitryuv/otdartlib

the split version seems to be faster, my bad assumption here :), just it hides the code intention to count newlines so it harder to read.

webzwo0i:
hmm yes, maybe I should put it a line above like: newlines = strIter.peek...
In general, because the commits in this pull request reflect the errors I made more than the actual changes, maybe I should close and open another pull request:
reverting of the old stuff/fixing the code for = op
additional checks for - and + ops
try-catch block for the call sites

just asking in case you feel it got to confusing

JohnMcLear:
I think it's good to have a history of your learning /discovery tied into a pull request.

PR #2525, opened 2015-02-22
Set correct API version in doc

ldidry:
The doc says current version is 1.2.9, but there is getPadID which uses version 1.2.10.


JohnMcLear:
tnx =)

ldidry:
You're welcome :-)

PR #2530, opened 2015-02-26
do not crash when encountering mismatched compositions. log the changese...

webzwo0i:
...ts and padid
If this happens than its either an old pad with broken changesets in it and we have to manually clean the database (or fix the cs and try to reimport the pad).
Or when it is a new changeset during normal pad editing and the client is the etherpad editor, than we probably have a new bug and need to fix the editor's code - so we need to know where to start debugging.

easysync/backend/frontend tests pass, but please check if I do the error handling (i.e. returning in async.series) correctly

### DIFF
--- a/src/node/handler/PadMessageHandler.js
+++ b/src/node/handler/PadMessageHandler.js
@@ -1629,10 +1629,14 @@ function composePadChangesets(padId, startNum, endNum, callback)
       changeset = changesets[startNum];
       var pool = pad.apool();
 
-      for(var r=startNum+1;r<endNum;r++)
-      {
-        var cs = changesets[r];
-        changeset = Changeset.compose(changeset, cs, pool);
+      try {
+        for(var r=startNum+1;r<endNum;r++) {
+          var cs = changesets[r];
+          changeset = Changeset.compose(changeset, cs, pool);
+        }
+      } catch(e){
+        console.warn("failed to compose cs in pad:",padId);
+        return;
       }
 
       callback(null);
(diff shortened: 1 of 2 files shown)


JohnMcLear:
This looks good to me..  @dmitryuv @marcelklehr did you want to have any input?

dmitryuv:
The error in composition is fatal, so it should be bubbled up in callback chain. There should be no "return" statement because it breaks ASYNC control flow.

webzwo0i:
I updated the pull request.
Do I need to "return callback(the-error)" or is it enough to just callback(the-error)? I am sure I understood the details in the past but not right now...

The reason there was another warn-statement with the padId is because it is not enough (for debugging issues with broken changesets) to know the changesets. You sometimes need the pool or the last atext to construct the text before the revision that makes trouble.
I would love if at some point other admins of epl instances could securely share their logs (without any padId/urls, changesets, file names, session credentials) - more logs->less debug times because hopefully reproduction is easier.

Anyways, I removed the cs texts and print the revision numbers instead. I think this should be enough.

The alternative (not printing even padId) would imo prevent us from fixing those issues. But I will think about this a little bit more, also considering those cloud-log services some people use to aggregate their logs

dmitryuv:
> Do I need to "return callback(the-error)" or is it enough to just callback(the-error)? I am sure I understood the details in the past but not right now...

The difference is that if you don't return, you'll continue executing current function that have second callback() call at the end. That's a language basics. The rest looks good now.

JohnMcLear:
bump @webzwo0i

JohnMcLear:
Thanks! :)

PR #2533, opened 2015-02-27
Trigger renumbering when deleting (via cut) the first item of a list

### LINKED ISSUE #2514: Deleting the first item of a numbered list causes all items to be number one
If I select (by triple-clicking; dragging to select does not trigger it unless the selection is due to undo) and delete (with the delete key or cut, not backspace) the first item in a numbered list, all the items become number one. I would expect the number of all the remaining items to decrease by one instead.

If I paste the item back, undo, remove text from an item, or add another item, then it becomes numbered correctly again. If I remove an empty item to form two lists, the list above the removal remains all number one, and the list below is numbered correctly.

JohnMcLear mentioned in IRC that deleting the first item isn't triggering a numbering redraw. I'm using [Etherpad-lite](https://aur.archlinux.org/packages/etherpad-lite/) 1.5.1, but as of this writing it is also reproducible on beta.etherpad.org. This is on Arch Linux with Firefox 35.

cjb:
Fixes #2514.

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -3313,6 +3313,14 @@ function Ace2Inner(){
     return [rep.lines.offsetOfIndex(lineRange[0]), rep.lines.offsetOfIndex(lineRange[1])];
   }
 
+  function handleCut(evt)
+  {
+    inCallStackIfNecessary("handleCut", function()
+    {
+      doDeleteKey(evt);
+    });
+  }
+
   function handleClick(evt)
   {
     inCallStackIfNecessary("handleClick", function()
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Wow this is crazy clean..  Didn't expect this..  Does it work on pad "viewers" as well as the authors screen?

cjb:
@JohnMcLear Yep, seems to work fine for both the author and viewers immediately.

JohnMcLear:
Wow excellent!  I will give it a test..

JohnMcLear:
Hrm, the caret appears to jump now to the last line on delete, are you seeing that?

cjb:
No, don't think I can reproduce that -- is it easy for you to share a screencast?

It's a little weird to call doDeleteKey() when no key was pressed down, so it might be related to that, but I wasn't able to find any misbehavior with it here.

PR #2538, opened 2015-03-03
Check author attribution and fix clearAuthorship test

webzwo0i:
Users can submit changes as another author. Clients/new users think someone else did the change. Another possibility is that an author which never exists in globalAuthor is introduced to the pad's pool.

As of now I only know of one exception to the rule "only accept changes of the client's authorId' namely clearAuthorship which uses an empty author (like epl itself does, when it introduces newlines to the end of the document).

### DIFF
--- a/src/node/handler/PadMessageHandler.js
+++ b/src/node/handler/PadMessageHandler.js
@@ -656,7 +656,12 @@ function handleUserChanges(data, cb)
           , op
         while(iterator.hasNext()) {
           op = iterator.next()
-          if(op.opcode != '+') continue;
+
+          //+ can add text with attribs
+          //= can change or add attribs
+          //- can have attribs, but they are discarded and don't show up in the apool
+          if(op.opcode == '-') continue;
+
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Hrm this might break the Etherpad CLI Client & load Test setting as that skips the requirement to use a real session, will test..

JohnMcLear:
Seems to be okay...

JohnMcLear:
@marcelklehr @dmitryuv good to merge?

JohnMcLear:
Actually I was right htis does appear to break the CLI client..

JohnMcLear:
@webzwo0i can you do 
``etherpad http://127.0.0.1:9001/p/t2est -a 'herrro'`` when you run this branch?

Set "loadTest" to true in your settings.json first

I seem to be having issues connecting..

JohnMcLear:
I have a feeling this will break the CLI Client.  I will test that and let you know.

JohnMcLear:
Tested and works fine :) Thanks!

PR #2541, opened 2015-03-06
fixed + support for value

cristo-rabani:


cristo-rabani:
It is annoying when name of function and even its comment
provides to mislead!

example:
getAtribute doesn't really do what is sugested by name. 
This function should have name: 'getAttributeValue' because this is what this function do.

In version 1.4 function removeAttributeOnLine doesn't use declared parameters. 
and even description misstell.

If something is not done or doesn't have full functionality it should have in comment a notice or even such of function should be prefixed by '_'.

And formatting of code is ugly and mark  ';' always should be used, code should be clear. Please take a look on meteor repository (https://github.com/meteor/meteor)


Additionaly: I think that probably jsHint will be good option. 
proposed settings:
"bitwise": true,
    "curly": true,
    "eqeqeq": true,
    "freeze": true,
    "immed": true,
    "newcap": true,
    "noarg": true,
    "noempty": true,
    "nonbsp": true,
    "quotmark": true,
    "undef": true,
    "unused": true,
    "strict": true,
    "trailing": true,
    "maxcomplexity": 12,
    "maxlen": 120,
    "globalstrict": true,
    "devel": true,
    "node": true,
    "browser": true,
    "jquery": true

Please consider, to enter a attribute object. 
Such object can have much more information than array of name value pairs.
for example attribute can have remove function or information about where is located. 

I have such object and i can make pull request with it.

JohnMcLear:
+1 an attribute object.  When I first started using getAttributeOnLine I expected it to return an object of applied attributes.  It is confusing.

RE linting code, yep we're going through and doing this in small pieces as we refactor parts of the code.

We would welcome a pull request for objects.

As far as this pull request goes.. ``removeAttributeOnLine`` isn't broken for me, I'm using it fine.  What problem are you facing with it?

cristo-rabani:
Hello, 
1. Take a look on function removeAttributeOnLine.
   Do you see if(hasMarker)? What if you haven't hasMarker ?
   //Etherpad allows for such situations, otherwise this 'if' would have been unnecessary. Correct?
   This function worked only for line with hasMarker!
2. I added support for unused parameter "attributeValue" declared in arguments of this function.
   All declared parameters should be supported. 

You sound to me like this:
```
function getSumOfArray(arr, length){
  return arr[1]+ arr[2]+arr[3];
}
```
 `getSumOfArray` isn't broken for me, I'm using it fine.
Function works for my all cases because I always have only 4 items in array and first always is equal 0

JohnMcLear:
1.  Can you please explain a situation where this was actually a problem / bug though?  Was it during plugin authorship or?  cc @marcelklehr 

2. Again, what does this actually fix?

Sorry about so many questions, I'm just a bit confused to the actual bug you are trying to solve here, perhaps you have a link to the issue that is at fault that I can refer to?

src/static/js/AttributeManager.js
```diff
@@ -153,37 +153,36 @@ AttributeManager.prototype = _(AttributeManager.prototype).extend({
...
-       if(attribs.length <= 4) ChangesetUtils.buildRemoveRange(this.rep, builder, loc, (loc = [lineNum, 1]))
-       else ChangesetUtils.buildKeepRange(this.rep, builder, loc, (loc = [lineNum, 1]), attribs, this.rep.apool);
-     }
+       if (!found) {
+           return;
+       }
+       ChangesetUtils.buildKeepToStartOfRange(this.rep, builder, [lineNum, 0]);
+       var list = _.chain(attribs).filter(function(a){return !!a[1];}).map(function(a){return a[0];}).value();
+       //if we have marker and any of attributes don't need to have marker. we need delete it
+       if(hasMarker && !_.intersection(lineAttributes,list)){
```
marcelklehr: Where does `lineAttributes` come from? What is the assumption here? (author attrib has a value, so assuming that you meant `attribs` instead of `lineAttributes` this if statement would never be true as there'd always be an intersection in the case of the author attribute)

src/static/js/AttributeManager.js
```diff
@@ -153,37 +153,36 @@ AttributeManager.prototype = _(AttributeManager.prototype).extend({
...
-     })
+       var builder = Changeset.builder(this.rep.lines.totalWidth());
+       var hasMarker = this.lineHasMarker(lineNum);
+       var found = false;
 
-     if(!foundAttrib) {
-       return
-     }
+       var attribs = _(this.getAttributesOnLine(lineNum)).map(function (attrib) {
+           if (attrib[0] === attributeName && (!attributeValue || attrib[0] === attributeValue)){
```
marcelklehr: Nice one! +1

src/static/js/AttributeManager.js
marcelklehr: Also, what If the line has no marker?, then you apply attribs to the first char of the line. That may be ok, as there'd probably be no attribs in there anyway (since there's no line marker), but it's misleading at least. The assumption for putting the `if(hasMarker)` in there was that there are no attributes to be removed from the line if there's no line marker.

marcelklehr:
In addition to that: Both the old and your new implementation blindly trust getAttributesOnLine to only return attribs that were set on the entire line (aka on the line marker). I just checked and getAttributesOnLine seems to accumulate *all* attribs set on any part of the line. This needs to be fixed, as well.

src/static/js/AttributeManager.js
cristo-rabani: lineAttributes is defined in this file on line 11
```
var lineMarkerAttribute = 'lmkr';
 
 // If one of these attributes are set to the first character of a 
 // line it is considered as a line attribute marker i.e. attributes
 // set on this marker are applied to the whole line. 
 // The list attribute is only maintained for compatibility reasons
 var lineAttributes = [lineMarkerAttribute,'list'];
```
And I make list of attributes, which are attached currently to the line.
In other place I seen that the attributes without values are deleted automatically but they can existed a while
And this is the reason why I checked !!a[1] in underscore filter.

But I see possibility of having local attributes in line which does not have marker.

Maybe is some logical reason to separate line attributes from local attributes (applied on fragment of line). But currently and and before this was never controlled.

cristo-rabani:
Yes, this method removes line attributes and local too.
I don't know if this is bug because this function worked in that way by a lot of time and changing this behavior can spoil some plugins. Maybe better will be add new argument "excludeLocalAttributes"

JohnMcLear:
Have you tested this functionality with plugins such as ep_headings2 ?  Those plugins provide front end tests so you can just hitup /tests/frontend to see if it still works...

cristo-rabani:
Yes, I made test with other plugins.
My function doesn't change any logic it works like before.
I only added support for unsupported cases like support for last parameter which was defined before but never used by this function. And for situation where we haven't marker or if we have less that 3 parameters or we want remove one of those three.

JohnMcLear if you don't want my pull requests, please tell me. 
I haven't time to banter.
And you don't understand that implementing functions with defined but don't supported parameters is very bad thing. (because all works for your)
 I do not wanna explain to you that opposite cases should be handled or documented. 
This function is called by other with parameters whose this function don't support!!!
And if someone doesn't know about this,  will be wasted time for debugging.

And you tells me, in answer that:
> `removeAttributeOnLine` isn't broken for me, I'm using it fine. What problem are you facing with it?

JohnMcLear:
@cristo-rabani okay cool.  I do want your pull requests. The reason I ask so many questions is because it was unclear in your initial PR what you are fixing so I try to get an answer for reference.  Often I know but we need a track record so things are verbose...  It's not me questioning you or code it's me trying to ensure others can review and understand what is being changed and why :)  

I understand if you don't have time to banter so if you are going to be contributing to Etherpad for a while it might make sense to give you admin on the repo (once you have gotten used to the coding conventions and used to explaining things before you merge) and then you will be able to whack your own stuff in.  I'm not controlling about this stuff I just try to ensure we're all being as coherant as possible so others find it easy to follow.  Also on our journey I might help improve your English and you might improve my JS ;)  

@marcelklehr this is your code so if you are happy with these changes please merge them :)

marcelklehr:
Try indenting a line and then outdenting it. You're left with the line marker.

JohnMcLear:
@cristo-rabani peep what @marcelklehr wrote, looks like your code breaks core functionality :(

cristo-rabani:
Hello,
Yes, there was a bug.
index is started from 1 not from 0.
I think that this should be correct now.

Could you fix travis configuration?

marcelklehr:
Nope, didn't fix it. sorry.

cristo-rabani:
Hello guys, 

Now It should be good. 
The problem was in variable 'lineAttributes'

I made my own list, I found list of line attributes in `setAttributeOnLine` method.
Please consider to make a getter or something else for those attributes. 

I checked it with list atribute, headings, headings 2.
I made mixed few attributes and always this function removes only attribute from parameter.
`countAttribsWithMarker` checks if we have only those that what it should be

JohnMcLear:
Looks good and tests okay.  The docs here were terrible and this PR was really poorly executed.  In the future please try to be more coherent with your PR's!  If in doubt just ask I'm happy to help :)

JohnMcLear:
Thanks!

PR #2549, opened 2015-03-13
Support CA bundle files

lid2000:
This adds support for the intermediate certificate files that some SSL providers require. Code adapted from http://www.benjiegillam.com/2012/06/node-dot-js-ssl-certificate-chain/

If you want to use a CA bundle, then add the path of the file to `settings.ssl.ca` in the settings.json file.

### DIFF
--- a/src/node/hooks/express.js
+++ b/src/node/hooks/express.js
@@ -40,10 +40,27 @@ exports.restartServer = function () {
     console.log( "SSL -- server key file: " + settings.ssl.key );
     console.log( "SSL -- Certificate Authority's certificate file: " + settings.ssl.cert );
     
+    if (settings.ssl.ca) {
+      var ca = [];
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Um, don't you just put int certs into your public cert?   We have int certs on beta.etherpad.org..  It' just a case of appending the int cert to the public cert file and expresss handles it fine..  No need for us to add any additional logic..  Please check that @lid2000

lid2000:
Yep you could do that - this way is just neater for me.

JohnMcLear:
As much as this might make things neater I feel like bundles can be just um, bundled..  I just don't wanna have to maintain some functionality that can be solved with a simple ``cat foo >> bar``

Thanks anyway though, if this issue comes up in the future at least we can point to this PR as a solution =)

PR #2559, opened 2015-03-20
fixes #2556 error toggling line attribute; clarified documentation

### LINKED ISSUE #2556: [Bug] Error in AttributeManager.toggleAttributeOnLine()
The toggle function does not work properly for me. 

The reason seems to be that this call: https://github.com/ether/etherpad-lite/blob/develop/src/static/js/AttributeManager.js#L198
should be 
`this.getAttributeOnLine(lineNum, attributeName)`

tm-linkwerk:


JohnMcLear:
tnx = )

Issue #2558, opened 2015-03-19
How to test if an attribute is set at a range


tm-linkwerk:
Hi,

I want to find out if an attribute is set at a certain range (in my case the selected text). 

There is a function `editorInfo.ace_getAttributeOnSelection(attributeName)` which name suggests that it should solve my problem but (seems to) return true only if the value of the attribute is true, and false otherwise. 

How can I get the information if an attribute is set to any value? Or even better, get a list of all attributes that are set to any value on a specified range?

JohnMcLear:
You can do it on the caret location/selected range such as this: 
https://github.com/JohnMcLear/ep_font_size/blob/master/static/js/index.js#L63

This will return the value of the attribute IE if you want the font size of this selection it will return the integer associated with that font size, not just a boolean..

There isn't a way to get a list of all attributes on a specified range at current.  I'd welcome this functionality though =)

tm-linkwerk:
That's the way I tried it, but got only boolean values, and looking at the source code of the function https://github.com/ether/etherpad-lite/blob/develop/src/static/js/ace2_inner.js#L2325 I also only found boolean returns.

JohnMcLear:
Try ``console.log`` what the ep_font_size plugin gets, use it as a point of reference.

tm-linkwerk:
The ep_font_size plugin uses only boolean values for attributes (like fs12=true) so it is not affected by the problem.

JohnMcLear:
I'd agree that hooks such as ``getAttributeValueOnSelection`` and ``getAttributesOnSelection`` would be useful if you wanna craft those in?

tm-linkwerk:
For a given position it turned out to be not as difficult to implement as suspected. So I added the function to the AttributeManager and created a pull request. 

But for a range, after thinking about it, I am not yet sure, what a function should reasonably return if the selected range overlaps attributes. So maybe I come back to this point later.

JohnMcLear:
Cool thanks :)

PR #2565, opened 2015-03-25
feature #2558 get attributes on position

tm-linkwerk:


JohnMcLear:
Looks good to me from initial looks, haven't tested.  I wonder if it's possible to do test coverage?  Perhaps put tests into a plugin or so?

tm-linkwerk:
Unfortunately I'm not yet familiar with testing in etherpad, though I would really appreciate to have test coverage. Shouldn't it be possible to add tests in etherpad itself? At first glance I didn't find any tests for AttributeManager in https://github.com/ether/etherpad-lite/tree/develop/tests.

JohnMcLear:
I guess what you could do is add some functionality to set the bold button to active based on the caret location (and if the font is bold), that way you would test against that functionality?  That's the only thing I can think to do

tm-linkwerk:
Hmm, I have no idea how to add that functionality to the core.

src/static/js/AttributeManager.js
```diff
@@ -123,6 +123,57 @@ AttributeManager.prototype = _(AttributeManager.prototype).extend({
...
+        return attributes;
+      }
+    }
+    return attributes;
+    
+  },
+  
+  /*
+    Gets all attributes at caret position 
+    if the user selected a range, the start of the selection is taken
```
JohnMcLear: Missing docs for what is returned See getAttributesOnPosition

JohnMcLear:
Thanks! :)

Issue #1803, opened 2013-06-10
Improve accessibility for blind people and other users of screen readers


lizzard:
Might there be some ways to usefully improve Etherpad Lite's accessibility? I know that people who browse with a screen reader can't use it. Etherpad is so useful that I hate to turn away from it as a collaboration tool.  Is this something you have discussed already in issues or in a forum? I looked, but didn't find anything. If I can help gather suggestions I'm happy to do that.

JohnMcLear:
Help us help you: https://www.bountysource.com/issues/1371778-improve-accessibility-for-blind-people-and-other-users-of-screen-readers

JohnMcLear:
Hey guys, I have been tasked with working on this so if anyone can jump in and help please say! :)

sathomas:
Hi John. I'm definitely interested in working on this. I'm swamped at work for the next week or so, but might have some time to help after that. I have no knowledge of the Etherpad code base, but, as a user, I do have some ideas on how to improve its accessibility.

One suggestion I'd make is to first worry about the single user case (saving the live collaboration case for the second phase.) Even though one of Etherpad's major features is collaboration, I suspect there are plenty of folks that use it, at least some of the time, alone. There are also some relatively "low hanging fruit" for the single user case. If you like, I can try to put together a quick list now or a more reasoned and thoughtful list next week.

sathomas:
Some initial thoughts:

* [CKEditor](http://ckeditor.com/demo#a11y) is really strong on accessibility and, in general, probably serves as a great model.
* Probably the biggest accessibility problem with Etherpad is that it traps the keyboard user in the main textarea. Once a user navigates to the textarea, there is no way (using the keyboard) to navigate out. A first step here would be to define a special key combination (I like `Esc`, but other online editors seem to have settled on `Alt-F10`) that shifts the focus from the main textarea to the first toolbar button. From there, the user should be able to `Tab` to other toolbar buttons (but see below).
* Whatever the key combination(s) that support this behavior, there needs to be a way to tell the keyboard user that they exist. Perhaps a pop-up that's "visible" to screen readers, though that does introduce it's own i18n issues.
* A related issue for those sites that embed Etherpad documents in iframes is how to let a keyboard user "escape" from the iframe and get back to the main page. That will probably require some kind of plugin that enable `postMessage` communication between the Etherpad iframe and the main page. Since Etherpad seems to use a lot of nested iframes, this might not be as trivial as it first appears.

[...]

PR #2566, opened 2015-03-25
Accessibility

JohnMcLear:
Merging on 7th of April.  Get your complaints in BEFORE then.

Based on the notes at: http://john.primarypad.com/ogf-24-03-15

Designed to resolve: https://github.com/ether/etherpad-lite/issues/1803

 - [x] Use Alt Keys to Access Editbar
 - [x] Use Arrow keys to Navigate Editbar
 - [x] Use Alt Keys to access Chat etc.
 - [x] Provide a usable screen reader experience (ARIA specs) (Mostly for visually impaired hombres)
 - [x] Provide fonts suitable for dyslexic users

New tasks as per: https://beta.etherpad.org/p/pvagner%27s_test
 - [x] Make Timeslider Accessible
 - [x] Show users part is not intuitive
 - [x] Additional Keypresses need to be documented in the WIki
 - [x] Hover on buttons is not consistent due to new titles
 - [x] Pressing on Import/Export button seems to lose focus
 - [x] Make Alt C work from anywhere
 - [x] Alt F9 should only work WITH Alt key
 - [x] Make Escape & Alt Keys work from anywhere (not just pad)
 - [x] Ensure left/right arrow keys don't steal focus when on inputs.
 - [x] High Contrast UI - Available through ep_themes
 - [x] Include Open Dyslexic - "Enable Dyslexic(lulz) friendly font" (prolly a pluigin)
 - [x] Fix Good/Bad & Smiley face instead of Good/Bad - Click event with more info including reading score etc. (Blahblahbblahbhlalalalal)
 - [x] Allow left-right arrow keys to cycle from start > end > start again infinitely.
 - [x] Orca and FF 3.14 says the name of the element on focus <-- I run unity which doesn't allow Orca to be easily updated.  I might have to resolve this bug once Orca stable lands on Ubuntu unless someone can jump in and resolve that runs Gnome 3  <-- getting a VM to help w/ this -- Orca test is on hold, doing JAWS first.
 - [x] Can't return to pad from Timeslider
 - [x] Use Control Shift 2 instead of Alt Shift A

 - [ ] Provide a nice experience for Showing Authors of a line (curent MVP uses Ctrl-Shift-2, try this!)
 - [ ] Test with JAWS
 - [ ] Write tests for keystrokes -- Is this possible?  Looks like it might just be!  (see bold.js)
 - [ ] Provide in app documentation?
 - [x] Change Font on Timeslider
 - [x] High Contrast on Timeslider (ep_themes)

TEST at http://beta.etherpad.org
DOCS at https://github.com/ether/etherpad-lite/wiki/Accessibility


JohnMcLear:
Please note Alt A is now Alt Shift A as per https://bugzilla.mozilla.org/show_bug.cgi?id=1149579

JohnMcLear:
Thinking this for a setting: https://d262ilb51hltx0.cloudfront.net/max/857/0*4QM6Tc6smxr9IWR-.png

On by default

"Keyboard shortcuts" would be a page to a rendered tab of HTML page w/ accessibility controls on built from the Wiki?

JohnMcLear:
Alt Shift A is now Control Shift 2...

PR #2582, opened 2015-04-07
Add esperanto translation from translatewiki.net

mschmitt:


JohnMcLear:
We need @siebrand to handle translatewiki merges :) Thanks though!

PR #2600, opened 2015-04-11
fix: only match /javascript/* for caching middleware

devoidfury:
This should correct the issue of fonts and images being corrupted, introduced in the express4 change.


JohnMcLear:
Thanks! :) Works 10/10

PR #2602, opened 2015-04-11
fix large etherpad import, fixes #2524 

### LINKED ISSUE #2524: Large .etherpad files cause Maximum range exceeded error
Pads with tends of thousands (50k or so) revisions when imported cause this error.  Not sure why yet, for now I have put a limit in to 50k keys in an export.

devoidfury:
Made setPadRaw async. The async library expects functions to be asynchronous, but they aren't if you just call the callback immediately, and this can lead to blowing the stack. Wrapped callbacks in `setImmediate` to fix.

Also, file import handling streamlined a bit.

This will probably need some testing to make sure I didn't break the whole thing.

#2524

### DIFF
--- a/src/node/handler/ImportHandler.js
+++ b/src/node/handler/ImportHandler.js
@@ -113,10 +113,8 @@ exports.doImport = function(req, res, padId)
         if(ERR(err, callback)) return callback();
         if(result.length > 0){ // This feels hacky and wrong..
           importHandledByPlugin = true;
-          callback();
-        }else{
-          callback();
         }
+        callback();
       });
     },
     function(callback) {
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Seems to work pretty well with both abiword enabled and disabled!  Nicely done!

JohnMcLear:
Doesn't matter so much if it breaks some fringe import events as we're gonna be breaking quite a few things over the coming days!

ldidry:
I applied the patch, and it's better, but it is still not perfect. I'm now able to import file up to 1.5Mio, but after that, it says the import was successful, but the pad is still empty :disappointed:

webzwo0i:
If the import was successful but you are not seeing the pad (rendering could take a moment with complex pads) then the error could be client-side. Do you see anything in the console output of your browser's debugger (via developer mode)?

ldidry:
No, I use curl to import the pad (I have a lot of pads to import), and I got an "ok" message. If I go to the pad with my browser, the result is still an empty pad.

If I try again to import the pad, I got the "ok" message instead of a "padHasData" message (which  happens when you try to import a pad on a pad which already has data). So I think the import doesn't work at all.

JohnMcLear:
We could probably add backend test coverage for this
----- Reply message -----

webzwo0i:
and your etherpad.log does not show anything? I would change "loglevel":"info" to "loglevel":"debug" and see if anything is at the end of the log. You should see lots of "[DEBUG] ueberDB - SET - pad:foo...:revs:"

PR #2603, opened 2015-04-11
cache template data, avoid blocking for template rendering

### LINKED ISSUE #642: eejs is using blocking API :: Wont fix as it doesn't actually effect performance noticably.
readFileSync blocks the event loop until the file is read so it should not be used on every request: https://github.com/Pita/etherpad-lite/blob/7cf4510bf545927edead70931befb7d62ca5863c/src/node/eejs/index.js#L104 
//cc @redhog

devoidfury:
This patch saves template lookups in memory to avoid reading from the disk on every template load or in-template `require` call. It will still hit the disk a few times the first time a template is referenced, but will not for subsequent requests (unless `exports.NO_CACHE = true`, which reverts to the old behavior, for development). 

There's still some logging in there, should make it easy to see what's going on.

Help to resolve #642.

### DIFF
--- a/src/node/eejs/index.js
+++ b/src/node/eejs/index.js
@@ -32,6 +32,8 @@ exports.info = {
   args: []
 };
 
+var require_cache = {};
+
 function getCurrentFile() {
   return exports.info.file_stack[exports.info.file_stack.length-1];
 }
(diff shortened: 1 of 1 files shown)


devoidfury:
Here's something to help test the sync `fs` methods:

```js
var fs = require('fs');
Object.keys(fs).forEach(function(name) {
  if (name.toLowerCase().indexOf('sync') === -1) return;
  var old = fs[name];
  fs[name] = function() {
    console.log('BLOCKING ' + name + (new Error()).stack.split('\n')[2].trim());
    return old.apply(fs, arguments);
  }
})
```

At the top of server.js works, but you'll get a *ton* of messages until it finishes loading.

JohnMcLear:
cc @redhog

JohnMcLear:
Should the ``minify:``and ``maxAge:`` settings from ``settings.json`` have an effect on this possibly disabling this caching?  I just don't want plugin authors templates to be cached if they are actively developing them...

redhog:
Why not have a separate option in settings.json?

Maybe we should group all these and add a comment like "These options should only be changed for development" or something?

devoidfury:
>Why not have a separate option in settings.json?
>
>Maybe we should group all these and add a comment like "These options should only be changed for development" or something?

I agree, that seems best.

We could go the 'dirtydb route', where it's disabled by default but has a warning log line and comments in settings.json. That would have less surprising behavior, but might be overlooked.

JohnMcLear:
I think we have another PR with a more elegant solution?  If so please close this.

PR #2609, opened 2015-04-13
Requirejs

redhog:
I have AMDified lots of code, and it all works. Better to merge often and merge early than have lots of conflicts later...


JohnMcLear:
I'm gonna be testing this today and creating a checklist of stuff left to do =)

redhog:
ping @JohnMcLear I have merged iframes-must-die into requirejs branch and it all works :)

JohnMcLear:
Niceish but that will break most plugins.  Something I wanted to avoid.  Can we chat about this next week?
----- Reply message -----

JohnMcLear:
How are you set for Wed pm?  I can be free then if you can?
----- Reply message -----

JohnMcLear:
Closing because it's way out of scope for now.

PR #2636, opened 2015-04-22
Adding support for providing intermediate CA certificates 

andreassolberg:
When running etherpad-lite with ssl through Node/expressjs, you would typically like to be able to provide intermediate certificates as part of the SSL handshake. This is supported by expressjs, but some extra config was needed.


src/node/hooks/express.js
```diff
@@ -46,6 +46,13 @@ exports.restartServer = function () {
       key: fs.readFileSync( settings.ssl.key ),
       cert: fs.readFileSync( settings.ssl.cert )
     };
+    if (settings.ssl.ca) {
```
JohnMcLear: wouldn't this error if !settings.ssl

PR #2643, opened 2015-04-29
timeslider: Fix export links

AlexanderS:
Names of the pads can contain more charaters than \w. So while
transforming the export links, we simply can allow all charaters
except the slash as pad names.


JohnMcLear:
Thanks

JohnMcLear:
Worth checking this for XSS, should be fine but just be careful.

PR #2649, opened 2015-05-05
When using setText(), replace the entire existing text of the pad.

xavidotron:
Fix setText() to replace the full pad text rather than replacing all except the last letter.  Previously, e.g., calling setText() with a pad's existing contents would make the pad one character longer.

### DIFF
--- a/src/node/db/Pad.js
+++ b/src/node/db/Pad.js
@@ -290,7 +290,7 @@ Pad.prototype.setText = function setText(newText) {
   var oldText = this.text();
 
   //create the changeset
-  var changeset = Changeset.makeSplice(oldText, 0, oldText.length-1, newText);
+  var changeset = Changeset.makeSplice(oldText, 0, oldText.length, newText);
 
   //append the changeset
   this.appendRevision(changeset);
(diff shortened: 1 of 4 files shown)


JohnMcLear:
Do me a favor and quickly run ``bin/backendTests.sh`` -- this looks fine but I want to ensure it's not affecting anything else.

xavidotron:
Thanks!  The tests did depend on the old behavior, so I added a commit to fix them.  The tests also revealed that my change broke setPadHTML(), which seems to depend on setText('') leaving a single newline character in the pad.  I don't understand the implementation of setPadHTML() enough to understand the failure, but changing that line to setText('\n') seems to fix the test.

JohnMcLear:
Okies, prolly worth running the frontend tests just to be sure nothing else got b0rked.

http://youretherpad.com/tests/frontend

Once they pass we should be good to merge :)

xavidotron:
Frontend tests pass without issue.  Thanks!

JohnMcLear:
Thanks :)

JohnMcLear:
@xavidotron looks like @LaKing found a bug with your commits, can you take a look at #2700 please?

JohnMcLear:
@LaKing can you please do me a favor and include a backend test that replicates your issue?  That way we can ensure it doesn't happen again :)

LaKing:
@JohnMcLear  I would need help on that. The bug is propably backend side, but it shows up on the frontend when using the test plugin. ...

PR #2657, opened 2015-05-07
Use the cookie parser middleware

simong:
When running etherpad with:
 - settings.requireSession = true;
 - settings.sessionKey = "some value";

the PDF export will *always* return 403 because the cookies never get parsed and the access check in `src/node/padaccess` boils down to:

```javascript
req.cookies = {};
checkAccess(.., req.cookies.sessionID, ...);
```

### DIFF
--- a/src/node/hooks/express/webaccess.js
+++ b/src/node/hooks/express/webaccess.js
@@ -6,6 +6,7 @@ var hooks = require('ep_etherpad-lite/static/js/pluginfw/hooks');
 var ueberStore = require('../../db/SessionStore');
 var stats = require('ep_etherpad-lite/node/stats');
 var sessionModule = require('express-session');
+var cookieParser = require('cookie-parser');
 
 //checks for basic http auth
 exports.basicAuth = function (req, res, next) {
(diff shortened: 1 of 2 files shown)


JohnMcLear:
@simong One thing to be aware of is we changed the sessionKey logic so it's no longer a requirement in ``settings.json`` -- Does the functionality still work as desired if sessionKey is not included in settings.json ?

simong:
Yes, If you don't specify a `sessionKey` you will simply not have any signed cookies, that is all.

JohnMcLear:
if you don't specify sessionKey the sessionKey is automatically generated..  You realize that right?

simong:
Ah, I didn't realise that. Regardless, I've tested this both with and without a sessionKey in settings.json, with and without a SESSIONKEY.txt and it works just fine. All it does is [add the `signedCookies`](https://github.com/expressjs/cookie-parser/blob/master/index.js#L40) on the request if it can successfully decrypt the cookies

JohnMcLear:
cool n1 ta =-)

PR #2665, opened 2015-05-15
Fixing documentation of collectContentPre

lpagliari:
On documentation of `collectContentPre`, the explanation about how to set a value to an attribute was incorrect (it said you need to use `"attributeName:value"`, but it is actually `"attributeName::value"`).


lpagliari:
I see that the build failed, but I've only changed documentation...

PR #2668, opened 2015-05-18
Tidy HTML before trying to convert it with abiword

simong:
We're running into quite a bit of issues in our production environment where users generate quite big documents (think lots of nested lists) and can't export them to PDF due to abiword not being able to handle the exported HTML.

This PR cleans up the HTML before handing it off to abiword which seems to help quite a bit

### DIFF
--- a/settings.json.template
+++ b/settings.json.template
@@ -10,12 +10,12 @@
   // favicon default name
   // alternatively, set up a fully specified Url to your own favicon
   "favicon": "favicon.ico",
-  
+
   //IP and port which etherpad should bind at
   "ip": "0.0.0.0",
   "port" : 9001,
 
-  /*  
+  /*
   // Node native SSL support
   // this is disabled by default
   //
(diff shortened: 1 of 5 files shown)


JohnMcLear:
Looks relatively clean, all tests pass with it enabled/disabled?

Can you please add a test for exporting HTML and make it expect to fail w/ bad pad data?

simong:
Can you give some more details about how to accomplish this? AFAICT, both backend and frontend tests are integration tests which require a running etherpad instance that can't be configured from inside the tests itself. Keep in mind that abiword needs to be enabled for these tests to work and that it's currently not installed on Travis CI.

Is there a place for unit-level tests? I could easily write some tests that test TidyHtml directly for example

simong:
I pushed some backend tests. Let me know if those are acceptable. (I also used log4js in TidyHtml)

JohnMcLear:
Should be good, I will merge and we can deal with any fallout as it lands.  Thanks!

JohnMcLear:
Might be worth putting some docs in for people too?  Perhaps the wiki is the right place for that

simong:
Thanks! I didn't realise I could edit the wiki, I've added some docs in https://github.com/ether/etherpad-lite/wiki/How-to-enable-importing-and-exporting-different-file-formats-with-AbiWord

JohnMcLear:
Yea the wonders of Github :) Thanks!

PR #2681, opened 2015-05-25
Allowing 'value' to be a localizable attribute on HTML elements.

lpagliari:
Currently only `title`, `innerHTML`, `alt`, and `textContent` can be localizable using html10n.js. But if you have a form with submit button, as the [ep_comments_page plugin] (https://github.com/JohnMcLear/ep_comments) has (check the comment when user suggested a change 
image
), the button won't be localizable.

I've added `value` as an attribute that is allowed to be localized, ran the backend and frontend tests, everything working fine.

### DIFF
--- a/src/static/js/html10n.js
+++ b/src/static/js/html10n.js
@@ -851,6 +851,7 @@ window.html10n = (function(window, document, undefined) {
        , "innerHTML": 1
        , "alt": 1
        , "textContent": 1
+       , "value": 1
        }
     if (index > 0 && str.id.substr(index + 1) in attrList) { // an attribute has been specified
       prop = str.id.substr(index + 1)


JohnMcLear:
Looks useful, need to check to see how this works w/ Accessibility and Screen readers though as some of those items depend on values afaik.

lpagliari:
@JohnMcLear do you mean testing with screen readers to check if this change didn't break anything? Or is your concern related to allowing developers to create plugins that might break things without they being aware of that? In this case, should we restrict the change of `value` attribute only when target element is an `input` with `type=submit`?

JohnMcLear:
Yes @lpagliari - screen readers :)

It might be that i18n is fine for all inputs so it's probably just worth checking it to see how it behaves.

lpagliari:
@JohnMcLear I've tested NVDA + Firefox + Windows: when mouse is over the localized button, the screen reader is reading the correct word (the localized one).

I've compared the behavior with and without changing the value of the button, and it was exactly the same (same word read by NVDA).

JohnMcLear:
Perfect thanks!

On 26 May 2015 19:01, Luiza Pagliari <notifications@github.com> wrote:

@JohnMcLear<https://github.com/JohnMcLear> I've tested NVDA + Firefox + Windows: when mouse is over the localized button, the screen reader is reading the correct word (the localized one).

I've compared the behavior with and without changing the value of the button, and it was exactly the same (same word read by NVDA).

PR #2684, opened 2015-05-30
Conver.js: Use current mysql client API

hinnerk:
The used mysql client API was outdated. This change uses the current API and was successfully tested on a database containing ~2500 pads.


JohnMcLear:
thanks

PR #2688, opened 2015-06-03
Slight tweaks to the startup script for Windows

JonTheNiceGuy:
This uses the path to the batch file (assuming it is being launched as a shortcut), and then opens the default browser with the path to Etherpad.

This does all assume that this is being run on a user's desktop rather than on a server. Perhaps a subsequent change for running in server-only mode should be proposed?


JohnMcLear:
This isn't the desired behaviour

PR #2690, opened 2015-06-10
Add new icon to fontawesome subset: comment

lpagliari:
I needed an icon for ep_comments_page that was not on the FontAwesome subset provided by Etherpad, so I added it using Fontello and the existing `config.json` on the project.


lpagliari:
Added how-to on wiki about this: https://github.com/ether/etherpad-lite/wiki/Adding-a-new-font-icon-to-fontawesome-etherpad

JohnMcLear:
Nice thanks

On 11 Jun 2015 18:37, Luiza Pagliari <notifications@github.com> wrote:

Added how-to on wiki about this: https://github.com/ether/etherpad-lite/wiki/Adding-a-new-font-icon-to-fontawesome-etherpad

DrHuxtable:
I tried to update this all locally and ended up my browser console crowing about missing font-awesome, and the ep_comments as icons doesn't seem to work at all, even with the JSON clause in my settings. I thought I'd pull down fresh, before I made mention, but even after that I'm having the same issue.

Version number: 1.5.6
Git sha: a0aa947

JohnMcLear:
@DrHuxtable try do a ``bin/cleanRun.sh``

JohnMcLear:
@lpagliari I am noticing the bin icon doesn't appear to be working any more for delete?

lpagliari:
Two things you'll need to do at this moment to make sure icons will work:

1. Use the latest version of develop branch of Etherpad, not a stable
version (this is because comment icon was merged yesterday to the code,
there's no stable version of it yet);

2. Use "minify":false on your settings (apparently minifying makes plugins
not able to find fonts on the correct folder).

#1 should be unnecessary when we have a new stable version of Etherpad
released; #2 is a bug and still need to be fixed.

Just as a side note: #2 is necessary even if you don't configure
ep_comments to use icons, otherwise the "delete" icon will not be displayed
correctly. Try to delete a sidebar comment, and you'll get the same 404
error message about fontawesome not found on your console. At least on my
environment I saw that happening.

Hope this helps.

lpagliari:
John, check #2 I've mentioned on the previous comment.

It seems that, when minifying, plugins are looking for the font on /font,
but they are on /static/font.

If disabling minify does not work for you, let me know and I'll check what
the cause is. I'm planning to look for a fix for this today.

JohnMcLear:
Doing a "Clear browsing data" appears to have sorted it.

JohnMcLear:
FWIW I'm still not seeing the alternative comment icon view thing as per my comment on your PR

DrHuxtable:
@JohnMcLear,@lpagliari: Thank you both. I'll update to the latest dev branch in an effort to help me run the comments-as-icons feature (and use bin/cleanRun.sh in conjunction), and I'll try temp adding a softlink '/font -> /static/font' locally to see if that will suffice for now on the icon front, as clearing my cache during my previous trials did not help. The last time I disabled minify, I could actually see the performance lags, so that's going to be my last ditch effort. Thanks again. I love the plugin, and I really like this comment icon idea and can't wait to get it folded into my setup.

PR #2695, opened 2015-06-17
On Import, when successful, loading circle didn't go away

goldquest:
After successfully importing a document, the loading circle didn't hide itself.

This can happen, if you try to access a variable, which was not declared before. Even if you soft-check it against null, when it was not declared before JS throws a ReferenceError (and doesn't return undefined).


JohnMcLear:
n1 tnx

goldquest:
np

PR #2696, opened 2015-06-19
Allow a pad to end with only a single newline.

xavidotron:
Currently, if a pad ends with fewer than two newlines, Etherpad will add newlines.  Ending with a single newline should be sufficient, and avoids unnecessary diffs when integrating with other systems.


JohnMcLear:
@xavidotron did you run the backend tests to see if this passes?

```
bin/backendTests.sh
```

My concern is the export tests might fail?  Should be okay though :) Let me know

xavidotron:
I ran the backend tests and they passed.

PR #2697, opened 2015-06-20
prefetch pad files from index page

JohnMcLear:
This is thanks to the excellent work by @kentonv at @sandstorm as per https://github.com/kentonv/etherpad-lite/commit/910c507882d93ccd5440a6af0a5d41b937718c30

Prefetching the pad files while on the index.html makes a lot of sense as it's your most likely next destination.  

In this branch I'm going to explore using prefetch to optimize Etherpad performance for the client.

The develop branch going from index to pad: 
- index load time, 51ms 32.7Kb
- index > pad load time, 740ms 1.5Mb?? <-- doesn't add up
- pad refresh load time, 345ms 102Kb

In this branch:
- index load time, 129ms 121Kb
- index > pad load time, 669ms 208Kb?? <-- doesn't add up
- pad refresh load time, 239ms 185Kb

### DIFF
--- a/src/node/hooks/express/specialpages.js
+++ b/src/node/hooks/express/specialpages.js
@@ -1,4 +1,5 @@
 var path = require('path');
+var async = require('async');
 var eejs = require('ep_etherpad-lite/node/eejs');
 var toolbar = require("ep_etherpad-lite/node/utils/toolbar");
 var hooks = require('ep_etherpad-lite/static/js/pluginfw/hooks');
@@ -95,6 +96,26 @@ exports.expressCreateServer = function (hook_name, args, cb) {
 *
 */
 function sendPadHeaderFiles(res, callback){
+  var plugins = hooks.plugins.plugins;
+  var pluginPaths = [];
+  // Go through each plugin and get the paths
+  Object.keys(plugins).forEach(function(key) {
+    if(key === "ep_etherpad-lite") return;
+    var plugin = plugins[key];
+    var client_hooks = plugin.parts[0].client_hooks;
+    Object.keys(client_hooks).forEach(function(k){
+      pluginPaths.push(client_hooks[k]);
+    });
+  });
+
+  var uniquePaths = pluginPaths.filter(function(item, pos) {
+    return pluginPaths.indexOf(item) == pos;
+  })
+
+  console.warn("paths", uniquePaths);
+
+  // Now we need to uniquify the array
+
   res.set('Link', '<static/js/require-kernel.js>; rel=prefetch \, \
     <javascripts/lib/ep_etherpad-lite/static/js/pad.js?callback=require.define>; rel=prefetch \, \
     <javascripts/lib/ep_etherpad-lite/static/js/ace2_common.js?callback=require.define>; rel=prefetch \, \
--- a/src/node/hooks/express/specialpages.js
+++ b/src/node/hooks/express/specialpages.js
@@ -119,7 +119,7 @@ function sendPadHeaderFiles(res, callback){
   var uniquePluginString = "";
 
   for (var i = 0; i < uniquePaths.length; i++) {
-    uniquePluginString += "<static/plugins/"+uniquePaths[i]+".js>; rel=prefetch\, ";
+    uniquePluginString += "</javascripts/lib/"+uniquePaths[i]+".js>; rel=prefetch\, ";
   }
 
   // console.log("unique plugin string...", uniquePluginString)
--- a/src/node/hooks/express/specialpages.js
+++ b/src/node/hooks/express/specialpages.js
@@ -103,9 +103,11 @@ function sendPadHeaderFiles(res, callback){
     if(key === "ep_etherpad-lite") return;
     var plugin = plugins[key];
     var client_hooks = plugin.parts[0].client_hooks;
-    Object.keys(client_hooks).forEach(function(k){
-      pluginPaths.push(client_hooks[k]);
-    });
+    if(client_hooks){
+      Object.keys(client_hooks).forEach(function(k){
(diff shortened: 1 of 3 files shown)


JohnMcLear:
Gah I tested this with a  bunch of plugins enabled..  Testing again without plugins would make sense..

Questions we need to answer.
  - [ ] Is it worth prefetching in the pad.html too?
  - [ ] Can we put this in core or should it be a plugin?  I'd vote for Core as it improves UX.  The negative is the increased traffic a user would see if they hit index.html and don't want to visit a pad.
  - [ ] Are we catching all the files we can prefetch?
  - [ ] Should we provide a mechanism / documentations so plugins can easily expose files to prefetch?  Doing an example with ep_page_view would make sense.

kentonv:
Random thoughts:

* Be sure to test on a network with some latency. This is all about removing network round trips so if you test against a localhost server you won't see much benefit. :)

* Prefetching plugin Javascript is the most important thing, since otherwise they are loaded serially. But will that need help from the plugins to declare their dependencies? In my case I simply brute-forced it by recording exactly the files that were normally loaded using my particular plugin set.

* It's probably preferable to use HTTP `Link` headers over `<link>` tags as a SPDY or HTTP/2 proxy can probably do nice optimization with it. I only used `<link>` in my hack because Sandstorm currently drops `Link` headers (something I should fix).

* You could prefetch on on the root page, but keep in mind that Sandstorm skips that page. For Sandstorm I'll need prefetching on the pad page.

* Also note that if you prefetch on the root page, you'll want to make sure that caching headers (specifically max-age) are set correctly so that the browser doesn't have to ping the server again to check that nothing has changed when the user browses to the pad page. (Maybe you've already done this; I didn't check.) Of course, setting a non-zero max-age always has the down side of causing staleness problems immediately after a server upgrade.

* If you prefetch directly on the pad page, then I think you don't have to set a cache max-age; the browser's behavior is that it will fetch the resource to cache, and then serve the resource from cache *once* during the page load, even if it is not otherwise cacheable.

* Notice the *once* part. Currently, at least with my plugin set, `ace2_common.js` and `ace2_inner.js` actually get loaded _twice_ during pad load. If the resource has no caching max-age, then only the first of these two will be served by the prefetch, and the second will go back and hit the server again. These files are quite large so this is pretty bad. I'm not sure why it happens but it seems like a bug!

JohnMcLear:
Okay new todo..

 - [ ] Look to see if there is a way for plugins to specify file paths or for us to parse them out so we can set a prefetch for them
 - [ ] Use header instead of link IE ``<meta http-equiv="Link" content="</images/big.jpeg>; rel=prefetch">``
 - [ ] Check to see if core does load ace2_* files twice

Notes:
* Yea we do set ``maxAge``, it's a value set in ``settings.json``

kentonv:
> Use header instead of link IE `<meta http-equiv="Link" content="</images/big.jpeg>; rel=prefetch">`

To be clear, I don't think there's any advantage of `<meta>` over `<link>`. The advantage of the HTTP `Link` header is that if it is actually delivered as an HTTP header (not a meta-equiv) then HTTP/2 bridges which don't know how to parse the HTML can act on it.

JohnMcLear:
@kentonv see anything wrong w/ me sending this from express?
```
    res.set('Link', '</images/big.jpeg>; rel=prefetch');
```
It doesn't appear to be working..

kentonv:
Honestly I've never used the header before so your guess is as good as mine.

Incidentally I am looking at some Chrome network traces of my HTML-level hack right now and seeing that only some resources are being served from prefetch and not others. I'm not sure I totally understand the rules here.

JohnMcLear:
Well the header doesn't appear to be properly handled by Chrome :\

prefetch is new to me too so just trying to help out where I can..

JohnMcLear:
http://stackoverflow.com/questions/30959966/how-do-i-properly-set-link-prefetch-headers-in-express-4

JohnMcLear:
@kentonv either chrome is lying to the net tab or prefetch headers aren't supported in chromium based browsers.

kentonv:
Well, their docs claim it's supported:

https://www.chromium.org/spdy/link-headers-and-server-hint/link-rel-subresource

But yeah, if you're having trouble, I say just stick with the HTML approach.

JohnMcLear:
Oho scumbag Google strikes again...

http://www.stevesouders.com/blog/2014/04/30/unexpected-prerender-in-chrome/

"Suspecting that these background requests might not show up in Net panel, I fired up tcpdump and repeated the test – again only typing “www.s” and NOT hitting return. I uploaded the pcap file to CloudShark and saw 86 HTTP requests!"

Crazy huh?

JohnMcLear:
I think I figured out a way to automate plugin file prefetching :)

JohnMcLear:
@kentonv branch now includes support for prefetching plugin client files... :)

JohnMcLear:
@kentonv do me a favor and checkout this branch and test it on yours without your hacks.

kentonv:
@JohnMcLear FYI I'll probably get a chance to look at this on Friday.

JohnMcLear:
Cool, no rush from me :)

kentonv:
Hmm this _currently_ doesn't help Sandstorm, but that's Sandstorm's fault: We aren't passing through the `Link` header. (We operate on a whitelist since we need to examine the security implications of each header before allowing it.)

We also don't pass through `Cache-Control`, which would be a good idea to implement as well so that the Ace files aren't double-fetched.

But I think once we fix these, this change should help!

JohnMcLear:
Etherpad is already super optimized so this gain is negligible because 80% of hit points is not the index page.

PR #2699, opened 2015-06-21
Add author information to the context for padCreate and padUpdate hooks.

xavidotron:
This gives the hooks access to information that's otherwise awkward to get.

Ran backend tests and they pass.

### DIFF
--- a/doc/api/hooks_server-side.md
+++ b/doc/api/hooks_server-side.md
@@ -110,6 +110,7 @@ Called from: src/node/db/Pad.js
 Things in context:
 
 1. pad - the pad instance
+2. author - the id of the author who created the pad
 
 This hook gets called when a new pad was created.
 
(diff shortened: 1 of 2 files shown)


JohnMcLear:
1. Please include test coverage for the author information.
2. Please test on pads that have no authors (IE blank pads or pads that don't exist).
Should work fine :)

xavidotron:
Is there an example of a test that registers a hook?  I tried to figure out how to write a test, but wasn't sure the best way to register a hook.

I updated the documentation in doc/api/hooks_server-side.md.  I don't think there's anything to update with the API endpoints, since this is just changing a server-side hook, not an endpoint.  Is there anywhere else I need to document this?

JohnMcLear:
Can't think of an example of a plugin that uses a hook, should work okay anyways.

JohnMcLear:
tnx

PR #2707, opened 2015-06-24
Fixing @font-face src path to be able to use font icons on plugins when minify is on

lpagliari:
When plugins use font icons from `fontawesome-etherpad` and `minify` is on, the icons cannot be displayed because the URL seems to be incorrect: it looks for the font at `http://localhost:9001/font/fontawesome-etherpad.*`, but the correct path is `http://localhost:9001/static/font/fontawesome-etherpad.*`.

To reproduce this, use `ep_comments_page` and check for the delete icon on the comment.
Should be:
image

But is:
image

This is the same change of 56ce8e8, but for `fontawesome-etherpad`, instead of `opendyslexic`.

### DIFF
--- a/src/static/css/pad.css
+++ b/src/static/css/pad.css
@@ -189,7 +189,7 @@ a img {
   left:0px;
   right:250px;
 }
-  
+
 .toolbar ul.menu_right {
   right:0px;
 }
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Are there any 404's if minify is set to false?

lpagliari:
No. When `minify` is off plugin works fine and icons are displayed correctly.

JohnMcLear:
GJ :) Tnx

PR #2710, opened 2015-06-28
Add an indicator of the sync status to the toolbar.

xavidotron:
Interestingly, there's already javascript in src/static/js/pad_editbar.js to change the visibility of elements to reflect the sync status.  However, there's no elements with the relevant classes to hide and show.  This commit adds such elements to provide an indication of whether your latest changes have finished saving or not.

### DIFF
--- a/src/static/css/pad.css
+++ b/src/static/css/pad.css
@@ -1291,4 +1291,10 @@ input[type=checkbox] {
   display:none !important;
 }
 
-
+.toolbar ul li.syncstatus {
+  display: none;
+  line-height: 32px;
+  margin-left: 10px;
+  font-size: 14pt;
+  color: #666;
+}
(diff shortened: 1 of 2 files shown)


JohnMcLear:
This wont get merged in it's current form, here is why.

1) You are adding an element to the toolbar without using the correct method (see settiings.json, toolbar.js)

2) No i18n implementation

3) No accessibility implementation

4) No requirement for this to be added to the UI.  <-- There is no open issue where people have complained about knowing if it's syncing or not.

I suggest you make this a plugin if you really want it.  If folks use it we can consider implementing it in core.

Hopefully I didn't miss anything here...

xavidotron:
Ok, abandoning this in favor of a plugin (https://github.com/xavidotron/ep_sync_status).

JohnMcLear:
GG

Issue #2700, opened 2015-06-21
mismatch error setting raw text in setDocAText / assertion errors


LaKing:
While updating codepad and checking compatibility a bug came up with the recent etherpad version.
It causes a client-freeze with an endless loading ...

Here is some sample code:
```javascript
var padManager = require("ep_etherpad-lite/node/db/PadManager");
var padMessageHandler = require("ep_etherpad-lite/node/handler/PadMessageHandler");

exports.padCreate = function(hook, context) {
context.pad.setText("TEST");
padManager.getPad(context.pad.id, null, function(err, value) {
    if (err) throw err;
	padMessageHandler.updatePadClients(value, function() {});
    });
};
```

The error message leads to ace2_inner.js line ~770. 
```javascript
    if (rep.alltext != atext.text)
    {
      dmesg(htmlPrettyEscape(rep.alltext));
      dmesg(htmlPrettyEscape(atext.text));
      throw new Error("mismatch error setting raw text in setDocAText");
    }
```
With the test data above rep.alltext is TES and a newline character, and atext.text is TEST. As it seems, the last character always disappears and a newline character always appears.
The pad content is just TES
 
Is this a bug, Could we fix this?  Or is my method for setting pad text at pad creation incorrect?

JohnMcLear:
You said this bug exists in the latest version, did you do regression testing?

LaKing:
Okay, I tryed to reproduce it in a more blank, enviroment, and the error didnt show up, ..

However, I still get issues on my development enviroment. Messages like:
Error: Failed assertion: line assembler not finished:Z:j>27s|1-j|2g+289+2$

I have the feeling this might be a timing-related issue. ...

LaKing:
Well, the blank enviroment was using the latest zip release, and there is no such problem.
Something broke between the latest release and the current git version.

Steps to reproduce the problem.
Install etherpad from the latest git release.
Install the unreleased ep_test ... here is a bash script that will do it.

```bash
#!/bin/bash

## set this directory according to your enviroment
tp="/srv/etherpad-lite/node_modules/ep_test"

## create the plugin directory
mkdir -p $tp

## ep.json with a server side hook
cat > $tp/ep.json <<EOF
{
    "parts": [ 
        {
            "name": "ep_test",
            "hooks": {
                "padCreate": "ep_test/index:padCreate"
            }
        }
    ]
}
EOF

## on pad-creation we set TEST as text, and update the pad.
[...]

PR #2712, opened 2015-06-30
Restore newline-adding to setText() if passed string does not end in '\n'.

xavidotron:
Also adds a test for the ending-in-'\n' case and update tests for the other case.

This is a followup to my previous pull request #2649 intended to fix issue #2700.


JohnMcLear:
tnx

LaKing:
Thank you guys.

Seems its fixed, That's all folks!

PR #2714, opened 2015-07-01
Added CSS fallback fonts

charakterziffer:
I’m not sure if this is the correct and only place where to change this. I just gave the present fonts some fallbacks (when possible) but didn’t add, remove or merge fonts.


JohnMcLear:
What browsers / OS did you test this on?

charakterziffer:
I tested it only in Firefox at Ubuntu. If I download my fork, the file I changed wasn’t changed. Maybe I have to accept my own pull request anyhow – I though I had pressed this green button …? When I edited the file locally, it seems to get ignored when I run run.sh. Because of that I wasn’t sure if it’s the correct file – but could find any other then this (but for timeslider.js, also in the static folder) where the font names appear.

So: I couldn’t get it work yet. Do you know any other place where the CSS font-family value is set?

charakterziffer:
Yeah, not it works! I have saved it online the wrong way (and it still ignores local editings …). How can I send you the file now?

It’s enough to test in one browser, because all other interpret such basic CSS the same. I only can’t see fonts like Impact, Wingdings, etc. because they are not (freely) available for my system and therefore I haven’t installed them. But I can see in the CSS that they would be there.

Sorry that I don’t know much about collaborating with GitHub – my knowlegde in CSS and typography is much better ☺

JohnMcLear:
You might need to put etherpad into development mode to see changes...

open settings.json and set minify to false and maxAge to 0

Do you have the ability to test on other OS / browsers?

charakterziffer:
I’ve now tested my cloned online fork also in Opera and Chromium – works fine. I have a friend with Windows, but I’m sure he won’t be very amused if I install anything on his PC. (And I’m not sure if I’m able to run my fork there … or is it enough to take my local cloned folder without having to install anything more and starting the start.bat?)

JohnMcLear:
I can test windows for you if you want?

charakterziffer:
That would be really great! In the best case you won’t see any differences but for the font setting Monospace, Serif and Sans-Serif, which honours your browser settings now – and maybe with Lucida, because now it uses Lucida Bright as a fallback (the fonts “Lucida” and “Lucida Serif” are normally not installed, you couldn’t get it before).

JohnMcLear:
Symbol doesn't work on Windows.

JohnMcLear:
Windings doesn't appear to work either

JohnMcLear:
Serif doesn't appear to work either..

JohnMcLear:
Looks like you didn't change those anyway so they must have been broken before...

JohnMcLear:
Symbol works in IE :|

charakterziffer:
Symbol has never worked before, because it doesn’t contain latin characters. But if you have Webdings and Wingdings installed, they should give you the same result as IE … just googled: Firefox and Opera seem to [not accept these non-standard fonts](http://stackoverflow.com/questions/7862569/wingdings-font-family-does-not-seem-to-work-on-firefox-and-opera). So they also never work before.

For Serif you should get the serif font defined in your browser, e.g. in Firefox [Settings > Content > Fonts & Colors > Advanced > Serif](https://support.mozilla.org/en-US/kb/settings-fonts-languages-and-pop-ups?redirectlocale=en-US&redirectslug=Options+window+-+Content+panel) That does not have to be the same font as in your other browsers.

PR #2718, opened 2015-07-03
Fix toolbar and chat positioning on mobile devices

lpagliari:
On mobile devices (or at least on some iOS ones), if user scrolls the toolbar is not displayed anymore, and chat does not stick to the bottom of the page:

image

Tested on a Safari on Simulator + on Chrome on iPhone 5S with iOS 8.3.

With this PR:
image
image

### DIFF
--- a/src/static/css/pad.css
+++ b/src/static/css/pad.css
@@ -1109,7 +1109,7 @@ input[type=checkbox] {
       vertical-align:top !important;
     }
     #chaticon {
-      position:absolute;
+      position:fixed;
       right:48px;
     }
     .popup {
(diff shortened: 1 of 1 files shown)


JohnMcLear:
This doesn't break overflow on responsive UI elements IE when desktop browser is resized to very small?

JohnMcLear:
@Pita btw @lpagliari is really familiar w/ EP, she tends to commit good stuff.

lpagliari:
Just tested with Chrome:

(with chat maximized)
image
image

.. and FF:

(with chat maximized)
image
image

They look ok, but I'm not sure if you were mentioning other elements.

PR #2724, opened 2015-07-11
Add a aceSelectionChanged hook to allow plugins to react when the cursor moves

xavidotron:
No current hook seems to address this use case; you can fake key-related movement with aceKeyEvent, but there doesn't seem to be a way to catch mouse-related cursor changes.

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -2894,6 +2894,11 @@ function Ace2Inner(){
       rep.selFocusAtStart = newSelFocusAtStart;
       currentCallStack.repChanged = true;
 
+      hooks.callAll('aceSelectionChanged', {
+        rep: rep,
+        documentAttributeManager: documentAttributeManager,
+      });
+
       return true;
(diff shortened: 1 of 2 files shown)


src/static/js/ace2_inner.js
```diff
@@ -2894,6 +2894,11 @@ function Ace2Inner(){
       rep.selFocusAtStart = newSelFocusAtStart;
       currentCallStack.repChanged = true;
 
+      hooks.callAll('aceSelectionChanged', {
+        rep: rep,
+        documentAttributeManager: documentAttributeManager,
```
Pita: why does it pass the attribute manager?

xavidotron: The plugin I intend to use this hook uses the attribute manager to look at the attributes of the current cursor to show the current style in a UI element.  Other hooks pass the documentAttributeManager in this way, so I thought that would be fine.

xavidotron:
Any status on this pull request?  Anything I need to do to get it merged?

JohnMcLear:
Sorry about the delay @Pita is too busy for us nowadays ;)

PR #2739, opened 2015-07-27
Avoid space removal when pasting text from word processor.

otetard:
Since bf380eea504662ea41aa431e30d7e30ad6a36cd3, some spaces were removed when pasting text from a word processor (at least using Libre Office). To avoid double space creations and space removal, we only remove line break which are tight to a space character.


otetard:
This PR fixes #2781.

JohnMcLear:
I have a feeling this will insert double spaces in Firefox but let's see..

otetard:
I’ve use this patch on a production Etherpad-lite instance for a few months and it works good with Mozilla Firefox (at least, there are no double spaces inserted).

PR #2721, opened 2015-07-07
Complementing #2718 + Fixing empty toolbar when on read-only & menu_left has no button

lpagliari:
I've noticed my previous fix on #2718 was woking on my Etherpad because I have ep_page_view installed. When this plugin is not present, the editbar is still not fixed on top of the page when using iOS devices. So one of the commits of this PR is to fix that.

The other commit is to address an issue when pad is read-only and no button is on editbar, which I've noticed while testing the first commit:

image

This time I've tested on an Etherpad with no plugins, on these 4 scenarios: regular pad, read-only pad, inside of an iframe (using https://github.com/ether/etherpad-lite-jquery-plugin), and on timeslider. I've also tested on Safari@iPhone, Chrome@iPad, Chrome@desktop, and FF@desktop.

I hope this is all we need this time.

Issue #2224, opened 2014-08-13
Cursor jumping around when typing


bertpareyn:
From time to time when typing the cursor jumps up a couple of lines. If you continue to type the text isn't actually added where the cursor jumped to, but where the cursor was before it jumped.

A second, possibly related, issue is that on a mobile device, when the etherpad is being edited (or scrolled), the cursor consistently jumps all the way to the top.

JohnMcLear:
This is probably an upstream bug.  Apple made a huge mess of contentEditable in iOS8, we have some open upstream bugs so let's wait to see if they are resolved then things should be fixed.

sathomas:
FWIW, Just checked the iOS 9 beta and the problem is still present.

sathomas:
BTW, CKEditor has also flagged this as an [issue](http://dev.ckeditor.com/ticket/12457)

sathomas:
@JohnMcLear Have you filed a bug/issue with webkit or Apple Radar?

JohnMcLear:
Afaik there was an ios bug filed in December 2015 but I didn't file it personally and don't have a reference.

alexzautke:
I've filed a bug with Apple regarding the issue #2293, but contentEditable divs are still broken :/

sathomas:
Since Apple bug reports aren't public (and duplicates help increase the priority), I went ahead and created a simple (no dependencies, no JavaScript) [test case](https://jsfiddle.net/sathomas/rv6ktgL0/8/embedded/result/) and filed a bug report with Apple. It's number 21321546

ebarry:
I also experienced this on the etherpad-lite installation on publiclab.org. I've seen the occasional hiccup on our installation over the years, but what happened yesterday was notable. 
i started working on a new pad, and it was fine for about half an hour until i pasted in a link. 
Then the cursor started losing its position. As i typed new letters, the characters would insert several characters to the left of the cursor location. This effect worsened until i stopped typing completely. Then, without touching the keyboard at all, I watched the blinking cursor move itself backwards in a line of text, one character for every blink.
Example http://pad.publiclab.org/p/test

lpagliari:
@ebarry This behavior usually happens when a plugin adds a line attribute but doesn't register it using [`aceRegisterBlockElements `](http://etherpad.org/doc/v1.6.0/#index_aceregisterblockelements). Do you have any plugin installed?

[...]

PR #2743, opened 2015-08-04
Removing style that makes editor scroll to the top on iOS without any action from the user

lpagliari:
tl;dr
- avoid focus to be moved to the top every second when on iOS;
- but move back the issue that shows toolbar in the middle of the screen when editing on iOS:
image

Long story:

This PR reverts part of #2721, because that change made scroll on iOS devices very unstable. Using `position:scroll` as it was before this PR made the editor scroll to the top every second or so, even without any action from the user. I'm not sure if this issue is related to #2224, so maybe fixing one might fix both.

Anyway, I thought it was much worse to keep moving the focus to the top of the editor than having a toolbar "flying" on the screen when editing the pad.

Thoughts about these issues? Any suggestion?

### DIFF
--- a/src/static/css/pad.css
+++ b/src/static/css/pad.css
@@ -217,8 +217,7 @@ li[data-key=showusers] > a #online_count {
   bottom: 0px;
   z-index: 1;
 
-  /* Required to fix toolbar on top/bottom of the screen on iOS: */
-  overflow: scroll;
+  /* Required to be able to scroll on iOS: */
   -webkit-overflow-scrolling: touch;
 }
 #editorcontainer iframe {


lpagliari:
This is a draft of the code I was using to avoid the "flying" toolbar on iOS. It was basically changing toolbar menus to `position: absolute` instead of `fixed` and moving the menu_right to the far bottom of the editor when virtual keyboard is supposed to be displayed (in this example, when an input or textarea has focus). The code is faaaaaar away from being perfect, it was only a partial hack to fix the issue for the most relevant scenario on the Etherpad instance I am using.

If anyone has any other suggestion I would be glad to give a try.

Here's the code:
```
var fixFlyingToobarOnIOS = function() {
  if (browser.ios) {
    var placeMenuRightOnBottom = $(".toolbar ul.menu_right").css('bottom') !== "auto";

    getNewCommentContainer().find('input, textarea')
    .on("focus", function() {
      $(".toolbar ul.menu_left").css("position", "absolute");
      $(".toolbar ul.menu_right").css("position", "absolute");

      if (placeMenuRightOnBottom) $(".toolbar ul.menu_right").css("top", $(document).outerHeight());
    })
    .on("blur", function() {
      $(".toolbar ul.menu_left").css("position", "");
      $(".toolbar ul.menu_right").css("position", "").css("top", "");

      if (placeMenuRightOnBottom) $(".toolbar ul.menu_right").css("top", "");
    });

    // When user changes orientation, we need to re-position menu_right
    if (placeMenuRightOnBottom) {
      var resizeTimer;
      $(window).on("resize", function() {
        clearTimeout(resizeTimer);
        resizeTimer = setTimeout(function() {
          var needToUpdateTop = $(".toolbar ul.menu_right").css("top") !== "";
          if (needToUpdateTop) $(".toolbar ul.menu_right").css("top", $(document).outerHeight());
        }, 500); // give enough time for the UI to change and editor be re-displayed
      });
    }

  }
```

Gared:
I have no iOS device right now. So I can only merge your pull request if you think this is good to go

lpagliari:
@Gared I've checked on my iPhone and iOS simulator, the problem is gone after removing `overflow: scroll;`.

PR #2752, opened 2015-08-11
linux environment variables in safeRun.sh

2-B:
If etherpad-lite is cloned to a server and the bin/safeRun.sh is used, then the variables should be excluded from git. So I created a configuration file where the variables are stored. Additional the logfile and the nodejs binary can be specified on that configuration too.

The new file is something like that (an example exists):
<pre>
# Handling Errors
# 0 silent
# 1 email
ERROR_HANDLING=0

# Your email address which should recieve the error messages
EMAIL_ADDRESS="no-reply@example.com"

# Sets the minimun amount of time betweens the sending of error emails. 
# This ensures you not get spamed while a endless reboot loop 
# It's the time in seconds
TIME_BETWEEN_EMAILS=600 # 10 minutes

# Node.js binary file
NODEJS=/usr/bin/nodejs

# Logfile (if not specified, define it as a parameter at program start).
LOG="$HOME/etherpad-lite.log"
</pre>

### DIFF
--- a/bin/cleanRun.sh
+++ b/bin/cleanRun.sh
@@ -3,6 +3,9 @@
 #Move to the folder where ep-lite is installed
 cd `dirname $0`
 
+#Parse settings.json
+eval $(python parsejson.py)
+
 #Was this script started in the bin folder? if yes move out
 if [ -d "../bin" ]; then
   cd "../"
@@ -38,4 +41,4 @@ bin/installDeps.sh $* || exit 1
 echo "Started Etherpad..."
 
 SCRIPTPATH=`pwd -P`
-node $SCRIPTPATH/node_modules/ep_etherpad-lite/node/server.js $*
+$NODEJS $SCRIPTPATH/node_modules/ep_etherpad-lite/node/server.js $*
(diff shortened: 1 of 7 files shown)


Gared:
You're totally right. These variables have to be outside of the script file, but I think these variables should better be placed inside settings.json to have only one config file.
What do you think about? @2-B

2-B:
This is a good objection. So I fully rewrote the pull request and set all variables in the settings.json.

JohnMcLear:
I have no idea what this is needed for..  What is the purpose here?

2-B:
I put the variables to the settings.json file, so e.g. you can define the node binary. Not on every linux system the node binary has the same name.

JohnMcLear:
This seems like a bad idea.  Looking at it for 5 seconds I can think of one major issue..  if the user has node installed elsewhere and git pulls this would break their deployment...  You use an absolute location for the binary instead of using which and just parsing the path...

Heaven only knows how this would behave on Windows...

I dunno this just seems like 200 lines of cruft to me which is fixed by a one line symlink...

----- Reply message -----

JohnMcLear:
Bump..

JohnMcLear:
Closing due to lack of activity

PR #2762, opened 2015-08-24
Accepting Arrays on 'exportHtmlAdditionalTags'

lpagliari:
To handle attributes stored as `['key', 'value']` (and not only `['key', 'true']`). This is useful for plugins that use the former way of storing attributes on the AttributePool.

Tags provided as arrays will be exported to HTML as `<key:value>` (we could use any other string, I've used `":"` just to follow the standards I saw on some of the plugins, like ep_font_color and ep_comments).

### DIFF
--- a/doc/api/hooks_server-side.md
+++ b/doc/api/hooks_server-side.md
@@ -357,7 +357,7 @@ Things in context:
 
 1. Pad object
 
-This hook will allow a plug-in developer to include more properties and attributes to support during HTML Export. An Array should be returned. If a value in this array is a string, the exported HTML will contain tags like `<tag_name>` for the content where attributes are `['tag_name', 'true']`; if a value in this array is a pair `['tag_name', 'value']`, the exported HTML will contain tags like `<tag_name:value>` for the content where attributes are `['tag_name', 'value']`.
+This hook will allow a plug-in developer to include more properties and attributes to support during HTML Export. An Array should be returned. If a value in this array is a string, the exported HTML will contain tags like `<tag_name>` for the content where attributes are `['tag_name', 'true']`; if a value in this array is a pair `['tag_name', 'value']`, the exported HTML will contain tags like `<span data-tag_name:"value">` for the content where attributes are `['tag_name', 'value']`.
 
 Example:
 ```
--- a/doc/api/hooks_server-side.md
+++ b/doc/api/hooks_server-side.md
@@ -357,7 +357,7 @@ Things in context:
 
 1. Pad object
 
-This hook will allow a plug-in developer to include more properties and attributes to support during HTML Export. An Array should be returned. If a value in this array is a string, the exported HTML will contain tags like `<tag_name>` for the content where attributes are `['tag_name', 'true']`; if a value in this array is a pair `['tag_name', 'value']`, the exported HTML will contain tags like `<span data-tag_name:"value">` for the content where attributes are `['tag_name', 'value']`.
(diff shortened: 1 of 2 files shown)


lpagliari:
@JohnMcLear can you point someone that could take a look on this?

JohnMcLear:
Hopefully @pita or @Gared can, I'm still completely swamped -_- sorry!

lpagliari:
@Pita @Gared Could you also take a look on #2743, please? It's been there for more than a month, and I have other commits on other plugins that depend on these PRs to be approved... Or at least give some thoughts about a better way to fix it.

Gared:
Is HTML like <tag_name:value> really valid? Can you please post references?

lpagliari:
@Gared do you mean the generated HTML? Is your concern with the ":" between key an value, or is it something else?

I've made this test using ep_font_color (which stores attributes as ["color":"red"], instead of ["color_red", "true"], for example, so it needs Etherpad to be more flexible when building pad HTML). If you export [this pad](https://beta.etherpad.org/p/LFP_color), you won't see any colored text; this is also what I get when requesting the HTML via API on my environment:

```
curl "http://localhost/api/1.2.11/getHTML?apikey=e14b6af40c1489e9940a81eed94690725ee58061f695c6a188750e1730688b95&padID=LFP"

{"code":0,"message":"ok","data":{"html":"<!DOCTYPE HTML><html><body>line 1<br>line 2 -- this is red<br>line 3<br></body></html>"}}
```

After applying the changes to ExportHTML.js, this is what I get when requesting pad HTML via API:

```
curl "http://localhost/api/1.2.11/getHTML?apikey=e14b6af40c1489e9940a81eed94690725ee58061f695c6a188750e1730688b95&padID=LFP"
{"code":0,"message":"ok","data":{"html":"<!DOCTYPE HTML><html><body>line 1<br><color:red>line 2 -- this is red</color:red><br>line 3<br></body></html>"}}
```

Does that help to explain? It wasn't clear to me what your concerns were to ask for references, so if you need more clarification let me know and I'll post it here.

Gared:
@lpagliari Yeah, my concerns are about the ":" in the HTML, because as far as I know this is valid XML, but not valid HTML code. So my question was if you know any project, which is using a colon in a HTML tag or if you know any documentation about it.

What about storing the information in a data attribute like this:
```
<div data-info="{color: 'red'}">
```
or
```
<div data-color="red">
```
because this would be valid HTML code.
What do you think about this idea? Maybe I missed a point?!

lpagliari:
Make sense to me. Will try to change this today or tomorrow. Thanks!

lpagliari:
Will use `<span>` instead of `<div>` to be less intrusive.

lpagliari:
@Gared we've submitted another commit on this PR with the suggestions you gave. Is that what you were expecting?

Gared:
@lpagliari I didn't had much time to have a look on your new commit, but so far I tried to use the hook "exportHtmlAdditionalTags" in a test plugin on your branch, but couldn't export any data in the HTML export. So maybe I will have to take a deeper look in the ep_font_color plugin to get an idea how this should work.
I'm not really familiar with this hook so this needs some time for me to understand the whole thing

lpagliari:
@Gared please take a look on [this fork](https://github.com/storytouch/ep_font_color/) of ep_font_color, the original one still does not have the changes. We were waiting for this Etherpad PR to be approved before submitting the PR.

If you want to take a look on the export HTML part, [this is the file you need](https://github.com/storytouch/ep_font_color/blob/master/exportHTML.js#L7).

And if you need any other support, please let me know.

Gared:
@lpagliari After spending some time with the old hook and your changes I think this is too complex to use. The name of the hook 'exportHtmlAdditionalTags' is more a registration of additional tags, but with your changes it's more than tags that are registered, because we add new data-attributes to the html.
I'll accept your changes, but I would recommend to make this more clear in this hook or to make this more simpler in a newer hook.
Anyway: Thanks for your work @lpagliari

lpagliari:
@Gared Thanks for the suggestion. I'll try to think about creating a separated hook, so it would be more clear what each hook does -- even if in the end both `exportHtmlAdditionalTags` and this new hook have their contents merged.

PR #2767, opened 2015-08-28
Revert "Adding missing code to make toolbar be fixed on top/bottom of the screen for iOS devices."

elijh:
This reverts commit 8a0a1a65b1e9594e5e3bee2a73c7690c0589a008.

This commit needs to be reverted because it causes desktop browsers to show a duplicate, and useless, scroll bar for the editor window. Seems cool to add support for iOS, but let us not mess up desktop browsers in the process.

### DIFF
--- a/src/static/css/pad.css
+++ b/src/static/css/pad.css
@@ -216,10 +216,6 @@ li[data-key=showusers] > a #online_count {
   right: 0px;
   bottom: 0px;
   z-index: 1;
-
-  /* Required to fix toolbar on top/bottom of the screen on iOS: */
-  overflow: scroll;
-  -webkit-overflow-scrolling: touch;
 }
 #editorcontainer iframe {
   height: 100%;


lpagliari:
@elijh can you check if removing only `overflow: scroll;` fixes the duplicate scroll bar, please?

Gared:
@elijh I can't see duplicate scroll bars on my browsers. Which OS and browser do you use?

elijh:
yes, removing `overflow: scroll` does remove the duplicate scrollbars for me. I tested firefox/iceweasel and chromium on Ubuntu 15.04, 14.10, and Debian Jessie. 

to clarify, you must have enough text to make the main editor window scroll before you see the duplicate scrollbars. For example (from current beta.etherpad.org):

screenshot from 2015-09-06 00 57 13

micah:
Is this solved by the pull request merged in #2743?

elijh:
yes, that also eliminates the troublesome line.

PR #2773, opened 2015-09-03
Fix #2772.

### LINKED ISSUE #2772: AttributeManager.removeAttributeOnLine not working properly when text style is already applied
To reproduce:

* create pad with 2+ lines;
* indent all lines;
* apply BOLD to all lines;
* outdent all lines;

Result: first line is ok; other lines have an extra "*" on the beginning of the line.

Example: https://beta.etherpad.org/p/lfp_removeAttributeOnLine

It seems that the issue is on `AttributeManager.removeAttributeOnLine`: when line has BOLD on it (or any other attribute that is not applied to the whole line), it is not removing the "*" used as marker for the line. __This seems to be breaking all plugins that use `setAttributeOnLine()`, like ep_headings2.__

Possible solution: create a new hook for plugins to provide which lineAttributes they would use on `setAttributeOnLine()`, so when we check if line still has line attributes, we only check for those that can be used like that.

lpagliari:
Create new hook `aceRegisterLineAttributes` to have a list of all line attributes used by Etherpad and the plugins.

This PR fixes #2772 but might impact other plugins that call `AttributeManager.removeAttributeOnLine()`. This will only happen if:
1. line has two line attributes (one being the attribute defined by plugin) __and__
2. user removes line attribute not defined by the plugin (remove list or outdent, for example).

To fix the plugin: simply define the hook `aceRegisterLineAttributes`.

### DIFF
--- a/doc/api/hooks_client-side.md
+++ b/doc/api/hooks_client-side.md
@@ -139,7 +139,14 @@ Called from: src/static/js/ace2_inner.js
 
 Things in context: None
 
-The return value of this hook will add elements into the "lineMarkerAttribute" category, making the aceDomLineProcessLineAttributes hook (documented below) call for those elements.
+The return value of this hook will add elements into the "lineMarkerAttribute" category, making the aceDomLineProcessLineAttributes hook (documented below) call for those elements. This hook is usually used with aceRegisterLineAttributes hook (documented below).
+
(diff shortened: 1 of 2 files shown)


lpagliari:
We found a less intrusive way to fix #2772. We will close this PR and create one with the other solution.

PR #2791, opened 2015-09-29
Fix off-by-one: The chatHead is in fact a valid message index.

### LINKED ISSUE #2751: Off-by-one in getChatHistory's error checks
*Edited to add revision information.*

It is impossible to request only the last chat message via the HTTP API, because getChatHistory does not accept the current chatHead as a start value. This has been tested with revision 02c022aab14f9087e22cebd4123336935cc089d8 (HEAD of develop at time of writing).

For example, having the following chat history: 
```json
[{"text":"zero","userId":"a.KoWAfVv5a5GZCDpI","time":1439230819559,"userName":"fread"},
 {"text":"one","userId":"a.KoWAfVv5a5GZCDpI","time":1439230820359,"userName":"fread"},
 {"text":"two","userId":"a.KoWAfVv5a5GZCDpI","time":1439230821236,"userName":"fread"},
 {"text":"three","userId":"a.KoWAfVv5a5GZCDpI","time":1439230822246,"userName":"fread"}]
```
getChatHead returns (as expected):
```json
{"code":0,"message":"ok","data":{"chatHead":3}}
```
I can request message two with `getChatHistory?start=2&end=2` and I can request messages two and three with `getChatHistory?start=2&end=3`, but there is no way to request only message three, since `getChatHistory?start=3&end=3` gives an error ("start is higher or equal to the current chatHead").

The following patch fixes this instance of the bug, but not knowing the etherpad code (or indeed Javascript) I am reluctant to make it into a pull request.

```javascript
--- a/src/node/db/API.js
+++ b/src/node/db/API.js
@@ -473,9 +473,9 @@ exports.getChatHistory = function(padID, start, end, callback)
     end = pad.chatHead;
     }
     
-    if(start >= chatHead && chatHead > 0)
+    if(start > chatHead)
     {
-      callback(new customError("start is higher or equal to the current chatHead","apierror"));
+      callback(new customError("start is higher than the current chatHead","apierror"));
       return;
     }
     if(end > chatHead)
```

Thanks in advance for your help.

fread:
Since @Gared seems to agree, I'm proposing this to fix #2751.

### DIFF
--- a/src/node/db/API.js
+++ b/src/node/db/API.js
@@ -473,9 +473,9 @@ exports.getChatHistory = function(padID, start, end, callback)
     end = pad.chatHead;
     }
     
-    if(start >= chatHead && chatHead > 0)
+    if(start > chatHead)
     {
-      callback(new customError("start is higher or equal to the current chatHead","apierror"));
+      callback(new customError("start is higher than the current chatHead","apierror"));
       return;
     }
     if(end > chatHead)


Gared:
@fread Thanks

PR #2798, opened 2015-10-07
Get git commit hash even if the repo only points to a bare repo.

ypid:
* Used by https://github.com/debops/ansible-etherpad


JohnMcLear:
Thanks @ypid ! :)

PR #2810, opened 2015-10-19
Add an appendText API

luser:
I wanted to write a script that could stream text into an Etherpad, and the getText/setText APIs don't work well for this purpose, because setText clobbers attributes, so if anyone else is trying to edit the pad at the same time it results in errors. I added an appendText method which does exactly what I need.

### DIFF
--- a/src/node/db/API.js
+++ b/src/node/db/API.js
@@ -308,6 +308,38 @@ exports.setText = function(padID, text, callback)
 }
 
 /**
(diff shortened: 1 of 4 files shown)


JohnMcLear:
Nice job!

Gared:
@luser Could you please add the documentation for this api in this file? https://github.com/ether/etherpad-lite/blob/develop/doc/api/http_api.md

PR #2820, opened 2015-10-29
Fix logic to check if any hook handled ACE key event. Fix #2818

### LINKED ISSUE #2818: Logic to find out if any plugin handled key event seems incorrect
I'm puzzled with a logic I found on `ace2_inner.js` (https://github.com/ether/etherpad-lite/blob/develop/src/static/js/ace2_inner.js#L3718):

```
specialHandled = (specialHandledInHook&&specialHandledInHook.length>0)?specialHandledInHook[0]:specialHandled
```

Documentation says " The return value should be true if you have handled the event.", so why do we check only if first plugin had handled the key event (`specialHandledInHook[0]`) instead of looking for all of the returned values? Was there any reason for this?

I'm currently having issues on my Etherpad because only the second plugin handles the key event, but even so Etherpad-core keeps processing the event, which breaks what I was expecting to happen.

If the current behavior is incorrect, just let me know and I'll submit a PR for it.

lpagliari:
Instead of checking if only the first hook had returned true, we check
all hooks.

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -3715,7 +3715,12 @@ function Ace2Inner(){
           documentAttributeManager: documentAttributeManager,
           evt:evt
         });
-        specialHandled = (specialHandledInHook&&specialHandledInHook.length>0)?specialHandledInHook[0]:specialHandled;
+
+        // if any hook returned true, set specialHandled with true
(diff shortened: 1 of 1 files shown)


JohnMcLear:
much nicer, thanks!

PR #2822, opened 2015-10-29
Allow plugins handle "esc" key event

joassouza:
PR #984 fixed an issue with "esc" on older versions of Firefox (15-19), but avoided the key event to be sent to plugins on aceKeyEvent.
This issue was fixed in Firefox versions 20 and above. 
With this commit we keep the fix for the older Firefox versions, besides to allow to handle "ESC" by plugins.

More details about this issue :
 http://blog.ffextensionguru.com/2013/01/11/changes-to-esc-key-behavior/

This was tested in Firefox 15, 20, 41 for Linux.


JohnMcLear:
tnx

PR #2834, opened 2015-11-16
Added variables for APIKEY, SESSIONKEY and plugin initialization file into settings.json

yurivict:
This allows to run etherpad-lite without touching files in its directory, which is good when it is installed by a package in the read-only location.


JohnMcLear:
What if the APIKey file is already used and is a fixed value?
Is this tested across multiple OS'?

yurivict:
If the new settings values aren't set behavior doesn't change, so the old files will work.

Not tested across multiple OS'.

JohnMcLear:
Closing in favor of the ability to provide environment variables into settings.json

PR #2839, opened 2015-11-22
Improve DB migration performance.

Blake-jne:
Achieve 10x speedup in migrating a 12mb dirtyDB file by disabling
ueberDB caching in this special case.  Add some progress messages
to the migration script and rename it, as nothing in it is tied
to mysql.


JohnMcLear:
n1 tnxd

Issue #2846, opened 2015-11-28
Using proper ellipsis in l10n files


ujdhesa:
Hello,

en.json uses `...` (three dots) instead of `…` (ellipsis). There’s even a (misspelled) string having only 2 dots instead of `...`.

PR #2847, opened 2015-11-28
Make use of proper ellipsis

ujdhesa:
Here you are a fix for issue https://github.com/ether/etherpad-lite/issues/2846


JohnMcLear:
We only accept localization changes through translatewiki

Issue #2858, opened 2015-12-18
Libre Office (soffice) is never checked for pdf export features


cmbirk:
In the `settings.json` file it says `soffice` can be used as a replacement for `abiword`, but in `src/static/js/pad_impexp.js` it only checks `abiword` and calls `$('exportpdfa').remove()` but never checks soffice.

I'm working on a pull request for this.

PR #2859, opened 2015-12-18
Check for SOffice before removing export links

cmbirk:
This is an initial pull request for #2858.

As this is my first PR for etherpad-lite core, I would appreciate any feedback.


JohnMcLear:
Wouldn't this need an additional into settings.json.template?

cmbirk:
The `soffice` setting is already in there

Issue #2516, opened 2015-02-11
Consider MySQL UTF8 issues


JohnMcLear:
MySQL might be better with a character-set of utf8mb4 and a collation of utf8mb4_bin.

I'm not an expert in this though so we need some validation..

The reason for this encoding type is that utf8_bin lies, it can't store all utf8 encoded characters that are available in the utf8 charset.

Reference:
- http://mzsanford.com/blog/mysql-and-unicode/
- http://dba.stackexchange.com/questions/8239/how-to-easily-convert-utf8-tables-to-utf8mb4-in-mysql-5-5

We need opinions on this from people who have ample MySQL and character encoding experience.

webzwo0i:
No expert but debug session revealed this to be the root cause for some crashes and inconsistencies in the database. With utf8 (as limited by https://tools.ietf.org/html/rfc3629) up to 4 of 1 byte long units are allowed. This synchronizes the expressible range with UTF-16, so Unicode code points up to U+10FFFF can be encoded. utf8mb4 is the MySQL name for this.

I migrated one installation:
The installation was at latin1(!!)... Can try this on beta if desired because our installation is really small and the used languages/features are not as diverse as it may be the case in other installations.

what I did:
stop etherpad
stop the db
cp -ab /var/lib/mysql mysql.bak
start the db
mysqlcheck -c -u user -p database-name #should return ok
mysql> alter database `database-name` CHARACTER SET = utf8mb4 COLLATE = utf8mb4_bin;
mysql> alter table `store` convert to character set utf8mb4 collate utf8mb4_bin;
mysql> alter table `store` change `key` `key` varchar(100) character set utf8mb4 collate utf8mb4_bin;
mysql> alter table `store` change `value` `value` longtext character set utf8mb4 collate utf8mb4_bin;

Somewhere in the internets I found that this method would not work and you have to manually convert to varbinary/longblob and than back to varchar/longtext.
For me this was not needed (hopefully ;)) and besides I could not find a good explanation why these additional steps are needed.

SHOW VARIABLES WHERE Variable_name LIKE 'character\_set\_%' OR Variable_name LIKE 'collation%';
stop the db
open /etc/mysql/my.cnf
in section [mysql]:
default-character-set=utf8mb4

in section [mysqld]:
init_connect ='SET NAMES utf8mb4'
init_connect ='SET collation_connection = utf8mb4_bin'
character-set-server=utf8mb4
collation-server=utf8mb4_bin

[...]

Issue #2522, opened 2015-02-18
JSON-PROBLEM shuts server down


Gared:
A few pads are invalid on my instance and every time these pads are called I get a JSON-PROBLEM-Exception:
```
[2015-02-18 20:13:48.167] [ERROR] console - JSON-PROBLEM (pad:49Xv36g3re93MSV):{"atext":{"text":"some text\n\n
[2015-02-18 20:13:48.173] [ERROR] console - gracefulShutdown...
[2015-02-18 20:13:48.173] [ERROR] console - Async Stacktrace:
    at /opt/eplite/etherpad-lite-dev/src/node/handler/PadMessageHandler.js:1306:5
    at /opt/eplite/etherpad-lite-dev/src/node/handler/PadMessageHandler.js:1042:16
    at /opt/eplite/etherpad-lite-dev/src/node/db/PadManager.js:171:10
    at Object.callback (/opt/eplite/etherpad-lite-dev/src/node/db/Pad.js:385:8)

SyntaxError: Unexpected end of input
    at Object.parse (native)
    at /opt/eplite/etherpad-lite-dev/src/node_modules/ueberDB/CacheAndBufferLayer.js:164:24
    at Query._callback (/opt/eplite/etherpad-lite-dev/src/node_modules/ueberDB/mysql_db.js:116:5)
    at Query.Sequence.end (/opt/eplite/etherpad-lite-dev/src/node_modules/ueberDB/node_modules/mysql/lib/protocol/sequences/Sequence.js:96:24)
    at Query._handleFinalResultPacket (/opt/eplite/etherpad-lite-dev/src/node_modules/ueberDB/node_modules/mysql/lib/protocol/sequences/Query.js:143:8)
    at Query.EofPacket (/opt/eplite/etherpad-lite-dev/src/node_modules/ueberDB/node_modules/mysql/lib/protocol/sequences/Query.js:127:8)
    at Protocol._parsePacket (/opt/eplite/etherpad-lite-dev/src/node_modules/ueberDB/node_modules/mysql/lib/protocol/Protocol.js:271:23)
    at Parser.write (/opt/eplite/etherpad-lite-dev/src/node_modules/ueberDB/node_modules/mysql/lib/protocol/Parser.js:77:12)
    at Protocol.write (/opt/eplite/etherpad-lite-dev/src/node_modules/ueberDB/node_modules/mysql/lib/protocol/Protocol.js:39:16)
[...]

PR #2862, opened 2015-12-21
set charset for mysql connection in settings.json

Gared:
See https://github.com/ether/etherpad-lite/issues/2516#issuecomment-164768811
Maybe we should output a warning on console if this parameter isn't set?!

This pull request is related to my change for ueberDB:
https://github.com/Pita/ueberDB/pull/87

### DIFF
--- a/settings.json.template
+++ b/settings.json.template
@@ -47,7 +47,8 @@
                     "user"    : "root",
                     "host"    : "localhost",
                     "password": "",
-                    "database": "store"
+                    "database": "store",
+                    "charset" : "utf8mb4"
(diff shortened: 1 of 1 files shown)


rohieb:
FWIW: I ran into issue #2522 on v1.5.7 by typing U+1F686 (🚆) into one of my pads. After ALTERing the MySQL DB charsets, the server didn't crash anymore, but still hung at "Loading". After I applied this patch, everything was back to normal, so: Works for me.

Gared:
@rohieb Thanks for your comment

PR #2863, opened 2015-12-22
Limit padname input field to 50 characters

gutobenn:
It seems good to limit the padname length in the input field. As I could check, the max length accepted is 50 characters.

If this config depends on other server/installation configs, please let me know.


Gared:
This is because of ueberDB and the maximum key length of some databases, but you're right it is limited to 50 characters and can not vary.

Gared:
Thx

PR #2866, opened 2015-12-31
Fix handleClientMessage_USER_* payloads not containing user info

tiblu:
Client hook handleClientMessage_USER_* did not contain user info.


tiblu:
Closing, managed to include same fix in https://github.com/ether/etherpad-lite/pull/2871

PR #2871, opened 2016-01-05
pad_userlist.js: BUGFIX: users() returning duplicates on several calls. FEATURE: usersOnline() returns only online users

tiblu:
pad_userlist.js:

* BUGFIX: users() returning duplicates on several calls. 
* BUGFIX: ``handleClientMessage_USER`` to contain user info.
* FEATURE: usersOnline() returns only online users. Useful when one wants to implement it's own online users list.


JohnMcLear:
Needs docs afaik

PR #2889, opened 2016-01-19
Use exec to switch to node process

reedacartwright:
At the end of run, `exec` should be used to switch to the node process.  That way node will take over the pid of `sh run.sh`, making it easier to monitor and daemonize the server. 

```
    exec [command [arg ...]]
             Unless command is omitted, the shell process is replaced with the
             specified program (which must be a real program, not a shell
             built-in command or function).  Any redirections on the exec com-
             mand are marked as permanent, so that they are not undone when
             the exec command finishes.
```

### DIFF
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -35,5 +35,5 @@ bin/installDeps.sh $* || exit 1
 echo "Started Etherpad..."
 
 SCRIPTPATH=`pwd -P`
-node $SCRIPTPATH/node_modules/ep_etherpad-lite/node/server.js $*
+exec node $SCRIPTPATH/node_modules/ep_etherpad-lite/node/server.js $*
 


Gared:
@JohnMcLear What do you think about this change?

JohnMcLear:
I feel like it will break the init etc. scripts

reedacartwright:
Can you explain this concern in more detail?   I've been running with this
patch for about 2 months and haven't seen any breakage.

JohnMcLear:
afaik the ubuntu init scripts look for a running node process, I dunno if this will break that or not..

reedacartwright:
Exec replaces the shell process with the node process.  Therefore,  the
node process will exist.  Using exec is the recommended way to start a
program at the end of a wrapper script.

JohnMcLear:
Then it should be good to go, I don't know about exec to confirm this..  We could merge into develop and see if anything explodes though?

Gared:
@reedacartwright Thx

Issue #2885, opened 2016-01-15
BUG: Timeslider not translated when EP is opened with lang parameter


tiblu:
**Reproduce:**

* Go to https://beta.etherpad.org/p/asd?lang=de
* Go to Timeslider page, by clicking the icon

**Result:**

* Timeslader page is in English

**Expected:**

* Timeslider page to be in German (de)

**Technical info:**

* Etherpad 1.5.7 (02c022a)

JohnMcLear:
Confirmed

JohnMcLear:
Confirmed

bhldev:
@tiblu please get latest develop and confirm, thanks

tiblu:
Thanks, fixed!

PR #2890, opened 2016-01-20
Timeslider Not Translated Issue 2885

bhldev:
Fixes Issue #2885


Gared:
@bhldev Looks good to me. Thanks

Issue #2672, opened 2015-05-19
Put the ability to chat etc. in Read Only under a setting flag


JohnMcLear:
Ie `readOnlyCanCommentAndChat`

rugk:
Currently the behaviour is strange: In all "usual" situations the chat window is hidden (and not showable in the UI) when accessing the read-only mode.

However if you previously pinned the chat window (in read-write mode), so that it is always visible it is also visible in ro mode. But when you close (aka "minimize") it there it gets completely hidden (again).

PR #2894, opened 2016-01-25
Feature: Read Only Chat Issue #2672

bhldev:
Feature Read Only Chat Issue #2672 and default settings in settings.json on the client side for future use


Gared:
I don't get the idea behind your code.
When is a user able to send a chat message?
As far as I can see the check is only on client side and not on server side, so a user can easily circumvent your check...

bhldev:
Added server side check, user can send a chat message by default
When readOnlyCanCommentAndChat is set to false in settings.json, chat is disabled in etherpad according to feature request #2672

bhldev:
Hello,
Can this be approved and merged or closed?
Thanks,
~ B

JohnMcLear:
Sorry been busy, is this still a feature you want?  What issue does this pull request solve?

Issue #2895, opened 2016-01-26
Problem with translation into German


dscso:
Firefox 44.0; Chomium similar problem
Server: Debian 8.3; Kernal 3.16.0-4-amd64
NodeJS Version: v0.10.29

"Neues Dokument" ≙ "New Pad" 
I'm not sure but @german_people: change "Dokument" to Pad
my pull request: https://github.com/ether/etherpad-lite/pull/2896
![image](https://cloud.githubusercontent.com/assets/6866008/12593458/7fb8e54c-c471-11e5-95f9-97ca886d033b.png)

akuckartz:
"Dokument" is word known by every German speaker. "Pad" is neither a German word nor one offen used by Germans.

dscso:
Yes you are right. But even the french translation of etherpad-lite uses "Nouveau pad" or the spanish translation: "Nuevo pad". So I think German is the only language which uses "Dokument" (german document) instead of pad.

thmsjkn:
To me a document (Dokument) is more static than a Pad. I would go with the term "Pad".

dscso:
Me too. Most popular languages are using "pad".

Gared:
I would also prefer "pad" for the german translation. Please use translatewiki.net to change these translations @dscso

Gared:
Fixed

PR #2896, opened 2016-01-26
alternative translation into german.

dscso:
my issue: https://github.com/ether/etherpad-lite/issues/2895
I know I should use translatewiki.net sorry for that


Gared:
Correct, please use translatewiki.net

PR #2909, opened 2016-02-09
Added URI encoding for New Pads

mikesterlingw:


JohnMcLear:
@gared afaik you wrote the padId validation stuff, this might be of interest to you :)

Gared:
This does not solve the problem if you want to create a pad with the character "$" for example.
I think we have to use a validation like here:
https://github.com/Gared/etherpad-lite/blob/83094e0dfd3361625ec14442d58c1e8ba935810b/src/node/db/API.js#L692

mikesterlingw:
@Gared What is the problem with creating a pad with a "$" character?

mikesterlingw:
Closing as issue is actually resolved and was a result of an error on my end

Issue #2836, opened 2015-11-17
TypeError: Cannot read property 'substring' of undefined - RESTART!


micah:
We have a very active set of pads, with several hundred users on them. Unfortunately, something is causing etherpad to restart on a regular basis, this is the log of when it does:

```
[2015-11-17 07:27:54.247] [WARN] client - Uncaught TypeError: Cannot read property 'data' of undefined -- { errorId: 'ZdIdZ8nXVLQTE9tVbPgb',
  msg: 'Uncaught TypeError: Cannot read property \'data\' of undefined',
  url: '[REDACTED]',
  linenumber: 3902,
  userAgent: 'Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.86 Safari/537.36' }
[2015-11-17 07:28:13.168] [WARN] client - TypeError: undefined is not an object (evaluating 'rowsPresent[index].data') -- { errorId: '7FGxdP5milNbRykoStKV',
  msg: 'TypeError: undefined is not an object (evaluating \'rowsPresent[index].data\')',
  url: '[REDACTED]',
  linenumber: 3902,
  userAgent: 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_1) AppleWebKit/601.2.7 (KHTML, like Gecko) Version/9.0.1 Safari/601.2.7' }
[2015-11-17 07:28:45.256] [WARN] client - SyntaxError: expected expression, got ')' -- { errorId: 'X45Sz4e8TZIuMmRRkoCA',
  msg: 'SyntaxError: expected expression, got \')\'',
  url: '[REDACTED]',
  linenumber: 1,
  userAgent: 'Mozilla/5.0 (Windows NT 6.1; WOW64; rv:38.0) Gecko/20100101 Firefox/38.0' }
[2015-11-17 07:28:46.691] [WARN] client - Uncaught Error: Failed assertion: mismatched apply: 6831 / 6841 -- { errorId: 'I07sb37FnB4hqN8MEU8S',
  msg: 'Uncaught Error: Failed assertion: mismatched apply: 6831 / 6841',
  url: '[REDACTED]',
  linenumber: 492,
  userAgent: 'Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.86 Safari/537.36' }
[...]

Issue #2802, opened 2015-10-12
Undoing causes disconnect


fcassin:
Hello Etherpad Team,

We've found a reproducible bug with the clear authorship action that causes a disconnect.
To reproduce this bug :
- Log in to a pad with user A, activate author colors and type something
- Log in to the same pad with user B, activate author colors and type something
- As user B, click on _clear authorship colors_
- As user B, click on _undo_

User B should be disconnected from the pad.

Here is a relevant stack trace : 

```
[2015-10-12 12:37:47.276] [WARN] console - Error: Can't apply USER_CHANGES, because Trying to submit changes as another author in changeset Z:151>0=2k*0|2=2*0=7*1|2=2*1=4$
    at /home/webapps/etherpad/src/node/handler/PadMessageHandler.js:683:25
    at /home/webapps/etherpad/src/node_modules/async/lib/async.js:610:21
    at /home/webapps/etherpad/src/node_modules/async/lib/async.js:249:17
    at iterate (/home/webapps/etherpad/src/node_modules/async/lib/async.js:149:13)
    at /home/webapps/etherpad/src/node_modules/async/lib/async.js:160:25
    at /home/webapps/etherpad/src/node_modules/async/lib/async.js:251:21
    at /home/webapps/etherpad/src/node_modules/async/lib/async.js:615:34
    at /home/webapps/etherpad/src/node/handler/PadMessageHandler.js:633:9
    at Object.exports.getPad (/home/webapps/etherpad/src/node/db/PadManager.js:161:5)
    at /home/webapps/etherpad/src/node/handler/PadMessageHandler.js:629:18
```

The issue seems to be in _PadMessageHandler.js_. The clear authorship message does not specify any author, thus not triggering the test on line 699.
However, when hitting the _Undo_ button, two authors are specified, triggerring the error message.

Regards,
François

webzwo0i:
Thank you for the detailed bug report.

clear authorship (undo that, respectively) is the only changeset that can has zero (deletion) or multiple (undo) authors at once. However we cannot simply disable the check, because this would introduce impersonation scenarios where an author can send changes as another author.  

Maybe we should do this server-side instead of sending changesets.

This also needs a test, but at the moment the test infrastructure cannot easily express scenarios that involve multiple clients.

webzwo0i:
also we might want that anybody can reimport the authorship colours and not only the client that deleted them

muxator:
[...]

PR #2911, opened 2016-02-15
Fix text lines mutator

webzwo0i:
This should address https://github.com/ether/etherpad-lite/issues/2836 but please do not merge this yet!
As you can see there are some "TODOs" left.

The textLinesMutator gets lines of text and a changeset and applies the latter to the former. It should work for attribution and normal text lines. Sometimes when there are no lines left but the changeset is not finished it will crash.

I am not sure I did this right, but I will try to solve the remaining TODOs. I want to ask if anyone has a good idea how to test this stuff. It is used in the timeslider and on clients before submitting changesets. What I need is working test data:
- take a running epl instance and dozens of html (from different revisions) and etherpad exports
- the html exports represent text lines
- randomly take two html exports and calculate composed changesets in both directions using the revisions from etherpad exports
- apply the composed changesets to the html export's lines and test for equality with the other html export

Does this sound reasonable to you? Any other ideas how to be sure this does not break anything?

### DIFF
--- a/src/static/js/Changeset.js
+++ b/src/static/js/Changeset.js
@@ -711,9 +711,14 @@ exports.textLinesMutator = (lines) => {
           curSplice[1] += L - 1;
           const sline = curSplice.length - 1;
           removed = curSplice[sline].substring(curCol) + removed;
-          curSplice[sline] = curSplice[sline].substring(0, curCol) +
-              linesGet(curSplice[0] + curSplice[1]);
-          curSplice[1] += 1;
+          const line = linesGet(curSplice[0] + curSplice[1]);
+          // if no line follows the splice
+          if (!line) {
+            curSplice[sline] = curSplice[sline].substring(0, curCol);
+          } else {
+            curSplice[sline] = curSplice[sline].substring(0, curCol) + line;
+            curSplice[1] += 1;
+          }
         }
       } else {
         removed = nextKLinesText(L);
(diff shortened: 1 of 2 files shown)


webzwo0i:
note to self: this breaks some pads with TypeError: rep.lines.atIndex(...) is undefined

JohnMcLear:
Be happy to see this land :)  Sounds like it's already fixing things?  Can I help by giving you access to beta.etherpad.org ?

muxator:
Hi, this is an old PR, but it seems important. Any chance to resume the work?

JohnMcLear:
@muxator I tried to bump @webzwo0i - this defo looks worth putting some time into..

JohnMcLear:
I spoke to @webzwo0i he's going to work on this when he can :) so we're waiting on him to review/merge.

webzwo0i:
Will submit the documentation from here in a separate PR. The tests I added to easysync_tests in this PR are probably nonsense (the last newline is special). textLinesMutator will crash on some crafted changesets but the way it is used in the timeslider seems to be save and every exception is handled well

webzwo0i:
Reopening so that I'm reminded I need to make a new PR with docs.
Also I see a related error that renders the timeslider useless. Steps to replicate:
1. create a pad
2. enable allowAnyoneToImport or write something
3. import an empty html file (or anything else that calls `setText("\n")`)
4. goto timeslider, disable "follow contents" and go backwards, it won't work between revision 2 and 1 (revision 1 in my case is "write something" from step 2. so it triees to apply the inverse of the changeset that is created with `setText("\n")`)
`timeslider.js?callback=require.define&v=33c0661d:3378 curSplice[sline] not populated, actual curSplice contents is  (3) [0, 2, undefined] . Possibly related to https://github.com/ether/etherpad-lite/issues/2802`

Coverage test needs to go into https://github.com/ether/etherpad-lite/issues/4190

rhansen:
I rebased onto current `develop` and split the documentation changes into a separate PR (#5199).

webzwo0i:
Thanks @rhansen for the cleanup work!!!
Before this can be merged, I need to finish up https://github.com/ether/etherpad-lite/issues/4190

webzwo0i:
Closed in favor of https://github.com/ether/etherpad-lite/pull/5253

PR #2912, opened 2016-02-19
made url relative

cxcv:
I have etherpad installed in a subdirectory

### DIFF
--- a/src/static/js/pad.js
+++ b/src/static/js/pad.js
@@ -872,7 +872,7 @@ var pad = {
       $.ajax(
       {
         type: 'post',
-        url: '/ep/pad/connection-diagnostic-info',
+        url: 'ep/pad/connection-diagnostic-info',
         data: {
           diagnosticInfo: JSON.stringify(pad.diagnosticInfo)
         },


Gared:
Can anyone confirm this?

michael-dev:
I have etherpad lite mod_proxied in an subdirectory as well and see in access.log requests for /ep/pad/connection-diagnostic-info, which cannot be proxied due to missing subdirectory prefix.

girgen:
Yes, this change is correct. If Etherpad is proxied, this fixed is absoutely requried for EP to work correctly. Please merge.

muxator:
I am a bit uncomfortable because it seems that there is already a mechanism in place to build correct (albeit absolute) urls when ep is proxied (see for example https://github.com/ether/etherpad-lite/blob/develop/src/static/js/pad.js#L194-L195).

This change makes the url relative (which is fine), but shouldn't we stick to a single strategy?

girgen:
Well, something needs to be done. We saw access to apache for `/ep/pad/connection-diagnostic-info` before changing that. Stick to single strategy sounds great, but it need need be applied here in that case. Seems it is not happening just now?

muxator:
The fix is beneficial. Merged, thanks.

@cxcv: next time, it would be appreciated if the commit message (and the issue itself) contained a short one-line summary of the problem, with a description in the body of the message, for example:

```
diagnostic-info: made url relative

After a failed attempt to reconnect to the server, the
client sends a diagnostic message to an absolute
endpoint.
This was prone to breakage when etherpad was reverse
proxied in a subdirectory.
```

webzwo0i:
@benschweizer @michael-dev @girgen 
Can you please elaborate on your configuration? Are you using the example config from (wiki)[https://github.com/ether/etherpad-lite/wiki/How-to-put-Etherpad-Lite-behind-a-reverse-Proxy]? 
How do you handle requests to /locales.json /pluginfw/plugin-definitions.json /admin or /jserror?

The change here breaks functionality for standard installations

benschweizer:
I cannot reproduce it anymore as the PR iss 2+ years old and the setup changed since then

girgen:
We use apache with just
```
ProxyPassMatch /etherpad/(.*) http://127.0.0.1:9001/$1
```

webzwo0i:
Thanks, appreciated!

Issue #2897, opened 2016-01-28
API client author comments


c-martinez:
Is it possible to use the http-API to insert comments, as if they were written by a user ?

I am writing a _bot_, but I would like him to look like any other author on the pad: have his name on the list of authors, have the text he writes coloured, etc. At the moment I'm using: `getText` / `setText`, but this feels really clumsy. Is there a better way? Would it be desirable to have such a feature in the api? How would one go about implementing this feature?

lpagliari:
@c-martinez take a look on [ep_comments](https://github.com/JohnMcLear/ep_comments#creating-comment-via-api)

c-martinez:
Thanks for the suggestion @lpagliari but I don't think this will do what I need. If I understand correctly, I still need to do `getHTML` / `setHTML` to add the comment tags to the main body of the text.

What I would really want is for my *bot* to be able to "type" on the pad itself and for other user's to see the text from the bot appear in the same way as it does when another user types.

Gared:
@c-martinez I would really appreciate to see this function in the HTTP API. It shouldn't be too difficult to implement this function. Maybe I have time to implement this the next days.

c-martinez:
Hi @Gared -- so it would be a useful feature then. I'll also give it a go at implementing it then. I thought it was better to ask before I started trying to implement it myself. Now that I know it is not yet there and it may be useful, I'll have a go at it.

Gared:
@c-martinez Yes, indeed. Please let me know if you need any help.

Gared:
@c-martinez Any update on this? Do you need any help?

c-martinez:
Nothing yet, sorry @Gared -- I was distracted by another project, but I should get back to it this week.

PR #2914, opened 2016-02-20
Insert text via API

c-martinez:
Implements feature suggested on issue #2897.

Not quite yet finished, but would like to check this is going in the right direction. Feedback welcome!


src/node/handler/APIHandler.js
```diff
@@ -56,6 +56,7 @@ var version =
   , "listSessionsOfAuthor"      : ["authorID"]
   , "getText"                   : ["padID", "rev"]
   , "setText"                   : ["padID", "text"]
+  , "insertText"                : ["padID", "text", "lineNum"]
```
Gared: Add your new API method only to the latest API version

c-martinez:
Ah, yes -- I started by adding it to every version while I was figuring it out. Removed.

JohnMcLear:
This needs backend tests plz :)

c-martinez:
Better?

Gared:
Looks like the pad is being destroyed if you remove the "lineNumber" parameter (I get javascript errors after calling the API without this parameter). Please make this parameter necessary.

Gared:
Also please set up a "real author" otherwise you'll get the error message "There is no author for authorId: API-bot"

JohnMcLear:
Thanks for test coverage @c-martinez -- Please also add coverage for the bits @Gared has discovered

c-martinez:
@Gared, thanks for feedback. lineNum now is a required parameter.

I'm adding authorID also as a required parameter -- but I'm not sure how to test this -- `createAuthor` doesn't seem to have tests either, and I'm not sure how to get a authorID to use in the tests. Any tips?

c-martinez:
Hi -- added a "real author" (via authorMapper -- not sure if this is the most logical way), and added tests for it as well.

Along the way, added tests for `createAuthor` and `createAuthorIfNotExistsFor`.

comments?

src/node/db/API.js
```diff
@@ -298,15 +298,91 @@ exports.setText = function(padID, text, callback)
...
+  var authorMangager = require("./AuthorManager");
+  // Check author exists
+  authorMangager.createAuthorIfNotExistsFor(authorMapper, false, function(err, authorID)
+  {
+    if(ERR(err, callback)) return;
+
+    //author does not exist
+    if(authorID == false)
+    {
+      callback(new customError("Author not given or given author does not exits","apierror"));
```
Gared: propably there is a misspelling in "exits". Maybe "exist"? :-)

src/node/db/API.js
```diff
@@ -298,15 +298,91 @@ exports.setText = function(padID, text, callback)
...
+        var Changeset = require("ep_etherpad-lite/static/js/Changeset");
+        var oldText = pad.text();
+        // Index of insert point
+        var oldLines = oldText.split("\n");
+        var insertPoint = 0;
+        if(lineNum < oldLines.length) {
+          for (var i = 0; i < lineNum-1; i++) {
+            insertPoint += oldLines[i].length + 1;
+          }
+        } else {
```
Gared: Please use a separate line for the brackets and "else" (code formatting)

Gared:
Looks good to me so far. Thx for your work :-)
Will do some tests the next days.

c-martinez:
Well spotted @Gared , thanks!

src/node/db/API.js
```diff
@@ -298,15 +298,93 @@ exports.setText = function(padID, text, callback)
...
+        var oldLines = oldText.split("\n");
+        var insertPoint = 0;
+        if(lineNum < oldLines.length) {
+          for (var i = 0; i < lineNum-1; i++) {
+            insertPoint += oldLines[i].length + 1;
+          }
+        }
+        else
+        {
+          insertPoint = oldLen;
```
Gared: Where is oldLen set? Seems to me that this variable is not defined at this point

src/node/db/API.js
```diff
@@ -298,15 +298,93 @@ exports.setText = function(padID, text, callback)
...
+  }
+
+  var authorMangager = require("./AuthorManager");
+  // Check author exists
+  authorMangager.createAuthorIfNotExistsFor(authorMapper, false, function(err, authorID)
+  {
+    if(ERR(err, callback)) return;
+
+    //author does not exist
+    if(authorID == false)
```
Gared: Why do you check the authorID? As far as I can see there is no case that the authorID can be 'false'

src/node/db/API.js
```diff
@@ -298,15 +298,93 @@ exports.setText = function(padID, text, callback)
...
+{
+  //text is required
+  if(typeof newText != "string")
+  {
+    callback(new customError("text is no string","apierror"));
+    return;
+  }
+
+  //check if lineNum is a number
+  if(!isNaN(parseInt(lineNum)))
```
Gared: Please do call "parseInt" only once

src/node/db/API.js
Gared: This seems to be the reason why you get an error if you specify a lineNum that is higher than the actual number of lines on this pad

Gared:
Please also documentate your new API method here:
https://github.com/ether/etherpad-lite/blob/9335ae00d0c889bf96482447990968ca3eef6806/doc/api/http_api.md

c-martinez:
Hi @Gared , thanks for comments. (Sorry, it looks like you spent quite a bit of time fixing issues which I should have seen before -- really thanks for that). I'll address them and send a new commit.

src/node/db/API.js
c-martinez: Good point

src/node/db/API.js
```diff
@@ -298,15 +298,82 @@ exports.setText = function(padID, text, callback)
...
+
+  if(typeof authorMapper != "string")
+  {
+    callback(new customError("Invalid authorMapper","apierror"));
+    return;
+  }
+
+  var authorMangager = require("./AuthorManager");
+  // Check author exists
+  authorMangager.createAuthorIfNotExistsFor(authorMapper, false, function(err, authorID)
```
Gared: In the callback you get the author object and not the authorID

src/node/db/API.js
```diff
@@ -298,15 +298,82 @@ exports.setText = function(padID, text, callback)
...
+      {
+        insertPoint = oldLen-1;
+      }
+
+      var assem = Changeset.smartOpAssembler();
+      assem.appendOpWithText('=', oldText.substring(0, insertPoint));
+      assem.appendOpWithText('+', newText);
+      assem.endDocument();
+      var typedChanges = Changeset.pack(oldLen, newLen, assem.toString(), newText);
+      pad.appendRevision(typedChanges, authorID);
```
Gared: Use here author.authorID

src/node/db/API.js
Gared: Is there any special reason why you set the parameter for "name" to "false"?

Gared:
@c-martinez Can you do these changes, please? I want to release a new version of etherpad with this feature.

c-martinez:
Hi @Gared -- yes, sorry, I've been distracted with other things. I'll make these changes later today.

c-martinez:
Done!

But I get the impression that I'm still doing something not-quite-right... I'm using `createAuthorIfNotExistsFor` but I don't think I should be creating a new author. Is there a checkIfAuthorExistsForMapper function (or something similar?)

Gared:
@c-martinez No, there is currently no such function, but I'm fine to use the author mapper to create an author if necessary. So I would merge your PR now.

c-martinez:
Would it be a potentially useful feature?

Gared:
@c-martinez I think you're right. If we use the authorMapper to create a new author there is no way to insert text as another author or create an author with a name. So I would prefer to replace the parameter ```authorMapper``` with ```authorID```. There should be a check if this authorID exists, but nothing more. What do you think about?

To answer your question: I don't think such a function would be useful, because the authorMapper is a random string created by an application using the API. In my opinion there is no reason to ask etherpad for an authorMapper.

c-martinez:
I guess using an authorID makes more sense then.

Gared:
@c-martinez Thanks for your quick changes, but I've noticed that there is a problem after applying a text to a pad:
1. Open a pad
2. Write something in the pad
3. Call insertText-API with the authorID of your current user in the Pad
4. Write something in the pad again
5. You get a disconnect in the pad
```
[WARN] client - Error: applySubmittedChangesToBase: no submitted changes to apply -- { errorId: 'pp8WgPgOyFStijvP4rXA',
  msg: 'Error: applySubmittedChangesToBase: no submitted changes to apply',
  url: 'http://localhost:9001/p/test',
  linenumber: 10465,
  userAgent: 'Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:45.0) Gecko/20100101 Firefox/45.0' }
```

It seems that there is a problem with the author and the change being applied to the pad

c-martinez:
@Gared -- I'll look into it.

c-martinez:
@Gared -- reproduced the issue you mentioned. It seems like the API call itself is working fine, but the browser session gets disconnected or something.

We could avoid this by not using a normal author to insert text via the API (i.e. do not have an authorID as a parameter). Instead we could create a 'special' author (via `createAuthorIfNotExistsFor` for instance), and always use the same author for API calls. This should solve the issue, right? What do you think?

Gared:
@c-martinez That would work, but I would really appreciate if we can add text as any author via the API.
I think there is a problem with the authorID in the change, because the added text via the API isn't colored and has no author attributes. This looks like the main problem to me.

fcassin:
The problem seems to be that the client receives an ACCEPT_COMMIT message but there was no submittedChangeset created beforehand. This is what causes the uncaught error.

If you have two users connected on the same pad when the API call is made, the user not "impersonated" by the API sees the changes immediatly, but with no author color. Only the impersonated user loses its session after the API call. Maybe a NEW_CHANGES message is required, instead of a ACCEPT_COMMIT message.

c-martinez:
Following @Gared suggestion on #2950, this should add the colour to the text according to the colour set by the user. (Maybe need to check if author already has attributes set for this pad).

JohnMcLear:
@c-martinez is this something you still need?

c-martinez:
Not really -- I am not using etherpad at the moment.

Issue #2740, opened 2015-08-01
Do not save new pad to database on initial GET request


Gared:
We should save a pad to the database after the first change on the pad text.
There are some users that create a pad to see what etherpad is, but never do any changes on these documents or crawlers that request urls that opens new pads.
What do you think about this idea? Maybe I have time to implement this.

bhldev:
Completed in #2923 

**Only saves initial key as read only, ex. saves the following**
{"key":"pad2readonly:DbNoSave4","val":"r.bbea3ea68384826a1456f6edc2ab9ed7"}
{"key":"readonly2pad:r.bbea3ea68384826a1456f6edc2ab9ed7","val":"DbNoSave4"}

**Does not save the following until first revision**
{"key":"pad:DbNoSave4:revs:1","val":{"changeset":"Z:bj>1|7=bi*0|1+1$\n","meta":{"author":"a.B7o63QPh4YbRdTs8","timestamp":1457914152598}}}
{"key":"pad:DbNoSave4","val":{"atext":{"text":"Welcome to Etherpad!\n\nThis pad text is synchronized as you type, so that everyone viewing this page sees the same text. This allows you to collaborate seamlessly on documents!\n\nGet involved with Etherpad at http://etherpad.org\n\nWarning: DirtyDB is used. This is fine for testing but not recommended for production. -- To suppress these warning messages change suppressErrorsInPadText to true in your settings.json\n\n\n","attribs":"|7+bi*0|1+1|1+1"},"pool":{"numToAttrib":{"0":["author","a.B7o63QPh4YbRdTs8"]},"nextNum":1},"head":1,"chatHead":-1,"publicStatus":false,"passwordHash":null,"savedRevisions":[]}}

PR #2923, opened 2016-03-14
Save Only On First Revision #2740

bhldev:
Creates setting to save only on first pad revision feature request #2740

### DIFF
--- a/settings.json.template
+++ b/settings.json.template
@@ -67,6 +67,7 @@
     "rtl": false,
     "alwaysShowChat": false,
     "chatAndUsers": false,
+    "savePadAtStart": true,
     "lang": "en-gb"
   },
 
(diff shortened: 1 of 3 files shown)


Gared:
This doesnt't look good to me. What happens if you write a chat message without editing the pad?
I would expect that the first entry to the database will be written when also something pad related should be written to the database (like chat messages, authors, etc.).
I think these change may have some ugly side effects on hooks or something like that. Needs to be tested...

bhldev:
Can't get the frontend tests to work with localhost, cross origin error
Also using dirtydb, real tests should use a real database
I will have to look at this later in the month, if someone else hasn't

JohnMcLear:
+1 @Gared this causes more problems than it solves?

bhldev:
Closing, will resubmit if/when I can think of a better solution

PR #2929, opened 2016-03-18
Pass through the "item" parameter to registerAceCommand callbacks.

xavidotron:
Passing through the item parameter makes it easier to have one registerAceCommand callback that handles multiple toolbar items in a plugin and makes the interface more similar between registerCommand() and registerAceCommand().


xavidotron:
Any chance of getting this pull request looked at?

Gared:
I think this should be finde. What do you think @lpagliari

lpagliari:
Yep, looks good to me. Sorry, @xavidotron , for taking so long.

PR #2938, opened 2016-03-25
Remove sessions script

JohnMcLear:
DO NOT MERGE YET

Contains two scripts:
- One that removes pointless sessions that clutter the database
- One that removes old pad data, this removes the ability to play back pads from the timeslider

Still to do before merge:
- [ ] Discover why so many database entries are made for sessions
- [ ] See if there is a way to remove pad database entries without breaking timeslider functionality

### DIFF
--- a/bin/removeOldPadData.js
+++ b/bin/removeOldPadData.js
@@ -1,0 +1,79 @@
+/*
+  This is a debug tool. It removes old pad data from a couch database
+*/
+
+//get the padID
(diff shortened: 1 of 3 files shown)


JohnMcLear:
I should keep an eye on https://github.com/ether/etherpad-lite/pull/3000 to see if we have any conflicts.

JohnMcLear:
Closing as it doesn't appear there is much need for this.

PR #2940, opened 2016-03-26
extend attributesonselection method

JohnMcLear:
Ready to test and hopefully merge

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -2338,8 +2338,17 @@ function Ace2Inner(){
   editorInfo.ace_setAttributeOnSelection = setAttributeOnSelection;
 
 
-  function getAttributeOnSelection(attributeName){
+  function getAttributeOnSelection(attributeName, prevChar){
     if (!(rep.selStart && rep.selEnd)) return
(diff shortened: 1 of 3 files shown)


src/static/js/ace2_inner.js
```diff
@@ -2338,8 +2338,17 @@ function Ace2Inner(){
...
 
 
-  function getAttributeOnSelection(attributeName){
+  function getAttributeOnSelection(attributeName, prevChar){
     if (!(rep.selStart && rep.selEnd)) return
+    var isNotSelection = (rep.selStart[0] == rep.selEnd[0] && rep.selEnd[1] === rep.selStart[1]);
+    if(isNotSelection){
+      if(prevChar){
+        // If it's not the start of the line
+        if(rep.selStart[1] !== 0){
```
lpagliari: What if line has line attribute (which makes it start with "*")? Is that ok to look the attributes there, or should you look the attributes at 2nd char on line text?

JohnMcLear: Hrm, would be interesting to try, in through it should try to put rep.selStart[1] as -1 in that instance which is not ideal :P

PR #2944, opened 2016-03-30
Create hook to register events that won't scroll editor after aceEditEvt

lpagliari:
One of the plugins I use triggers edit events under the hood that might change the text of the pad. This makes the scroll position to change at the end of the event, but I need to prevent this behavior. (Just to give a more concrete example, the plugin paginates the pad -- in batches, to avoid freezing the editor when changing long pads)

One possible solution was to add a new flag to callstack and check for it when going to call `scrollSelectionIntoView`, but I thought that would be too intrusive, so instead I've created a new hook where plugins can register edit events that won't scroll the viewport when event is complete.

Thoughts? Concerns?


JohnMcLear:
Please provide example source code in docs and if possible link to a plugin that leverages this behavior.

lpagliari:
I won't be able to link to the plugin because it is a private repo, but I included the example in docs as you suggested, @JohnMcLear

PR #2947, opened 2016-04-14
Added new API call to get all comments from a pad

tfsamaral:
- Increased API version to 1.2.14 with new function ("getPadComments")
- System checks if the comments plugin is installed, if not returns a new "pluginerror" with a message, so the user knows the plugin is not installed
- If everything goes well the user gets the list of comments through the call


lpagliari:
@tfsamaral ep_comments have an API to get comments: https://github.com/JohnMcLear/ep_comments/blob/master/index.js#L160

PR #2950, opened 2016-04-19
Api typing

fcassin:
This a proposed solution to the issue encountered in @c-martinez pull request.

To work around the issue, I created a new message type : API_APPEND.
The user impersonated by the API reacts to the changes as if it were made by another user, adding the changes correctly to his/her own pad.
Unfortunately, there is still an issue with the color not being displayed.

P.S: I don't know if I could have proposed these changes on @c-martinez 's original pull request, is it possible ? Sorry if I created a needless duplicate.


c-martinez:
I think this works.

@Gared -- what do you think ?

fcassin:
I still need to check for API_APPEND messages in the messageQueue handling, should be a quick change, I'll look into it today.
If I have a bit more time I'd like to get the colors displaying properly.

Gared:
I'm not familiar with the message types, but this might work. To make this message type reusable and more general I would suggest to name it "API_CHANGE".
Regarding the author colors: I noticed that the attributes of the changeset operation are missing. You have to add the attributes on this method:
```
assem.appendOpWithText('+', newText, '*0');
```
This shows a color, but not the right one. You have to replace the "0" with the appropiate attribute number, but I have to dig deeper in the code to exactly know what to do here.

JohnMcLear:
Is there still any want/need for this or can it be closed?

c-martinez:
The original motivation for this PR is explained in #2897 -- Personally I do not have a concrete use case for it anymore because I am not using etherpad these days.

JohnMcLear:
Okay well I'm gonna close now and if someone else brings it up we can consider implementing.

fkmjec:
Hello, is this still on the table? I would really like to use this feature.

JohnMcLear:
@fkmjec It's not being actively worked on but we're happy for you to pick it up if it's something you are really interested in :)

PR #2970, opened 2016-05-07
Remove [ and ] from the regexp used by the URL filter.

xavidotron:
These characters are rarely used in URLs, and including them leads to
mislinkifying when editing various formats, such as wiki markup formats that
use [] around links.


JohnMcLear:
I'd like to see test coverage for this please :)  https://github.com/ether/etherpad-lite/blob/develop/tests/frontend/specs/urls_become_clickable.js would be the place to add it!

xavidotron:
Added a frontend test and ensured that frontend tests pass. Thanks!

JohnMcLear:
tnx

PR #2972, opened 2016-05-09
Improve french clearcolors message

ldidry:
A lot of users are confused by the clearcolors message. They use it
instead of using the preference to hide the paternity colors and then
complain about losing it.

This commit improve the french clearcolors message by saying "Don't you
want to just hide the colors instead? Use the preference choice." to
avoid this mistake.


Gared:
Use https://translatewiki.net to change the translation

Issue #2963, opened 2016-05-05
Font Awesome not loadable when installed in a subdirectory


rugk:
I'm using [nginx as a proxy with proxy_rewrite from a subdirectory](https://github.com/ether/etherpad-lite/wiki/How-to-put-Etherpad-Lite-behind-a-reverse-Proxy#from-subdirectory-url). It is nearly exact the config you suggested.

However when using a plugin which requires Font Awesome (see https://github.com/JohnMcLear/ep_comments/issues/94) I encountered a loading issue: Nginx seem to have issues to rewrite the URL correctly, because the browser actually tried to load `http://<host>/static/font/fontawesome-etherpad.woff`whereas it should load `http://<host>/SUBDIR/static/font/fontawesome-etherpad.woff`.

It would be nice if this can be fixed. In the meantime I added this workaround in the nginx config:
```
# rewrite etherpad font
# workaround for https://github.com/ether/etherpad-lite/issues/2963
location /static/font/fontawesome-etherpad {
        rewrite /static/font/fontawesome-etherpad(.*) /SUBDIR/static/font/fontawesome-etherpad$1 redirect;
}
```

rugk:
Actually I found out that the [this loading is correct](https://github.com/ether/etherpad-lite/blob/develop/src/static/css/pad.css#L1265), so this seems to be more an ep_comments issue.

JohnMcLear:
Please close your issues if they aren't needed any more....

rugk:
Yeah, sorry. When hopping from one issue to the other one I forgot this...

PR #2977, opened 2016-05-13
Fix not-loading of Font Awesome

rugk:
Fixes https://github.com/ether/etherpad-lite/issues/2963

Please confirm it also works in a usual Etherpad installation.


rugk:
No, it does not fix this problem as I noticed, but somewhere there have to be the bug...

Issue #374, opened 2012-01-29
exporting authorship colours


dora71:
This is a feature request:
It would be nice that when exporting to PDF, OpenDocument or Microsoft Word, not only to export the plain text, but also the authorship colours.
Perhaps it is possible to choose whether or not to export the colours by (un)checking a box.

JohnMcLear:
Bounty for this: https://www.bountysource.com/issues/1371946-exporting-authorship-colours

JohnMcLear:
https://github.com/Gared/ep_export_authors

PR #2994, opened 2016-05-30
Export author color

JohnMcLear:
For #374


JohnMcLear:
@Gared can you please test and merge this if it passes?

dcht00:
Hi all, is there maybe any status update on this?

muxator:
Hi @dcht00, I am not working on this right now. I do not know if someone else is having a look, but the feature looks interesting for sure.
I would welcome contributions on this.

Some practical details:

The feature was developed right on top of 1.6. The contributor could test & eventually fix that version (5cd781d09fd4).
Afterwards, it could be forward ported to current develop (1.7+) and merged there.

JohnMcLear:
@muxator is there still any interest for this?  Looks like it is a relatively healthy PR despite conflicts.

muxator:
This PR has many conflicts, and probably it will have more once we merge #3786.

Should be reworked.

JohnMcLear:
https://www.npmjs.org/package/ep_export_authors <-- this plugin does it anyway.  Closing as it's not something that needs to be in core.

Gared:
But in my plugin I create random author colors, becauase I can't load the original one from database

JohnMcLear:
Yeah I spent a while on this and noticed most of this code is in core...  Not sure why we don't expose it...  ?

Gared:
I think this would be good to expose in plugins, but as far as I know we can't do it right now, because the hook is not async

Issue #2985, opened 2016-05-16
Strange cookies are set and bypass cookie_path redirection 


rugk:
Two cookies are set:
- `test` (content: `test`) - I know this is probably an plugin... :expressionless:
- `token`

At least of the first one I want to get rid of.

Although I have set this in my nginx config, these two cookies do seem to byass this and set their path to `/`:

``` nginx
proxy_cookie_path / /etherpad/;
```

All other cookies (`prefs`, `language`, `express_sid`) are fine.

JohnMcLear:
Try disabling plugins ........

rugk:
Okay, now I disabled all plugins and I can see that this is *not* an issue with a plugin.

I could also identify the requests where this is set. You can see them in this gif:
![cookieetherpad](https://cloud.githubusercontent.com/assets/11966684/15714727/fb8d8368-281b-11e6-8f90-fb5ad8ff958b.gif)

So to summarize:
* the `test`cookie is first transmitted by the *second* socket.io request after the request to locales.json.
* the `token`request is first transmitted by the POST request to socket.io

I could reproduce it with the latest dev version (a7405825a8fa011bdea2a50fd6b0f2756abd674d).

Another thing I noticed:
When the connection breaks (because I e.g. restart Etherpad-lite) and Etherpad tries to connect to the Etherpad server again, the `token` cookie is restored:
![cookkierestore](https://cloud.githubusercontent.com/assets/11966684/15715409/a77d8af4-281e-11e6-8333-780ca716e08a.gif)


Reproduce:
1. Load an Etherpad
2. Delete the cookies `test` and `token`
3. Cause Etherpad to close the current websocket connection and having to reconnect (e.g. by restarting the Etherpad server)
4. When the reconnection succeeds the  `token` is set again

----

I also found the places in the source code where this cookies are presumably set:
* test: https://github.com/ether/etherpad-lite/blob/a3188cf3574234f17f927499c2f739c2bd5fd9d3/src/static/js/pad.js#L505
* token: https://github.com/ether/etherpad-lite/blob/a3188cf3574234f17f927499c2f739c2bd5fd9d3/src/static/js/pad.js#L194

----

So to come back to this issue about these cookies, here is what I think should be changed:
1. I don't like test cookies. Can Etherpad life without setting this cookie? Please remove it.
[...]

PR #3000, opened 2016-06-08
Improve cookies

### LINKED ISSUE #2969: Cookies set do not use the secure flag
When using HTTPS all cookies set by Etherpad should get the [secure flag](https://www.owasp.org/index.php/SecureFlag), which prevents that this cookies are leaked in insecure (HTTP) requests.

I tried adding `proxy_set_header X-Forwarded-Proto $scheme;` to my nginx config, but NodeJS still does not add the secure flag.
So I think the easiest way to solve this would be to add a setting to the `settings.json`:  `"cookiesAddSecureFlag": true;`

This way web admins can configure this manually.

The same also applies to the [`HttpOnly`](https://www.owasp.org/index.php/HttpOnly) flag, ~~but AFAIK you need to read all cookies via JavaScript, don't you?~~ Okay, for `express_sid` you already do this. What about other cookies such as `token`? (For what is this cookie used anyway?)
And if I'm already complaining about these cookies: Why does Etherpad set a cookie `test` with the value `test`?

Gared:
Fixes #2969 
- [x] Remove "test" cookie
- [x] Add secure flag to browser created cookies if pad accessed through https
- [x] Make express-cookie scheme dependent


rugk:
Also partially fixes https://github.com/ether/etherpad-lite/issues/2985, does not it? At least the test cookie get's away. Taht's good! :smiley: 
Still the path value of `token` is set incorrectly and alos this `debug` in the local storage is unnecessary.

Gared:
@rugk Yeah, this fixes this ticket partially. One step after another :-)

Gared:
@rugk Can you please test this ticket with your reverse proxy?

src/static/js/pad_cookie.js
```diff
@@ -79,6 +80,10 @@ var padcookie = (function()
       alreadyWarnedAboutNoCookies = true;
     }
   }
+  
+  function isHttpsScheme() {
+    return window.location.protocol == "https:";
```
rugk: I think this won't work for reverse proxys as they usually forward requests over HTTP. You should honor the [`X-Forwarded-Proto`](https://stackoverflow.com/questions/13111080/what-is-a-full-specification-of-x-forwarded-proto-) header - and probably also the standardized [`Forwarded`](https://tools.ietf.org/html/rfc7239) header, but AFAIK the unofficial `X-Forwarded-Proto` is still more common.

src/static/js/pad_utils.js
```diff
@@ -53,13 +53,16 @@ function createCookie(name, value, days, path){ /* Used by IE */
   if(!path){ // IF the Path of the cookie isn't set then just create it on root
     path = "/";
   }
+  
+  //Check if we accessed the pad over https
+  var secure = window.location.protocol == "https:" ? ";secure" : "";
```
rugk: Same as above. This does not honor reverse proxys.

rugk:
I can do that soon. However I already commented on some source code lines where I see that this won't work, so could you please fix this first?

rugk:
For now I already [added this header](https://github.com/ether/etherpad-lite/wiki/How-to-put-Etherpad-Lite-behind-a-reverse-Proxy/_compare/bdb85b0188370354676fa47ca406fbabd9a588e8...16d5a32c0707ade0cbbf5d1b3045793949b1b906) to the config.

Gared:
@rugk Thanks for adjusting the config. The comments in my code are the part which is executed directly in the browser. So the browser knows if it is talking http or https with the server. The cookie which is set on server-side is for express.

rugk:
Ah okay, so I'll test it soon.

rugk:
At first I had some issues with caching: You should consider loading the files with `file.css?v=0.3`.

So but now when I see it working:
* test is gone.
* SecureFlag is set for `prefs` and `token`, but **not** for `express_sid`, `io` and `lang`.

Additionally I get a "cookies could not be found" in the browser when Etherpad is loading (that's before the cookies with SecureFlag are set), but it disappears quickly. Anyway it is a message, which is at least confusing.

And `express_sid`is still the only cookie with HTTPOnly flag.

DanielBickley:
@rugk @Gared 

You're getting the "cookies could not be found message" because the cookie test is failing on "prefs" basically because it doesn't exist at that point. You could conduct the test after the "prefs" cookie is actually attempted to be created, or you could do what I do and just create the test cookie, and if found to exist, then promptly delete the cookie.

rugk:
Any news about this PR? @Gared could @DanielBickley's advice help you?

PR #2957, opened 2016-05-03
Dependency updates

Gared:
Updated these dependencies:

Package | Old version | New version
----------- | -------------- | --------------
socket.io |1.3.7 | 1.4.5
resolve | 1.1.7 | 1.3.7
express | 4.12.3 | 4.13.4
express-session | 1.11.1 | 1.13.0
clean-css | 3.1.9 | 3.4.12
uglify-js | 2.4.19 | 2.6.2
log4js | 0.6.22 | 0.6.35
cheerio | 0.19.0 | 0.20.0
ejs | 2.3.1 | 2.4.1
graceful-js | 3.0.6 | 4.1.3
semver | 4.3.3 | 5.1.0
unorm | 1.3.3 | 1.4.1
jsonminify | 0.2.3 | 0.4.1
measured | 1.0.0 | 1.1.0
mocha | 2.2.4 | 2.4.5
supertest | 0.15.0 | 1.2.0

Todo as another PR:
- async
- npm

PR #3004, opened 2016-06-14
bump some deps that have warnings from deps

JohnMcLear:


JohnMcLear:
We could probably go through and bump a load more stuff too :)

Gared:
@JohnMcLear You mean like here? https://github.com/ether/etherpad-lite/pull/2957

JohnMcLear:
Yea like that but with new bumps :)

JohnMcLear:
don't merge this btw, something is broken with ueberdb2 @ 0.3.5...

PR #3027, opened 2016-07-23
Download JQuery over HTTPS

rugk:
Otherwise anyone can MITM the connection and add rogue code there.

You should be more careful with downloading things from remote servers.


rugk:
I seems your Travis-CI test fails, because some secrets are missing. In any case this PR does not cause this issue.

lpagliari:
@rugk Thanks!

PR #3045, opened 2016-08-24
Use an EEJS template for HTML export instead of inlining it in the JS code.

danfuzz:
The semantics of all the substitutions are identical to what they were before. I _did_ take the liberty of formatting the CSS to be a bit more readable (at the cost of adding a little bit of whitespace).

**Note:** I did _not_ fix the pre-existing misspellings in the CSS, e.g. `eigth`, `twelth`, and `sixthteenth`. Other than whitespace, I left the export document contents alone.

I checked both front-end and back-end tests, and they all seem to be okay when run on my local machine. (Two of the back-end tests come back as "pending." I assume that's okay because that's how they also report when I run on a clean checkout of the `develop` branch.)

### DIFF
--- a/src/node/handler/ExportHandler.js
+++ b/src/node/handler/ExportHandler.js
@@ -76,7 +76,7 @@ exports.doExport = function(req, res, padId, type)
       }
       else if(type == "txt")
       {
-        exporttxt.getPadTXTDocument(padId, req.params.rev, false, function(err, txt)
+        exporttxt.getPadTXTDocument(padId, req.params.rev, function(err, txt)
         {
           if(ERR(err)) return;
           res.send(txt);
--- a/src/node/handler/ExportHandler.js
+++ b/src/node/handler/ExportHandler.js
@@ -92,7 +92,7 @@ exports.doExport = function(req, res, padId, type)
           //render the html document
           function(callback)
           {
-            exporthtml.getPadHTMLDocument(padId, req.params.rev, false, function(err, _html)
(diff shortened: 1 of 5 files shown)


danfuzz:
I just noticed that I _did_ inadvertently change the result if `noDocType` is passed as `true` to `getPadHTMLDocument`: In the original, that would _also_ cause it to not emit the `<head>` section. But that said, the only two call sites to `getPadHTMLDocument` pass `noDocType` as `false`, so my PR is still semantics-preserving with respect to the actual uses.

If you think I should update the PR, then please let me know. If anything, I'd be inclined to just remove the `noDocType` argument and thereby keep the code a bit simpler.

danfuzz:
(I squashed the branch to remove my separate whitespace fix commits.)

lpagliari:
Hi, @danfuzz . Good point about `noDocType`. I've noticed the export to TXT has also the same extra (unused and always `false`) var for doc type: `exporttxt.getPadTXTDocument(padId, req.params.rev, false, function(err, txt)`.

Would you mind removing `noDocType`, as you suggested? I'll merge your PR when you're done.

And thanks a lot for the code!

danfuzz:
Thanks for the feedback. Done!

It wasn't clear to me if you wanted me to change `getPadTXTDocument()` or just `getPadHTMLDocument()`. I made both changes, but kept them as separate commits in case you only wanted the latter.

lpagliari:
Oh, that's even better! Thanks a lot, @danfuzz !!

danfuzz:
You're welcome! Glad to be of service.

PR #3048, opened 2016-08-30
Account for any top margin on the first line.

danfuzz:
Without this change, a top margin on any element on the first line of pad
content would throw off the alignment of line numbers. The default stylesheet
doesn't define any elements with top margins, but plugins might. (This is also
explained in a code comment.)

In order to see the problem, add the following clause to `iframe_editor.css`
(before incorporating this commit):

    #innerdocbody > :first-child {
      margin-top: 100px;
    }


danfuzz:
This passed all local FE test cases.

lpagliari:
Thanks, @danfuzz !

PR #3051, opened 2016-09-08
Fix typos in ordinal names.

danfuzz:
- `eigth` -> `eighth`
- `twelth` -> `twelfth`
- `sixthteenth` -> `sixteenth`
- `fixteenth` -> `sixteenth`

PR #3053, opened 2016-09-09
Fix timeslider lists / indents, but also make CSS `@import` work

danfuzz:
The main point of this PR was to harmonize the CSS related to lists and indentation on the pad editor and timeslider views, which were quite different. With that goal in mind, I extracted the pad version of the CSS (which looks to be the better-maintained version, and certainly more consistent with what's done in HTML export) into its own file. This file then got referenced via `@import` in both `iframe_editor.css` (the pad editor CSS). Then, for `timeslider.css` I removed its corresponding (but incorrect) CSS, again replacing it with an `@import`.

With all that done, I then had to go into the guts of `Minify.js` to get it to handle `@import`s when minifying CSS. It took me two attempts to get that fully right, which you can see in the commit history. (If this offends your source control sensibilities, I'd be happy to rebase, etc. Let me know.)

The end result has been tested both with and without CSS minimization, and it seems to work in both configurations. Backend and frontend tests pass on my local box (with two backend tests marked "pending" which is the same as how they show up in a clean branch for me).

**Note:** This PR incorporates some of the same typo fixes that I also addressed in https://github.com/ether/etherpad-lite/pull/3051. If you merge that PR, then this one will have to be rebased/merged; likewise vice versa.

**Note:** My editor auto-stripped EOL whitespace in a couple of places.

### DIFF
--- a/src/node/utils/Minify.js
+++ b/src/node/utils/Minify.js
@@ -409,7 +409,8 @@ function compressJS(content)
 function compressCSS(filename, content, callback)
 {
   try {
-    new CleanCSS({relativeTo: ROOT_DIR}).minify(content, function (errors, minified) {
+    var base = path.join(ROOT_DIR, path.dirname(filename));
+    new CleanCSS({relativeTo: base}).minify(content, function (errors, minified) {
       if (errors) {
         // On error, just yield the un-minified original.
         callback(null, content);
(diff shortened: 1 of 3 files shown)


danfuzz:
Here's a trio of screenshots, editor view, timeslider before this PR, and timeslider with this PR. It shows a numbered list example, but to be clear this PR also addresses similar problems with the other kinds of lists / indents.

**Pad Editor**
pad-editor

**Timeslider Without This PR**
timeslider-before

**Timeslider With This PR**
timeslider-after

Gared:
Sorry for the late response. This looks good to me

PR #3055, opened 2016-09-12
Ensure that all lines in the pad are marked with class `ace-line`.

danfuzz:
Without this change, lines that haven't ever been edited will have either
an empty class or, in the case of list start lines, a class that begins
with a space (because the `ace-line` before the space never got added).

### DIFF
--- a/src/static/js/domline.js
+++ b/src/static/js/domline.js
@@ -1,5 +1,5 @@
 /**
- * This code is mostly from the old Etherpad. Please help us to comment this code. 
+ * This code is mostly from the old Etherpad. Please help us to comment this code.
  * This helps other people to understand this code better and helps them to improve it.
  * TL;DR COMMENTS ON THIS FILE ARE HIGHLY APPRECIATED
  */
@@ -135,7 +135,7 @@ domline.createDomLine = function(nonEmpty, doesWrap, optBrowser, optDocument)
             }
             postHtml += '</li></ol>';
           }
-        } 
+        }
         processedMarker = true;
       }
       _.map(hooks.callAll("aceDomLineProcessLineAttributes", {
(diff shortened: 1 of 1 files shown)


danfuzz:
PR passes both front-end and back-end tests on my machine.

danfuzz:
As an example, with the following pad content:

pad-view

(where the second line is a numbered list), upon initial load, the DOM for the two lines is as follows _without_ this PR:

```
<div id="magicdomid2" class="">...</div>
<div id="magicdomid3" class=" list-start-number1">...</div>
```

But _with_ this PR, the corresponding DOM is:

```
<div id="magicdomid2" class="ace-line">...</div>
<div id="magicdomid3" class="ace-line list-start-number1">...</div>
```

This (latter) is notably what the DOM is upon initial entry of the document (from scratch), and also what the DOM turns out to be if you make edits on the lines.

JohnMcLear:
What's the point of this?  Seems like it fixes nothing.

danfuzz:
Okay. Seems like the status quo violates the (implicit) design of the system, but YMMV. (I've long since given up trying to work on this codebase, but left this PR open in case it would be of use.)

muxator:
Confirmed.
This commit tried to guarantee an invariant:
- on first load, either a line has an empty class, or - if it needs to have a class because, for example, is an ordered list - its class attribute starts with a space and then has the relevant class
- after editing it, a line ends up having an `ace-line` class prepended to its class list

This sounds like something performs string concatenation on an uninitialized attribute. The fix is sound.

@danfuzz, if you don't mind, I would commit this one-liner with the original metadata (name, date, ...).

Thanks

danfuzz:
@muxator Please go right ahead.

muxator:
Committed 2a876e5e9ba9. Clearly the world is a safer place, now :)

PR #3059, opened 2016-09-20
Added LibreJS support

### LINKED ISSUE #2913: Add LibreJS support
Since Etherpad is free software, it should be labelled as such.

Making it compatible with [LibreJS](https://www.gnu.org/software/librejs/) would be a good way to do it, all it needs to be done is add some [magnet links](https://www.gnu.org/software/librejs/free-your-javascript.html) (paragraph 3.2.2.1) pointing to the correct license.

paper-pauper:
The source code should now be 100% LibreJS compliant via the web labels method (the least invasive, no modification to any of the JavaScript files required).

I could only test the homepage because LibreJS breaks when the callback is passed, but that's not Etherpad's fault and it's something they should look into.

Closes #2913, all changes released under the same license as etherpad-lite (Apache 2.0)

### DIFF
--- a/src/node/hooks/express/specialpages.js
+++ b/src/node/hooks/express/specialpages.js
@@ -16,6 +16,13 @@ exports.expressCreateServer = function (hook_name, args, cb) {
     res.send(eejs.require("ep_etherpad-lite/templates/index.html"));
   });
 
+  //serve javascript.html
+  args.app.get('/javascript', function(req, res)
+  {
+    res.send(eejs.require("ep_etherpad-lite/templates/javascript.html"));
+  });
+
+
   //serve robots.txt
   args.app.get('/robots.txt', function(req, res)
   {
--- a/src/templates/admin/index.html
+++ b/src/templates/admin/index.html
@@ -20,5 +20,6 @@
         </ul>
       </div>
     </div>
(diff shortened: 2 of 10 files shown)


JohnMcLear:
"I could only test the homepage because LibreJS breaks when the callback is passed, but that's not Etherpad's fault and it's something they should look into."

Huh?  So other pages are broken in this PR?

paper-pauper:
There are no broken pages in this PR - try it yourself :smile:

It's LibreJS that breaks, but as I said, it's something they should look into, because as far as I know everything was set up correctly. As a proof of that, there are no blocked scripts, only these errors given by LibreJS:

```
error parsing: b8a4179684707f366ec7f0e06c9eedccc4677eb9
http://localhost:9001/javascripts/lib/ep_etherpad-lite/static/js/pad.js?callback=require.define

error parsing: 205296743f94aa3aa1c7d2ed860fb278b53c68d8
http://localhost:9001/javascripts/lib/ep_etherpad-lite/static/js/ace2_common.js?callback=require.define

error parsing: 3fd3f3096ff407c26a846d6b9dd4433377ef1410
http://localhost:9001/socket.io/socket.io.js
```

In short, Etherpad would not work only for people using LibreJS (as it already doesn't). Only this time it's LibreJS that needs to be patched in order to fix this issue.

PR #3067, opened 2016-10-28
Add additional link for demoing

neynah:


lpagliari:
@neynah I get an alert saying "Somehow, you seem to have attempted to launch a package you have not installed." when I click on the "Try Etherpad" link. Is there a more straightforward way for users to try Etherpad @ Sandstorm?

One more suggestion: there are several other sites that run Etherpad (check [this wiki page] (https://github.com/ether/etherpad-lite/wiki/Sites-that-run-Etherpad)), so maybe adding a link to this page on the `README` + adding Sandstorm to the list would be a better way to include Sandstorm on Etherpads docs. What do you think?

neynah:
There's a temporary issue that affects people who have new versions of Etherpad installed (compared to the available version on the app market). We'll be fixing it over the next few days, however, it does work for those using the available Etherpad version from the App Market; you can test this by opening the link in incognito. We are fixing this in the next few days so I'll keep you posted.

Thanks for the suggestion. I will also add a link to that wiki.

Issue #2948, opened 2016-04-14
Frontend travis sauce test doesn't run


tfsamaral:
After unzipping the "sauce.zip", just gives an error "Missing required arguments USERNAME, API_KEY".

Like so: 
**Archive:  /tmp/sauce.zip
      inflating: /tmp/NOTICE.txt
      inflating: /tmp/license.html
      inflating: /tmp/Sauce-Connect.jar
Missing required arguments USERNAME, API_KEY**

I noticed it when trying to do a pull request

JohnMcLear:
Only me and Pita appear to be authorized to do this, I can't remember how it's setup.. ;\

jainendra:
Please close this issue if resolved

lpagliari:
not resolved yet

rhelmer:
I can get into the github and travis configs, looks like the `SAUCE_USER` and `SAUCE_ACCESS_KEY` are not set. I'll see if I can't figure it out.

rhelmer:
OK well I got it a little closer - I made a new sauce labs account (etherpad-lite, but we can probably get them to merge and just use etherpad). Now the travis script appears to be hanging instead of erroring out, so there is further work to be done.

@lpagliari I suggest we disable the sauce labs integration (maybe just comment it out) until it's functional, so contributors don't get confused by their PRs failing. What do you think?

lpagliari:
Thanks for working on this, @rhelmer ! [My understanding](https://github.com/ether/etherpad-lite/issues/3179#issuecomment-299856724) was that we needed to use JWT to enable PRs to run SauceLabs + Travis:

> Note that due to security restrictions, the Sauce Labs addon is not available on pull request builds unless you use the JWT Addon.

From [here](https://docs.travis-ci.com/user/sauce-connect/).

But I'm fine with disabling the integration until we fix this -- lately most of the improvements of Etherpad come from external PRs anyway.

muxator:
This was dealt with in #3230 (59ba183c82eb) disabling the Sauce integration.

BTW, in the meantime `tests/frontend/travis/sauce_tunnel.sh` rotted, because it now tries to download a file that gives an HTTP/404. I'll add a comment in the code anyway.

muxator:
For context: the Saucelab frontend tests were enabled again by John and me with 61f0b5db0bfe, and the following work.

PR #3119, opened 2017-01-12
Create plugin to run tests for line attribs + fix #3118

### LINKED ISSUE #3118: Line attributes not preserved when dragging and dropping line
bug_dnd-et

**How to reproduce:**
- create a pad with 2+ lines
- set a line attrib on line X
- set a different line attrib on line Y
- select the entire line X (including line break)
- move it to the beginning of line Y

**Expected:** line X is saved with its original line attribs
**What actually happens:** line X is saved with line attribs of line Y
**Browsers tested:** Chrome, Firefox, Safari

I'm already working on the fix for this.

lpagliari:
Besides fixing #3118, this PR also includes the plugin [ep_test_line_attrib](https://github.com/storytouch/ep_test_line_attrib) when running tests on Travis. This was necessary because there was no way to run tests with the current Etherpad structure, due to the fact that there are no line attribs defined by Etherpad besides the one with lists -- and code was already working properly with lists, but was failing for other line attribs.

This plugin will also allow other tests for plugin-related features in the future.

Does anyone have any concerns about adding this plugin to the test?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js:
 install:
   - "bin/installDeps.sh"
   - "export GIT_HASH=$(git rev-parse --verify --short HEAD)"
+  - "npm install ep_test_line_attrib"
 before_script:
   - "tests/frontend/travis/sauce_tunnel.sh"
 script:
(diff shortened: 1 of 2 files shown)


lpagliari:
Any idea how to solve the `Missing required arguments USERNAME, API_KEY` error message when building on Travis? I know #2948 is about this issue, but it is still happening. Ideas? @JohnMcLear @Pita 

As this PR is changing the tests on Travis, it would be necessary to run them to be sure I didn't break anything.

PR #3122, opened 2017-01-12
readme: added curl as dependency

tpltnt:
Hi there,

`curl` was missing as a dependency on FreeBSD. I adjusted the readme.

Cheers,
tpltnt


lpagliari:
Thanks!

PR #3123, opened 2017-01-17
Update called from for two hooks

Sjord:
`expressCreateServer` and `expressConfigure` are called from [`express.js`](/ether/etherpad-lite/blob/develop/src/node/hooks/express.js#L83-L84), not
from `server.js`.


lpagliari:
Thanks!

PR #3126, opened 2017-01-19
Add settings to scroll on edition out of viewport

joassouza:
By default, when there is an edition of a line, which is out of the
viewport, Etherpad scrolls the minimum necessary to make this line
visible. This makes that the line stays either on the top or the bottom
of the viewport. With this commit, we add a setting to make possible to
scroll to a position x% pixels from the viewport. Besides of that, we
add a setting to make an animation of this scroll.
If nothing is changed on settings.json the Etherpad default behavior is
kept.

Etherpad Default
eto

Google Docs
gdo

Microsoft Word
scroll_ms_word

### DIFF
--- a/settings.json.template
+++ b/settings.json.template
@@ -121,6 +121,21 @@
   /* Privacy: disable IP logging */
   "disableIPlogging" : false,
 
+  /*
+  * By default, when caret is moved out of viewport, it scrolls the minimum height needed to make this
+  * line visible.
+  */
+  "scrollWhenFocusLineIsOutOfViewport": {
+    /*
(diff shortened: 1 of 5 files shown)


lpagliari:
Great work, @joassouza !!

There are some other scenarios that also scroll the viewport: when user uses any of the arrow keys to navigate through the text; if user types something on a line out of viewport (at least on my tests here the scroll happened until X% of the screen, but there was no animation); etc.

Can you handle those scenarios on your PR, please? Thanks!!

joassouza:
Sure @lpagliari. I'm gonna work on it.

joassouza:
I'm close this one to create a new one

PR #3131, opened 2017-01-26
Remove deprecated comment

Sjord:
The session key is currently stored in SESSIONKEY.txt, so it is no longer reset
every time the server starts.


Gared:
You're right. Thx :-)

PR #3160, opened 2017-04-04
wrong path of generate.js

pageye123:
The generate.js is in the `/bin/doc/` directory.
And the origin cmd will cause error.
image


lpagliari:
Great, @pageye123 , thanks!!

PR #3164, opened 2017-04-09
add capability to use the login user name as the author name in pads

weilaudm:
Added useLoginForName to the padOptions in settings.json to allow turn on login name as author name capability
This requires that the requireAuthentication field be set to true, in order to work
If requireAuthentication is false, the useLoginForName parameter has no affect.  Can't use Login name is its not require/there


src/node/handler/PadMessageHandler.js
```diff
@@ -1004,7 +1004,14 @@ function handleClientReady(client, message)
   var historicalAuthorData = {};
   var currentTime;
   var padIds;
-
+  var loginUsr = null;
+  
+  if(settings.requireAuthentication){
+    var authStr = client.conn.request.headers.authorization;
+    loginUsr = new Buffer(authStr.split(' ')[1], 'base64').toString('ascii').split(':')[0];
+  }
```
lpagliari: You are assuming `authorization` header has an specific format -- and more than that: it is not `undefined`. This is not always true: I can set `requireAuthentication: true` on `settings.json` and still get no `client.conn.request.headers.authorization`.

Please provide a safer way to get `loginUsr` (one that won't throw an exception, I mean).

src/static/js/pad.js
```diff
@@ -280,6 +280,12 @@ function handshake()
         $('#options-chatandusers').parent().hide();
         $('#options-stickychat').parent().hide();
       }
+      
+      
+      if(clientVars.padOptions['useLoginForName'] && clientVars.userName!= null){
```
lpagliari: Please fix style here: `clientVars.userName!= null` (add spaces; use `!==` instead of `!=`)

JohnMcLear:
Resolved merge conflicts but I don't like this PR.  You can achieve the same thing with a plugin no?

JohnMcLear:
Doesn't https://www.npmjs.com/package/ep_httpauth_author already do this?  I'm gonna close because I think it does..  If I'm wrong I will re-open.

PR #3185, opened 2017-05-05
[fix] Do not show a 'pad deleted' message when pad is copied

### LINKED ISSUE #3183: Client with copied pad receives a "pad deleted" message
How to reproduce:

- open a browser with a pad
- call the api to copy the pad
- result: a message is displayed on the browser saying the pad was removed

[Code](https://github.com/ether/etherpad-lite/blob/develop/src/node/db/Pad.js#L467-L469) that makes this happen:
```
Pad.prototype.copy = function copy(destinationID, force, callback) {
(...)
  //kick everyone from this pad
  // TODO: this presents a message on the client saying that the pad was 'deleted'. Fix this?
  padMessageHandler.kickSessionsFromPad(sourceID);
```

Does anybody know if there was a reason for `Pad.prototype.copy` to call `kickSessionsFromPad`? It does not make sense to me at all. I suspect that `Pad.prototype.copy` code was based on `Pad.prototype.remove`, and in this case it makes total sense to tell clients the pad is no longer available. But for copy...

I'll implement a fix for this, unless someone gives a good reason to kick everyone from the copied pad.

lpagliari:
Message should be shown when pad is deleted or moved, not when it is copied.

Tests performed:

* pad is deleted using API => shows message to client
* pad is moved to another name using API => shows message to client
* pad is copied to another name using API => **does not** show message to client

Fix #3183

### DIFF
--- a/src/node/db/Pad.js
+++ b/src/node/db/Pad.js
@@ -464,10 +464,6 @@ Pad.prototype.copy = function copy(destinationID, force, callback) {
   }
   else force = true;
 
-  //kick everyone from this pad
-  // TODO: this presents a message on the client saying that the pad was 'deleted'. Fix this?
-  padMessageHandler.kickSessionsFromPad(sourceID);
-
   // flush the source pad:
   _this.saveToDatabase();
 


JohnMcLear:
Has merge conflicts :)  Please resolve.

lpagliari:
Crap! Yeah, @muxator , you're right.

Issue #3179, opened 2017-04-26
BUG: Using EP on same domain, but over different protocols causes "Warning: it appears that your browser does not have cookies enabled. "


tiblu:
![image](https://cloud.githubusercontent.com/assets/7405383/25445387/ef13c0a4-2ab6-11e7-900d-1ef12a1c1b7f.png)

# Reproduce

* Access EP instance over HTTPS
* Access EP instance over HTTP on the same domain

# Result

* Alert:
> Warning: it appears that your browser does not have cookies enabled. EtherPad uses cookies to keep track of unique users for the purpose of putting a quota on the number of active users. Using EtherPad without cookies may fill up your server's user quota faster than expected.
* Etherpad renders fine, but prefs are lost.

# Expected

* No error

# Root cause

* When visiting over HTTPS ``prefs=..;secure`` cookie is set.
* When visiting over HTTP, EP tries to write ``prefs=`` NON ``secure`` cookie, but fails cause browser denies overwriting ``secure`` cookie with the same name. Thus error is displayed.

# Introduced by

* https://github.com/ether/etherpad-lite/commit/06ff0230474f9c3b5baa4e5827b897f42fe16213

# Solution

I know this probably does not bother a lot of people as HTTPS should be the standard protocol and never use over HTTP. For us it only causes problems in testing environments. If we talk about solutions only idea I have is:

* To have differently named cookies for HTTPS and HTTP. 
    * HTTPS - ``prefsSecure`` (or whatever)
    * HTTP - ``prefs``

The shortcoming of this solution is that prefs are different for HTTP and HTTPS depending which User opens. But taking into account, that prefs are lost anyway for this bug, it's OK.

I will PR this if people find this solution OK.

lpagliari:
@Gared what do you think?

tiblu:
@lpagliari  @Gared Forgot to mention that I only tested this on Chrome. Not sure what other browsers think of trying to write non-secure cookie with the same name on the same domain.

lpagliari:
@tiblu I'm not sure if @Gared will have time to take a look on this for now, but for me it makes sense to fix as you said. Could you implement the fix for this, please? Thanks!

Gared:
@tiblu Sorry for the late response.
I think this makes sense to fix this issue as you described. It's the best solution for this problem right now.

tiblu:
@Gared @lpagliari Created a PR, but Travis seems to hang due to some kind of build script issue.

[...]

PR #3187, opened 2017-05-08
#3179 - Using EP on same domain, but over different protocols causes "Warning: it appears that your browser does not have cookies enabled.

tiblu:
PR to fix https://github.com/ether/etherpad-lite/issues/3179


Gared:
Looks good to me. Thanks @tiblu for your work!

Issue #3201, opened 2017-06-10
More Modern Monospaced Fonts


Samurai88:
Would it be nice to have several more monospaced fonts to choose from?

Aside courier, which is the currently the only monospaced font available which I am forced to use if I want to use monospaced font.

Especially handy when pasting parts of code for example.

00SteinsGate00:
Overall some more modern fonts would be nice. For a non-monospaced font [Montserrat](https://www.fontsquirrel.com/fonts/montserrat) is very pretty and free.

lpagliari:
Hi, guys, could you submit a PR with the fonts you want? Here's an example of a commit that added a font to Etherpad: https://github.com/ether/etherpad-lite/commit/41bb4ef094ebd22c5d0068dd9e0ca13816519c81.

00SteinsGate00:
Thanks for the help! I actually got **Montserrat** running and working here, somewhat ready for a Pull request.

One issue remains though. When you download the **Montserrat** Font you do not receive one big font file but numerous (22 I think) ranging from something like *Hairline* over *Regular* to *Bold*.

I chose *Light* because it just looks nice (had a hard time deciding between *Light* and *ExtraLight*), but when I press the **Bold** or **Italic** Button obviously nothing happens. Not sure if that is good behaviour. Any suggestions?

Maybe there is a way to combine more than one font file to one?

<img width="419" alt="bildschirmfoto 2017-06-21 um 19 09 17" src="https://user-images.githubusercontent.com/19982919/27397014-32a00358-56b5-11e7-9522-af3bbaee507c.png">

Montserrat is published under the [SILOpen Font Licence v1.10](https://www.fontsquirrel.com/fonts/montserrat) by the way.

lpagliari:
I'm not an expert on the subject, but I've just found this, which seems to address the behaviour you need: https://stackoverflow.com/a/28339483/7884942

00SteinsGate00:
Genius! Sorry you had to do the Googleing for me 😅 
I'll be right at it!

00SteinsGate00:
Sorry for the double commit confusion. Somehow the package.json was altered by Atom and I accidentally included that in the commit. Instead of doing some weird stuff with removing stuff from an already pushed commit, I thought it is just cleaner to do an all new commit on a new branch.

Thanks for the help with the font by the way 😄

PR #3206, opened 2017-06-21
Added RobotoMono as a more modern and pretty monospaced font

00SteinsGate00:
I added [RobotoMono](http://www.1001freefonts.com/roboto_mono.font) as a new monospaced font as  addressing #3201 

The font is licensed under the GPL license so it shouldn't be a problem. The default monospaced font simply called `monospaced` looks really dated. It might be worth considering switching to this more modern font as a default option, of course only if desired.

bildschirmfoto 2017-06-21 um 23 37 07

I tried to stay in alphabetical order where possible, though on some parts of the code the ordering is simply mixed up.


00SteinsGate00:
By the way @Samurai88, what do you think?

Samurai88:
Hi!
Im am in "spectator" mode and just watching the conversation without 
polluting it with my talks, which I usually do

I myself like the Liberation Mono font alot which I found online 
somewhere for my DreamWeaver.

And the other problem keeps me at the moment from using Etherpad:

When setting a font size, then creating new text or copy pasting 
anything, in other words any new text creation, defaults back to the old 
font size. Really annoying. Im using FireFox 50.

This is what I think at the moment :)

Cheers
Sam

ir. Sam Eftegarie

/architect—ingenieur—ontwerper—onderzoeker/

lpagliari:
Thanks, @00SteinsGate00 ! I'm closing #3201 too

PR #3214, opened 2017-07-05
Add credentials arg to CLI module

patrickeasters:
As I was working on containerizing this for use in OpenShift, I realized that overriding the location of the credentials.json file didn't work properly. There is code to override the value in [Settings.js](https://github.com/ether/etherpad-lite/blob/develop/src/node/utils/Settings.js#L318), but it looks like Cli.js never actually pulls the value for credentials out of the argument list. Adding support for that here.


lpagliari:
Thanks, @patrickeasters

PR #3227, opened 2017-07-25
Fix AbiWord wikilink and capitalization

HairyFotr:
Going through translatewiki.net for this case would be very hard, I hope a PR is ok.

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -71,7 +71,7 @@
   "pad.importExport.exportword": "Microsoft Word",
   "pad.importExport.exportpdf": "PDF",
   "pad.importExport.exportopen": "ODF (Open Document Format)",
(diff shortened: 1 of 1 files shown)


lpagliari:
Sorry, @HairyFotr , this does not work because the locale files are generated avery few weeks, so your changes would be overwritten on the next locale update. You need to go through the languages on translatewiki.net :-/

HairyFotr:
I've noticed that English, as the source language, is not editable on translatewiki. Should I instead make a PR with just the English fix, and then (I'm guessing) translatewiki will mark all other translations as "outdated", and they'll eventually get fixed?

lpagliari:
Hum, good point. Yes, I believe that would work. If you want, just update this PR here.

HairyFotr:
Ok, updated.

lpagliari:
Tks

PR #3229, opened 2017-07-27
DO NOT MERGE just testing travis integration

rhelmer:
Seeing if sauce labs works now.


JohnMcLear:
Afaik it only works if me or Pita submit the PR..  Can't remember why ;\

Issue #2864, opened 2015-12-24
Import a PDF file not working


quydang:
Importing a text file breaks abiwords capability to import PDF files until service restart.

STR.
1. Create pad
2. Import Text file
3. Attempt to import PDF file

Import will fail.  This is probably caused by reasons modification to the import logic

JohnMcLear:
What are the steps to replicate here?

quydang:
Hi @JohnMcLear :First, I created new pad on https://beta.etherpad.org/ . I edited content of pad, export current pad to HTML, text, etherpad, PDF,DOC file.
Then, I import PDF file, it work. But, after importing text file, I can not import PDF file again.
In my local etherpad, I have same error. I can restart service etherpad for import PDF file again. But, after import text file, the error occur.
I just test and replicate the error. I don't know where root cause is. Please take a look.
Thanks

JohnMcLear:
Unable to replicate this locally..  beta.etherpad.org does seem to have an issue w/ importing pdf, prolly caused by abiword service..

JohnMcLear:
Okay so TLDR;

Importing a text file breaks abiwords capability to import until service restart.

vnc-rainer-schuth:
I am also having this bug.
Related to #2782

jainendra:
Does anyone have a fix for this or any follow up?

jainendra:
@JohnMcLear @quydang Any fix or info

JohnMcLear:
I'm unable to replicate this bug on latest develop.  Plase re-open if it still exists for you.

PR #3231, opened 2017-07-28
FIX-2864: Add useAbiword flag

jainendra:
This is a fix for the issue - #2864
Which is import failure of PDF after importing a text file

### DIFF
--- a/src/node/handler/ImportHandler.js
+++ b/src/node/handler/ImportHandler.js
@@ -56,10 +56,14 @@ exports.doImport = function(req, res, padId)
     , pad
     , text
     , importHandledByPlugin
-    , directDatabaseAccess;
+    , directDatabaseAccess
+    , useAbiword;
 
   var randNum = Math.floor(Math.random()*0xFFFFFFFF);
   
+  // setting flag for whether to use abiword or not
+  useAbiword = (abiword != null);
+
   async.series([
     //save the uploaded file to /tmp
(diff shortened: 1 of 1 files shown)


jainendra:
At present if anyone tries uploading a PDF and doc will work until someone uploads a text file.
There was a logic error, abiword object was being used as a flag. abiword variable gets assigned when ImportHandler.js is loaded now when a text file gets uploaded we change the abiword variable value to false so as to not use the abiword plugin. Now on second attempt when we try to upload a PDF or a Doc uploading fails as abiword was assigned to false.

src/node/handler/ImportHandler.js
```diff
@@ -56,10 +56,15 @@ exports.doImport = function(req, res, padId)
...
     , text
     , importHandledByPlugin
-    , directDatabaseAccess;
+    , directDatabaseAccess
+    , useAbiword;
 
   var randNum = Math.floor(Math.random()*0xFFFFFFFF);
   
+  // setting flag for whether to use abiword or not
+  if(abiword)useAbiword = true;
```
Gared: Please use a shorter syntax like
```
useAbiword = abiword != null;
```

jainendra: Okay I'll update it

Gared:
Thanks for your work! @jainendra

PR #3232, opened 2017-07-31
Make APIKEY and SESSIONKEY file customizable

michael-dev:
Running multiple instances sometimes requires different api- and session-keys for security reasons.


JohnMcLear:
Docs need adding to README or so?

michael-dev:
See https://github.com/ether/etherpad-lite/pull/3360

PR #3241, opened 2017-08-04
Update socket.io to 1.7.3

djmaze:
That in turn upgrades engine.io to 1.8.2. This fixes a crash for me when running behind a traefik reverse proxy.

The error is as described here: https://github.com/socketio/socket.io/issues/2779


SvenAbels:
We are impacted by this as well (inc traefik crash)

SvenAbels:
I can confirm that this push resolves the problem.

Gared:
@djmaze Thx for this PR

thomasf:
it would be nice to have a release version with this fix.

PR #3266, opened 2017-10-16
Catch SIGTERM for graceful shutdown

### LINKED ISSUE #3265: SIGTERM not honoured
Etherpad ignores `SIGTERM`. It [processes `SIGINT`](https://github.com/ether/etherpad-lite/blob/7ecf240d9e46ac41259100bf83242ca7b8fc205b/src/node/hooks/express/errorhandling.js#L51) signal. This potentially causes an unexpected loss of data or data corruption when used with **Docker** as PID 1 and makes the shutdown unnecessarily slow.

* host os and version: https://github.com/tvelocity/dockerfiles/tree/master/etherpad-lite (Debian GNU/Linux 8 (jessie), ETHERPAD_VERSION 1.6.1, nodeJS=v7.6.0)
* steps to reproduce: `docker stop etherpad`
* what you expected to happen: Graceful shutdown
* what actually happened: Graceful shutdown time times out and container (the nodeJS process) is shut down forcefully by SIGKILL. Outgoing database connections and incoming client connections can't be gracefully closed. Internal buffers might not be written to disk (don't know whether this is an issue here, not an Etherpad insider).

C.f.
* https://github.com/nodejs/node-v0.x-archive/issues/9131
* https://joseoncode.com/2014/07/21/graceful-shutdown-in-node-dot-js/

Rillke:
Shut down database connection and exit the node process
when SIGTERM is encountered. This is especially important
when nodejs is run as PID1, e.g. in a docker container.

Shutting down connections to clients (browsers) is beyond
this patche's scope.

Resolves #3265

### DIFF
--- a/src/node/hooks/express/errorhandling.js
+++ b/src/node/hooks/express/errorhandling.js
@@ -49,5 +49,8 @@ exports.expressCreateServer = function (hook_name, args, cb) {
     //sigint is so far not working on windows
     //https://github.com/joyent/node/issues/1553
     process.on('SIGINT', exports.gracefulShutdown);
+    // when running as PID1 (e.g. in docker container)
+    // allow graceful shutdown on SIGTERM c.f. #3265
+    process.on('SIGTERM', exports.gracefulShutdown);
   }
 }


Rillke:
Note this also affects "normal" non-docker installations: When SIGTERM is received, 2 additional log lines display: "graceful shutdown ..." and "db successfully closed." Though, I think it is generally a good idea to perform graceful shutdowns, closing connections and writing buffers to disk.

I am by no means a UNIX or Etherpad professional so please double-check.

Rillke:
Agreeing to and signing Apache Contributor license.

lpagliari:
Thanks!

PR #3267, opened 2017-10-19
fix to https://github.com/ether/etherpad-lite/issues/2486

ilmartyrk:
Here is my fix to this issue, some plugins need to be updated also, soon I'll submit my fixes to them also


lpagliari:
@ilmartyrk did you find any issues with your solution?

ilmartyrk:
@lpagliari I didn't think through that returning true from each hook ended up in a value 'true true true', just added an updated to get pass that and created a new pull request

Issue #3129, opened 2017-01-25
HTML Export exports invalid HTML


renebrain:
In our project we we're able to "break" the etherpad export to html. We have a testcase put on your server under the following link: [https://beta.etherpad.org/p/bugWithExportBugBug](https://beta.etherpad.org/p/bugWithExportBugBug) 

The Magical Dom gets exported as invalid html. List Elements are floating without an existing ul's and there are some random ul elements hanging around in the html.

Is it a bug in the export oder maybe in the paste functionality? The text was pasted from outlook initially.

renebrain:
Anything?

JohnMcLear:
Confirmed.  Export HTML OL / UL is pretty wildly broken.

https://github.com/ether/etherpad-lite/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+lists

JohnMcLear:
Closed because not specific enough

Issue #2997, opened 2016-06-07
PDF / Word export doesn't work with nested lists in the end of a pad


raimann:
If you write in a pad, and the last line ist a nested list on the second level, the Export doesn't work.

I added a few screenshot to show the problem. It's written in the pad, where the export works and where not:
this works:
![bildschirmfoto 2016-06-07 um 16 35 02](https://cloud.githubusercontent.com/assets/7721945/15861870/035a4e08-2cce-11e6-99dc-32d260dfaea0.png)

this doesn't work
![bildschirmfoto 2016-06-07 um 16 34 37](https://cloud.githubusercontent.com/assets/7721945/15861869/0359dd6a-2cce-11e6-9e3c-a9d020a0cf01.png)

this works:
![bildschirmfoto 2016-06-07 um 16 34 05](https://cloud.githubusercontent.com/assets/7721945/15861868/03593f2c-2cce-11e6-9978-d7d44a5aae0a.png)

lpagliari:
Do you get an error message? Or the line is exported incorrectly?

JohnMcLear:
I have a feeling this guy is using a third party instance and isn't the site admin..

raimann:
I installed Etherpad on different servers, to use it with a LMS. I'm the site admin of these installations. ;-)
But the screenshots are from another site, not my own, where the problem also exists.

JohnMcLear:
Please export the .etherpad file so I can import and replicate.

JohnMcLear:
I can replicate this locally...

JohnMcLear:
The HTML is wrong..

```
<body>
<ul class="bullet">
        <li>hello</li>
        <ul class="bullet">
                <li>world</li>
        </ul>
        </li>
        </ul></body>
</html>
```

JohnMcLear:
This bug has existed for years..  It was introduced so far back I can't get such an old version of Etherpad to install...

JohnMcLear:
https://github.com/ether/etherpad-lite/blob/develop/src/node/utils/ExportHtml.js#L474 is the offending line.  I don't have time to dig much deeper than this but I expect a catch to see if we are still in a nested list would solve the problem.  @webzwo0i did a lot of work on this so he might be able to get a fix in quicker than me!

skupfer:
Still relevant ... Cannot export my pads. This isn't related to the last line only because the last 20 lines are text

```
[...]

Issue #2630, opened 2015-04-21
Problem in ExportHTML.js regarding creation of HTML for lists


shanukus98765:
My view in Etherpad:
![etherpad_list2](https://cloud.githubusercontent.com/assets/6191415/7251085/8a470574-e845-11e4-9512-84c83a7041f5.png)

This is the so-called HTML associated with it.
<xmp><ol class=\"number\"><li>Numbering1</li><ul class=\"bullet\"><li>bullet1</li><li>bullet2</li><li>bullet3</li></ul><li>Numbering2</li><ol class=\"number\"><li>A</li><li>This is part of A</li><li>This is again part of A</li><ul><ul class=\"indent\"><li>This is a new paragraph and it is not part of the list.</li></ul></ul></ul></ul><br>
</xmp>

The output generated thus, looks like this:
<ol class=\"number\"><li>Numbering1</li><ul class=\"bullet\"><li>bullet1</li><li>bullet2</li><li>bullet3</li></ul><li>Numbering2</li><ol class=\"number\"><li>A</li><li>This is part of A</li><li>This is again part of A</li><ul><ul class=\"indent\"><li>This is a new paragraph and it is not part of the list.</li></ul></ul></ul></ul><br>

This was given by the function "getHTMLFromAtext" within ExportHTML.js
I tried this out with multiple cases and closing tags seem to be missing from a variety of edge cases such as.
1. If you start a 'bullet list' within a 'numbered list' of 2nd level.
2. If you immediately start with an indented paragraph after finishing a list.

webzwo0i:
ups, probably I introduced/did not handle this when I changed export a little bit last year.
Thanks, I will try to reproduce this and get it fixed.

shanukus98765:
Thanks. This would be of great help. I will cross check for more edge cases.

JohnMcLear:
@webzwo0i if you follow the repo I can assign this to you I think :)

shanukus98765:
@webzwo0i: I had mentioned 2 cases where the "getHTMLFromAtext" within ExportHTML.js is generating wrong HTML output.

As far as I can see, putting indentation as a <ul> tag is causing problems. If you immediately start with an indented paragraph after finishing a list, the list tag is not being closed and an extra <ul> tag comes in for indentation.

PR #3268, opened 2017-10-20
 getLineHTMLForExport - Fixes #2486 but breaks plugins 

### LINKED ISSUE #2486: getLineHTMLForExport produces multiple copies of a line

ilmartyrk:
The key of this fix is fixing the wrong usage of `context.lineContent` and  `getLineHTMLForExport`.
Previously all `getLineHTMLForExport` hooks were expected to return lines updated by plugins and all of these lines were concatet into one string resulting mutliple lines with same text but different html wrappers.
The correct use would be to update `context.lineContent` inside each of these hooks and return true if a line was changed. Because the context is always passed on each plugin will modify the same value with each one adding its own updates rather than modifing the same line from orignal value. 

This fix will need updating each plugin using `getLineHTMLForExport` to work properly!!!

I updated my previous fix cause I found that returning true from multiple plugins made  `lineContentFromHook` value into 'true true true'. But I must add that this fix cannot work if there are updated and not updated plugins in use at the same time.


JohnMcLear:
I'd merge this but it would make too much work for me.  I'd be more inclined to merge if you can also patch my plugins for me please? :)

Gared:
Could we use ```aCallAll``` instead of ```callAll``` to make it possible for plugin developers to do db queries for example?

ilmartyrk:
Fixes html export and issues related to that:
#3129
#2997
#2630
#2486

ilmartyrk:
@Gared  I think it's ok, it doesn't change anything in this fixes behaviour so I added it

ilmartyrk:
@JohnMcLear Could you merge this now?

JohnMcLear:
@ilmartyrk can you resolve the merge conflicts please?

ilmartyrk:
Thank you @JohnMcLear !!!

muxator:
If releasing this change breaks some plugins we should probably have an upgrade strategy (for example, we could invite the authors to issue a new release for their plugin, starting with the ones for which ilmartyrk has already wrote a pull request).

Current Etherpad version is **1.6.6**. Since this is a change that somewhat breaks compatibility, I would propose to bump Etherpad version to **1.7.x** (a plugin with a sane dependency spec could make good use of this).

ilmartyrk:
@muxator I have to say that I downloaded all the plugins, that I could find from the official plugins lisy, that had html export hook and most of them were not working anyway. But I believe that 1.7.x would be the correct way to move forward because update isn't backwards compatible.

ilmartyrk:
@muxator Sorry for not reacting, now I finally created a new test that will blow up in the previous version and works fine in the updated one. This test shows how bad the final result was when two different types of lists were inserted one after another. Hope that you are good with it.

muxator:
@ilmartyrk, thank you for your work.

The new test is clear, and - as desired - breaks before #3268 (fe08d2a1db5b), and works afterwards. Perfect.

For clarity, the test we are talking about is contained in https://github.com/ether/etherpad-lite/pull/3408/commits/06cf3ffaec0eca69d765e69ef0bc8a2157e21402 and is currently part of #3408.

Some nits:

1. I would like to keep this change separate from PR #3408, which is unrelated. For example: it could be applied to the current develop head cherry-picking it to another branch, and a **new PR** created from there. If you are off put by this request, **I can do it by myself**, keeping your name & commit data. But, since it's your work, maybe you want to take care of it. **Just tell me**.

2. The info that this test (and therefore, #3268) address the case of _two different types of list inserted one after another_ is probably worth mentioning in `specs/api/pad.js`, either in the comments or in the test text itself.

3. I would normally not want to remove an existing test, but it probably makes sense to remove it this time, for the following reasons (that should probably be included in the commit message for the new test):
   - the old list test was introduced in cadb83ac5af0 with an explicit attempt to import an invalid HTML and see if it could recover from it
   - in 5967e085b7bc the code base finally managed to export a meaningful HTML (but we do not know if other bugs were introduced in pursuing this)
   - that behaviour seems aimed at attaining resiliency, but correctness should be addressed before that

muxator:
@ilmartyrk, I went on and committed 718b17552421 on your behalf (merged with b466acd8810c), so this PR is deservedly covered by a test.

Thank you again for the help!

raybellis:
The documentation for `getLineHTMLForExport ` is now out of date with respect to this change.

http://etherpad.org/doc/v1.7.0/#index_exporthtmladditionaltags

It's also unclear now how (for example) one might even fix the `ap_export_authors` plugin, since the `author.chars` property is indexed w.r.t the `.text` property of the context and not the `lineContent`.

Issue #3233, opened 2017-07-31
Improved readibility in default font size and highlight colors


joaobarcia:
Hi all, thanks for the great project. I use etherpad with several remote teams, but one of the major issues I've faced when trying to make others adopt the tool is the initial readability.

### Issue: Readability for default font settings could be improved
- font-size is 12px
- color palette seems to be generated by the array at src/node/db/AuthorManager.js

![capture1](https://user-images.githubusercontent.com/6717014/28777989-a5bdb8fa-75fd-11e7-870d-4658b5a6be80.PNG)

### Proposal: Increase font size and improve highlight colors
- change font-size to 16px (why? [source 1](https://www.smashingmagazine.com/2011/10/16-pixels-body-copy-anything-less-costly-mistake/), [source 2](https://ux.stackexchange.com/questions/211/is-there-an-optimal-font-size), [source 3](https://www.imarc.com/blog/best-font-size-for-any-device))
- change colour palette so that all colors have contrast >15 and lightness >80% (i.e. http://jxnblk.com/colorable/demos/text/?background=%2312d1ad&foreground=%23000000 to http://jxnblk.com/colorable/demos/text/?background=%239FF9E9&foreground=%23000000)

![capture2](https://user-images.githubusercontent.com/6717014/28778663-96b78f36-7600-11e7-9de8-5973ae11d6d6.PNG)

---

Happy to create a PR if the community agrees with the proposal.

Cheers

jainendra:
I agree with this, currently etherpad doesn't have good readability

Dhwty:
I feel like this sort of a change should be optional -- something that can be enabled in the settings -- rather than the only choice. I say this because I work with a low-resolution screen and a dark theme, and this would be nightmarish for my ability to read my documents.

jainendra:
I agree with @Dhwty. @joaobarcia Why don't you do something like @Dhwty mentioned and raise a PR

joaobarcia:
Great, will try to find some time to work on it the coming week.

@Dhwty , just to make sure I come up with a good proposal it would be nice to understand your user case as well. Why do you think it would impact your ability to read documents? The letters get too big in a low resolution screen? And about the dark theme, is it the theme of your os/browser, or does etherpad have themes that I didn't know of? And how would the highlights affect your readibility?

Thanks

Dhwty:
[...]

PR #3280, opened 2017-11-23
Improved readibility

lmagniez:
In order to solve the issue of having a good readibility on etherpad-lite, I changed as it was requested the set of user color (I made sure every color have a sufficient brightness), and changed the pad's font-size to 16px. 
Issue's link: [issue](https://github.com/ether/etherpad-lite/issues/3233)

### DIFF
--- a/src/node/db/AuthorManager.js
+++ b/src/node/db/AuthorManager.js
@@ -25,7 +25,7 @@ var customError = require("../utils/customError");
 var randomString = require('ep_etherpad-lite/static/js/pad_utils').randomString;
 
 exports.getColorPalette = function(){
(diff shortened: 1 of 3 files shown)


lpagliari:
Awesome work, @lmagniez !! Thanks a lot for it!

@Dhwty does this PR impact your work, as you've mentioned on the original issue?

I'm still feeling these should be features enabled by props on `settings.json` -- one for the colors, one for the font-size / line-height... @lmagniez could you do that, please? The default style would be 12px / 16px, and the customised one 16px / 21px -- I'm trying to keep the same ratio for both configs here, not sure if this is the best approach. Thoughts?

If we use 16px / 16px, we won't have the spacing between lines, which make reading more comfortable.

### 12px / 16px:
image

### 16px / 16px:
image

### 16px / 21px:
image

Dhwty:
> I'm still feeling these should be features enabled by props on settings.json -- one for the colors, one for the font-size / line-height...

If this change goes into settings.json, rather than being a hard-coded change in font size and colour, it's not going to impact me at all. Thanks for the @.

lpagliari:
we can define on `settings.json.template` that the improved styles would be used, and let other people avoid them if they want. At least we make more people to use the improvements of @lmagniez without they needing to do anything when starting their Etherpad instance for the first time

vvalembois:
Sorry for the late reply, and thank you for your suggestions. We will work on it with @lmagniez and @lisonferez when we will be free.

lpagliari:
Thanks, @vvalembois . This will be an excellent improvement for Etherpad!

JohnMcLear:
I really dislike this PR.  Sorry!

slel:
Thaanks!!

JohnMcLear:
Had a number of complaints about this already...

1. Line height wasn't bumped to match font-size
2. I increased line height this has broken numbering and formatting.

rohieb:
As a short feedback to this PR, I have reverted it and related changes on my local instance for now, since I find 16pt too big to read and I cannot turn the font size back to 12pt otherwise except hard-reverting it. However I found that only on my Hi-DPI laptop, on the desktop it looked fine…

muxator:
This change effectively breaks the alignment in Etherpad 1.6.4+ (see for example #3378).

As the feature seems interesting, maybe we can revert the changes on iframe_editor.css (I have them ready to commit) and open a separate line of work where these parameters would be configurable, as suggested by @lpagliari and others.

muxator:
Reverted back to old appearance with 9daade0b95bb.

We can build a more comprehensive approach in a different line of work.

fungi:
Any chance to get a new release for this sometime soonish? We're trying to stick to released versions in production deployments, but have gotten a number of complaints about the font size and line number misalignment since upgrading to deal with the recent vulnerabilities.

muxator:
Released in [1.6.6](https://github.com/ether/etherpad-lite/releases/tag/1.6.6).

PR #3282, opened 2017-11-28
Add settings to scroll on edition out of viewport 

joassouza:
By default, when there is an edition of a line, which is out of the
viewport, Etherpad scrolls the minimum necessary to make this line
visible. This makes that the line stays either on the top or the bottom
of the viewport. With this PR, we add a setting to make possible to
scroll to a position x% pixels from the viewport. Besides of that, we
add a setting to make an animation of this scroll and allow to scroll if the 
caret is at the last line of the viewport. Both of them are optional.
If nothing is changed on settings.json the Etherpad default behavior is
kept.
_Etherpad Default_

_Google Docs_

_Microsoft Word_

This PR has been used/tested for one year or so in production. 
We've tested this PR on differents browsers versions, some of them
screen shot 2017-11-28 at 11 09 45 am

ps: When there is a selection and user either presses arrow keys or does an edition, the default behavior of browser is kept.

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -2924,18 +2924,22 @@ function Ace2Inner(){
       // the viewport, but it is not in the bottom of the viewport, so it should not scroll.
       // This case only happens when it has a plugin of pagination, like ep_page_view
       var edgeLinesVisibleOnViewport = getVisibleLineRange();
-      var lastLineCompletelyVisibleOfViewport = edgeLinesVisibleOnViewport[1];
+      var lastLineVisibleOfViewport = edgeLinesVisibleOnViewport[1];
 
-      // it is possible a line is partially visible and user places the caret in this line
-      var caretIsInThelastLineVisibleOfViewport = rep.selEnd[0] === lastLineCompletelyVisibleOfViewport;
-      var caretIsInThelastLinePartiallyVisibleOfViewport = rep.selEnd[0] === lastLineCompletelyVisibleOfViewport + 1;
-      var caretIsInTheLastLineOfViewport = caretIsInThelastLineVisibleOfViewport || caretIsInThelastLinePartiallyVisibleOfViewport;
-      var linesOfPad = rep.lines.length();
-      var lastLineVisibleIsInTheBottomOfViewport = linesOfPad > lastLineCompletelyVisibleOfViewport + 1;
+      // it is possible that the last line visible is partially visible and user places
+      // the caret at the line before of this line
+      var caretIsInThelastLineVisibleOfViewport = rep.selEnd[0] === lastLineVisibleOfViewport;
+      var caretIsInThelastButOneLineVisibleOfViewport = rep.selEnd[0] === lastLineVisibleOfViewport - 1;
+      var caretIsInTheLastLineOfViewport = caretIsInThelastLineVisibleOfViewport || caretIsInThelastButOneLineVisibleOfViewport;
+
+      // avoid scrolling when caret is at the last line of the pad but this line is not in the bottom
+      // of the pad. E.g. when pad has only one line and caret is on that line.
+      var lastLineIsBelowOfTheBottomOfViewport = lastLineOfPadIsBelowOfViewportBottom(rep);
 
       // avoid scrolling to the end of pad, when user press shortcut to select all (Cmd + A)
-      var hasSelection =  rep.selStart[0] !== rep.selEnd[0];
-      if(caretIsInTheLastLineOfViewport && lastLineVisibleIsInTheBottomOfViewport && !hasSelection)
+      var hasSelection = rep.selStart[0] !== rep.selEnd[0];
+
+      if(caretIsInTheLastLineOfViewport && lastLineIsBelowOfTheBottomOfViewport && !hasSelection)
       {
 
         var win = outerWin;
@@ -2951,6 +2955,16 @@ function Ace2Inner(){
     }
     return false;
     //console.log("%o %o %s", rep.selStart, rep.selEnd, rep.selFocusAtStart);
+  }
+
+  function lastLineOfPadIsBelowOfViewportBottom(rep){
+    var lastLineNumber = rep.lines.length() - 1;
+    var lastLine = rep.lines.atIndex(lastLineNumber);
+    var lastLinePosition = getLineEntryTopBottom(lastLine);
+    var viewportHeight = getInnerHeight() - getEditorPositionTop();
+    var lastLineIsBelowOfTheBottomOfViewport = lastLinePosition.bottom > viewportHeight;
+
+    return lastLineIsBelowOfTheBottomOfViewport;
   }
 
   function doCreateDomLine(nonEmpty)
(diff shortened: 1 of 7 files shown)


src/static/js/caretPosition.js
```diff
@@ -0,0 +1,236 @@
...
+
+  if (node.length === 0){
+    while (node.length === 0 && node.nextSibling) {
+      node = node.nextSibling;
+    }
+  }
+
+  var newRange = new Range();
+  newRange.setStart(node, offset);
+  newRange.setEnd(node, offset);
```
lpagliari: Is it possible that the offset would be inside a more internal node on the DOM? For example:

```
<span id='outer'>
  123
  <span id='inner'>
    456
  </span>
  789
</span>
```

`node` = `#outer`
`offset` = `5`

What would happen in this case?

src/static/js/caretPosition.js
```diff
@@ -0,0 +1,236 @@
...
+  return previousLineTop;
+}
+
+function caretLineIsFirstBrowserLine(caretLineTop, rep)
+{
+  var caretRepLine = rep.selStart[0];
+  var lineNode = rep.lines.atIndex(caretRepLine).lineNode;
+  var firstRootNode = getFirstRootChildNode(lineNode);
+
+  // to get the position of the node we the position of the first char
```
lpagliari: need to review this comment, there's a word missing here

src/static/js/caretPosition.js
```diff
@@ -0,0 +1,236 @@
...
+    height: rect.height,
+    top: rect.top
+  }
+  return linePosition;
+}
+
+// here we have two possibilities:
+// [1] the line before the caret line has the same type, so both of them has the same margin, padding
+// height, etc. So, we can use the caret line to make calculation necessary to know where is the top
+// of the previous line
```
lpagliari: you are assuming that only line attributes change CSS styles in a more drastic way (margins, paddings, height, etc). This is not necessarily true:

image

Is your code handling this scenario?

src/static/js/caretPosition.js
```diff
@@ -0,0 +1,236 @@
...
+    return getLastRootChildNode(node.lastChild);
+  }
+}
+
+// here we have two possibilities:
+// [1] The next line is part of the same rep line of the caret line, so we have the same dimensions.
+// So, we can use the caret line to calculate the bottom of the line.
+// [2] the next line is part of another rep line. It's possible this line has different dimensions, so we
+// have to get the exactly dimension of it
+exports.getBottomOfNextBrowserLine = function(caretLinePosition, rep)
```
lpagliari: same note of `getPositionTopOfPreviousBrowserLine`

src/static/js/ace2_inner.js
```diff
@@ -2922,6 +2931,66 @@ function Ace2Inner(){
...
+  // Some plugins might set a minimum height to the editor (ex: ep_page_view), so checking
+  // if (caretLine() === rep.lines.length() - 1) is not enough. We need to check if there are
+  // other lines after caretLine(), and all of them are out of viewport.
+  function isCaretAtTheBottomOfViewport()
+  {
+    // computing a line position using getBoundingClientRect() is expensive.
+    // (obs: getBoundingClientRect() is called on caretPosition.getPosition())
+    // To avoid that, we only call this function when it is possible that the
+    // caret is in the bottom of viewport
+    var caretLine = rep.selStart[0];
```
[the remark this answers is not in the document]
joassouza: this function is called in 3 cases: 
When user presses arrow up, arrow down or make an edition. In these three cases we don't keep the selection `(rep.selStart !== rep.selEnd)`

src/static/js/ace2_inner.js
```diff
@@ -4018,51 +4115,34 @@ function Ace2Inner(){
...
+              // rep line on the top of the viewport
+              var arrowUpWasPressedInTheFirstLineOfTheViewport = percentageScrollArrowUp && arrowUp && isCaretAtTheTopOfViewport();
+              if(arrowUpWasPressedInTheFirstLineOfTheViewport){
+                var win = outerWin;
+                var pixelsToScroll = getPixelsToScrollWhenUserPressesArrowUp();
+                // by default, the browser scrolls to the middle of the pad. To avoid the twist made
+                // when we apply a second scroll, we made it immediately (without animation)
+                scrollYPageWithoutAnimation(win, -pixelsToScroll);
+              }else{
+                scrollNodeVerticallyIntoView();
```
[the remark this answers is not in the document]
joassouza: I totally agree but this is code it was already there

src/static/js/caretPosition.js
joassouza: You'd get an error. This function is called twice
`getPositionOfRepLineAtOffset(firstRootChildNode, 1)`
`getPositionOfRepLineAtOffset(lastRootChildNode.node, lastRootChildNode.length)`
In both cases is not possible to get the inner node as the example.

src/static/js/caretPosition.js
joassouza: actually it's not the same but it's quite similar

joassouza:
To make clear we use two definitions of line: browser line and rep line. To make clear, this is representation of it.
image

src/static/js/ace2_inner.js
```diff
@@ -4838,8 +4835,8 @@ function Ace2Inner(){
       }
     }
     // if near edge, scroll to edge
-    var scrollX = getScrollX();
-    var scrollY = getScrollY();
+    var scrollX = scroll.getScrollX();
+    var scrollY = scroll.getScrollY();
```
lpagliari: are these used anywhere?

src/static/js/caretPosition.js
joassouza: It does. We consider the height of the current line to validate if it has space to another line of the same height. When it changes the height of a part (e.g. a word, as the example) it changes the height of the all line.

src/static/js/scroll.js
```diff
@@ -0,0 +1,351 @@
+/*
```
lpagliari: AWESOME! Much better having all the code related to scrolling on a separated file 🎉

src/static/js/ace2_inner.js
```diff
@@ -4014,55 +4017,36 @@ function Ace2Inner(){
...
+              // rep line on the top of the viewport
+              var innerHeight = getInnerHeight();
+              if(scroll.arrowUpWasPressedInTheFirstLineOfTheViewport(arrowUp, rep)){
+                var pixelsToScroll = scroll.getPixelsToScrollWhenUserPressesArrowUp(innerHeight);
+
+                // by default, the browser scrolls to the middle of the viewport. To avoid the twist made
+                // when we apply a second scroll, we made it immediately (without animation)
+                scroll.scrollYPageWithoutAnimation(-pixelsToScroll);
+              }else{
+                scroll.scrollNodeVerticallyIntoView(rep, innerHeight);
```
lpagliari: Interesting: this entire `if`-`else` block is calling functions from `scroll` module. It looks like something that would make more sense if extracted into a function of it, isn't it?

src/static/js/caretPosition.js
lpagliari: awesome! 👍

src/static/js/ace2_inner.js
joassouza: good idea!

PR #3301, opened 2018-01-02
Select formatting button on selection

joassouza:
Now when the caret or selection is in a text with formatting, we select the style button applied.
jan-02-2018 18-32-18

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -75,6 +75,9 @@ function Ace2Inner(){
   var EDIT_BODY_PADDING_TOP = 8;
   var EDIT_BODY_PADDING_LEFT = 8;
 
+  var FORMATTING_STYLES = ['bold', 'italic', 'underline', 'strikethrough'];
+  var SELECT_BUTTON_CLASS = 'selected';
+
   var caughtErrors = [];
 
   var thisAuthor = '';
@@ -2911,6 +2914,9 @@ function Ace2Inner(){
       rep.selFocusAtStart = newSelFocusAtStart;
       currentCallStack.repChanged = true;
 
+      // select the formatting buttons when there is the style applied on selection
+      selectFormattingButtonIfLineHasStyleApplied(rep);
+
       hooks.callAll('aceSelectionChanged', {
         rep: rep,
         callstack: currentCallStack,
@@ -2937,6 +2943,24 @@ function Ace2Inner(){
   function isPadLoading(eventType)
   {
     return (eventType === 'setup') || (eventType === 'setBaseText') || (eventType === 'importText');
+  }
+
+  function selectFormattingButtonIfLineHasStyleApplied (rep) {
+    _.each(FORMATTING_STYLES, function (style) {
+      var hasStyleOnRepSelection = hasAttributeOnSelectionOrCaretPosition(rep, style);
+      var $formattingButton = parent.parent.$('[data-key="' + style + '"]').find('a');
+      $formattingButton.toggleClass(SELECT_BUTTON_CLASS, hasStyleOnRepSelection);
+    })
+  }
+
+  function hasAttributeOnSelectionOrCaretPosition(rep, attributeName) {
+    var hasNotSelection = (rep.selStart[0] == rep.selEnd[0] && rep.selEnd[1] === rep.selStart[1]);
+    var hasAttrib = getAttributeOnSelection(attributeName);
+    if (hasNotSelection) {
+      var attributesOnCaretPosition = documentAttributeManager.getAttributesOnCaret();
+      hasAttrib = _.contains(_.flatten(attributesOnCaretPosition), attributeName);
+    }
+    return hasAttrib;
   }
 
   function doCreateDomLine(nonEmpty)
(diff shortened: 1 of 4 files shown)


src/static/js/ace2_inner.js
```diff
@@ -2922,6 +2930,36 @@ function Ace2Inner(){
...
+  }
+
+  function selectFormattingButtonIfLineHasStyleApplied (rep) {
+    _.each(FORMATTING_STYLES, function (style) {
+      var hasStyleOnRepSelection = hasAttributeOnSelectionOrCaretPosition(rep, style);
+      ifAttribIsFormattingUpdateStyleButtonState(style, hasStyleOnRepSelection);
+    })
+  }
+
+  function hasAttributeOnSelectionOrCaretPosition(rep, attributeName) {
```
lpagliari: this looks like a function that should be on `AttributeManager`

src/static/js/ace2_inner.js
```diff
@@ -2922,6 +2930,36 @@ function Ace2Inner(){
     //console.log("%o %o %s", rep.selStart, rep.selEnd, rep.selFocusAtStart);
   }
 
+  function ifAttribIsFormattingUpdateStyleButtonState(attribName, hasStyleOnRepSelection) {
+    if (attributeIsFormattingStyle(attribName)) {
```
lpagliari: this checking is duplicated when `ifAttribIsFormattingUpdateStyleButtonState()` is called from `selectFormattingButtonIfLineHasStyleApplied` -- we are looping through the `FORMATTING_STYLES`, so `attribName` surely is a formatting style. To avoid this, you could turn `ifAttribIsFormattingUpdateStyleButtonState` into a function that simply toggles the class on the button, and extract the checking (`if (attributeIsFormattingStyle(attribName))`) to the other place where it is called -- on `toggleAttributeOnSelection`.

src/static/js/ace2_inner.js
```diff
@@ -2471,12 +2474,14 @@ function Ace2Inner(){
...
     if (selectionAllHasIt)
     {
+      ifAttribIsFormattingUpdateStyleButtonState(attributeName, false); // italic, bold, ...
       documentAttributeManager.setAttributesOnRange(rep.selStart, rep.selEnd, [
         [attributeName, '']
       ]);
     }
     else
     {
+      ifAttribIsFormattingUpdateStyleButtonState(attributeName, true); // italic, bold, ...
```
lpagliari: these two calls of `ifAttribIsFormattingUpdateStyleButtonState` can be simplified into a single one, passing `! selectionAllHasIt` as the second parameter:
```
    if (selectionAllHasIt)
    {
      documentAttributeManager.setAttributesOnRange(rep.selStart, rep.selEnd, [
        [attributeName, '']
      ]);
    }
    else
    {
      documentAttributeManager.setAttributesOnRange(rep.selStart, rep.selEnd, [
        [attributeName, 'true']
      ]);
    }
    ifAttribIsFormattingUpdateStyleButtonState(attributeName, !selectionAllHasIt); // italic, bold, ...
```

lpagliari: BTW, it is even possible to simplify the code even more:
```
var attributeValue = selectionAllHasIt ? '' : 'true';
documentAttributeManager.setAttributesOnRange(rep.selStart, rep.selEnd, [[attributeName, attributeValue]]);
ifAttribIsFormattingUpdateStyleButtonState(attributeName, !selectionAllHasIt); // italic, bold, ...
```

src/static/js/ace2_inner.js
```diff
@@ -2472,20 +2472,10 @@ function Ace2Inner(){
...
-    {
-      ifAttribIsFormattingUpdateStyleButtonState(attributeName, true); // italic, bold, ...
-      documentAttributeManager.setAttributesOnRange(rep.selStart, rep.selEnd, [
-        [attributeName, 'true']
-      ]);
-    }
+
+    var attributeValue = selectionAllHasIt ? '' : 'true';
+    documentAttributeManager.setAttributesOnRange(rep.selStart, rep.selEnd, [[attributeName, attributeValue]]);
+    if (attribIsFormattingStyle(attributeName)) updateStyleButtonStateWhenAttribIsFormatting(attributeName, !selectionAllHasIt); // italic, bold, ...
```
lpagliari: styling: can you use curly braces on the `if` as we have on the other `if`s on this module, please?

src/static/js/ace2_inner.js
```diff
@@ -2930,36 +2920,22 @@ function Ace2Inner(){
     //console.log("%o %o %s", rep.selStart, rep.selEnd, rep.selFocusAtStart);
   }
 
-  function ifAttribIsFormattingUpdateStyleButtonState(attribName, hasStyleOnRepSelection) {
-    if (attributeIsFormattingStyle(attribName)) {
-      var $formattingButton = parent.parent.$('[data-key="' + attribName + '"]').find('a');
-      $formattingButton.toggleClass(SELECT_BUTTON_CLASS, hasStyleOnRepSelection);
-    }
+  function updateStyleButtonStateWhenAttribIsFormatting(attribName, hasStyleOnRepSelection) {
```
lpagliari: this function needs to be renamed too, now you don't check if attrib is a formatting one :)

lpagliari:
Thanks, @joassouza ! 🎉

Issue #3299, opened 2018-01-01
unix socket activation


dvzrv:
This project could benefit from implementing [unix domain socket](https://en.wikipedia.org/wiki/Unix_domain_socket) activation, alongside the provided IP socket based setup.
Unix sockets are particularly useful, if etherpad-lite is run on the same machine as the web server (which very often is the case). Instead of communicating with a TCP/IP socket, a unix network socket is used, involving less overhead and context switches.

This approach could be settable in the configuration file (to be used instead of IP socket).

In such a scenario, instead of redirecting user traffic to the local TCP/IP port, it is pointed towards the inode representing the socket (e.g. `/run/etherpad-lite/etherpad-lite.sock`).
Etherpad-lite would be started by some other process accessing the socket and not started as a fixed system job.

Someone on the [AUR](https://aur.archlinux.org/) has created [a patched version of etherpad-lite](https://aur.archlinux.org/cgit/aur.git/tree/socket-activation.patch?h=etherpad-lite-systemd) a while ago.
I'm just wondering, whether it's really that easy...

rhelmer:
This does look really easy, with the `systemd` and `autoquit` modules.

The only added complexity really is making it configurable, but that shouldn't be too hard.

I don't have time to do this myself right now but happy to mentor or review, looks like a really useful feature.

PR #3312, opened 2018-01-30
Unix domain socket support.

iia:
Based on this [issue](https://github.com/ether/etherpad-lite/issues/3299) this PR tries to add a new feature with which Etherpad Lite can be reached locally with a UNIX domain socket and also enable Systemd socket activation. The implementation adds a new dependency to the **autoquit** module.

To test the implementation:

1. Enable the **unixSocketActivation** configuration option.

2. Add Systemd service and socket files,

- Service File (/etc/systemd/system/etherpad-lite.service)
```
[Unit]
Description=Etherpad Lite, the collaborative editor.
After=syslog.target network.target
Requires=etherpad-lite.socket

[Service]
Type=simple
User=<USER>
Group=<GROUP>
WorkingDirectory=<INSTALL_DIRECTORY>
ExecStart=<NODEJS_BIN_PATH> <INSTALL_DIRECTORY>/src/node/server.js
StandardOutput=syslog
StandardError=syslog
SyslogIdentifier=etherpad-lite
Restart=no

[Install]
WantedBy=multi-user.target
```

- Socket File (/etc/systemd/system/etherpad-lite.socket)
```
[Unit]
Description=Etherpad Lite socket for UNIX socket activation.
PartOf=etherpad-lite.service

[Socket]
ListenStream=/tmp/etherpad-lite.socket
Accept=False

[Install]
WantedBy=sockets.target
```

3. Enable the socket with the following commands,

```
sudo systemctl --system daemon-reload
```
```
sudo systemctl start etherpad-lite.socket
```

4. Test Etherpad Lite functionalities,

- HTTP GET request

```
curl --unix-socket /tmp/etherpad-lite.socket -X GET http:/
```

- HTTP REST API: Create a pad

```
curl --unix-socket /tmp/etherpad-lite.socket \
-d '{"apikey":"<API_KEY>", "padID":"<PAD_ID>", "text": "<TEXT>"}' \
-H "Content-Type: application/json" \
-X POST http:/api/1/createPad
```

- HTTP REST API: Set text of the created pad

```
curl --unix-socket /tmp/etherpad-lite.socket \
-d '{"apikey":"<API_KEY>", "padID":"<PAD_ID>", "text": "<TEXT>"}' \
-H "Content-Type: application/json" \
-X POST http:/api/1/setText
```

- HTTP REST API: Get text from the created pad

```
curl --unix-socket /tmp/etherpad-lite.socket \
-d '{"apikey":"<API_KEY>", "padID":"<PAD_ID>"}' \
-H "Content-Type: application/json" \
-X POST http:/api/1/getText
```

### DIFF
--- a/src/node/hooks/express.js
+++ b/src/node/hooks/express.js
@@ -5,17 +5,11 @@ var fs = require('fs');
 var path = require('path');
 var npm = require("npm/lib/npm.js");
 var  _ = require("underscore");
-var errorHandling = require("./express/errorhandling");
+require('systemd');
 require('autoquit');
 
 var server;
 var serverName;
-
-/*
- * Check here,
- * https://cgit.freedesktop.org/systemd/systemd/tree/src/systemd/sd-daemon.h
- */
-var systemdListenFDStart = 3;
 
 exports.createServer = function () {
   console.log("Report bugs at https://github.com/ether/etherpad-lite/issues")
@@ -23,21 +17,6 @@ exports.createServer = function () {
   serverName = "Etherpad " + settings.getGitCommit() + " (http://etherpad.org)";
   
   console.log("Your Etherpad version is " + settings.getEpVersion() + " (" + settings.getGitCommit() + ")");
-
-  if(settings.unixSocketActivation) {
-    if(!parseInt(process.env.LISTEN_FDS, 10)) {
-      console.error('No listen file descriptors provided by Systemd. It seems like Etherpad was not invoked by Systemd.');
-      errorHandling.gracefulShutdown();
-    }
-
-    if (parseInt(process.env.LISTEN_FDS, 10) < 1) {
-      console.error('No listen file descriptors provided by Systemd.');
-      errorHandling.gracefulShutdown();
-    } else if (parseInt(process.env.LISTEN_FDS, 10) > 1) {
-      console.error('Too many file descriptors provided by Systemd. Only one file descriptor expected.');
-      errorHandling.gracefulShutdown();
-    }
-  }
 
   exports.restartServer();
 
@@ -133,6 +112,6 @@ exports.restartServer = function () {
      * { timeOut: <TIMEOUT_IN_SECONDS> }
      */
     server.autoQuit();
-    server.listen({fd: systemdListenFDStart});
+    server.listen('systemd');
   }
 }
--- a/src/node/hooks/express.js
+++ b/src/node/hooks/express.js
@@ -5,9 +5,17 @@ var fs = require('fs');
 var path = require('path');
 var npm = require("npm/lib/npm.js");
 var  _ = require("underscore");
+var errorHandling = require("./express/errorhandling");
+require('autoquit');
 
 var server;
 var serverName;
+
+/*
+ * Check here,
+ * https://cgit.freedesktop.org/systemd/systemd/tree/src/systemd/sd-daemon.h
+ */
+var systemdListenFDStart = 3;
 
 exports.createServer = function () {
   console.log("Report bugs at https://github.com/ether/etherpad-lite/issues")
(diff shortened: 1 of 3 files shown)


lpagliari:
@rhelmer can you take a look on this, please?

JohnMcLear:
Bump @rhelmer

muxator:
Hi,

we are already _unknowingly_ "supporting" unix sockets. A user found a minor bug because of this: #3437.

This configuration snippet actually results in etherpad listening to a unix socket:
```javascript
// settings.json
{
  "ip": "",                               // <-- empty string
  "port" : "/somepath/etherpad.socket",   // <-- path to a Unix socket
  // ...
}
```

This is due to [`express.listen()`](http://expressjs.com/en/4x/api.html#app.listen_path_callback), which in turn just [re-exposes `net.server.listen()`](https://nodejs.org/api/net.html#net_server_listen) and its overloads, including the one that binds on a Unix socket.

This PR makes it **explicit**, and this is certainly desirable.
Let me propose what should be added before merging:

## Important
1. [ ] suitable examples for binding Etherpad to a Unix socket should be added in `settings.json.template`
2. [ ] the timeout for `autoQuit()` should be configurable in `settings.json` (and documented in `settings.json.template`)
3. [ ] if one does not want to use automatic quitting, a value of `0` for the timeout should completely bypass the feature (since, [reading autoquit's source](https://github.com/rubenv/node-autoquit/blob/master/lib/autoquit.js), there is no way of starting autoquit in a disabled state)
4. [ ] once this PR is ready, a last commit should be added to make stricter validation of `settings.ip` and `settings.port` in order to make mandatory the use of the new configuration options (making impossible the bypass described above).

## Nice to have
1. [ ] the examples with service and socket files are really useful, and worth putting in a documentation file (maybe a new one)?
2. [ ] The [wiki about reverse-proxying](https://github.com/ether/etherpad-lite/wiki/How-to-put-Etherpad-Lite-behind-a-reverse-Proxy) is worth at least an example about using Etherpad as an upstream bound to a unix socket.

muxator:
This would be a good moment to work on this PR, since we are at the start of a new cycle.

iia:
Hi @muxator,

Already looking into the requests.

Cheers.

muxator:
Hi @iia, give a voice if you need help.

iia:
Hi,

Sorry for the late reply. While trying to follow the instructions of this PR on the branch of this PR I couldn't make the API POST to work. GET seems to work OK.

I'm getting a, **405: Only the HEAD or GET methods are allowed.**

I suspect it has something to do with the order of things in ep.json hooks and how they are loaded.

Any ideas?

Cheers.

muxator:
I have a doubt now: this change is systemd-specific.

Would Etherpad still run if executed in an environment without systemd? For Example, on Windows, or on Alpine Linux, of in a container (which has no notion of PID 1).

Considering [we are already unknowingly supporting unix domain sockets](https://github.com/ether/etherpad-lite/pull/3312#issuecomment-412340758), I am starting to think if a different approach could make sense:

1. document the possibility of using unix domain socket on supported platforms, without introducing a new configuration parameter
2. deduce at runtime (from the value of the listen config parameter) if we are binding to a network or unix socket
3. keep the helpful message logs at startup which indicate in which mode Etherpad is running
4. discuss the useful autoquit functionality in a different PR

**n.b.**: this is not a comment about systemd, but about platform support

JohnMcLear:
-1 Autoquit.

PR coming for changes requested.

muxator:
We are good to go with the already existing Unix socket support.

Merging #3790, which receives [my considerations above](https://github.com/ether/etherpad-lite/pull/3312#issuecomment-412340758), and closing this.

PR #3314, opened 2018-01-31
better sanitize jsonp

rhelmer:
r? @JohnMcLear


src/node/hooks/express/apicalls.js
```diff
@@ -18,7 +18,7 @@ var apiCaller = function(req, res, fields) {
     apiLogger.info("RESPONSE, " + req.params.func + ", " + response);
 
     //is this a jsonp call, if yes, add the function call
-    if(req.query.jsonp)
+    if(req.query.jsonp && isVarName(response))
```
marcelklehr: Shouldn't this be rquire'd, too? :D (👋 )

rhelmer: Indeed :) My local testing apparently wasn't good enough and our automated tests didn't catch it.

PR #3318, opened 2018-02-04
Don't send COMMIT-MESSAGE when socketio connection is not active

anugu-chegg:
**BUG:**
If socket.io connection is lost and user is entering some input in the pad at the same time, then pad becomes unresponsive and shows a 'Disconnect modal' after few sec even after the socket.io connection is reestablished. 
+
Data entered in the pad during the loss of connection won't be sent to the server

**SOLUTION:**
Check if socket.io connection is active before sending any COMMIT-MESSAGE to the server. If not, wait for few sec before retrying to send the COMMIT-MESSAGE to the server.

### DIFF
--- a/src/static/js/collab_client.js
+++ b/src/static/js/collab_client.js
@@ -181,20 +181,26 @@ function getCollabClient(ace2editor, serverVars, initialUserInfo, options, _pad)
     }
 
     var sentMessage = false;
-    var userChangesData = editor.prepareUserChangeset();
-    if (userChangesData.changeset)
-    {
-      lastCommitTime = t;
-      state = "COMMITTING";
-      stateMessage = {
-        type: "USER_CHANGES",
-        baseRev: rev,
-        changeset: userChangesData.changeset,
-        apool: userChangesData.apool
-      };
-      sendMessage(stateMessage);
-      sentMessage = true;
-      callbacks.onInternalAction("commitPerformed");
+    if (getSocket().realConnected) {
+        var userChangesData = editor.prepareUserChangeset();
+        if (userChangesData.changeset)
+        {
+          lastCommitTime = t;
+          state = "COMMITTING";
+          stateMessage = {
+            type: "USER_CHANGES",
+            baseRev: rev,
+            changeset: userChangesData.changeset,
+            apool: userChangesData.apool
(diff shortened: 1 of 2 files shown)


JohnMcLear:
@lpagliari can you test this please?

muxator:
@anugu-chegg, thanks for analyzing the problem and for taking the time of writing a fix.

I have done a very basic test of your patchset (everything on `localhost:9001`) cutting the connection via `tcpkill -9 -i any port 9001`, and **the robustness really improved greatly**!

In a single case, however, I was _still_ able to lose changes: instead of describing it, I would like to show you, but unfortunately I was not recording the screen at the time. I will try again in the following days.

Coping with these king of failures and simulating them is really challenging: any help will be appreciated in trying to replicate the remaining issue (if it's really still there).

muxator:
I was finally able to break it while recording. This is the patched version going under attack :)

short-small

anugu-chegg:
@muxator The reason they were out of sync is because the pad on the right did not reconnect. Can you try this once again and you might want to look at the browser's console for errors if there are any.

As for testing, you can use this command `while :; do sleep 5; kill -9 $(lsof -i:9001 | awk '$1 == "node" { print $2 }'| uniq); bin/run.sh --root; done` which restarts etherpad server after 5s. During the restart you need to add as much text as possible into both the pads.

muxator:
Using the `kill -9` approach I did not encounter any issues. The code works flawlessly and definitely needs to be merged.

With the more brutal approach (`tcpkill -9 -i any port 9001`) it is relatively easy to break a browser's connectivity and make it stop trying. I would like to see some testing (and future patches) on this.

One last thing, @anugu-chegg: it seems that [socket.realConnected](https://github.com/ether/etherpad-lite/pull/3318/files#diff-eeb5b9a97d3e8bdc960c1edab4afa376R205) is a leftover of previous commits. Could you please confirm, and eventually take the time to polish the patch series?

anugu-chegg:
@muxator Yes, you are right. I have removed the leftover code from previous commits.

muxator:
Released in [1.6.6](https://github.com/ether/etherpad-lite/releases/tag/1.6.6).

PR #3326, opened 2018-02-12
Update ueberdb dep to 0.3.6

ldidry:
Ueberdb2 0.3.6 has been released, introducing support for new databases (crate.io, rethinkdb), connections pools for PostgreSQL and have some dependencies updates.

Etherpad should use the new version of ueberdb2.


lpagliari:
Thanks!!

PR #3330, opened 2018-02-25
use apt instead of apt-get in the README

sevu:


lpagliari:
Doesn't `apt-get install` work the same way of `apt install`?

sevu:
It does. I thought it's the more modern way to just write apt.

PR #3331, opened 2018-02-25
specify python version in shebang

sevu:
This would fail e.g. on Archlinux otherwise


lpagliari:
@sevu I'm not familiar with python, but what if I have python v3 instead of v2?

sevu:
Then the script won't work because the print statement is a function in v3

0ip:
Better use env (`#!/usr/bin/env PYTHONUNBUFFERED=1 python2`)  these days :-)

PR #3336, opened 2018-03-07
Etherpad in cluster mode

anugu-chegg:


JohnMcLear:
Would break a lot of things, needs a rewrite to not be redis specific

PR #3338, opened 2018-03-08
Fix .doc export with LibreOffice (soffice)

ldidry:
When using LibreOffice to convert pads to doc, we got `Error: no export filter for /tmp/xxxx.doc` (tested with LO 5 and 6). Maybe it's a regression from LO. Anyway, converting HTML to odt, then to doc works.

### DIFF
--- a/src/node/utils/LibreOffice.js
+++ b/src/node/utils/LibreOffice.js
@@ -35,7 +35,21 @@ var queue = async.queue(doConvertTask, 1);
  * @param  {Function}   callback    Standard callback function
  */
 exports.convertFile = function(srcFile, destFile, type, callback) {
-  queue.push({"srcFile": srcFile, "destFile": destFile, "type": type, "callback": callback});
+  // soffice can't convert from html to doc directly (verified with LO 5 and 6)
+  // we need to convert to odt first, then to doc
+  // to avoid `Error: no export filter for /tmp/xxxx.doc` error
+  if (type === 'doc') {
+    queue.push({
+      "srcFile": srcFile,
+      "destFile": destFile.replace(/\.doc$/, '.odt'),
+      "type": 'odt',
+      "callback": function () {
+        queue.push({"srcFile": srcFile.replace(/\.html$/, '.odt'), "destFile": destFile, "type": type, "callback": callback});
+      }
+    });
+  } else {
+    queue.push({"srcFile": srcFile, "destFile": destFile, "type": type, "callback": callback});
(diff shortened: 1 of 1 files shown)


src/node/utils/LibreOffice.js
```diff
@@ -35,7 +35,18 @@ var queue = async.queue(doConvertTask, 1);
  * @param  {Function}   callback    Standard callback function
  */
 exports.convertFile = function(srcFile, destFile, type, callback) {
-  queue.push({"srcFile": srcFile, "destFile": destFile, "type": type, "callback": callback});
+  if (type === 'doc') {
+    queue.push({
+      "srcFile": srcFile,
+      "destFile": destFile.replace(/\.doc/, '.odt'),
```
lpagliari: I don't know if `srcFile` could be something like `"test.html.html"`, but in this case this `replace` would not work as you're expecting. You can use the `$` regex mark of EOL to fix that:
```
"test.html.html".replace(/\.html/, '.odt')
> "test.odt.html"
"test.html.html".replace(/\.html$/, '.odt')
> "test.html.odt"
```

src/node/utils/LibreOffice.js
```diff
@@ -35,7 +35,18 @@ var queue = async.queue(doConvertTask, 1);
...
+      "type": 'odt',
+      "callback": function () {
+        queue.push({"srcFile": srcFile.replace(/\.html/, '.odt'), "destFile": destFile, "type": type, "callback": callback});
```
lpagliari: same here

src/node/utils/LibreOffice.js
```diff
@@ -35,7 +35,18 @@ var queue = async.queue(doConvertTask, 1);
  * @param  {Function}   callback    Standard callback function
  */
 exports.convertFile = function(srcFile, destFile, type, callback) {
-  queue.push({"srcFile": srcFile, "destFile": destFile, "type": type, "callback": callback});
+  if (type === 'doc') {
```
lpagliari: as this is a very specific case, it would be good to have some comments explaining the reason we have this `if/else`. Would you mind adding it here, @ldidry ? You can add a reference to this PR, it would make it easier for other devs to know why this code needed to be added.

ldidry:
Thank you for your suggestions, they're indeed very good :slightly_smiling_face: 
(and now integrated)

lpagliari:
Thanks, @ldidry !

PR #3342, opened 2018-03-09
Add checkPad variant to check all pads/revisions

caugner:


caugner:
It would be ideal to extract common parts of `checkPad.js` and `checkAllPads.js` to `src/node/utils`.

PR #3352, opened 2018-03-25
Fix line number display when plugin add padding-top (like ep_page_view)

ldidry:
Line numbering display was broken when using page view since https://github.com/ether/etherpad-lite/commit/a3765d97853ecfe3fa608cd3b191a33cabe311ec.

The first number took the line height plus 100px, due to ep_page_view padding-top CSS rule.

This fixes the problem.


renebrain:
For us non pluign user it screws the entire line count, so we removed that line locally on our side. I guess there are errors getting thrown so ethepad goes back to default line heights/margins/Paddings

PR #3360, opened 2018-04-04
Document --credentials, --apikey and --sessionkey; remove -k (duplicate)

michael-dev:
This pull requests adds some basic documentation for --credentials, --apikey and --sessionkey.
It removes the -k option, as it is currently used for both apikey and sessionkey.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ If cloning to a subdirectory within another project, you may need to do the foll
 # Next Steps
 
 ## Tweak the settings
-You can initially modify the settings in `settings.json`. (If you need to handle multiple settings files, you can pass the path to a settings file to `bin/run.sh` using the `-s|--settings` option. This allows you to run multiple Etherpad instances from the same installation.)  Once you have access to your /admin section settings can be modified through the web browser.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -70,7 +70,7 @@ If cloning to a subdirectory within another project, you may need to do the foll
 # Next Steps
 
 ## Tweak the settings
-You can initially modify the settings in `settings.json`. (If you need to handle multiple settings files, you can pass the path to a settings file to `bin/run.sh` using the `-s|--settings` option. This allows you to run multiple Etherpad instances from the same installation.)  Once you have access to your /admin section settings can be modified through the web browser.
+You can initially modify the settings in `settings.json`. (If you need to handle multiple settings files, you can pass the path to a settings file to `bin/run.sh` using the `-s|--settings` option. This allows you to run multiple Etherpad instances from the same installation. Similarly '--credentials' can be used to give an settings override file, '--apikey' to give a different APIKEY.txt file and '--sessionkey' to give a non-default SESSIONKEY.txt.)  Once you have access to your /admin section settings can be modified through the web browser.
```
nashe: Can we stick to using ` around arguments, as in setting's example?

src/node/utils/Cli.js
```diff
@@ -40,12 +40,12 @@ for ( var i = 0; i < argv.length; i++ ) {
   }
 
   // Override location of settings.json file
-  if ( prevArg == '--sessionkey' || prevArg == '-k' ) {
```
muxator: Removing an argument ruins backward compatibility.
Was this argument so broken that removing it is not going to break any installation?

I am 100% ok merging the doc changes, but would like some details on this change.

PR #3364, opened 2018-04-05
Fix line-wrapping in WebKit

### LINKED ISSUE #3270: Bad line-wrapping in WebKit browsers
The line

```js
$$INCLUDE_CSS("../static/css/iframe_editor.css");
```

from ace2.js, adds the style `whitespace: nowrap` to `body`. In Blink-based browsers, this style gets overridden by the `.noprewrap` class, but for whatever reason this class is not applied when viewed from Safari, WebKit Nightly, or presumably Epiphany or other similar browsers. Effectively, this makes pads unreadable from an iphone.

Surely there's no good reason to add `whitespace: nowrap` to `body` for the entire document, this only  makes sense from a text editor rendered in an iframe (and even then... does it really make sense?)

caitp:
Re-add `white-space: pre-wrap;` to `body.doesWrap` (This possibly breaks older versions of Mozilla browsers, but makes etherpad-lite documents more readable on iOS devices).

Fixes #3270

### DIFF
--- a/src/static/css/iframe_editor.css
+++ b/src/static/css/iframe_editor.css
@@ -43,7 +43,7 @@ body.grayedout { background-color: #eee !important }
 }
 
 body.doesWrap {
-  /* white-space: pre-wrap; */
+  white-space: pre-wrap;
 
   /*
     Must be pre-wrap to keep trailing spaces. Otherwise you get a zombie caret,


JohnMcLear:
@muxator can't we just close this?

PR #3384, opened 2018-04-23
WIP: Linkable lines

CsatariGergely:
I would like to add linkable lines with this PR. It is work in progress, but I would like to ask some advice. 

- Linking works for me when I click the numbers, but it does not work when I copy a link and paste it to a new window. This is a bit disturbing as the main use case would be to send these links in chats or emails.
- The height of the lines is somehow not calculated correctly, but this seems to be problematic with the currect development branch also

Any help is appreciated.

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -5369,7 +5369,7 @@ function Ace2Inner(){
   function initLineNumbers()
   {
     lineNumbersShown = 1;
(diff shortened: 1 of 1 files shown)


JohnMcLear:
I swear I implemented this as a plugin but I can't find it else I would refer to that..

JohnMcLear:
Boom, I knew I wrote this: https://github.com/JohnMcLear/ep_scrollto

PR #3391, opened 2018-05-10
Update settings.json.template

Wikinaut:
Harmonizing the database name.

"store" is the table name - not to be used here!

Database name must not contain "-", but can contain "_".

See https://github.com/ether/etherpad-lite/wiki/How-to-use-Etherpad-Lite-with-MySQL

### DIFF
--- a/settings.json.template
+++ b/settings.json.template
@@ -45,12 +45,14 @@
                  },
 
   /* An Example of MySQL Configuration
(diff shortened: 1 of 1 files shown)


muxator:
Merged, thanks

Wikinaut:
There will be more coming, as I resumed my work on epl.

PR #3407, opened 2018-06-02
Correctly display plugin list in troubleshooting tab

### LINKED ISSUE #3402: Plugin list is not correctly displayed
The template `templates/admin/plugins-info.html` calls `replace` to display one plugin per line:
https://github.com/ether/etherpad-lite/blob/f289611ba93a45cd7141f849c7ae896048f9861b/src/templates/admin/plugins-info.html#L30

However, this function replaces only the __first__ occurrence of `, `:
etherpad_replace

nashe:
It aims to fix #3402.


muxator:
Merged, thanks.

PR #3408, opened 2018-06-05
added goToRevisionEvent hook, regexp support for pre and post values in ep.json - WONT MERGE

ilmartyrk:
Added this hook to run custom functions in timeline view. For example currently ep_resize plugin does not work in that view and some of the content is displayed outside of the frame. That happens when during the editing process text has been longer than it is in the final version. Fixed htmlExport test: https://github.com/ether/etherpad-lite/issues/3425


ilmartyrk:
Added basic regexp support to ep.json post and pre values. That means that if a plugin should be run after all other plugins then, it can be done by using:  `pre: ["ep_*"]` . If there is a conflict for example `pre: ["ep_*", post: ["ep_etherpad-lite/static"]` then value that is stricter is used in that case all other plugins will be loaded before except for ´ep_etherpad-lite/static´

muxator:
I would prefer to have PRs submitted as a linear series (== without merges and _sync_ commits) on top of current develop.

I understand this is a new requirement, still without an explicit guideline.

The reason for which I am going to ask this is that bisecting a bug when the history is full of spurious commits becomes extremely slow, and impairs maintainability in the long run.

muxator:
As promised, here is some [documentation for pull requests](https://github.com/ether/etherpad-lite/blob/develop/CONTRIBUTING.md#pull-requests).

Drop a line if not clear / need help.

ilmartyrk:
@muxator, thank you

muxator:
If we want to merge this PR, it should be rebased on current revision.

JohnMcLear:
Note that the plugin work can't be merged due to conflicts with the async changes.

I will try pick out what I can.

JohnMcLear:
Both of these contributions are now in separate PRs so this can be closed.

Issue #3383, opened 2018-04-19
Etherpad cannot handle its own shortcut keys with Firefox Nightly


masayuki-nakano:
Firefox is now trying to conform to the latest keypress event spec, that is, keypress event should be dispatched only when a key press produces a character input (e.g., typing 'KeyA', 'Space', etc) and Enter key press (Enter, Shift+Enter and Control+Enter for compatibility both with backward and the other browsers). I.e., 'Tab', Shift + 'Space', Ctrl + 'A' etc won't cause keypress event even on Firefox too. You can test this with setting "dom.keyboardevent.keypress.dispatch_non_printable_keys_only_system_group_in_content" in about:config to true on Firefox 60 or later (60 is currently Beta). If you use Firefox Nightly, you also need to ensure that your Etherpad URI is not included in "dom.keyboardevent.keypress.hack.dispatch_non_printable_keys".

Then, you can test this with the following steps:

0. Download Firefox Nightly or Firefox Beta, install it and launch it. <https://www.mozilla.org/en-US/firefox/channel/desktop/>
1. Open an Etherpad.
2. Type someting.
3. Type 'Tab' key.

Then, the paragraph should be indented.  However, nothing happens.

We will include major public Etherpad URLs into the blacklist. So, be careful at testing.

Bug at Mozilla: <https://bugzilla.mozilla.org/show_bug.cgi?id=1455059>

muxator:
Thanks for reporting this [before May 9](https://developer.mozilla.org/en-US/Firefox/Releases/60).

This will not just apply to `tab` key, but also to other combinations (like `ctrl`-`b` for bold), right?

Could you please point to some documentation about how to keep up with this change? I guess a lot of products will need to handle this.

masayuki-nakano:
> This will not just apply to tab key, but also to other combinations (like ctrl-b for bold), right?

Right.

> Could you please point to some documentation about how to keep up with this change? I guess a lot of products will need to handle this.

Here is an unofficial document:
https://www.fxsitecompat.com/en-CA/docs/2018/non-printable-keys-will-soon-stop-firing-keypress-event/

There are some points:

- Firefox will stop dispatching "keypress" event when user presses non-printable keys (e.g., 'Tab', 'ArrowLeft', 'Escape', etc) and non-printable key combinations (e.g., Ctrl+'B', etc) except Enter/Shift + Enter/Ctrl + Enter for compatibility with backward and the other browsers.
[...]

PR #3413, opened 2018-06-28
Use keydown instead of keypress on Firefox.

manta12:
This related with issue #3383.

Firefox will not fire the keypress when target key is the non-printable keys. (like Enter / Tab / Delete... etc) For detail, see https://bugzilla.mozilla.org/show_bug.cgi?id=1440189.

This behavior is same as safari / chrome, So etherpad-lite will not need to special handling of these key when UA is Firefox. This PR will use keydown event when target key is no-printable keys.

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -3708,8 +3708,8 @@ function Ace2Inner(){
       return; // This stops double enters in Opera but double Tabs still show on single tab keypress, adding keyCode == 9 to this doesn't help as the event is fired twice
     }
     var specialHandled = false;
-    var isTypeForSpecialKey = ((browser.msie || browser.safari || browser.chrome) ? (type == "keydown") : (type == "keypress"));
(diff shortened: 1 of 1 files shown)


muxator:
Hi @manta12, thanks for this PR.

I would like to confirm if this change would indeed be compatible with currently supported Firefox versions.
Before merging this, would you please check it on Firefox ESR 52 and 60 (if not already done)?

Ideally this should be a task for automated testing, or a QA team, but I am not sure we have someone in charge.

Thank you again.

muxator:
Merged, thanks.

Fixes #3383.

muxator:
I just noted that this PR (merged in 2be873e3c708) breaks the frontend tests (`http://<yourhost>/tests/frontend/`), because they were full of special cases that sent `keypress` when executed in Firefox:

```js
    if(inner$(window)[0].bowser.firefox || inner$(window)[0].bowser.modernIE){ // if it's a mozilla or IE
      var evtType = "keypress";
    }else{
      var evtType = "keydown";
    }
```

Taking out the special check for Firefox goes back to 100% success rate for frontend tests in Firefox **61** (when applying only over this change).

PR #3437, opened 2018-07-28
Fix misparse of port when binding Unix socket

muhmuhten:
The hostname:port of URIs used in Minify are currently bogus and refer to localhost only for historical reasons; there's no reason to retain them and omitting them avoids generating an invalid URI when "port" is not an integer.

For context: settings.port is passed to express's `listen`, which treats it as a filename for a Unix domain socket if it doesn't look like a port number. This is useful for starting a server to be reverse-proxied on a multi-user system without requiring fuss around port numbers or authentication.

Currently, etherpad-lite runs without error when configured to listen on a Unix domain socket; however, when a pad is accessed, the client JS fails with "Uncaught Error: The module at "ep_etherpad-lite/static/js/rjquery" does not exist." due to `pad.js` and `ace2_common.js` being generated incorrectly:

when the port is actually a path e.g. `"etherpad.sock"`, the generated URIs are of the form `"http://localhost:etherpad.sock/static/js/rjquery.js"`. Accordingly, the file searched for is `:etherpad.sock/static/js/rjquery.js`, rather than the expected `static/js/rjquery.js`. Since such a file doesn't exist, the required code doesn't get included in the bundle.

A workaround is to remove the reference to the port. In my patch, I've also replaced extraneous references to `localhost` with `invalid.invalid` to hopefully save future readers a wild goose chase looking the requests being made over the loopback interface before understanding the significance of the comment on `requestURIs`...

### DIFF
--- a/src/node/hooks/express/static.js
+++ b/src/node/hooks/express/static.js
@@ -17,11 +17,12 @@ exports.expressCreateServer = function (hook_name, args, cb) {
 
   // Setup middleware that will package JavaScript files served by minify for
   // CommonJS loader on the client-side.
+  // Hostname "invalid.invalid" is a dummy value to allow parsing as a URI.
   var jsServer = new (Yajsml.Server)({
     rootPath: 'javascripts/src/'
-  , rootURI: 'http://localhost:' + settings.port + '/static/js/'
+  , rootURI: 'http://invalid.invalid/static/js/'
   , libraryPath: 'javascripts/lib/'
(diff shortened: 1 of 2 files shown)


muhmuhten:
Yes, that's essentially the configuration I'm using.

I've amended the commit to include the detailed explanation in the PR in the commit message and added a brief comment at both "invalid.invalid" sites.

muxator:
Great! Pulled in, thanks.

PR #3473, opened 2018-08-24
Add --writer option to soffice convert command

ldidry:
If you edit `src/templates/export_html.html` to remove the `<meta name="changedby" content="Etherpad">` tag[1], PDF export with soffice has a bug: the first word of the pad is deleted and a blank page is inserted as first page (the pad's text begins on the second page).
The `--writer` soffice option avoids that bug.

[1] you may want to delete that tag since it is inserted as a comment in .doc or .odt soffice export.

### DIFF
--- a/src/node/utils/LibreOffice.js
+++ b/src/node/utils/LibreOffice.js
@@ -63,6 +63,7 @@ function doConvertTask(task, callback) {
         '--invisible',
         '--nologo',
         '--nolockcheck',
+        '--writer',
         '--convert-to', task.type,
         task.srcFile,
         '--outdir', tmpDir


muxator:
Neat. Thanks!

BTW: I was not able to find the .odt or .doc comment caused by `<meta name="changedby" content="Etherpad">`. Could you help me (in another issue)?

ldidry:
@muxator I used Centos7, Etherpad 1.6.5, LibreOffice 5.2 (or 5.3) and read the output with LibreOffice 6.1, but also reproduced on Debian 9, Etherpad 1.6.5, LibreOffice 5.2 (or 5.3).
I think you need to enable "Track Changes" (in menu: "View" > "Track Changes") to see it.

Wikinaut:
I confirm, that the patch solves the issue.

PR #3474, opened 2018-08-25
installDeps.sh: don't nuke all installed plugins on update

rohieb:
When this script is run after an update, and something fails during the npm
run, the installed plugins will also be deleted. This is probably not what is
intended here.

### DIFF
--- a/bin/installDeps.sh
+++ b/bin/installDeps.sh
@@ -116,7 +116,6 @@ echo "Ensure that all dependencies are up to date...  If this is the first time 
   cd ep_etherpad-lite
   npm install --no-save --loglevel warn
 ) || {
-  rm -rf node_modules
   exit 1
 }
 


muxator:
Hi @rohieb, thanks for the remark. I believe there is a subtler bug here.

My impression is that the original script author wanted to delete:
* `/<repository_root>/node_modules/ep_etherpad-lite/node_modules` (the one containing the runtime dependency of the application)

and ended up deleting this instead:
* `/<repository_root>/node_modules` (the one containing both the application AND the plugins)
probably because he got caught in a problem with bash subshell scoping.

Let me explain, and please correct me if I am wrong. In the following example, the author possibly tought that the error handling code would start from the last directory he moved in:
```bash
# this starts in <repository_root>
echo "Ensure that all dependencies are up to date..."
(
  mkdir -p node_modules
  cd node_modules
  [ -e ep_etherpad-lite ] || ln -s ../src ep_etherpad-lite
  cd ep_etherpad-lite
  false # simulate a failure of "npm install"
) || {
  # will print:
  #   <repository_root>
  # whereas the author _possibly_ intended:
  #   <repository_root>/node_modules/ep_etherpad-lite
  echo "I am here: $(pwd)"
  exit 1
}
```

Whereas upon error the old scope is thrown away and the cwd becomes the original one.

A shrunk-down example would be:
```bash
#!/usr/bin/env sh

cd /etc
(
  cd /var
  false
) || {
  echo "I am here: $(pwd)" # prints "/etc" and not "/var"
}
```

If so, I think that a proper fix would be replacing `rm -rf node_modules` with `rm -rf src/node_modules`.

Tell me what you think.

rohieb:
Ah, that could posisbly be the case. I'm not very experienced in Node modules, is there a reason to delete `<repo>/node_modules/ep_etherpad-lite/node_modules` instead? If so, this subtler bug should be corrected instead of not deleting anything.

anoymouserver:
Instead of deleting the whole `ep_etherpad-lite/node_modules` folder, it would probably be better to run [`npm prune`](https://docs.npmjs.com/cli/prune) in `ep_etherpad-lite`, which deletes all packages that are not in the `package.json`.

rohieb:
Just tried it for confirmation, plugins are installed in `<repo>/node_modules`, not in `<repo>/src/node_modules`.
@muxator: sounds good to me, feel free to just push to master instead :)

muxator:
Committed the modified version in c09e638fc324, closing this.

Thanks.

PR #3476, opened 2018-08-27
travis-ci: updated nodejs version

anoymouserver:
should fix the current error
```
$ bin/installDeps.sh
Your npm version "2.15.1" is too old. npm 3.10.x or higher is required.

The command "bin/installDeps.sh" failed and exited with 1 during .
```


muxator:
Merged, thanks

PR #3477, opened 2018-08-27
url encode pad name

anoymouserver:
URL encode entered pad names to allow special characters, which are currently causing some unexpected behaviour.

Some examples:
`test.html?q=abc` - currently creates a pad with the name `test.html`
`test.html/abc` - results in the error `Cannot GET /p/test.html/abc`


muxator:
Merged, thanks.

Issue #3435, opened 2018-07-28
Test authorship_of_editions.js complains


muxator:
#3420 fixed an UI bug and introduced a test to verify it.

Problem is, I have difficulties running successfully the test in my environment.

| Component | Version |
|-------|-------|
|Etherpad | 58c315476954 (1.6.6+)|
| Browser | Firefox **61** & Chromium **67** |
| Node | **6.9.0** (should not matter) |
| Url for replicating | http://localhost:9001/tests/frontend/?grep=authorship_of_editions.js |

This screencast shows an example of execution.
What puzzles me is that there are two authors in the initial screen, and after reload the modifications ("x", "y", "z" at position 2,2) are not from another author.

![test-bug](https://user-images.githubusercontent.com/4067621/43351500-368b1b5c-9213-11e8-8348-2e9a234b7f2a.gif)

Supposing I have understood correctly the test case, when driven manually Etherpad behaves well:  authorship colors are maintained. This seems a problem in the test code, if at all.

@lpagliari, what is wrong with me?

lpagliari:
sorry, @muxator, I didn't have time to take a look on this. Will do so and let you know.

muxator:
No problem @lpagliari, thank you!

muxator:
Hi, Luiza.

Since this was not a blocker and **1.7** was due, I released it. Hopefully we will make it for the next point release. :)

lpagliari:
@muxator I can confirm the test does not work on Firefox on Mac too -- although it passes on Chrome and on all browsers on our Travis CI (not sure why...).

Anyway, I'm working on a fix right now, will submit a PR soon.

lpagliari:
Some details about the issue: the test on `authorship_of_editions.js` tried to make changes with 2 different authors, and then checked if each change was assigned to the correct user. The author switch was made by reloading the pad -- which on Chrome was enough to set a different author.

The error happened on Firefox because the even after reloading the pad the author id was not reset. Weirdly the tests on this browser passed on Travis CI.

The fix was to force the cookie expiration, which in turn would force a new author id to be set.

PR #3480, opened 2018-09-01
tests: make sure author is switched on authorship test

lpagliari:
Original implementation of authorship test assumed a simple pad reload would switch the author -- and it actually did on Chrome. But other browsers might keep the author id on cookie.

To force the author switch, expire the cookie before reloading the pad.

Fixes: #3435


muxator:
Merged, thanks

PR #3482, opened 2018-09-12
unable to convert with running soffice process

P-D-:
needed to add this parameter, because soffice already ran in headless mode for a different purpose
found here https://stackoverflow.com/questions/30349542/command-libreoffice-headless-convert-to-pdf-test-docx-outdir-pdf-is-not
Env: Ubuntu Server 14.04.5
LibreOffice 4.2.8.2 420m0(Build:2)

### DIFF
--- a/src/node/utils/LibreOffice.js
+++ b/src/node/utils/LibreOffice.js
@@ -63,6 +63,7 @@ function doConvertTask(task, callback) {
         '--invisible',
         '--nologo',
         '--nolockcheck',
+        '-env:UserInstallation=file:///tmp/LibreOffice_Conversion_${USER}',
         '--writer',
         '--convert-to', task.type,
         task.srcFile,
--- a/src/node/utils/LibreOffice.js
+++ b/src/node/utils/LibreOffice.js
@@ -60,10 +60,10 @@ function doConvertTask(task, callback) {
     function(callback) {
       var soffice = spawn(settings.soffice, [
         '--headless',
-        '--invisible',
-        '--nologo',
-        '--nolockcheck',
-        '-env:UserInstallation=file:///tmp/LibreOffice_Conversion_${USER}',
+       //'--invisible',
+        //'--nologo',
+        //'--nolockcheck',
+        '-env:UserInstallation=file://'+tmpDir+'/LibreOffice_Conversion_${USER}',
         '--writer',
         '--convert-to', task.type,
         task.srcFile,


muxator:
Thanks for finding this out.

But now I went from zero to three doubts:

1. `/tmp` is hardcoded. It should probably be [os.tmpdir()](https://nodejs.org/api/os.html#os_os_tmpdir)
2. the [stack overflow question](https://stackoverflow.com/questions/30349542/command-libreoffice-headless-convert-to-pdf-test-docx-outdir-pdf-is-not) you linked was in the context of a bash shell, in which `${USER}` would be expanded to the current user name. I am not sure that this would still be the case in the js code. Could you please verify that your patch does not create a literal `/tmp/LibreOffice_Conversion_${USER}` instead of `/tmp/LibreOffice_Conversion_yourname`?
3. at this point, why not compute a unique dir name for `env:UserInstallation` at startup time using something that behaves like the [mktemp utility](https://www.gnu.org/software/autogen/mktemp.html) in order to be sure that different Etherpad instances would not collide? I see that [the `tmp` package in npm](https://www.npmjs.com/package/tmp) could be useful. This would both be more general and get us rid of points 1 & 2.

P-D-:
Thanks for your answer,

1. changed the patch. Use now os.tmpdir()
2. `${user}` creates `/tmp/LibreOffice_Conversion_myname`, works on Linux, but maybe not with M$
3. As far as I see, this would be the best. I encounter the same problem with other plugins that uses soffice (ep_export_documents_images (the problem with soffice was solved there already in the same way, as I saw in the meantime), ep_export_cp_html_image and ep_document_import_hook). Anyway I don't know much about nodejs. I had an error message which I could use to find the problem. I know a bit js and that's it. Sorry, but I think this will not be "my task". I could test some patches and give feedback, but thats it...

muxator:
Let's try to address this more generally in #3508.

Leaving the PR open for now, because a part of it will be implemented for sure (we will want to launch a dedicated LibreOffice instance per each Etherpad instance via `-env:UserInstallation=file://<somepath>`).

JohnMcLear:
Just trying to understand this.  Is this PR suggesting libreoffice can only convert/render one file at once?  My understanding was that you can have multiple exports happening at the same time or is that not the case?

JohnMcLear:
Bump @P-D-

P-D-:
Hi,

the problem seems to be, that libreoffice is already running for a different task. Conversion fails because tmp-file is used by the "other" libreoffice. 

Greetings Peter

JohnMcLear:
@P-D- I tried to replicate by spamming export and couldn't.  Do you have steps to replicate?

JohnMcLear:
Bump @P-D-

P-D-:
Hi John,
the problem (as far as I could recall it) is not the conversion of more than one document. The problem is, that if there is a libreoffice already running, for what ever reason, conversion fails because it is blocked somehow. 
Maybe this is resolved in the meantime but I can not check this right now.
Greetings Peter

JohnMcLear:
Does this impact ALL users or just if you are running libreoffice as the same user running Etherpad?

P-D-:
I think just users that run libreoffice and etherpad....

Am 14.05.20 um 12:09 schrieb John McLear:
> Does this impact ALL users or just if you are running libreoffice as the same user running Etherpad?
>

PR #3495, opened 2018-10-06
installDeps: no need to download jquery at starting time. Remove curl…

### LINKED ISSUE #3494: Do we still need to download jquery on start?
In `installDeps.sh` we have a bunch of code (from 2011) that checks for the existence of jQuery and eventually downloads it:

https://github.com/ether/etherpad-lite/blob/4121add1b8b2babfb71363b3deeb254c5aa3f05f/bin/installDeps.sh#L138-L140

In the readme, we also [list curl as a requirement](https://github.com/ether/etherpad-lite/tree/4121add1b8b2babfb71363b3deeb254c5aa3f05f#uber-quick-ubuntu).

Question is, since 1239ce7f2848 (year 2012) jquery is vendorized in `/src/static/js/jquery.js`.

Unless some old-time user has a better historical memory, I am going to remove it from the code & the requirements in the doc.

muxator:
… dependency

In installDeps.sh we have a bunch of code (from 2011) that checks for the
existence of jQuery and eventually downloads it, but since 1239ce7f2848 (year
2012) jquery is vendorized in /src/static/js/jquery.js.

Also, stop mentioning curl as a dependency in the README.md

Fixes #3494

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,10 +19,10 @@ git clone https://github.com/ether/etherpad-lite.git && cd etherpad-lite && bin/
 ```
 
 ## GNU/Linux and other UNIX-like systems
-You'll need gzip, git, curl, libssl develop libraries, python and gcc.  
-- *For Debian/Ubuntu*: `apt install gzip git curl python libssl-dev pkg-config build-essential`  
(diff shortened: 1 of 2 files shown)


muxator:
I do not see any potential regressions.
I'll go ahead and merge this.

PR #3507, opened 2018-10-30
Allow to import files using soffice

ldidry:
Currently, you have to use Abiword to convert odt, pdf and doc files to be able to import those files.
This add soffice as convertor too.

Replace https://github.com/ether/etherpad-lite/pull/3339 (I can't update that PR anymore).

### DIFF
--- a/src/node/handler/ImportHandler.js
+++ b/src/node/handler/ImportHandler.js
@@ -34,9 +34,18 @@ var ERR = require("async-stacktrace")
   , log4js = require("log4js")
   , hooks = require("ep_etherpad-lite/static/js/pluginfw/hooks.js");
 
-//load abiword only if its enabled
-if(settings.abiword != null)
-  var abiword = require("../utils/Abiword");
+var convertor = null;
+var exportExtension = "htm";
+
+//load abiword only if its enabled and if soffice is disabled
+if(settings.abiword != null && settings.soffice === null)
+  convertor = require("../utils/Abiword");
+
+//load soffice only if its enabled
+if(settings.soffice != null) {
+  convertor = require("../utils/LibreOffice");
+  exportExtension = "html";
+}
 
 //for node 0.6 compatibily, os.tmpDir() only works from 0.8
 var tmpDirectory = process.env.TEMP || process.env.TMPDIR || process.env.TMP || '/tmp';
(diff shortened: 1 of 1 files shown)


muxator:
Pulled, thanks.

While testing it, I have found another problem: if there is already another copy of LibreOffice running, the converted file is not created when LibreOffice exits, but a bit later.

Thus, the conversion functions generates an error:

```
Converting Error: { [Error: ENOENT: no such file or directory, rename '/tmp/upload_8e9203784bd79d7ca9bb251396bede41.html' -> '/tmp/etherpad_import_3723561530.html']
  errno: -2,
  code: 'ENOENT',
  syscall: 'rename',
  path: '/tmp/upload_8e9203784bd79d7ca9bb251396bede41.html',
  dest: '/tmp/etherpad_import_3723561530.html' }
Error: ENOENT: no such file or directory, rename '/tmp/upload_8e9203784bd79d7ca9bb251396bede41.html' -> '/tmp/etherpad_import_3723561530.html'
```

ldidry:
Do you have a similar behavior when exporting multiple files with LO? (I think you should, but it's better to be sure)

It seems to be caused by the [queue system](https://github.com/ldidry/etherpad-lite/blob/8ed921b4ac57737508eb49b433bcf9a22f09cd9c/src/node/utils/LibreOffice.js#L26) used to prevent system overload (since the LO utils spawns a new LO process each time it's used).

We may use [unoconv](https://github.com/dagwieers/unoconv) to mimic Abiword behavior (one listening process, and use the same process for convertions), like I did in [ep_unoconv](https://framagit.org/framasoft/Etherpad/ep_unoconv/) (abandonned plugin since some time after I wrote it, LO became native in Etherpad).

What do you think?

muxator:
> Do you have a similar behavior when exporting multiple files with LO?

How do I do this? Not via the web interface, I suppose. Any hints?

> It seems to be caused by the queue system used to prevent system overload (since the LO utils spawns a new LO process each time it's used).

It's possible, I have done some debugging and posted my impressions in #3508.

> We may use unoconv to mimic Abiword behavior (one listening process, and use the same process for convertions)

We should definitely use a single LibreOffice process, and not launch one for each conversion.
I am not sure about pulling in another dependency (even when optional), because I think that we could find a way of doing this directly. My 2 cents in #3508.

In the meantime, thanks for this much needed improvement :)

ldidry:
> How do I do this? Not via the web interface, I suppose. Any hints?

Set libreoffice in `settings.json`, then export two pads in odt, doc or pdf format at the same time (with curl maybe? Easier to do it at the same time than switching tab and press another button). But it does not really matter, that was just for curiosity.

PR #3509, opened 2018-11-02
Try/catch to skip invalid JSON in dirtyDB

ahzf:
I noticed, that a dirtyDB can end up having invalid lines of JSON. Sadly this will also break this little clean'up tool. The patch just ignored invalid JSON.


muxator:
Merged, thanks

Issue #3516, opened 2018-11-23
Generate "unique" pad names


drebs:
Hello, and thanks for the great piece of software!

Currently, etherpad-lite relies on the user's browser to generate a random pad name. The random function currently used is `Math.random()` which [does not provide cryptographically secure random numbers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random) and the generated pad name is not long enough (10 chars would lead to ~60-bit security if a PRNG was used). Because of those, the current solution is not safe against collisions.

If the expected behavior is to generate unique pad names, and that is delegated to the user's browser (i.e. there's no server-side check to ensure the generated name does not collide), i suggest the following:

* use a PRNG instead of `Math.random()` and ensure uniform distribution when selecting chars.
* choose the name length to achieve a specific number of bits of security.

The following is a suggestion:

```diff
diff --git a/src/templates/index.html b/src/templates/index.html
index 872367c9..f5dd4468 100644
--- a/src/templates/index.html
+++ b/src/templates/index.html
@@ -187,12 +187,22 @@
 
             function randomPadName()
             {
-                var chars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
-                var string_length = 10;
+                // the number of distinct chars (64) is chosen to ensure that
+                // the selection will be uniform when using the PRNG below
+                var chars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-+";
+                // the length of the pad name is chosen to get 120-bit security:
+                // log2(64^20) = 120
+                var string_length = 20;
+                // make room for 8-bit integer values that span from 0 to 255.
+                var randomarray = new Uint8Array(string_length);
+                // use browser's PRNG to generate a "unique" sequence
+                window.crypto.getRandomValues(randomarray);
                 var randomstring = '';
                 for (var i = 0; i < string_length; i++)
                 {
-                    var rnum = Math.floor(Math.random() * chars.length);
+                    // instead of writing "Math.floor(randomarray[i]/256*64)"
+                    // we can save some cycles.
+                    var rnum = Math.floor(randomarray[i]/4);
                     randomstring += chars.substring(rnum, rnum + 1);
                 }
                 return randomstring;
```

Might be related (not enough info on the motivation of the request): #3418

[...]

PR #3517, opened 2018-11-23
index.html: generate unique pad names

drebs:
Etherpad-lite relies on the user's browser to generate a random pad
name, but the current solution is not safe against collisions. In order
to generate unique pad names, the following modifications are made:

* use a PRNG instead of Math.random() and ensure uniform distribution
  when selecting chars.

* choose the pad name length to achieve a specific number of bits of
  security.

Closes: #3516

### DIFF
--- a/src/templates/index.html
+++ b/src/templates/index.html
@@ -187,12 +187,23 @@
 
             function randomPadName()
             {
-                var chars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
-                var string_length = 10;
+                // the number of distinct chars (64) is chosen to ensure that
+                // the selection will be uniform when using the PRNG below
+                var chars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-_";
+                // the length of the pad name is chosen to get 120-bit security:
+                // log2(64^20) = 120
+                var string_length = 20;
+                // make room for 8-bit integer values that span from 0 to 255.
+                var randomarray = new Uint8Array(string_length);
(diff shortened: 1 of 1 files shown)


src/templates/index.html
```diff
@@ -187,12 +187,22 @@
 
             function randomPadName()
             {
-                var chars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
-                var string_length = 10;
+                // the number of distinct chars (64) is chosen to ensure that
+                // the selection will be uniform when using the PRNG below
+                var chars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-+";
```
muxator: The `+` is a URL reserved character.
According to [RFC 3986](https://tools.ietf.org/html/rfc3986#section-2.3) the set of unreserved characters is:

```
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
```

We should replace `+` with one of (`.`, `_`, `~`).
Should we go with `_`?

muxator:
The original commit message of this PR mentioned that this change closes #3519 (which does not exist).

Supposing it is a typo (should really have been #3516) I have updated the commit message and the comment in this issue.

Merging after this small edit, thanks.

muxator:
@Wikinaut: I understand the preoccupation with generating human-friendly sequences (or more precisely, non adversarial). If we find a more elegant solution that is really easy on the eye, I will be ok with merging that one.

In this specific case, a randomly generated sequence is probably meant to be copied & pasted anyhow, and this gives me confidence to merge this PR, which is an improvement in itself.

Leaving open the ticket in case we want to comment.

Issue #2214, opened 2014-08-01
Problem migrating from dirtydb to sqlite


jparyani:
I'm having problems using bin/migrateDirtyDbtoMysql.js to migrate to sqlite. The script seems to run indefinitely without doing anything. Looking at the script, there doesn't appear to be anything particular to Mysql in it, but is it not intended to be used with sqlite?

If it helps, a dirty.db file that reproduces this for me is up at http://pastebin.com/HuU2KR22.

stefan2904:
(Wow, this issue is from 2014 ...)

I have the same problem ... 

I am trying to migrate from dirtydb to MySQL
using `node bin/migrateDirtyDBtoRealDB.js`

The script runs for ~3 hours (for 600k records) and after it reports successful flushing, the `store` table in MySQL is still empty. No errors are displayed.

stefan2904:
Quickfix, which did the job for me: disable the cache.

In `dbWrapperSettings` set `"writeInterval": 0`

In addition, I needed to increatse node's maximum memory size by calling `node --max-old-space-size=4096 bin/migrateDirtyDBtoRealDB.js` ([source](https://stackoverflow.com/a/38560292/1518225)).

muxator:
@stefan2904, after your experience, would you advise to add `"writeInterval": 0` to the script?

https://github.com/ether/etherpad-lite/blob/c17f5e5f0450d14a00318158df0f65a43233cd6b/bin/migrateDirtyDBtoRealDB.js#L13-L15

Even if it's a workaround, it prevents a behavior that sounds definitely buggy, and costed at least three hours of your time...

stefan2904:
@muxator, I am not entirely sure. It fixed the problem for me, but I am not sure what was wrong in the first place (except too little ram, which was fixed with the mentioned `node` option) and I also don't know what the implications of disabling the cache are (except higher database load).

I can only report the following: without setting `writeInterval` to 0 (= cache enabled) it flushed forever (>3 hours) and then nothing changed in the database. After setting it to 0 (= cache disabled) if wrote every record directly to the database, which worked (and took ~15 minutes). 

But since the script apparently works for other people without the change, I am not sure what was going on  - maybe a ueberDB maintainer knows? Or maybe other people had the same/similar problem, but did not report it? (Or maybe the script is not needed that much and thus no one noticed a problem.)

So maybe adding the line as remark/comment to the code makes sense?

[...]

PR #3519, opened 2018-11-27
migrateDirtyDBtoRealDB: memory and buffer  #2214

stefan2904:
`bin/migrateDirtyDBtoRealDB.js`

---

(my) solution for #2214:

* the script ran out of memory -> added comment to give the script more memory in 4a41241
* ueberDB never write to the actual database -> disable write buffer in 055642e, write to the db directly

---

Since this is a maintenance script, there are no tests. I hope that's indented/alright.


muxator:
merged, thanks.

Issue #3529, opened 2018-12-25
Dutch translation error


RickvLeeuwen:
Hi,
When opening the main page where you can create or open a pad, the Dutch text says:
_Pad maken **op** openen met de naam:_
This should be: 
_Pad maken **of** openen met de naam:_

Please see pull request https://github.com/ether/etherpad-lite/pull/3530

muxator:
Fixed by 02b3d427711805bf6f96d4ac6ba5e24d264db3f1, I guess.
Closing this.

RickvLeeuwen:
Indeed, thank you!

PR #3530, opened 2018-12-25
Fix Dutch translation error

RickvLeeuwen:
'op' should be 'of' here

### DIFF
--- a/src/locales/nl.json
+++ b/src/locales/nl.json
@@ -9,7 +9,7 @@
 		]
 	},
 	"index.newPad": "Nieuw pad",
-	"index.createOpenPad": "Pad maken op openen met de naam:",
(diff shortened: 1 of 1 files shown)


lpagliari:
Hi @RickvLeeuwen,
Thanks for the fix!

The translations on Etherpad [are made via TranslateWiki](https://github.com/ether/etherpad-lite/wiki/Translating), so the misspelling needs to be fixed there, not on the locale files here on the repo. Otherwise your fixes would be overwritten the next time the translations were pulled from there.

Could you help us and fix the misspelling there, please?

RickvLeeuwen:
Okay Will do. But have to wait before I can do translations there. They need to review me first..

muxator:
See #3529, closed by 02b3d427711805bf6f96d4ac6ba5e24d264db3f1.

PR #3536, opened 2019-01-10
set authorColors in HTML exports

raybellis:
This patch unconditionally adds colours to HTML exports by filling in the `authorColors` object which is not currently supplied to `getHTMLFromAtext()` during export (but is when the same function is used to produce an HTML diff).

If I knew how to make this optional I would.   The patch is shown here in case it can be merged as is, or for reference for other people that I know would like this functionality.

### DIFF
--- a/src/node/utils/ExportHtml.js
+++ b/src/node/utils/ExportHtml.js
@@ -29,6 +29,7 @@ var _encodeWhitespace = require('./ExportHelper')._encodeWhitespace;
 function getPadHTML(pad, revNum, callback)
 {
   var atext = pad.atext;
+  var authorColors;
   var html;
   async.waterfall([
   // fetch revision atext
@@ -49,12 +50,24 @@ function getPadHTML(pad, revNum, callback)
     }
   },
 
+    //get the authorColor table
+  function(callback){
+    pad.getAllAuthorColors(function(err, _authorColors){
+      if(err){
+        return callback(err);
+      }
+
+      authorColors = _authorColors;
+      callback();
+    });
+  },
+
   // convert atext to html
 
 
   function (callback)
   {
-    html = getHTMLFromAtext(pad, atext);
+    html = getHTMLFromAtext(pad, atext, authorColors);
     callback(null);
   }],
   // run final callback


raybellis:
It appears that this messes up PDF export, I think because the function `getHTMLFromAtext()` adds a `<style>` block to the body of the HTML (instead of to the `<head>`) which is not legal HTML.

JohnMcLear:
@raybellis did you want to get this properly submitted / fixed up is should we leave it as reference and close it?

raybellis:
@JohnMcLear We still need the feature (which is what got me started working on the Etherpad code in the first place) but it somewhat dropped off my priority list.   I had been planning to revisit it.

JohnMcLear:
@raybellis did you look at https://github.com/ether/etherpad-lite/pull/2994 ?

The difference here is that I added a new endpoint to support author colors I think?  I might be misunderstanding the functionality tho.

muxator:
> @JohnMcLear did #2994 work with PDF export / had a tunable to turn it off?

If so, we may want to work on that one (which currently has a lot of conflicts).
Otherwise, I would be ok with bringing this up to a minimum level of usability (decided by us) and merge this.

JohnMcLear:
@raybellis what's the shortcoming of https://www.npmjs.org/package/ep_export_authors ?

raybellis:
I don't recall now - I'll need to revisit it.

raybellis:
One thing I do recall is that there was a substantial difference in the rendering of HTML exports done from within the UI to those available via the API, with the latter being inferior (and perhaps not even fixable via the plugin interface?).

The problem is that my use case is performing a bulk export of old pads into a static format so that we can purge them from our MySQL DB and get back some performance.

JohnMcLear:
I'm +1 bringing this into core if you are willing to put a bit of time in.

But I'm more +1 keen on discovering WHY you are having to prune your database?  I run some pretty big sites (1M+ pads w/ 1K rev per pad) and I don't prune, I just optimize.  Do you have a specific requirement for running a pruned database?

I like lots of fidelity and Etherpad provides that, in fact it's a feature, not a bug.  We had a legal case where we had to provide every revision of a pad and "who changed what" character by character as per a EU law requirement.  That's why some people use Etherpad ( no other online editor does this ).

I'm not saying you are wrong btw, I will help in your use case as much as possible, for me, it feels like doing the job are you doing is a step backwards but obv if you are running on a RasPi or a tiny hard disk environment then I'm gonna understand the situation :D

PR #3537, opened 2019-01-11
simplify version table

raybellis:
This PR vastly shortens (and simplifies) the version table within `handler/APIHandler.js` by building each version's entry incrementally based off the previous version.

The resulting table has been validated by comparing the "before" and "after" output of the following loop on both versions of the code (albeit with an intermediate "sort" step to account for the different insertion order)

```
for (let v in version) {
        let m = version[v];
        for (let [k, a] of Object.entries(m)) {
                console.log(v, k, a);
        }
}
```

The patch also fixes a few typos, and removes a duplicate definition of `getChatHistory` which in each applicable version was defined with two different parameter lists, but where only the second would be used.


raybellis:
p.s. this turns out to be similar to #1638 although it's up-to-date with the latest API revision.

muxator:
I have edited the commit message to include the description contained in this PR, that was already adherent to the [contribution guidelines](https://github.com/ether/etherpad-lite/blob/ae494095932568a5909366e53c76b6cb9dad090a/CONTRIBUTING.md#pull-requests), in particular:

> * contain meaningful and detailed commit messages in the form:
>
> ```
> submodule: description
>
> longer description of the change you have made, eventually
> mentioning the number of the issue that is being fixed, in
> the form: Fixes #someIssueNumber
> ```

Issue #3540, opened 2019-01-16
Move to async/await. Get rid of callback hell


muxator:
Currently (1.8pre), Etherpad code base is heavily based on callbacks. While this made sense at the time (~2011) JS ecosystem has moved forward, and better solutions exist.

We should investigate the possibility of migrating the code base to an `async/await` paradigm, in order to make the code easier on the eye of future contributors.

This move was anticipated in #3424. It is now time to start discussing it concretely.

As usual, the first concern is maintaining stability: this change is going to be wide, but should be as shallow as possible, ideally not containing any new functionality work. After this, hopefully the code base will be easier to reason about, less buggy, and new features will be easier to implement.

**Edit**: the Pull Request that implements this change is #3559.

muxator:
Ray has completed the fist stage of his amazing work at https://github.com/raybellis/etherpad-lite/tree/async.

The review is being done at https://github.com/raybellis/etherpad-lite/tree/async-PR. That branch is a linear variant of Ray's work, rebased on the latest develop on which some adaptation are being made.

async-PR is a mutable branch, i.e. it will be periodically force-pushed until it is ready to be merged.

muxator:
Branch [raybellis/async-PR](https://github.com/raybellis/etherpad-lite/tree/async-PR) is now deployed on https://beta.etherpad.org/ to allow widespread testing.

It will be periodically updated while the branch is audited.

raybellis:
Please note that I've added a couple of extra commits to the `async` branch which relate to import and export.

muxator:
Thanks, imported.

I am (slowly) going through the series, trying to:
- separate out purely syntactic changes from substantial ones (to ease later review and catch eventual bugs)
- split changes where possible (for example https://github.com/raybellis/etherpad-lite/commit/2075e6fd6acda938d605cc4f94c8522f342c4a8e in async contained a [condition inversion](https://github.com/raybellis/etherpad-lite/commit/1948ed9f93668a5af032123eea2767e535ab0807) and the part that strictly [introduced promises consumption](https://github.com/raybellis/etherpad-lite/commit/f2300e60be36fe8b0c5b6c9cf69f787fcf87668c ), so they can be examined separately.

This will take some time.

Wikinaut:
Clarification request/question:

From the above I understand that the branch https://github.com/raybellis/etherpad-lite/tree/async-PR is now stable enough to try and test it on own instances.

Is my view correct?

muxator:
[...]

PR #3559, opened 2019-02-21
Migrate from callbacks to async/await

muxator:
This PR represents the line of work that refactors Etherpad to use `async`/`await` calls instead of callbacks, vastly simplifying the code, as per #3540.

This represents a vast change, and by its nature requires a lot of testing, hence its WIP status. It is currently deployed on https://beta.etherpad.org. It will be merged as soon as deemed sufficiently stable.

**It has the priority over other work**: present and future PRs, in case of conflicting changes, will need to rebase themselves over this work.

The development is by contributor @raybellis.
The branch will be periodically force-pushed if needed, please update often.

### DIFF
--- a/bin/checkAllPads.js
+++ b/bin/checkAllPads.js
@@ -1,144 +1,133 @@
 /*
-  This is a debug tool. It checks all revisions for data corruption
-*/
+ * This is a debug tool. It checks all revisions for data corruption
+ */
 
-if(process.argv.length != 2)
-{
+if (process.argv.length != 2) {
   console.error("Use: node bin/checkAllPads.js");
   process.exit(1);
 }
 
-//initialize the variables
+// initialize the variables
 var db, settings, padManager;
-var npm = require("../src/node_modules/npm");
-var async = require("../src/node_modules/async");
+var npm = require('../src/node_modules/npm');
+var async = require('../src/node_modules/async');
 
-var Changeset = require("../src/static/js/Changeset");
+var Changeset = require('../src/static/js/Changeset');
 
 async.series([
-  //load npm
+  // load npm
   function(callback) {
     npm.load({}, callback);
   },
-  //load modules
+
+  // load modules
   function(callback) {
     settings = require('../src/node/utils/Settings');
     db = require('../src/node/db/DB');
 
-    //initialize the database
+    // initialize the database
     db.init(callback);
   },
-  //load pads
-  function (callback)
-  {
+
+  // load pads
+  function (callback) {
     padManager = require('../src/node/db/PadManager');
-    
-    padManager.listAllPads(function(err, res)
-    {
+
+    padManager.listAllPads(function(err, res) {
       padIds = res.padIDs;
       callback(err);
     });
   },
-  function (callback)
-  {
-    async.forEach(padIds, function(padId, callback)
-    {
-        padManager.getPad(padId, function(err, pad) {
+
+  function (callback) {
+    async.forEach(padIds, function(padId, callback) {
+      padManager.getPad(padId, function(err, pad) {
+        if (err) {
+          callback(err);
+        }
+
+        // check if the pad has a pool
+        if (pad.pool === undefined ) {
+          console.error("[" + pad.id + "] Missing attribute pool");
+          callback();
+
+          return;
+        }
+
+        // create an array with key kevisions
+        // key revisions always save the full pad atext
+        var head = pad.getHeadRevisionNumber();
+        var keyRevisions = [];
+        for (var i = 0; i < head; i += 100) {
+          keyRevisions.push(i);
+        }
+
+        // run through all key revisions
+        async.forEachSeries(keyRevisions, function(keyRev, callback) {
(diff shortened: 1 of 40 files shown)


muxator:
Hi, the PR is almost ready to be integrated. I just need a review from @Ray.

While writing async, some boolean expressions where made stricter (e.g. `==` -> `===`, `!=` -> `!==`). This is generally a very good thing, but on a code base with lots of technical debt like ours, this can resurface latent bugs (this happened at least once).

1. For this reason, I have tried to extract & centralize as many such changes as possible in a single commit, namely d520e280bfed0dab6eadaf5a60a053b69417293a. These changes were all present in Ray's original work.

2. While doing this, I noticed that in some files this transformation was not symmetrical (some places were made stricter, others were not). I've done another commit (8ecd0dd9e858b9c80acc72547bc540e277d21045) which tentatively makes the situation even. None of these changes were part of Ray's work, yet I think we could integrate them.

I need a review on points 1 and 2. If Ray recognizes 1 as his beloved brainchild, and thinks that 2 may complete it, I'll fold the two changes and merge the PR.

raybellis:
I think your `author === null` check may result in a similar to the bug that I caused inadvertently with the SessionManager.

If you like, I'm happy to audit all calls to `db.get*` and ensure that they work with either `undefined` or `null` values, given that we can't be sure which will be returned by ueberDB.

I was going to suggest that just testing for `if (!value)` would suffice, but on reflection that may cause issues if the result value can legally be 0 or "", so an explicit check for `val === null || val === undefined` may be preferred.   The alternative is to explicitly require that loose checks are annotated to indicate that they are intentionally loose.

muxator:
That's true: using `author === null` could introduce regressions.

However, in its current status on `async` (`async-PR` closely tracks it) [AuthorManager.js has only strict checks](https://github.com/raybellis/etherpad-lite/blob/d03da2c8eb36f94a53621566e7cc0cfef8a493e5/src/node/db/AuthorManager.js) (`X !== null` or `X === null`) except for [the one in `mapAuthorWithDBKey()`](https://github.com/raybellis/etherpad-lite/blob/d03da2c8eb36f94a53621566e7cc0cfef8a493e5/src/node/db/AuthorManager.js#L91).

The main pain point is probably ueberDB, but I do not want to digress.

On commenting lax comparison: in async-PR, I agree, indeed I had already put such a comment in [SessionManager.js](https://github.com/raybellis/etherpad-lite/commit/8a97d5951c5bf771f82abd90a1dfb02fa472f9f2#diff-c6aae0f724e78a93dd699ad297010f52R88):
```js
/*
   * In some cases, the db layer could return "undefined" as well as "null".
   * Thus, it is not possible to perform strict null checks on group2sessions.
   * In a previous version of this code, a strict check broke session
   * management.
   *
   * See: https://github.com/ether/etherpad-lite/issues/3567#issuecomment-468613960
   */
if (!group2sessions || !group2sessions.sessionIDs) {
```

I think we have to decide what to do with AuthorManager, as well. All the original comparisons were lax, but if you have a look at d520e280bfed0dab6eadaf5a60a053b69417293a (which, I remember, just extracts changes made on async), all but one were made strict. Could this introduce latent bugs, as well?

> If you like, I'm happy to audit all calls to db.get* and ensure that they work with either undefined or null values, given that we can't be sure which will be returned by ueberDB.

That would be welcome: in that case we could reach a more coherent situation.

If you do not have time, or if we want to postpone this, I can also make a ugly proposal in the name of pragmatism:
- put lax comparisons wherever the old code base was already lax. This would bring us nearer to bug-for-bug compatibility
- put comments explaining our reasoning
- merge the PR so we can start moving forward

If you think it makes sense, I can compare _develop_ with the asynchronous version, and worsen its purity in the name of compatibility.

raybellis:
I've got time to do that audit.     How (and where) would you prefer to see any relevant commits made?

raybellis:
For reference, here's a comparison of ueberDB2's behaviour with `dirty` vs `mysql`:

DirtyDB
```
get undefined (random) key
-> undefined
set explicit value and verify
-> pass
set value to undefined
-> undefined
set value to null
-> undefined
```

MySQL
```
get undefined (random) key
-> null
set explicit value and verify
-> pass
set value to undefined and get
-> undefined
set value to null and get
-> null
```

raybellis:
I found an alternate solution - see 32cebc4298ad9261f068dc635065c2044014d5b4

muxator:
Nice. Pulled in with ac7663c33728.

muxator:
It's done.

I think it's worth citing from the commit message:

> Merge pull request #3559 from raybellis/async-PR
>
> With this commit, that closes #3540, we pay the first big slice of our technical debt. In this line of work **we streamlined the code base**, reducing its size by 15-20% and **making it more understandable at the same time**.
>
> The changes were audited and tested collaboratively and are deemed sufficiently stable for being merged.
>
>Known issues:
>- plugin compatibility is still not perfect
>- the error handling path needs to be improved
>
>This is an important day for Etherpad: thanks, Ray!
>

PR #3573, opened 2019-03-12
Await padManager.getPad in getPadLines

davidmehren:
`padManager.getPad` is an `async` function, therefore it needs to be `await`ed. Otherwise the timeslider cannot load old revisions.


muxator:
Nice catch, thanks.

Before merging, may I ask if you get it by visual inspection, or because you got any runtime error? If so, I would happily add a test case.

thegcat:
We noticed this when trying to see the history of a pad. There was no error in the interface, but the pad content shown was only the most recent version and not the history.

muxator:
Merged, thanks

Issue #3580, opened 2019-03-21
Redundant callback argument in Pad.prototype.copy()?


muxator:
@raybellis, is this `callback` argument is a leftover that can be removed?

https://github.com/ether/etherpad-lite/blob/7433d749f00cd389b6e4f8f1f9b1aa0c0a82dd23/src/node/db/Pad.js#L402

I am reviewing the code with the typescript compiler hoping to find some errors.

raybellis:
Yes, it's redundant, and I'm surprised it never came up in testing!  When executed it ought to trigger an "undefined variable" error.

raybellis:
The `return` in line 397 is also redundant - it's immediately after a `throw`.

muxator:
Incorporated this and other fixes in #3585.

I'll close the issue once that one is merged.

muxator:
Closed by #3585

PR #3585, opened 2019-03-26
Some fixes caught by the Typescript compiler

muxator:
These are assorted fixes found while during an experimental conversion to Typescript.

The most notable fixes are:

- `PadMessageHandler.handleMessage()` got the wrong `padId` for read only pads. This would lead to a crash.
- a redundant `callback` parameter to `pad.remove()` in `Pad.prototype.copy()` could cause a crash at runtime.
- in case of error, `restoreRevision()` would crash instead of writing a log message

### DIFF
--- a/src/node/db/API.js
+++ b/src/node/db/API.js
@@ -260,7 +260,7 @@ exports.getHTML = async function(padID, rev)
   }
 
   // get the html of this revision
-  html = await exportHtml.getPadHTML(pad, rev);
+  let html = await exportHtml.getPadHTML(pad, rev);
 
   // wrap the HTML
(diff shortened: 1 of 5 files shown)


muxator:
@raybellis, https://github.com/ether/etherpad-lite/pull/3585/commits/8808b4193f0f9b16e8ae0aaf9ce97453f745e406 fixes an overlook, confirmed?

raybellis:
Cofirmed, yes!    I found the inconsistency of "ID" vs "Id" quite confusing at times :(

muxator:
Merging this.

Aldo, closes #3580

PR #3587, opened 2019-03-29
remove devdeps from prod builds

alemairebe:
While trying to build Etherpad for ARM , it fails due to devdeps , so I removed them from prod docker builds.
This also generates smaller builds. (mine is around 50Mb)
My docker images are at https://gitlab.com/alm-docker/etherpad/container_registry 
I enabled multiarch builds


muxator:
Very useful! Merged, thanks.

slel:
Follow-up at #3588 to fix typo.

Issue #3088, opened 2016-11-23
Chat bubble shown on bottom right even though chat disabled


kernelfreak:
Hello friends,

I am trying to work with etherpad and currently facing some problem. I have disabled chat from settings.json as well as installed disable-chat plugin from admin-plugins. Even after that, whenever a pad is opening, for a few seconds on bottom right corner, I see a 0 which at hover says chat(alt+c). How can I remove this 0. Thank you.

JohnMcLear:
The solution would be to hide chat in CSS and only show it if showChat is set to true or undefined.  pad.js would handle this.

This is a low hanging fruit for someone.

PR #3595, opened 2019-04-15
Chat bubble: by default hide in CSS

Tristramg:
The current behaviour is to show the chat bubble and hide if chat is
disabled.

Because of this, the bubble appears wrongfully for a short time.

With this PR, by default it is hidden and displayed only if chat is
enabled.

Fixes: #3088


muxator:
Thanks, deployed on https://beta.etherpad.org, though before deploying it I was not able to trigger #3088, at least on that instance.

muxator:
Merged, thanks (with minor modification for unifiying formatting).

PR #3597, opened 2019-04-16
Fix a few vulnerabilities

Tristramg:
Two can not be fixed yet:
- https://github.com/swagger-api/swagger-node/pull/579
- https://github.com/nodejs/node-gyp/pull/1718


Tristramg:
Wouldn’t it be useful to also commit a package-lock.json? If you agree I will a make a follow-up PR

muxator:
Thanks for the contribution.
At least for the `clean-css` upgrade, we cannot proceed like this.

As you can see in [clean-css 4.2.1 readme](https://github.com/jakubpawlowicz/clean-css/tree/v4.2.1#important-40-breaking-changes):

>clean-css 4.0 introduces some breaking changes:
>
>  - `root`, `relativeTo`, and `target` options are replaced by a single `rebaseTo` option - this means that rebasing URLs and import inlining is much simpler but may not be (YMMV) as powerful as in 3.x;

The change affects Etherpad:
https://github.com/ether/etherpad-lite/blob/357780d573a055979ec3917d8b3a95ea9d55fa6e/src/node/utils/Minify.js#L418

The dummy fix (replacing `relativeTo` -> `rebaseTo`) does not work. I started working on this but had to suspend due to lack of time.

Would you be so kind to work on this on a different PR?

Tristramg:
Oh indeed. The only problem I see locally is with background on the `/` where the jpeg hasn’t the right route.
I tried to understand what happens, but I’m not a JS developper and I got lost here. Sorry, I don’t think I’ll be able to continue to focus on etherpad.

muxator:
Closing this PR since it is not going to progress.


The issues mentioned in this comment thread are going to be tracked in #3598 and #3616: refer there for status updates.

PR #3603, opened 2019-05-04
POC: remove dependency on require-kernel

muxator:
This is a proof-of-concept for getting rid of `ep-require-kernel`, which is unsupported and makes browsers cry, because:

> Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user’s experience.

It uses **browserify** to bundle the static js modules, which also results in a slightly lower load time and less round trips to the server (from 39 to 30 requests in a plain Etherpad installation with no skin).

This branch is currently ~deployed on https://beta.etherpad.org~.

What's working:
- pad editing
- chat
- the skin system

What's still not working:
- plugins (they do not load because there is no CommonJS loader in the browser now)
- the timeslider (easy to solve)
- the js bundle must be generated maually running `cd src/static/js ; ./create-bundle.sh`

A definitive solution would be probably use different technologies, but this work demonstrates that it is achievable.
It would be interesting to see if some results can be incorporated in https://github.com/storytouch/ep_webpack/, or if something from that plugin can be upstreamed.

### DIFF
--- a/bin/installDeps.sh
+++ b/bin/installDeps.sh
@@ -100,7 +100,7 @@ echo "Ensure that all dependencies are up to date...  If this is the first time 
   cd node_modules
   [ -e ep_etherpad-lite ] || ln -s ../src ep_etherpad-lite
   cd ep_etherpad-lite
-  npm install --no-save --loglevel warn
+  npm install --save --loglevel warn
 ) || {
   rm -rf src/node_modules
   exit 1
--- a/src/node/utils/Minify.js
+++ b/src/node/utils/Minify.js
@@ -28,7 +28,6 @@ var CleanCSS = require('clean-css');
 var uglifyJS = require("uglify-js");
 var path = require('path');
 var plugins = require("ep_etherpad-lite/static/js/pluginfw/plugins");
-var RequireKernel = require('etherpad-require-kernel');
 var urlutil = require('url');
 
 var ROOT_DIR = path.normalize(__dirname + "/../../static/");
@@ -253,8 +252,6 @@ function getAceFile(callback) {
     if (!settings.minify) {
       founds = [];
     }
-    // Always include the require kernel.
-    founds.push('$$INCLUDE_JS("../static/js/require-kernel.js")');
 
     data += ';\n';
     data += 'Ace2Editor.EMBEDED = Ace2Editor.EMBEDED || {};\n';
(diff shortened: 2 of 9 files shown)


lpagliari:
@muxator I was finally able to take a look on your POC, it looks pretty nice.

`ep_webpack` does something very similar to what is done on `create-bundle.sh`, but it bundles the files on running time due to the fact that it was implemented as an Etherpad plugin.

One thing we would need to know is the list of plugin dependencies, to they could be bundled together. On `ep_webpack` this is done on [`bundler.js`](https://github.com/storytouch/ep_webpack/blob/master/bundler.js): it generates a `ep_webpack/static/js/index.js` file with all the `exports` used by all plugins:
```
    exports.f0 = require("ep_myplugin1/static/js/index");
    exports.f1 = require("ep_myplugin2/static/js/other_index");
    exports.f2 = require("ep_myplugin2/static/js/index");
```

Note: we find out the list of all files used by the plugins by checking the values on `plugins.parts`.

If we could build this `ep_webpack/static/js/index.js` file, then it would be a matter of including it on the `create-bundle.sh`.

One more thing: on https://github.com/storytouch/ep_webpack/issues/11#issue-441942068, you've mentioned (correctly) that:

> ep_webpack is not able to dynamically recompile bundles

When do you think this would be useful, besides on a development environment? If you were thinking about users installing plugins on the Admin page, I believe we could handle that [by implementing `pluginInstall` and `pluginUninstall` hooks](https://github.com/storytouch/ep_webpack/blob/master/index.js#L5-L8), by forcing the `ep_webpack/static/js/index.js` file to be re-generated and then re-running `create-bundle.sh`.

lpagliari:
I won't be able to work on this on the next few days, but I'll try to play a little bit with your POC next week.

JohnMcLear:
This looked like really good progress, how are you guys getting on with this @lpagliari ?

JohnMcLear:
Bump guys, I think we should be pushing on this if we can :)

muxator:
This will NOT be part of **1.8.1**, no time for this now.

JohnMcLear:
+1 not 1.8.1 but I'd love to see it land in 1.9 (19) 💃

SamTV12345:
@muxator Do you have some time to move this forward? Possibly with rspack as a modern solution for webpack.?

Issue #3604, opened 2019-05-08
[Chrome] Ok button on index page not aligned


CupCakeArmy:
I was trying out the new theme (which btw is absolutely awesome!) and I have some style issues on chrome. (Firefox and Safari work fine)

<img width="400" alt="Screenshot 2019-05-08 at 13 27 48" src="https://user-images.githubusercontent.com/14948823/57372305-8790f280-7195-11e9-9d35-953b8d60ff4c.png">

The OK button is bad positioned.

```css
button[type=submit] {
    border-top-right-radius: 3px;
    border-bottom-right-radius: 3px;
    left: 305px;
    border: none;
    top: 0; // <-- This fixes the issue
}
```

This fixes the issue without breaking safari or firefox.

seballot:
Hi ! just tested locally, it fixes the issue, thanks @CupCakeArmy !

CupCakeArmy:
PR is opened

CupCakeArmy:
I've opened a PR

muxator:
Deployed on https://beta.etherpad.org, but does not work for me.

See detail on the PR #3606, let's continue the conversation there.

PR #3606, opened 2019-05-08
Fix chrome alignment

CupCakeArmy:
Fix for issue #3604

### DIFF
--- a/src/static/skins/colibris/index.css
+++ b/src/static/skins/colibris/index.css
@@ -64,6 +64,7 @@ button[type=submit] {
   color: #64d29b;
   background: #586a69;
   border: none;
+  top: 0;
 }
 
 #button:hover,


muxator:
That's weird, does not work for my browser.

```
$ chromium-browser --version
Chromium 73.0.3683.103 Built on Ubuntu , running on Ubuntu 19.04
```

Just deployed on https://beta.etherpad.org, cleared `<basedir>/var/minified*` and restarted (so everything is clean).

@CupCakeArmy, any hints?

seballot:
The changes is not deployed properly, do not why...
in https://beta.etherpad.org I got the following code from `index.css`
image

Change `top: 0` is not there

muxator:
Thanks @seballot. Indeed, I there is a small overlook in the PR. The `top: 0` statemente was added to the wrong block :)

https://github.com/CupCakeArmy/etherpad-lite/blob/ee383664adec624a9f443a9d774195fa92a95f5f/src/static/skins/colibris/index.css#L77-L83

@CupCakeArmy, would you mind:
- fixing the PR and
- transcribing the problem description in the commit message togheter with the refrence to issue #3604?

Thanks a lot!




**p.s.**: I have manually fixed in https://beta.etherpad.org/static/skins/colibris/index.css and the change seems ok, indeed.

CupCakeArmy:
This is weird... it is fixed for me in https://beta.etherpad.org/.
I've put the `top: 0;` in the `#padname`, which is the id of the ok field. Something with the deployment must be wrong. Is there any way to test this locally?

CupCakeArmy:
PR is updated

muxator:
Folded the commits, updated the commit message and pulled in.

Thanks.

PR #3614, opened 2019-06-16
Docker image will now be built based on Alpine

tabroughton:
I've build the image based on node:alpine, had to install curl separately.  I haven't tested on anything other than dirty db but works ok.  not much to it and not sure if it would cause any issues to plugins etc but it does reduce the image from 1GB to about 160MB which seems good for the liteness of etherpad-lite

```
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
tom/etherpad-alpine        latest              f8xxxxx        23 seconds ago      158MB
tom/etherpad             latest             ddxxxxx        19 hours ago        985MB
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,7 +6,7 @@
 #
 # Version 0.1
 
-FROM node:latest
+FROM node:alpine
 LABEL maintainer="Etherpad team, https://github.com/ether/etherpad-lite"
 
 # git hash of the version to be built.
@@ -16,6 +16,9 @@ ARG ETHERPAD_VERSION=develop
 # Set the following to production to avoid installing devDeps
 # this can be done with build args (and is mandatory to build ARM version)
 ARG NODE_ENV=development
+
+# add curl to alpine
+RUN apk add --no-cache curl
 
 # grab the ETHERPAD_VERSION tarball from github (no need to clone the whole
 # repository)


muxator:
Hi @tabroughton, thanks for the proposal.

1. I am confident we could get away without using `curl`: it has not been needed since a long time. For reference, see commit https://github.com/ether/etherpad-lite/commit/9aaa02ec7034, which also updated the documentation. Is it possible that you were mislead by some old documentation laying around the Internet? If, instead, you _had_ to add it, I would be interested in knowing why;
2. I am all for slimming down the docker image (who isn't? :) ). Could you please test some of the non-trivial cases? Ideally, at least `postgresql`, and/or whichever driver/plugin require a compiled binary. The reason is that alpine is `musl`-based, and we may want to test at least once in a lifetime if there could be some `glibc`-specifics in our dependencies.

Thank you & I hope we can move this forward.

pierreprinetti:
@muxator 

## Regarding 1. curl
The current Dockerfile does not actually use the code in the current working directly; instead, it uses `curl` to pull it from Github.

https://github.com/ether/etherpad-lite/blob/cb55905d9beb8a6d887a1b9157d36be2e27a2e0e/docker/Dockerfile#L23-L37

Unless you want to streamline the Dockerfile and let it use the code that is locally available, moving to Alpine means either installing `curl` or turning to `wget`

muxator:
@tabroughton: in Etherpad 1.8.0 we switched the base docker image to `node:10-buster-slim`. 

https://github.com/ether/etherpad-lite/blob/1.8.0-beta.1/docker/Dockerfile#L9

Size of the image for curent develop (7e44dc56) is 296MB.

muxator:
Closing, we are moving in another direction.
Please reopen if building on Alpine is absolutely needed.

Issue #3620, opened 2019-07-02
Importing unordered lists with a 'space' between the UL and LI adds an extra list entry - AWAITING REVIEW / MERGE


tombriles:
Here is a very simple example:

`<html><body><ul> <li>FOO</li></ul></body></html>`

There will be two bullets when this is imported.

muxator:
Reopening, because #3622 had to be reverted because introduced a regression (#3728).

JohnMcLear:
Input of:
```
<html><body><ul> <li>FOO</li></ul></body></html>
```

Turns into:
```
<!doctype html><html><body><ul class="bullet"><li>&nbsp;</li><li>FOO</ul></body></html>
```

Expected clean output would be:
```
<!doctype html><html><body><ul class="bullet"><li>FOO</li></ul></body></html>
```

- [ ] Understand the addition of the blank additional li.
- [ ] Understand the lack of closed </li>

JohnMcLear:
Well, even 
```
var html = '<html><body><ul><li>FOO</li></ul></body></html>';
```
Fails to properly import.  It misses the closing ``</li>``

JohnMcLear:
Turns out the trailing </li> is optional, I mean, I hate that but whatever.

JohnMcLear:
This problem occurs in OL too..  It's hard to explain but I will try...

Essentially the list _begins_, then you have content(in this case `` ``).  You have three choices.  
1) Exit out of the list
2) Assume it's a line item.  
3) Assume it's none list content within a list, Etherpad doesn't understand that concept.

To replicate this, open a pad with contents:
```
foo
bar
late
```

Make this a numbered list, select the line "bar", press Shift Tab.

Note that "late" is now 1), it should be 2)

Google Docs gets this right(ish[shift tab isn't supported]) but they support the concept of a non list item within a list.

So the TLDR here is that contentcollector needs to support non list items within lists. 

This is a rewrite of list handling logic in contentcollector, it's a big job.  contentcollector has been touched historically and maybe we did have this working at some point, I can maybe look into that but I don't think the rewrite of contentcollector list logic is doable with my day-to-day life as it is.

JohnMcLear:
Aha, I knew this felt familiar..  Here I talked about it back in 2014.. https://github.com/ether/etherpad-lite/issues/2194

JohnMcLear:
[...]

PR #3622, opened 2019-07-03
Issue 3620 fix

tombriles:
This pull request Fixes: #3620

### DIFF
--- a/src/static/js/contentcollector.js
+++ b/src/static/js/contentcollector.js
@@ -406,12 +406,13 @@ function makeContentCollector(collectStyles, abrowser, apool, domInterface, clas
         cc: this,
         state: state,
         tname: tname,
-        node:node,
-        text:txt,
+        node: node,
+        text: txt,
         styl: null,
         cls: null
       });
-      var txt = (typeof(txtFromHook)=='object'&&txtFromHook.length==0)?dom.nodeValue(node):txtFromHook[0];
+      var txt = (typeof(txtFromHook)=='object' && txtFromHook.length==0)
+        ? dom.nodeValue(node).trim() : txtFromHook[0].trim();
 
       var rest = '';
       var x = 0; // offset into original text
--- a/tests/frontend/specs/importexport.js
+++ b/tests/frontend/specs/importexport.js
@@ -52,125 +52,127 @@ describe("import functionality", function(){
     return exportresults
   }
 
-  xit("import a pad with newlines from txt", function(done){
-    var importurl = helper.padChrome$.window.location.href+'/import'
-    var textWithNewLines = 'imported text\nnewline'
-    importrequest(textWithNewLines,importurl,"txt")
+  it("import a pad with newlines from txt", function(done){
+    var importurl = helper.padChrome$.window.location.href+'/import';
+    var textWithNewLines = 'imported text\nnewline';
+    importrequest(textWithNewLines, importurl, "txt");
     helper.waitFor(function(){
-      return expect(getinnertext()).to.be('<span class="">imported text</span>\n<span class="">newline</span>\n<br>\n')
-    })
-    var results = exportfunc(helper.padChrome$.window.location.href)
-    expect(results[0][1]).to.be("imported text<br>newline<br><br>")
-    expect(results[1][1]).to.be("imported text\nnewline\n\n")
-    done()
-  })
-  xit("import a pad with newlines from html", function(done){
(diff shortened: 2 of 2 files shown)


muxator:
Thank you very much!

Commit 891f000e1e8e69ca9abf7c51f9a9eb6fe71290ef (Test for Issue-3620: space in the HTML for an unordered list) is hard to follow: probably the introduction of the new test was intermixed with reformatting & reorganization of the code.

It should be split so it can be audited, thanks.

tombriles:
You're asking me to remove the formerly 'pending' tests that are now passing?

muxator:
Hi @tombriles, the fact that you took care of updating the tests before fixing the issue is commendable, thanks for that.

The thing is, before merging a PR I inspect the changes and try to assess how it changes its behaviour. But when looking at a big patch, both a trivial formatting change, a smart refactoring, or a new test that shows a failure to be fixed in the next commit, all appear the same. This makes it difficult to tell what is the intent of the change, and ensure no other bugs are introduced.

In the case of 891f000e1e8e69ca9abf7c51f9a9eb6fe71290ef, it is difficult to tell apart the syntactic changes (reformattings, whitespace changes, added trailing semicolons) from the "real stuff" (the valuable work that was done on the tests).

Some examples of purely syntactical changes that, if are to be maintained (which is fine), should go in dedicated commits:
- added spaces or semicolons:
  [here](https://github.com/ether/etherpad-lite/pull/3622/commits/891f000e1e8e69ca9abf7c51f9a9eb6fe71290ef#diff-59cfbc66b5f9988c447ac566f93ab93eL93):
  ```diff
  - var htmlWithBullets = '<html><body><ul class="list-bullet1"><li>bullet line 1</li>...</html>'
  + var htmlWithBullets = '<html><body><ul class="list-bullet1"> <li>bullet line 1</li>...</html>'
  ```
  or [here](https://github.com/ether/etherpad-lite/pull/3622/commits/891f000e1e8e69ca9abf7c51f9a9eb6fe71290ef#diff-59cfbc66b5f9988c447ac566f93ab93eL111):
  ```diff
  - importrequest(htmlWithBullets,importurl,"html")
  + importrequest(htmlWithBullets,importurl,"html");
  ```
- [reformattings](https://github.com/ether/etherpad-lite/pull/3622/commits/891f000e1e8e69ca9abf7c51f9a9eb6fe71290ef#diff-59cfbc66b5f9988c447ac566f93ab93eL96):
  ```diff
  -      [...] expect(getinnertext()).to.be('\
  -<ul class="list-bullet1"><li><span class="">bullet line 1</span></li></ul>\n\
  +      [...] expect(getinnertext()).to.be(
  +        '<ul class="list-bullet1"><li><span class="">bullet line 1</span></li></ul>\n' +
  ```

What is left after that is the real work I would lke to see: was the code reorganized? Was a new test added? Was at some point replaced `xit()` with `it()`?

I am sure there is a reason for each of these actions. I simply need them to be explained by the commits.

Thanks.

muxator:
I cannot pull in this PR in its current state, but I'll leave it open since the fix is useful.

JohnMcLear:
@muxator you did the right thing here.  I will try bisect the fix and send through a cleaner PR.

JohnMcLear:
Right I read through all of this and actually it can be merged.  While I agree it's not ideal to merge so much code after doing a decent review it's all of the same fixes I would have put in.  

Obviously in the future we would prefer if @tombriles can ensure each major change is a separate commit.

I'm merging.  Again @muxator you did the right thing in pushing back and we ask @tombriles to provide smaller commits in the future.  Still, awesome work guys and thanks :D

muxator:
These changes caused a very bad problem with spaces and the cursor jumping around with both Firefox and Chrome (see #3728).

I had to revert it via #3729.

src/static/js/contentcollector.js
```diff
@@ -406,12 +406,13 @@ function makeContentCollector(collectStyles, abrowser, apool, domInterface, clas
...
-        node:node,
-        text:txt,
+        node: node,
+        text: txt,
         styl: null,
         cls: null
       });
-      var txt = (typeof(txtFromHook)=='object'&&txtFromHook.length==0)?dom.nodeValue(node):txtFromHook[0];
+      var txt = (typeof(txtFromHook)=='object' && txtFromHook.length==0)
+        ? dom.nodeValue(node).trim() : txtFromHook[0].trim();
```
JohnMcLear: .trim() being the cause of the fault / error here :(

PR #3631, opened 2019-08-02
adding example run with volume

RaymondCavallaro:
adding example run with volume


muxator:
Thanks.

Pulled in #3668 instead, which is practically a duplicate, but with merge conflicts already resolved.

Closing, feel free to comment there if needed.

PR #3633, opened 2019-08-07
NODE_ENV controls run-time behaviour, thus needs to be set by ENV

larso:
Setting the NODE_ENV to production does not change the run-time behaviour of etherpad-lite:

> [2019-08-07 10:29:49.914] [WARN] console - Etherpad is running in Development mode.  This mode is slower for users and less secure than production mode.  You should set the NODE_ENV environment variable to production by using: export NODE_ENV=production

Using the ENV command in the Dockerfile resolves this.


muxator:
Merged, thanks

PR #3634, opened 2019-08-07
Fix Unicode bug in HTML export

jomority:
Currently there is a bug in the HTML exporter.

Steps to reproduce:
- Create a new pad
- Add any character with a Unicode value above 0x10000, for example `👍`.
- Export the pad to HTML

The character will not be encoded correctly for HTML. It will be represented by two character references, which refer to surrogate code units. Therefore a browser cannot display the character correctly.

After applying this commit, the character is exported as just one character reference and gets displayed properly.

### DIFF
--- a/src/node/utils/ExportHelper.js
+++ b/src/node/utils/ExportHelper.js
@@ -77,7 +77,7 @@ exports._analyzeLine = function(text, aline, apool){
 
 
 exports._encodeWhitespace = function(s){
-  return s.replace(/[^\x21-\x7E\s\t\n\r]/g, function(c){
-    return "&#" +c.charCodeAt(0) + ";";
+  return s.replace(/[^\x21-\x7E\s\t\n\r]/gu, function(c){
+    return "&#" +c.codePointAt(0) + ";";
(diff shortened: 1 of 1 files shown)


muxator:
Hi, I am trying to replicate the fix.
I am using the web interface ("Export current Pad as HTML") as in the following screenshot:

etherpad-3634

But the export stays the same :(

etherpad-3634-02

Any advice?
- Etherpad 161a38efd248
- Nodejs 10.14.1
- Firefox 68.0.1

jomority:
Hi, that is very strange.

I just tried it in a new Ubuntu 18.04 VM with just build-essential, nodejs 10.14.1  and etherpad 161a38e installed and it seems to work:


image

I don't have any idea why it does not work in your environment.

muxator:
Hi @jomority, thanks for the reply.

I'll do another check ASAP and merge this if it turns out it was just a setup fault (which I believe).

muxator:
...aand it had to be me :)

The patch works well.
Merged, thanks.

Issue #3043, opened 2016-08-23
Use referrer policy


rugk:
See
* https://w3c.github.io/webappsec-referrer-policy/
* http://caniuse.com/#feat=referrer-policy

You can use it in HTML (`<meta name="referrer" content="no-referrer">`) or as [a HTTP header](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-header)

lpagliari:
@rugk can you submit a PR with a fix for this issue, please? This is an open source project, which means you can contribute with some code too. That would help us much more than only opening the issue. Thanks!!

PR #3044, opened 2016-08-23
Add Referrer-Policy (no-referrer)

rugk:
Fixes https://github.com/ether/etherpad-lite/issues/3043
Fixes https://github.com/ether/ep_hide_referrer/issues/17
Fixes https://github.com/ether/ep_hide_referrer/issues/10
Fixes https://github.com/ether/etherpad-lite/issues/1603

Makes https://github.com/ether/ep_hide_referrer obsolete.

PR #3636, opened 2019-08-25
change referrer policy. Stop sending referers as much as possible

ahmadine:
Before this patch I managed to send a bunch of referrers to third party websites with Firefox. Unfortunately, meta name="referrer" isn't enough.
I added "noopener" because that's also what Nextcloud does. One may want to check another pull request from 2015 which may be more complete, but it was denied with a plugin proposal: #2498
References to earlier issues: #1603, #3043

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 1.8
+* SECURITY: change referrer policy so that Etherpad addresses aren't leaked when links are clicked (discussion: https://github.com/ether/etherpad-lite/pull/3636)
+
 # 1.8-beta.1
 * FEATURE: code was migrated to `async`/`await`, getting rid of a lot of callbacks (see https://github.com/ether/etherpad-lite/issues/3540)
 * FEATURE: support configuration via environment variables
--- a/src/node/hooks/express.js
+++ b/src/node/hooks/express.js
@@ -76,6 +76,15 @@ exports.restartServer = function () {
     // https://github.com/ether/etherpad-lite/issues/2547
     res.header("X-UA-Compatible", "IE=Edge,chrome=1");
 
+    // Enable a strong referrer policy. Same-origin won't drop Referers when
+    // loading local resources, but it will drop them when loading foreign resources.
+    // It's still a last bastion of referrer security. External URLs should be
+    // already marked with rel="noreferer" and user-generated content pages are already
+    // marked with <meta name="referrer" content="no-referrer">
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy
+    // https://github.com/ether/etherpad-lite/pull/3636
+    res.header("Referrer-Policy", "same-origin");
+
     // send git version in the Server response header if exposeVersion is true.
     if (settings.exposeVersion) {
       res.header("Server", serverName);
--- a/src/node/utils/ExportHtml.js
+++ b/src/node/utils/ExportHtml.js
@@ -285,7 +285,10 @@ function getHTMLFromAtext(pad, atext, authorColors)
         var url = urlData[1];
         var urlLength = url.length;
         processNextChars(startIndex - idx);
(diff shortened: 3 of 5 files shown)


slel:
Isn't that what this plugin is for: https://github.com/ether/ep_hide_referrer

ldidry:
@slel We shouldn’t have to use a plugin to avoid pad’s URLs leaking to third party websites.

ldidry:
@muxator Would you consider merging this PR for 1.8.0? I think it’s important.

muxator:
Personally, I am sensitive to the issue: my Firefox always runs with the referrer disabled, or with [`network.http.referer.spoofSource = true`](https://wiki.mozilla.org/Security/Referrer). In summary: I would prefer a more secure default like the one proposed here.

Regarding the PR: I would merge it if it was in a more complete form:
- mention somewhere in the documentation that no referrer is leaked (and that this is a change that was impelmented starting from *1.8*)
- in the code & the commit message: explain what was done (at least a link to the html spec for `noreferrer` and `noopener`)
- it would also make sense, on the server side, to send the [Referrer-Policy header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy). Something like:
  ```http
  Referrer-Policy: strict-origin-when-cross-origin
  ```

Finally, we have to think if we want to implement this change without offering a configuration option to restore the old behaviour. I do not see a reason to do it, but you never know what's out there among your users.

ldidry:
@ahmadine would you need help to comply to @muxator requested changes?

muxator:
Gentle bump :)

ahmadine:
@slel It's already an Etherpad semantic to remove the referrers, #3044 has been merged, but Firefox seems to ignore that. This patch made Firefox remove them as well.

@muxator Of course, I will work on this patch more, this or next week. I found more places where the referrer may leak later, but this one is the most urgent.

ahmadine:
I rebased my commit on @joelpurra's work. Below I will paste the content of the commit message, because it is quite informative about the issue and I think it should be more visible to people who will be interested in that issue.

```
    Work towards the greatest level of referrer security.
    Pull request with discussion: https://github.com/ether/etherpad-lite/pull/3636
    
    What's already there:
    * meta name=referrer: already done in 1.6.1: https://github.com/ether/etherpad-lite/pull/3044
      https://caniuse.com/#feat=referrer-policy
      https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-delivery-meta
      (Chrome>=78, Firefox>=70, Safari>=13, Opera>=64, ~IE[1], ~Edge[1])
    
    The previous two commits (by @joelpurra) I backported in this batch:
    * a rel=noreferrer: a pull request denied before: https://github.com/ether/etherpad-lite/pull/2498
      https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer
      https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types
      (Firefox>=37, I can't find more info about support)
    
    This commit adds the following:
    * a rel=noopener: fixing a not-so-well-known way to extract referer
      https://html.spec.whatwg.org/multipage/links.html#link-type-noopener
      (Chrome>=49, Firefox>=52, Safari>=10.1, Opera>=36, !IE, !Edge)
    * Referrer-Policy: same-origin: the last bastion of referrer security
      https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy
      (Chrome>=61, Firefox>=52, Safari>=11.1, Opera>=48, !IE, !Edge)
    
    meta name=referrer wasn't enough. I happened to leak a few referrers with my Firefox browser,
    though for some browsers it could have been enough.
    
    [1] IE>=11, Edge>=18 use a different syntax for meta name=referrer, making it most probably
        incompatible (but I may be wrong on that, they may support both, but I have no way to test
        it currently). The next Edge release will be based on Chromium, so for that the Chrome version
        applies.
```

muxator:
Pulled in, thanks.

Issue #3629, opened 2019-07-31
Docker runs etherpad as root user


Cronixx:
Hey, i've noticed etherpad/node is running under the root user account in the default docker container.
Is there any special reason it has to be like that?

#1315 mentions it would be advisable to have ep run as a special etherpad user

zem:
+1 pse fix

Fnjn:
It's common practice to run as root in container just for simplicity. Unless you want to run many different services in a single container, which is generally not advisable.

muxator:
Merged #3643. Thanks @pierreprinetti for reworking the commit.

Fnjn:
> Using the root user in a container is bad practice

@pierreprinetti I don't think it's possible that a root user in a container can escalate to be root user on the host machine.

The example in docker official document [Best practices for writing Dockerfiles](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) does not switch to unprivileged user.

pierreprinetti:
@Fnjn 
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user

> If a service can run without privileges, use USER to change to a non-root user.

PR #3643, opened 2019-09-30
docker: Run as unprivileged user

pierreprinetti:
Processes in containers should not run as root.

This change creates an unprivileged user in the Docker container, and
runs the main process using that user.

## Tested:
* with a volume attached
* with MySQL

## Caveats

A container built to run as non-root will not have access to volumes created with a container built to run as root.

Fixes https://github.com/ether/etherpad-lite/issues/3629

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -61,5 +61,16 @@ RUN for PLUGIN_NAME in ${ETHERPAD_PLUGINS}; do npm install "${PLUGIN_NAME}"; don
 #   https://stackoverflow.com/questions/31528384/conditional-copy-add-in-dockerfile#46801962
 COPY ./settings.json /opt/etherpad-lite/
 
+# Follow the principle of least privilege: run as unprivileged user.
+#
+# Running as non-root enables running this image in platforms like OpenShift
+# that do not allow images running as root.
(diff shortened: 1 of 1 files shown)


pierreprinetti:
Rebased.

docker/Dockerfile
```diff
@@ -61,5 +61,12 @@ RUN for PLUGIN_NAME in ${ETHERPAD_PLUGINS}; do npm install "${PLUGIN_NAME}"; don
 #   https://stackoverflow.com/questions/31528384/conditional-copy-add-in-dockerfile#46801962
 COPY ./settings.json /opt/etherpad-lite/
 
+RUN \
+    echo 'nobody:x:65534:65534:nobody:/:' > /etc/passwd && \
```
muxator: Would be a problem calling it "etherpad" instead of "nobody"?

I am thinking about the output of a `ps` command, that in this case would be more informative.

muxator: The sources in [this comment](https://github.com/ether/etherpad-lite/issues/3629#issuecomment-545798318) about why it is a bad practice could be incorporated in the commit message (and in the dockerfile itself)

Issue #3623, opened 2019-07-05
Add docker install instructions to readme.


dessalines:
There are no instructions on how to do a docker install on the README.

muxator:
Hi @dessalines, personally I think this is a bit out of scope: an advanced user would know how to do it, while a newcomer would blindly copy the (soon outdated) instructions without understanding them.

But then, maybe you are right and a short section with the instructions (maybe with an external link for who wants more information) could be useful.

Feel free to propose a change and let's see.

PR #3644, opened 2019-10-01
docker: Enable env var settings by default

pierreprinetti:
By leveraging the templating mechanism in `settings.json`, this change
allows a Docker client to run a prebuilt image and change some basic
configuration settings, like the instance name or, more importantly, the
database coordinates.

Also addresses https://github.com/ether/etherpad-lite/issues/3623

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 node_modules
-settings.json
+/settings.json
 !settings.json.template
 APIKEY.txt
 SESSIONKEY.txt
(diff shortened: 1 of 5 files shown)


pierreprinetti:
Tested with:
• dirtyDB
• Postgres

muxator:
Great. My only concern is with the default settings.

A lot of people will run the container without setting `ADMIN_PASSWORD`, and that would result in a container with an administrative account with a well-known password.

https://github.com/pierreprinetti/etherpad-lite/blob/docker_env_settings/docker/settings.json#L338-L342:
```json
  "users": {
    "admin": {
      "password" : "${ADMIN_PASSWORD:changeme1}",
      "is_admin": true
    },
```

muxator:
I have made a proposal in #3648 to address the concern raised in the previous comment. Once that is implemented, the work on this can be resumed.

muxator:
Merged, thanks.

PR #3647, opened 2019-10-18
Support for clean & safe error handling on IE 11

aaron-costello:
Added pad_utils sanitization for clean and safe error handling on browsers that do not encode the path of the URL.

### DIFF
--- a/src/templates/pad.html
+++ b/src/templates/pad.html
@@ -401,7 +401,6 @@
 
               var plugins = require('ep_etherpad-lite/static/js/pluginfw/client_plugins');
               var hooks = require('ep_etherpad-lite/static/js/pluginfw/hooks');
-              var padutils = require('../static/js/pad_utils');
 
               plugins.baseURL = baseURL;
               plugins.update(function () {


muxator:
Thanks.

The patch was applied on **1.7.5** and applies cleanly on 1.8 as well.
Rebasing there & force pushing on this PR to keep history auditable.

src/templates/pad.html
```diff
@@ -400,6 +401,7 @@ <h2 data-l10n-id="pad.share.emebdcode"></h2>
 
               var plugins = require('ep_etherpad-lite/static/js/pluginfw/client_plugins');
               var hooks = require('ep_etherpad-lite/static/js/pluginfw/hooks');
+              var padutils = require('../static/js/pad_utils');
```
muxator: Is this require() needed?

muxator: I guess it's not (only the first one seems to be needed).

aaron-costello: Hi @muxator ,

Apologies I didn't notice that was still in there, that 2nd one is a result of testing and is not effective. Only the first 'require' is needed.

muxator:
The PR cannot be rebased on `develop` (where patches should be directed).

I'll manually apply this & close the PR.

muxator:
I think this change broke the frontend code.

This is what my developer tools show when loading a pad in etherpad **1.8.0-beta.1** (I am using Firefox 70):
```
ReferenceError: require is not defined
    <anonymous> https://beta.etherpad.org/p/issue3647:559
```

@aaron-costello, do you see this as well?


I have left a comment in the commit that implements this PR:
https://github.com/ether/etherpad-lite/commit/5879037ddca4ab9a4002adf90fc7ce6c9f82f01b#diff-af61109c6bb3440a9cb12337af45894eR364

PR #3656, opened 2019-10-24
docker: Fix the comments about settings.json copy

pierreprinetti:
The "conditional copy trick" was removed in #3644, but I failed to
update the corresponding comment.


pierreprinetti:
(rebased)

PR #3668, opened 2019-11-15
docker: Add Run with volume example

pierreprinetti:
Supersedes https://github.com/ether/etherpad-lite/pull/3631


pierreprinetti:
Rebased.

muxator:
Pulled in this PR instead of #3631, a duplicate which was older and thus wiser :)
This one had no conflicts, and so it won.

Comments by @RaymondCavallaro are welcome.

PR #3670, opened 2019-11-25
Fix missing 'require()' dependency in previous security fix

aaron-costello:
Original: https://github.com/ether/etherpad-lite/commit/5879037ddca4ab9a4002adf90fc7ce6c9f82f01b

### DIFF
--- a/src/templates/pad.html
+++ b/src/templates/pad.html
@@ -360,7 +360,11 @@
         <% e.end_block(); %>
 
         <% e.begin_block("scripts"); %>
+        
+        <script type="text/javascript" src="../static/js/require-kernel.js"></script>
+          
         <script type="text/javascript">
+            var padutils = require('../static/js/pad_utils').padutils;
             // @license magnet:?xt=urn:btih:8e4f440f4c65981c5bf93c76d35135ba5064d8b7&dn=apache-2.0.txt
             (function() {
               // Display errors on page load to the user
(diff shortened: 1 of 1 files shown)


muxator:
Thanks for the fix!

There are some merge conflicts. In order to fix them, I would have to force push on _your_ default branch. This can be a problem if you are relying on that repo for a specific reason, or an advantage if you are using it only for contributing to Etherpad. :)

Is it ok for you?

**p.s.**: btw, I suspect this is the reason for which I had to manually port your original PR #3647 on top of develop.

muxator:
Tested the change right now, and just moving the `<script>` block as proposed in this change is not enough.

I think the reason is that to be operational, `require-kernel` needs to be configured, and this happens at a later time:

https://github.com/ether/etherpad-lite/blob/ba38ed3bbae20bd195bd7fef983ab0ff5e387395/src/templates/pad.html#L402-L406

@aaron-costello, what do you see in your browser console?
I see this (Firefox 70, etherpad ba38ed3bbae2 + your patch):
```
Error: Attempt to retrieve the root module "/../static/js/pad_utils.js" but no root URI is defined.
```

muxator:
Hi @aaron-costello, still no luck :(

It's puzzling, because from visual inspection your fix (https://github.com/ether/etherpad-lite/pull/3670/commits/71bc1940c5bc022fdb7601ac48aae4a8017da7a7) should definitely work!
But I tried it on the latest develop and nonetheless I saw this:

```
Error: The module at "/../static/js/pad_utils" does not exist.
```

The **good news** is that I think I have found a way to make it work, taking inspiration from later in the file. I am closing this PR and committing 695c2d2e84b3 in the meantime. Please confirm if this works for you, too.

If we find a way to make your change work (which I like more) I'll happily accept that.

Thanks for helping chasing the bug!

PR #3676, opened 2019-12-01
docker: Set the home directory for the user

pierreprinetti:
Before this change, the docker user had home in a directory it had no
permissions on. The inability of creating a cache directory in `$HOME`
prevented npm to work properly.

Additionally, the `node_modules` in the base working directory had its
owner set to root, preventing further changes.

With this change, the `etherpad` user has a home directory.
Additionally, `npm i` is now run by `etherpad` rather than the root
user; this way, it is possible to dynamically change the `node_modules`
content in day 2 operations.

Note that while switching to the `useradd` builtin, a conflict was
discovered with the GID 65534 that was previously used. This change is
changing the `etherpad` user's UID to 5001 to avoid said conflict. As a
consequence, a `chmod -R 5001:5001` must be run prior to attaching
volumes created from previous Etherpad versions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,9 +19,19 @@ ARG ETHERPAD_PLUGINS=
 # this can be done with build args (and is mandatory to build ARM version)
 ENV NODE_ENV=development
 
+# Follow the principle of least privilege: run as unprivileged user.
+#
+# Running as non-root enables running this image in platforms like OpenShift
+# that do not allow images running as root.
+RUN useradd --uid 5001 --create-home etherpad
+
+RUN mkdir /opt/etherpad-lite && chown etherpad:etherpad /opt/etherpad-lite
+
+USER etherpad:etherpad
+
 WORKDIR /opt/etherpad-lite
 
-COPY ./ ./
+COPY --chown=etherpad:etherpad ./ ./
 
 # install node dependencies for Etherpad
 RUN bin/installDeps.sh && \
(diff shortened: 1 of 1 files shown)


muxator:
I guess something like a `chown -R in /opt/etherpad-lite`after the `COPY` statement  is needed, because docker build complains:

```
$ docker build --tag muxator/etherpad .
[...]
Step 8/13 : COPY ./ ./
 ---> f882b96b2ccb
Step 9/13 : RUN bin/installDeps.sh
 ---> Running in 31ae5510747f
Copy the settings template to settings.json...
cp: cannot create regular file 'settings.json': Permission denied
```

It seems that the `COPY ./ ./` statement populates `/opt/etherpad-lite` with root-owned files, even if `USER etherpad:etherpad` was already executed.

```
$ docker run -ti --rm 769477de80fd bash
etherpad@2da8f587ec10:/opt/etherpad-lite$ ls -ld /opt/etherpad-lite/
drwxr-xr-x 1 root root 326 Dec  1 18:56 /opt/etherpad-lite/
etherpad@2da8f587ec10:/opt/etherpad-lite$ ls -l /opt/etherpad-lite/
total 104
-rw-rw-r-- 1 root root 26442 Dec  1 18:24 CHANGELOG.md
-rw-rw-r-- 1 root root  8217 Oct 23 22:08 CONTRIBUTING.md
-rw-rw-r-- 1 root root  1424 Dec  1 18:55 Dockerfile
[...]
```

pierreprinetti:
Interesting: Podman somehow made it work.

However, it seems to work with Docker now: https://travis-ci.org/ether/etherpad-lite/jobs/619350446

muxator:
Queued, thanks

PR #3703, opened 2020-03-06
Fix toolbar on mobile

### LINKED ISSUE #3697: Default skin needs polish
I love the new default skin but it needs polish, especially around when the browser width is < 400px as you lose formatting buttons etc.  Afaik this was all handled before so should be handled with any new versions with a copy/paste :)

pkrasicki:
Fixes #3697

etherpad mobile toolbar fix

### DIFF
--- a/src/static/css/pad.css
+++ b/src/static/css/pad.css
@@ -66,7 +66,7 @@ a img {
   padding-left: 1px;
   z-index: 2;
   overflow: hidden;
-  float: left
+  float: left;
 }
 .toolbar ul li {
   float: left;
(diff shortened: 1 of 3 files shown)


muxator:
There are some remaining issues with the default skin (Colibris).
This is what I see when I resize my *Firefox 74* window while running Etherpad bb868beb9cf6 with this modification on top:

1. when the window is large everything is ok:


2. narrowing the window, some icons disappear (this is the original intended behaviour by @seballot, that this patch wanted to change, and we are probably going to change in a different way to solve #3697):


3. after narrowing some more pixels, I @pkrasicki's patch kicks in, but it's too late :), and I guess some alignments are off:


@pkrasicki, from your screenshot, I think you tested on the old skin. Is it so? If so, it would be good to find a way that is compatible with both old & new skin. Let's discuss on #3697.

pkrasicki:
I tested with the default settings and current develop branch. I didn't know there was any other skin included by default. Can you tell me more? Where are the files located for that other skin?

pkrasicki:
Or I can abandon this if https://github.com/ether/etherpad-lite/pull/3709 is going to be merged.

pkrasicki:
@muxator I can't reproduce your issue with default settings, but there is some strange behaviour in responsive design mode that I would have to look into.

muxator:
> I didn't know there was any other skin included by default. Can you tell me more? Where are the files located for that other skin?

Hi @pkrasicki, support for multiple skins was introduced in Etherpad **1.7.5**, along with an experimental skin (_Colibris_, contributed by the community). As starting point, have a look at the [changelog for 1.7.5](https://github.com/ether/etherpad-lite/blob/1.8.0/CHANGELOG.md#175).

A skin can be selected with the `skinName` configuration parameter in `settings.json`. For a description have a look at the [relevant lines in `settings.json.template`](https://github.com/ether/etherpad-lite/blob/1.8.0/settings.json.template#L72-L83).

**Colibris** became the default for new installations in **1.8.0** (i.e., it was put in `settings.json.template`, from wich settings.json is copied when a new instance starts for the first time).

If your install is not recent, chances are that your `settings.json` does not contain the `skinName` parameter, and thus your setup continues to run with the original appearance.

pkrasicki:
Fixed. It works with both themes now.

etherpad-toolbar-mobile-no-skin

etherpad-toolbar-mobile-colibris

muxator:
We are probably going to merge #3709 instead, due to considerations regarding what happens with more than two rows of toolbars.

Thanks @pkrasicki for this proposal!

PR #3708, opened 2020-03-12
Fixes #3707

### LINKED ISSUE #3707: Adding plugin items to toolbar then removing the plugin causes server to crash w/ no explanation
For example see https://github.com/JohnMcLear/ep_align/issues/13
Add these to settings.json
Restart Server
Works
Remove ep_align plugin
Restart Server
Dead server.

JohnMcLear:
https://github.com/ether/etherpad-lite/issues/3707

Also reports the error to the server admin / plugin dev via console.warn.

I can't see any security considerations here but haven't done thorough testing.

### DIFF
--- a/src/node/utils/toolbar.js
+++ b/src/node/utils/toolbar.js
@@ -68,7 +68,7 @@ ButtonsGroup.prototype.addButton = function (button) {
 };
 
 ButtonsGroup.prototype.render = function () {
-  if (this.buttons.length == 1) {
+  if (this.buttons && this.buttons.length == 1) {
     this.buttons[0].grouping = "";
   }
   else {
@@ -80,7 +80,7 @@ ButtonsGroup.prototype.render = function () {
   }
 
   return _.map(this.buttons, function (btn) {
-    return btn.render();
+    if(btn) return btn.render();
   }).join("\n");
 };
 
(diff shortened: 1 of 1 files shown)


src/node/utils/toolbar.js
```diff
@@ -80,7 +80,7 @@ ButtonsGroup.prototype.render = function () {
   }
 
   return _.map(this.buttons, function (btn) {
-    return btn.render();
+    if(btn) return btn.render();
```
JohnMcLear: This is probably not needed.

src/node/utils/toolbar.js
```diff
@@ -68,7 +68,7 @@ ButtonsGroup.prototype.addButton = function (button) {
 };
 
 ButtonsGroup.prototype.render = function () {
-  if (this.buttons.length == 1) {
+  if (this.buttons && this.buttons.length == 1) {
```
JohnMcLear: Again maybe not needed but maybe worth having just for some weirdness

JohnMcLear:
Any reason we can't merge @muxator ?

muxator:
Sure we can merge it.

If possible, next time please do a PR from a personal fork, so it can eventually be reworked without interfering with the main repo. Or directly commit to master if you think there is no need for review. :)

PR #3709, opened 2020-03-12
ui: refs #3697 improve toolbar on mobile

seballot:


muxator:
@seballot, I have problems editing this PR, probably because it was not done on a dedicated branch in your repo. Next time try to propose your work in a dedicated branch and let's see if I am able to work on it.

However, I have **accepted these changes** verbatim in the main repo, **at 0d61d6bb13e2**, with the only modification of the commit message, in order to describe the previous and the current behaviour, and the fact that we expect to implement a "+" button in the future.

This PR, thus, is being rejected for technical problems.

Thanks.

PR #3713, opened 2020-03-15
3546

### LINKED ISSUE #3546: API handler doesn't set a statusCode for incorrect function names
The REST API handler returns a 404 for an incorrect API number, and 401 for an incorrect API key, but doesn't change the status code if the specified function does not exist in the API.

JohnMcLear:
Fixes #3546


muxator:
Thanks for the PR.

Accepting a rewritten version in a0579c90dbcc2677ef690d3149c1a28e3e39a73b because per the [contribution guidelines](https://github.com/ether/etherpad-lite/blob/a0579c90dbcc2677ef690d3149c1a28e3e39a73b/CONTRIBUTING.md#pull-requests) PRs should be linear and should describe the reason for the change.

This was a trivial fix so it was easier to directly do it.

PR #3714, opened 2020-03-15
toolbar: explicitly add the "star" button for rw pads that "lost" it

### LINKED ISSUE #3702: The "star" button disappears for all pads after visiting a readonly pad
When I use a read-only link the little star-icon disappears. It does not just disappear from one pad. All the old pads and all the newly generated pads no longer have the star in the toolbar.
When I restart the etherpad-service the little star appears again.

JohnMcLear:
Fix #3702

Not a super elegant fix but does the job and should work fine.  I don't see why we re-render the toolbar and don't just restrict savedrevision call and hide the button with CSS but whatever.


JohnMcLear:
Bump @muxator

muxator:
I tried to replicate the problem yesterday, but I was not able to understand it very clearly.

Are you confident that the fix fixes a real problem? I am blind here. If you are, please let's describe it a bit better & rebase & pull in.

JohnMcLear:
1. Create a pad http://whatever.com/p/foo, note you have a star in the toolbar
2. Click the drop down to get the read only URL, copy and paste that into the browser window and visit that pad.
3. Visit that pad URI.
4. Note that the star has dissapeared.
5. Restart server, hit f5, star has come back.

It's due to the way Etherpad "creates" a toolbar for each visitor.

muxator:
Ok thanks for the description.

I have tried it. There is a thing that is not clear.

What happens now:
a. create a pad http://whatever.com/p/foo. I have a star in the toolbar
b. visit the read only URL for "foo"
c. revisit the original pad URI (not read only) http://whatever.com/p/foo. The star is there (before this PR it disappeared)
d. for the other RW pads the star is there, and that's ok (it disappeared before).

This behaviour sounds correct to me: "read only" behaves as a property of the specific view and not of the pad in itself. But from point 3 of your comment it seems - instead - that visiting the read only link of a pad should change its nature, and the star should not be there even when visiting the "normal" pad URL.

What's the correct behaviour?

JohnMcLear:
The behavior in this PR is correct.

"that visiting the read only link of a pad should change its nature, and the star should not be there even when visiting the "normal" pad URL."

No, that's not hte behavior, it should not "stick", once you return to the RW URL the star should be present.

muxator:
Thanks for the explanation.

Merged after rewording the commit message.

Issue #2107, opened 2014-03-14
Uncaught Error: Failed assertion: Invalid changeset (checkRep failed) 


SkoricIT:
Hey guys. We are using stable and have the problem that some pads randomly stop working and throw an uncaught error in the console.

```
Uncaught Error: Failed assertion: Invalid changeset (checkRep failed) 
```

Example:

```
https://etherpad.tugraz.at/p/l3tsbet
```

When this happens, the "loading" overlay blocks any action. It's unlikely to be a copy&paste issue because it sometimes happens to entirely handwritten pads. 

An interesting thing is, that the timeslider (opened by appending /timeslider to the url) always works without problems.

```
https://etherpad.tugraz.at/p/l3tsbet/timeslider
```

Right now we are manually fixing the pads by exporting+importing with HTML (losing all changesets). Any idea whats wrong?

akosiaris:
Hello, we 've just had this occur. We run 1.5.7, that is the last released version. Backend is a MySQL database. I have no indications as to user actions that might have caused this. 

Pad in question: https://etherpad.wikimedia.org/p/iOS-iteration-planning

Getting the pad data out via the timeslider trick works fine. The pad however will never load. 

Anything in the database can be dumped and provided for debugging if it helps.

webzwo0i:
hi, I had this discussion in the morning. 
08:47 < webzwo0i> mutante: i debugged the pad. you normally should not do this, but if you have a database backup (and after making export/etherpad to have a backup of 
                  the pad) you can change three database entries and the pad should be working fine again. the three mysql commands are
08:47 < webzwo0i> mysql> update `store` set `value` = replace(`value`,'????','??') where `key` like "pad:iOS-iteration-planning";
08:47 < webzwo0i> mysql> update `store` set `value` = replace(`value`,'????','??') where `key` like "pad:iOS-iteration-planning:revs:7200";
08:48 < webzwo0i> mysql> update `store` set `value` = replace(`value`,'????','??') where `key` like "pad:iOS-iteration-planning:revs:7300";
08:49 < webzwo0i> can you check if your database is running utf8mb4 charset or utf8?
08:51 < webzwo0i> oha ups please dont apply the mysql-commands. maybe i was a little bit to fast :-) need to check something first
09:08 < webzwo0i> mh nope should be fine, please test... it would be good to know if you ran latest release or something else and which plugins you have enabled
[...]

PR #3717, opened 2020-03-15
Check pad deltas - Ready for Review / Merge @muxator

JohnMcLear:
As per https://gist.github.com/marcelklehr/a78d293571e7f06e3cf9

Requested in https://github.com/ether/etherpad-lite/issues/2107


JohnMcLear:
Mheh I didn't rebase but I'm pretty sure the showChat stuff is good to merge so all good to just use this.

muxator:
So this PR just need to add `bin/checkPadDeltas.js`, right?

muxator:
So I went ahead and downloaded [Marcel's original version](https://gist.github.com/marcelklehr/a78d293571e7f06e3cf9) from the url you mentioned, and then put yours on top.

This results in the following diff, which is difficult to understand: https://github.com/ether/etherpad-lite/pull/3717/commits/0d3e825d996c6f4b9892a3bd1b93c17f2a0b6cf9

Could you explain it? From the initial comment i thought it was similar or equal to the file you linked, but there are many differences.

Any pointers? Thanks

JohnMcLear:
Afaik I used async / await that's pretty much all, I think I had to do some polish because something was broken, remember stuff like pad.getPadAuthors was b0rked in 1.7 or so

muxator:
Pulled in, thanks.

Issue #3710, opened 2020-03-14
UnhandledPromiseRejectionWarning when importing a file with unknown (or no) extension


YSelfTool:
This happens in 1.8:

```
[ERROR] console - (node:27731) UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_CALLBACK]: Callback must be a function
   at makeCallback (fs.js:136:11)
   at Object.rename (fs.js:578:14)
   at doImport (/opt/etherpad-lite/src/node/handler/ImportHandler.js:104:16)
   at process._tickCallback (internal/process/next_tick.js:68:7)
```

This happens when importing a file without known extension in the file name (e.g. no extension at all). As a workaround, you have to set the file name and let it end with an extension, e.g. `.txt`.

---
**Edit** by @muxator: this happens not only when no extension is given, but for every unknown extension (including none) when `allowUnknownFileEnds` in `settings.json` is `true`. Changed the title accordingly.

JohnMcLear:
Is this unexpected behavior?  Is this a new bug or was this behavior like this in previous versions?

I'm not against simply rejecting files without any extension....

JohnMcLear:
Ah okay so this was due to the async work cc @raybellis

https://github.com/ether/etherpad-lite/blob/develop/src/node/handler/ImportHandler.js#L104

This is the offending line, if you upload a whatever.md etc to Etherpad you will see this error.

Do you have time to take a look @raybellis 

For me for plugin development I just hack around it.

JohnMcLear:
https://github.com/ether/etherpad-lite/pull/3718/files is my hacky workaround. @raybellis

raybellis:
The non-hacky fix is to replace the call to `fs.rename(src, dst, cb)` with `await fsp_rename(...)`

raybellis:
Ah, I see tudorconstantin already beat me to it :)

muxator:
Fixed pulling in #3722 by @tudorconstantin.
Thanks everyone.

PR #3718, opened 2020-03-16
Import handler

JohnMcLear:


muxator:
For fixing #3710, pulled in #3722 instead of this one.
Thanks.

Issue #3616, opened 2019-06-23
Regex Denial of Service requires a semver major update to clean-css


muxator:
Etherpad 1.7.5 uses `clean-css@3.4.19`, which contains a Regular Expression Denial of Service.
A fix would require updating to `clean-css@4.2.1`, as shown by `npm audit`:

```
$ npm audit
                                                                                
                       === npm audit security report ===                        
                                                                                
# Run  npm install clean-css@4.2.1  to resolve 1 vulnerability
SEMVER WARNING: Recommended action is a potentially breaking change
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ clean-css                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ clean-css                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ clean-css                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/785                       │
└───────────────┴──────────────────────────────────────────────────────────────┘

[...]
```

The [documentation for clean-css 4.2.1 explains the breaking change]((https://github.com/jakubpawlowicz/clean-css/tree/v4.2.1#important-40-breaking-changes):

>clean-css 4.0 introduces some breaking changes:
>
>  - `root`, `relativeTo`, and `target` options are replaced by a single `rebaseTo` option - this means that rebasing URLs and import inlining is much simpler but may not be (YMMV) as powerful as in 3.x;

And this change affects Etherpad:
https://github.com/ether/etherpad-lite/blob/357780d573a055979ec3917d8b3a95ea9d55fa6e/src/node/utils/Minify.js#L418

muxator:
Also, it is to be noted that (apart from the difficulty in making `@import` statements work in clean-css `4.x` (see https://github.com/ether/etherpad-lite/pull/3721#issuecomment-600919716), the library is not actively maintained at the moment.

See this comment by the author, who is doing releases just because of kindness:
> [...] The project is not actively maintained anymore and I'm looking into handing it over. [...]
>
> Btw I can release the next version if there's a need.

_Originally posted by @jakubpawlowicz in https://github.com/jakubpawlowicz/clean-css/issues/1105#issuecomment-577150311_

jakubpawlowicz:
[...]

PR #3721, opened 2020-03-16
Update clean-css 3.4.19 -> 4.2.3

JohnMcLear:
Fixes https://github.com/ether/etherpad-lite/issues/3616


muxator:
I am not being able to validate this. Indeed, it seems to be the same problem I got at the time.

On my machine, I modified `compressCSS()` inside `Minify.js` to print the contents of `minified.styles`. Then I took the minified css from the console output, and grepped for it in the filesystem.

- before this patch, grep finds that text inside `var/minified*` as expected.
- after this patch, I cannot find a match.

Superficially, the application appears to work in both case, but from what I see this could introduce a regression.

@JohnMcLear, can you verify it?

p.s.: obviously, in both cases `minify` is `true` in `setings.json`

JohnMcLear:
Good shout, lemme check what's going on.  I wonder if the minified content is being stored in cache and therefor minified into the file system is now redundant..   Gimmea few hours

JohnMcLear:
Lol did you forget to visit a pad for it to create the minified CSS?  works fine for me :D  I see the files.

muxator:
Yes, I cleared the `var/minified*` files before each test, and then visited a pad before inspecting their contents.

What changes for me is the contents of the files, verified through grepping. Did you try this?

JohnMcLear:
No, sorr,y I didn't check the contents of the files.  My bad..  Gimme a few hours

JohnMcLear:
It looks like this PR is working as expected, I'm not sure why you are experiencing the behavior you are but it's worth that we documented it.  My "guess" is that the variable name is being shortened or so but looking through the files I'm not sure that's the case.  It's possible I don't understand how this library works but as the results seem right I'm not convinced..  I guess one thing to check would be file size difference...

I will do that now..

muxator:
I am going to document the steps.

JohnMcLear:
cleancss fix branch:
pad: 1.2Mb in 2.5s with minify on
pad: 2.3Mb in 2.02s with minify off

develop branch:
pad: 1.1Mb in 2.5s with minify on
pad: 2.3Mb in 2s with minify off

I'd want someone to validate these load times tho, I found my browser to not be a suitable tool as they fluctuated..  Still we know there will be a minor performance hit from this change.

So

1) It's defo minimizing
2) It appears to be delivering all the correct files properly
3) I am not aware of any other voodoo going on

muxator:
I agree it's puzzling.

1. > It's defo minimizing

   I am not a native speaker, sorry. I do not understand what "defo" means :)

2. > It appears to be delivering all the correct files properly

   I am really not sure of this. If you have a look at the **contents** of the `var/minified_*` files you'll see that the minified CSS statements generated by `compressCSS()` do not seem to end up there. I'll write a simple procedure for replicating it. I have already done it various times

3. > I am not aware of any other voodoo going on

   Etherpad appears to load correctly, anyway. But then, I suspect that this is just the result of another bug masking this one (maybe the minimization never really worked?)

To move this forward: among the [breaking changes of clean-css 4.0](https://github.com/jakubpawlowicz/clean-css#important-40-breaking-changes) there is this one:

* `root`, `relativeTo`, and `target` options are replaced by a single `rebaseTo` option - this means that rebasing URLs and import inlining is much simpler but may not be (YMMV) as powerful as in 3.x;

Maybe we could try to use `{rebaseTo: base}` instead of `{relativeTo: base}` in `src/node/utils/Minify.js`?

muxator:
@JohnMcLear, do you ever see your browser requesting any of the `minified_` files? I guess Etherpad is continuing to work because those files (which are generated by `src/node/utils/caching_middleware.js`) are **never** used.

Just to be sure, I am running with `settings.minified: true` and `NODE_ENV=production`.

slel:
@muxator "defo" is for "definitely, see
https://www.urbandictionary.com/define.php?term=Defo

JohnMcLear:
@muxator note the results from https://github.com/ether/etherpad-lite/pull/3721#issuecomment-600291631
Note the page size is reduced by ~50% with minify set to true.  -- How we can claim minification isn't working when we can see that reduction boggles my noggin.. :D

muxator:
I have more data.

- skin: **Colibris** (since it is the default now)
- run mode: **production**

In the current development version (826122932380 as of now), the minimization infrastructure **works**, my bad. I was confused by another error introduced by this PR, please read on.

With this PR, **the minimization on the Colibris-specific files regresses**. Etherpad remediates serving the un-minified files, and not concatenating the CSSes.

You can see this looking at the `# of requests` column in the following table and having a look at the errors in the logs (after applying the patch at the end of this message):

Version | Minify | # of requests | total size | Errors minifying? | Finish time
---|---|---|---|---|---
826122932380 (latest on develop)| false | 92 | 2.67 MB / 2.70 MB | no | 5.79 s
826122932380 (latest on develop)| true | **29** | 1.20 MB / 1.21 MB | **no** | 3.99 s
This PR | true | **88** | 1.27 MB / 1.30 MB | **yes** | 7.67 s

An example of the error that can be found in the logs if you apply the patch given at the end of the message can give guidance to fix the problem:

```
[2020-03-19 01:08:41.977] [ERROR] console - CleanCSS.minify() returned an error on skins/colibris/pad.css (base CSS path: /mnt/md/hot-data/documenti/codice/etherpad/etherpad-lite/src/static/skins/colibris): Ignoring local @import of "src/layout.css" as resource is missing.,Ignoring local @import of "src/pad-editor.css" as resource is missing.,Ignoring local @import of "src/components/buttons.css" as resource is missing.,Ignoring local @import of "src/components/popup.css" as resource is missing.,Ignoring local @import of "src/components/chat.css" as resource is missing.,Ignoring local @import of "src/components/sidediv.css" as resource is missing.,Ignoring local @import of "src/components/gritter.css" as resource is missing.,Ignoring local @import of "src/components/table-of-content.css" as resource is missing.,Ignoring local @import of "src/components/toolbar.css" as resource is missing.,Ignoring local @import of "src/components/users.css" as resource is missing.,Ignoring local @import of "src/components/form.css" as resource is missing.,Ignoring local @import of "src/components/import-export.css" as resource is missing.,Ignoring local @import of "src/plugins/brightcolorpicker.css" as resource is missing.,Ignoring local @import of "src/plugins/comments_page.css" as resource is missing.,Ignoring local @import of "src/plugins/font_color.css" as resource is missing.,Ignoring local @import of "src/plugins/set_title_on_pad.css" as resource is missing.,Ignoring local @import of "src/plugins/tables2.css" as resource is missing.,Ignoring local @import of "src/plugins/embedded_hyperlinks.css" as resource is missing.,Ignoring local @import of "src/plugins/author_hover.css" as resource is missing.
[2020-03-19 01:08:46.652] [ERROR] console - CleanCSS.minify() returned an error on css/iframe_editor.css (base CSS path: /mnt/md/hot-data/documenti/codice/etherpad/etherpad-lite/src/static/css): Ignoring local @import of "lists_and_indents.css" as resource is missing.
```


Patch to introduce more logging so the errors can be seen:
```diff
diff --git a/src/node/utils/Minify.js b/src/node/utils/Minify.js
--- a/src/node/utils/Minify.js
+++ b/src/node/utils/Minify.js
@@ -418,6 +418,7 @@ function compressCSS(filename, content, 
     new CleanCSS({relativeTo: base}).minify(content, function (errors, minified) {
       if (errors) {
         // On error, just yield the un-minified original.
+        console.error(`CleanCSS.minify() returned an error on ${filename} (base CSS path: ${base}): ${errors}`);
         callback(null, content);
       } else {
         callback(null, minified.styles);
@@ -425,6 +426,7 @@ function compressCSS(filename, content, 
     });
   } catch (error) {
     // On error, just yield the un-minified original.
+    console.error(`Unexpected error minifying ${filename} (base CSS path: ${base}): ${error}`);
     callback(null, content);
   }
 }
```

JohnMcLear:
Mheh, balls, will try look at this today.

JohnMcLear:
This PR, not using the new skin = works fine.  So it's the new skin that's causing the issue.  Will see who/how that got broken.  I wasn't involved in skins afaik.

JohnMcLear:
Related to https://github.com/jakubpawlowicz/clean-css/issues/1112 ??  

Because this is skin related I'm gonna bow out at this point, I have no idea how the skin stuff was implemented and I think my efforts are better spent on core Etherpad stuff :)

muxator:
This is not related only to Colibris: I only have reported data for Colibris because that is the default skin.

The problem happens even with the legacy skin:
```
[2020-03-21 02:01:13.945] [ERROR] console - CleanCSS.minify() returned an error on css/iframe_editor.css (base CSS path: <BASEPATH>/src/static/css): Ignoring local @import of "lists_and_indents.css" as resource is missing.
```

It really has to do with how `@import` is managed in clean-css **4.x**.

muxator:
Closing

PR #3722, opened 2020-03-16
Fix allowUnknownFileEnds in 1.8.0 when importing files with unknown extension

tudorconstantin:
`fs.rename()` is callback based, we need `fsp_rename()` which is declared above and is promised based, in order to catch the `unhandled promised rejection ...` warning

PS: I haven't tested this at all. I was looking at what kind of issues are open for this project and this one seems the obvious fix. Also, sorry if this PR is useless, feel free to close it, especially if it causes more work for integrating it than it helps.


muxator:
Queued, thanks.

This closes #3710 and replaces #3718.

PR #3730, opened 2020-03-18
ui: Initialize color picker for user's color - Ready for Review / Merge.  Merge this after merging the other testing code though because I want to check the tests pass. @muxator

### LINKED ISSUE #3617: Initial author color selecter is not set on current author color
When opening the color selecter on the first time, the selected color is not the current color.

See the screenshot : The initial author color is blue, but the color selecter is on grey-ish. It should be on blue.

Capture d’écran du 2019-06-28 11-30-40

orblivion:
Initialize the color picker with user's current color. Was previously initialized with grey in farbtastic library. Fixes #3617

### DIFF
--- a/src/static/js/pad_userlist.js
+++ b/src/static/js/pad_userlist.js
@@ -809,6 +809,7 @@ function closeColorPicker(accept)
 function showColorPicker()
 {
   previousColorId = myUserInfo.colorId;
+  $.farbtastic('#colorpicker').setColor(myUserInfo.colorId)
 
   if (!colorPickerOpen)
   {
--- a/tests/frontend/specs/change_user_color.js
+++ b/tests/frontend/specs/change_user_color.js
@@ -1,0 +1,99 @@
+describe("change user color", function(){
+  //create a new pad before each test run
+  beforeEach(function(cb){
+    helper.newPad(cb);
+    this.timeout(60000);
+  });
+
+  it("Color picker matches original color and remembers the user color after a refresh", function(done) {
+    this.timeout(60000);
+    var chrome$ = helper.padChrome$;
+
+    //click on the settings button to make settings visible
+    var $userButton = chrome$(".buttonicon-showusers");
+    $userButton.click();
+    
+    var $userSwatch = chrome$("#myswatch");
+    $userSwatch.click();
+
+    var fb = chrome$.farbtastic('#colorpicker')
+    var $colorPickerSave = chrome$("#mycolorpickersave");
+    var $colorPickerPreview = chrome$("#mycolorpickerpreview");
+
+    // Same color represented in two different ways
+    const testColorHash = '#abcdef'
+    const testColorRGB = 'rgb(171, 205, 239)'
+
+    // Check that the color picker matches the automatically assigned random color on the swatch.
+    // NOTE: This has a tiny chance of creating a false positive for passing in the
+    // off-chance the randomly assigned color is the same as the test color.
+    expect($colorPickerPreview.css('background-color')).to.be($userSwatch.css('background-color'))
+
+    // The swatch updates as the test color is picked.
+    fb.setColor(testColorHash)
+    expect($colorPickerPreview.css('background-color')).to.be(testColorRGB)
+    $colorPickerSave.click();
+    expect($userSwatch.css('background-color')).to.be(testColorRGB)
+
+    setTimeout(function(){ //give it a second to save the color on the server side
+      helper.newPad({ // get a new pad, but don't clear the cookies
+        clearCookies: false
+        , cb: function(){
+          var chrome$ = helper.padChrome$;
+
+          //click on the settings button to make settings visible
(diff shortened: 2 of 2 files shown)


JohnMcLear:
Awesome, any chance you could also add some test coverage?  See https://github.com/ether/etherpad-lite/tree/develop/tests/frontend/specs for other test specs

Run tests by visisting /tests/frontend

Cheers :D

orblivion:
I'm having an interesting time re-running tests after making changes to the files. I try restarting the server, restarting my browser, etc. It stays blank a lot of the time. Any advice on getting this to be consistent?

JohnMcLear:
Do you have minify to false, maxAge to 0 and running etherpad in development mode (it will show on startup)?

orblivion:
I've been in dev mode. I changed those two settings and it seems to be working. thank you!

orblivion:
I guess it's not helping consistently.

orblivion:
It seems to correlate to whether I currently have the file open in vim.

orblivion:
I went off of the user name test and decided to add basic coverage for the color since I didn't see it anywhere. It includes coverage for my fix here.

I should note - it seems that the database doesn't reset per test runs, and the same goes for the user name test. This easily leads to a false positive for passing. (Though, I did delete the other test files while I was testing to speed it up. Maybe there's a database reset directive in one of those files?)

JohnMcLear:
It makes sense for the database to keep things persistent, but yeah, I assume the author value is being reassigned, I'm not sure of a method to get a "new" author, perhaps faking some browser property or deleting some session / cookie on the client side?  It's easy to programmatically delete cookies, did you try that?

orblivion:
So, there's a `clearCookies` function in `helper.js`, and it's called by default as part of `newPad` (there's an option to not call it). However `clearCookies` does not work as named. Firstly, it seems you have to set each cookie with a past expiration date (as you do correctly elsewhere within these tests). Secondly, for some reason the cookie clearing doesn't seem to take effect in due time within `newPad`. It seems to need a delay in there; even a timeout of 1 seems to do the trick.

I made a WIP to fix both of these. Now it works as expected. Except now it breaks a bunch of other tests. Do you have any thoughts on what to try?

https://github.com/ether/etherpad-lite/compare/develop...orblivion:fix-clear-cookies-tests?w=1

My "false positive" concern here and in the name change would be fixed by fixing the helper as described above. At this point it seems to be orthogonal to this PR. Could we make that a separate fix and move on with this PR?

JohnMcLear:
I doubt @muxator would be happy to see MORE setTimeouts land in the source :D 

I'm not sure cookies aren't cleared "before" newPad though, you would have thought that would be the case.  But maybe it's supposed to be 

Do a test, test cookies etc.  Clear cookies..  Either way ``window.document.cookie = ""`` is wrong and your fix is better.  I guess one solution would be to make clearCookies async or do a callback IE ``clearCookies(function(){if(!padname....`` or ``await clearCookies()...``

Can you keep looking into it today?  I have to work on this other problem and I'm starting super late today ;\

JohnMcLear:
Okay wtf.  I can replicate your testing issue and it's horrible.  I'm gonna debug that for you.  Good find!

JohnMcLear:
Closer.  The spec framework for some reason is using the .swp file for the specs...  Fixing, assuming I can.

JohnMcLear:
Fixed in https://github.com/ether/etherpad-lite/pull/3745

Feel free to pull that and test.  I'm gonna go back to working on fixing broken tests now.

orblivion:
To be clear - what you replicated was related to editing with vim? Not the cookie issue?

orblivion:
Confirmed that merging in #3745 fixes the vim problem :+1:

orblivion:
Strange, I wasn't able to replicate my own problem with the cookies that supposedly required the timeout. But, I wrote a test that would catch it if it were to ever happen again, and I put in the other part of the cookie fix, and the result is now in my working branch for this issue:

https://github.com/ether/etherpad-lite/compare/develop...orblivion:fix-clear-cookies-tests

After confirming that I didn't break any more tests, I'll make a new PR for that.

JohnMcLear:
The new tests should be merged into develop today so just continue on this tomorrow with a pull of ether/develop into your branch.

orblivion:
Sounds good. I ended up trying the existing tests already and it seems to not break anything new. Interestingly none of the `importexport.js` tests seem to pass for me irrespective of my change.

orblivion:
(Firefox 68 esr on Debian on Qubes)

JohnMcLear:
Interesting RE importexport tests not working.  Same in Chrome/Edge et al?  What database backend?  Does import/export work in the UI?  Thanks!

@muxator btw this can be reviewed/merged..  The conversation has switched to something else

muxator:
The _peculiar_ caching behaviour from Etherpad hit me when testing, too. I had to try three times. :smile: 

Pulled in, thanks.

PR #3734, opened 2020-03-20
Bump ueberdb - Ready for Review / Merge @muxator

JohnMcLear:
Bumping UeberDB to use latest version which includes a UTF8 database check on startup.


muxator:
Pulling in.

The new ueberDB version (**0.4.1**) includes the following changes:

- support for connection pooling on Postgres
- support for CockroachDB (via the postgresql driver)
- the mysql driver checks at startup that dbSettings.charset in settings.json and the actual charset of the DB is the same
- setting of a new default so that instances without an explicit configuration are advised to move towards utf8mb4

Moreover, the repository was transferred under https://github.com/ether/ueberDB

Issue #3568, opened 2019-03-01
HTTP API needs POST for large parameters in NodeJS 8.14+


raybellis:
NodeJS 8.14.0 introduced a limit of 8KB to the size of HTTP headers that can be received.

This prevents HTTP API methods such as `setText` from working if the text content exceeds that size.

A possible solution would be to support the `POST` method in addition to `GET`, with any JSON object found in the HTTP body augmenting the URL parameters.

raybellis:
On further inspection, the API already has _some_ support for `POST`, but it's not documented.

raybellis:
Potential fix in branch `http-post` (commit 882b93487fe2bbd47a1cbab1b7d7b4b16bae5015) which allows all backend tests to pass 

If the preference is to allow _only_ body parameters in an HTTP POST then please advise, and I'll amend the patch and tests accordingly.    Also, the EP docs make no mention of HTTP POST support nor how to use it.   I'll wait to submit a PR until I hear back.

muxator:
I did not write the API spec, and it is on the wild for a long time now, so the following is just


I think it is more reasonable to maintain backwards compatibility as much as possible, and so retaining GET seems sensible.

raybellis:
In which case I think you should be able to cherry pick that commit, although the API documentation does need an update to mention POST and the need to use it if the data is too large.

muxator:
(sorry I hit CTRL+ENTER and posted an half backed comment).

Apart from the documentation (that should be updated), would it be possible to print a WARN in the logs when receiving a GET with a too big payload on node 8.14+? That would give an operator some hints about what needs to be done to maintain a working installation.

If, instead, it is the node runtime that discardeds the request at a low level, it would never hit the application, and nothing could be done.

raybellis:
I'm a little fuzzy on the details right now, but ISTR it was the latter.

The question on `GET` vs `POST` was not about deprecating the former, but whether the `POST` handler should work as my patch currently does, and _combine_ the parameters supplied in the body with those in the query string (with the former taking precedence) or alternatively completely ignore any that are in the query string and only use those found in the `POST` body.

The disadvantage of the "body only" approach is that it then becomes mandatory to put things like the `padID` field in the body (for `POST` requests).

muxator:
The approach taken in your patch (`GET` parameters are read, and eventually overridden by `POST` parameters of the same name) is the most sensible.

[...]

PR #3737, opened 2020-03-20
Backend test for utf8 encoding - Ready for Review/Merge @muxator

JohnMcLear:
This PR introduces better handling for databases that don't have proper UTF8 encoding.

It includes a backend Test for UTF8 encoding which works by reading a [properly formed HTML document that includes lots of weird characters](https://github.com/ether/etherpad-lite/pull/3737/files#diff-9511b7171d0f22a41e2d6000d4c8934a) which are then imported into Etherpad using The API method ``setHTML``.    This can be merged separate to #3737 as one test is native to Etherpad and the other native to UeberDB.

If the database is improperly configured these tests should crash the Etherpad instance (intentionally).  Along with https://github.com/ether/ueberDB/commit/a5db6c41e17a1cb06623e8d76a1bddf7e52118a8 // https://github.com/ether/etherpad-lite/pull/3734 this will also provide a test a ``init`` of database.

This PR also fixes ``setText`` and ``setHTML`` syntax in the pad backend tests.  Previously if you used ``.get`` with ``setText`` or ``setHTML`` those tests would fail if the ``text`` or ``html`` variables were beyond the capacity of ``get``

This PR also (sorry!) should resolve the issue of uploading "emojis" through the UI.  It does this by replacing the ``isAscii`` check with 

```
    let bytelike = unescape(encodeURIComponent(text));
    text = decodeURIComponent(escape(bytelike));
```

It works as well as copy/pasting and the setHTML.  Basically the following characters don't work in Etherpad at all.

🏋️‍♀️ 🏋🏻‍♀️ 🏋🏼‍♀️ 🏋🏽‍♀️ 🏋🏾‍♀️ 🏋🏿‍♀️ 🏋️‍♂️ 🏋🏻‍♂️ 🏋🏼‍♂️ 🏋🏽‍♂️ 🏋🏾‍♂️ 🏋🏿‍♂️ 🤼‍♀️ 🤼‍♂️ 🤸‍♀️ 🤸🏻‍♀️ 🤸🏼‍♀️ 🤸🏽‍♀️ 🤸🏾‍♀️ 🤸🏿‍♀️ 🤸‍♂️ 🤸🏻‍♂️ 🤸🏼‍♂️ 🤸🏽‍♂️ 🤸🏾‍♂️ 🤸🏿‍♂️ ⛹️‍♀️ ⛹🏻‍♀️ ⛹🏼‍♀️ ⛹🏽‍♀️ ⛹🏾‍♀️ ⛹🏿‍♀️ ⛹️‍♂️ ⛹🏻‍♂️ ⛹🏼‍♂️ ⛹🏽‍♂️ ⛹🏾‍♂️ ⛹🏿‍♂️ 🤺 🤾‍♀️ 🤾🏻‍♀️ 🤾🏼‍♀️ 🤾🏾‍♀️ 🤾🏾‍♀️ 🤾🏿‍♀️ 🤾‍♂️ 🤾🏻‍♂️ 🤾🏼‍♂️ 🤾🏽‍♂️ 🤾🏾‍♂️ 🤾🏿‍♂️ 🏌️‍♀️ 🏌🏻‍♀️ 🏌🏼‍♀️ 🏌🏽‍♀️ 🏌🏾‍♀️ 🏌🏿‍♀️ 🏌️‍♂️ 🏌🏻‍♂️ 🏌🏼‍♂️ 🏌🏽‍♂️ 🏌🏾‍♂️ 🏌🏿‍♂️ 🏇 🏇🏻 🏇🏼 🏇🏽 🏇🏾 🏇🏿 🧘‍♀️ 🧘🏻‍♀️ 🧘🏼‍♀️ 🧘🏽‍♀️ 🧘🏾‍♀️ 🧘🏿‍♀️ 🧘‍♂️ 🧘🏻‍♂️ 🧘🏼‍♂️ 🧘🏽‍♂️ 🧘🏾‍♂️ 🧘🏿‍♂️ 🏄‍♀️ 🏄🏻‍♀️ 🏄🏼‍♀️ 🏄🏽‍♀️ 🏄🏾‍♀️ 🏄🏿‍♀️ 🏄‍♂️ 🏄🏻‍♂️ 🏄🏼‍♂️ 🏄🏽‍♂️ 🏄🏾‍♂️ 🏄🏿‍♂️ 🏊‍♀️ 🏊🏻‍♀️ 🏊🏼‍♀️ 🏊🏽‍♀️ 🏊🏾‍♀️ 🏊🏿‍♀️ 🏊‍♂️ 🏊🏻‍♂️ 🏊🏼‍♂️ 🏊🏽‍♂️ 🏊🏾‍♂️ 🏊🏿‍♂️ 🤽‍♀️ 🤽🏻‍♀️ 🤽🏼‍♀️ 🤽🏽‍♀️ 🤽🏾‍♀️ 🤽🏿‍♀️ 🤽‍♂️ 🤽🏻‍♂️ 🤽🏼‍♂️ 🤽🏽‍♂️ 🤽🏾‍♂️ 🤽🏿‍♂️ 🚣‍♀️ 🚣🏻‍♀️ 🚣🏼‍♀️ 🚣🏽‍♀️ 🚣🏾‍♀️ 🚣🏿‍♀️ 🚣‍♂️ 🚣🏻‍♂️ 🚣🏼‍♂️ 🚣🏽‍♂️

Try copy / pasting them into any pad, it will format them incorrectly.  This is related to how Etherpad "thinks" about characters as a specific length and tbh I feel right now it's outside of the scope of what I'm trying to achieve.

At first I thought it was Etherpad being a bit old fashioned, then I had the same experience in Atom, Google Docs.  And then other Document editors and IDEs, and discovered, it's a bit of a mess out there...  

So I think my fix is "good enough" but now we have a set of characters we know we can test against so another developer can continue the struggle further to support


JohnMcLear:
@muxator ready for review :)

src/node/handler/ImportHandler.js
```diff
@@ -180,6 +167,8 @@ async function doImport(req, res, padId)
 
   if (!req.directDatabaseAccess) {
     text = await fsp_readFile(destFile, "utf8");
+    let bytelike = unescape(encodeURIComponent(text));
```
muxator: Quick question without having studied this code yet: is the `unescape()`/`escape()` roundtrip avoidable?

I ask this because those functions seem to be on the way of deprecation ([source on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape)):

> All of the language features and behaviours specified in this annex have one or more undesirable characteristics and in the absence of legacy usage would be removed from this specification.
> Programmers should not use or assume the existence of these features and behaviours when writing new ECMAScript code.

muxator:
Nice work, @JohnMcLear, thanks! I'll have a look ASAP.

src/node/handler/ImportHandler.js
JohnMcLear: ``escape`` & it's sister aren't going anywhere any time soon.  This gives us a few years (easily) to focus on handling the characters gracefully inside of the editor.

RE roundtrip, absolutely you could catch non processable characters and drop the import entirely.  For me, in this PR, it was about making import work.  Import is a small amount of processing handled by Etherpad instances and doing a roundtrip here is not computationally expensive..  That said, a DDoS attack using this as a vector is not ideal.  Pushing 4Gb via post to the upload endpoint and doing the roundtrip would not be enjoyable.  But Etherpad has much larger and more exploitable DDoS endpoints than this (think pad export URIs) so it's IMHO not a huge concern/worry.

If you want it to reject "characters" for the current implemented charset from settings.json we can do that but imho this serves it's purpose and our users well.  Imagine the computational power requirements of processing an entire document beyond what we already are doing here though?  .escape and .unescape are, despite being deprecated, optimized.  We can use that to our advantage instead of trying to bake in our own unoptimized character by character check.

muxator: > RE roundtrip, absolutely you could...

Whoa. When one is aiming to sanitize input doing an encode/decode roundtrip is actually a **sound strategy**. I am not thinking about performances right now.

My question was about why use two functions (`encode/decodeURIComponent` + `unescape/escape`) instead of a single one. I am sure there is a reason. Maybe this covers more cases? If so, what are they?

JohnMcLear: Good question, I don't know the answer!  See what happens if you just do decode.  I encoded then decoded because I figured it'd let less noise through and also create a saner output.  Also that's what the SO article did..   https://stackoverflow.com/a/2670211

JohnMcLear:
@muxator bump :D

muxator:
Tonight

muxator:
I am having a lot of **emojons** looking at [this](https://github.com/ether/etherpad-lite/blob/e929692f95f9ed20f825b948779430010a33aee4/tests/backend/specs/api/emojisLess.html)

🤣

muxator:
Wow, still going through...

JohnMcLear:
Are you testing character by character or something? :D

muxator:
Change by change of the original PR.

muxator:
Ok I am done.

I do not know how to use the review interface so I am posting here.
The work can be reduced in three parts. I need an answer on part **3**:

1. 117deb394582: _tests: in backend tests, use POST instead of GET for setText() and setHTML()_

   This switches some API calls from GET to POST on the client side.

   I see you were surprised by the fact that they were GETs in the first place, but this was just how the Etherpad API has always been. :smile:

   For context: the change that **supports both GET & POST in the API** was introduced in fc661ee13af1 (see that commit message and #3568 for the reason behind it). It landed in **1.8.0**.

   No problems here, just for your info.

2. 3964ee8c4613: _tests: backend tests for utf8 & emojis support_
   This is **good work**, much needed. Perfect.

3. ee0e4d3a429a: _ImportHandler: instead of failing (badly) when a non Ascii code is uploaded just convert it then convert it back_
   We already talked about `encode/decodeURIComponent` + `unescape/escape`, that the latter is deprecated, but that we are going to keep it because it helps with utf8 sanitization. I am ok with that.

   I have another doubt about this change, which may impact import/export: please open ee0e4d3a429a and view it in a side by side diff.

   Before the change, we were doing utf8 "sanitization" (actually, failing the upload) only if `useConvertor` was `false`.

   After this change, we replaced that block with a more forgiving one, but moved it inside a block where there is no condition on `useConvertor` (e.g., we are doing this even when using LibreOffice).

   **Is it possible that this could cause problems when importing via LibreOffice**, as we were seeing recently on another thread?

If we are safe with **3** I'll merge this PR right away.

Thanks.

muxator:
I stripped 3 (import/export) from this PR, and moved it in #3796, please comment there, @JohnMcLear.

All the other changes (the utf8 tests and the change from GET to POST) are merged.

JohnMcLear:
<3 looks good.

src/node/handler/ImportHandler.js
rhelmer: > `escape` & it's sister aren't going anywhere any time soon. This gives us a few years (easily) to focus on handling the characters gracefully inside of the editor.
[...]
> cc @rhelmer as he might know the schedule for deprecation of escape and unescape.

Hm. I agree that it is probably not going anywhere soon, let me research a bit more though and see what the recommended way to do this is from standards-y folks.

PR #3743, opened 2020-03-22
Testing travis ci sauce labs - Ready for Review / Merge @muxator

JohnMcLear:
Suggest doing a squash when merging.

Intentionally commited to ether repo.

- [x] SC latest reference file 404'd.  Used explicit sauce labs version file.
- [x] Attempt to remove wd (this can't be removed)
- [x] Make sure this works for @muxator and @lpagliari -- Okay so Travis runs the tests and these settings are set in Travis -  ``SAUCE_ACCESS_KEY / SAUCE_USERNAME`` -- Email sent to @muxator with credentials 
- [x] Check SL is running OS' we want (See changes to remote_runner.js and now we have more coverage!  This was a nightmare to do(and took hours) though because saucelabs OS / platform / browser names don't match what's required in their own docs [platform configuration tool] for some reason...)  -- I tried to make this future oriented with using 'latest' where possible.
- [x] Make tests that have been broken pass.  I'm not fixing tests that fail due to colibris tho.
- [ ] Understand if tests pass who and how the results are passed - Afaik it returns all tests passed even if some tests fail?!  <-- I suggest merging and doing this as an additional task.  The reason being is that until things are in develop it's hard to go between a working / broken state.

### DIFF
--- a/tests/frontend/travis/remote_runner.js
+++ b/tests/frontend/travis/remote_runner.js
@@ -67,41 +67,39 @@ var sauceTestWorker = async.queue(function (testSettings, callback) {
   });
 }, 5); //run 5 tests in parrallel
 
-// Firefox
+// 1) Firefox on Linux
 sauceTestWorker.push({
     'platform'       : 'Linux'
   , 'browserName'    : 'firefox'
-  , 'version'        : ''
+  , 'version'        : 'latest'
 });
 
-// Chrome
+// 2) Chrome on Linux
 sauceTestWorker.push({
     'platform'       : 'Linux'
   , 'browserName'    : 'googlechrome'
-  , 'version'        : ''
+  , 'version'        : 'latest'
 });
 
-/*
-// IE 8
+// 3) Safari on OSX 10.15
 sauceTestWorker.push({
-    'platform'       : 'Windows 2003'
-  , 'browserName'    : 'iexplore'
-  , 'version'        : '8'
-});
-*/
-
-// IE 9
-sauceTestWorker.push({
-    'platform'       : 'Windows XP'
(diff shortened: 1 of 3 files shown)


tests/frontend/travis/remote_runner.js
```diff
@@ -67,41 +67,39 @@ var sauceTestWorker = async.queue(function (testSettings, callback) {
...
-*/
 
-// IE 9
+// 4) IE 10 on Win 8
 sauceTestWorker.push({
-    'platform'       : 'Windows XP'
+    'platform'       : 'Windows 8'
   , 'browserName'    : 'iexplore'
-  , 'version'        : '9'
+  , 'version'        : '10.0'
```
muxator: Would it work if we put **IE 11** here? I know it took a lot of time to find these combinations.

IE 10 is practically out of support.

JohnMcLear: I intentionally chose IE10 and Edge to get opposite ends of the spectrum from MS.  You could bring another blob in for IE11 should you wish.  We're supposed to have 5 concurrent tests run but afaik it only does 4(saucelabs is weird) so it wouldn't slow things down at all..

muxator: Fine. Let's keep IE 10.

muxator:
Pulled in, thanks for the hard work.

Please not I have deleted the original `testing-travis-ci-sauce-labs` in the main repo. A re-clone will be advisable, see https://github.com/ether/etherpad-lite/pull/3745#issuecomment-602875166.

JohnMcLear:
Ack.

PR #3745, opened 2020-03-22
Fix tests during edit - Ready for Review / Merge @muxator

JohnMcLear:
When the test framework requested bold.js it would serve the swap file before the bold.js file causing developers to get confused as to why their changes to spec files weren't being represented upon page refresh.  To fix I remove swap files from the list of files used by Etherpad.

Also....   

Tests: What's changed and why?

Now our Sauce Labs tests worked it seemed like a suitable time to fix tests which over the years have become broken due to browser changes.

 - [x] Reduce duplicate code.  https://github.com/ether/etherpad-lite/pull/3745/files#diff-1f128f89d031a5ed781ca8474c46c7cb -- Fixing all browser inconsistency on testing keypress/down events.
 - [x] Comment out tests that are broken (scroll.js) and create separate issue to resolve them.
 - [x] Fix issue where firefox keeps line breaks in .text() values.

Also you might want to squash these commits.

### DIFF
--- a/tests/frontend/specs/helper.js
+++ b/tests/frontend/specs/helper.js
@@ -20,7 +20,7 @@ describe("the test helper", function(){
     });
 
     it("gives me 3 jquery instances of chrome, outer and inner", function(done){
-      this.timeout(5000);
+      this.timeout(10000);
 
       helper.newPad(function(){
         //check if the jquery selectors have the desired elements
--- a/tests/frontend/specs/helper.js
+++ b/tests/frontend/specs/helper.js
@@ -136,7 +136,15 @@ describe("the test helper", function(){
       helper.selectLines($startLine, $endLine, startOffset, endOffset);
 
       var selection = inner$.document.getSelection();
-      expect(cleanText(selection.toString())).to.be("ort lines to t");
+
+      /*
+       * replace() is required here because Firefox keeps the line breaks.
+       *
+       * I'm not sure this is ideal behavior of getSelection() where the text
+       * is not consistent between browsers but that's the situation so that's
+       * how I'm covering it in this test.
+       */
+      expect(cleanText(selection.toString().replace(/(\r\n|\n|\r)/gm,""))).to.be("ort lines to t");
 
       done();
     });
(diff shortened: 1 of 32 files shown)


JohnMcLear:
Not sure why the sauce labs commit is in? :D  Anyway, this is the fix for the issue experienced in https://github.com/ether/etherpad-lite/pull/3730

muxator:
> Not sure why the sauce labs commit is in? :D Anyway, this is the fix for the issue experienced in #3730

I came here to say it... by mistake this series was based on the saucelab enablement branch #3743.
I'll have to rebase it. It's a bit nasty to do this in the public repository, but this is a secondary branch.

If possible, I'd prefer PR based on private repos so they can be meddled with.

JohnMcLear:
Yea see comment on other PR, it's because I'm working on sauce labs which means I have to commit to ether/ to get the travis runner to init

muxator:
After this one, #3743 and #3749 are merged it would be better if you start over from a new clone, because otherwise there will be dragons ahead. I'll ping you.

JohnMcLear:
Yea that's fine, let's get these in so we can keep moving on.  Once these are done nearly all future commits will land from ``JohnMcLear/``

JohnMcLear:
Can this get merged in or do you want a new clone?  I don't see any conflicts but I'm mindful of # of commits.  (afaik you can squash these on merge)

muxator:
> do you want a new clone?

Not for this PR. It's not worth the effort since I'll have to rewrite the branch and delete it anyway. 

_After_ everything it's merged, I advise you to re-clone the repo on your dev machine because otherwise you'll likey will see fun stuff happening. :)

tests/frontend/specs/helper.js
```diff
@@ -136,7 +136,7 @@ describe("the test helper", function(){
       helper.selectLines($startLine, $endLine, startOffset, endOffset);
 
       var selection = inner$.document.getSelection();
-      expect(cleanText(selection.toString())).to.be("ort lines to t");
+      expect(cleanText(selection.toString().replace(/(\r\n|\n|\r)/gm,""))).to.be("ort lines to t");
```
muxator: This is logically part of the next commit where you do the same in other places, right?

I'll move it there and comment from that point.

muxator: To be clear, I am referring to this: 6a7bcfdb6552f06829bcdba5ace3fca484d0a38a

JohnMcLear: Yea.

JohnMcLear: I'm terrible at committing, I drank too much of the "commit often" juice.. 💯   Sorry about that!  At least I show my work tho, I like to show how dumb I can be / how my brain goes about solving / sorting things so when I have to fix something (in this case 6 years later?) I can see what I was trying to do and why I did something weird.

I do a lot of weird.  But that's okay 🍡

muxator: No problem. I'm just trying to make sure we won't get lost in another 6 years... :)

muxator:
Geez, that's a lot of work you have done! I'll continue tomorrow.

JohnMcLear:
@muxator thanks, lemme know if you get stuck, I have some more tests to merge / send PR for and I don't want to work against a volatile code base.

muxator:
I am taking those, with just a minimum rewrite.

I still have some doubts, but I understand you are still working on it and I do not want to block your progress.

Pulled in, thanks.

PR #3747, opened 2020-03-23
Fix minify cache ms

Chocobozzz:
`if-modified-since` header does not include milliseconds, so https://github.com/ether/etherpad-lite/blob/develop/src/node/utils/Minify.js#L219 will always be `false`

### DIFF
--- a/src/node/utils/Minify.js
+++ b/src/node/utils/Minify.js
@@ -201,6 +201,7 @@ function minify(req, res)
   statFile(filename, function (error, date, exists) {
     if (date) {
       date = new Date(date);
+      date.setMilliseconds(0);
       res.setHeader('last-modified', date.toUTCString());
       res.setHeader('date', (new Date()).toUTCString());
       if (settings.maxAge !== undefined) {


JohnMcLear:
What issue does this solve?

Chocobozzz:
Minification is not cached for [uncached routes](https://github.com/ether/etherpad-lite/blob/develop/src/node/hooks/express/static.js#L12) (for example `/static/css/pad.css`, `/static/js/html10n.js`, `/static/js/require-kernel.js` etc)

JohnMcLear:
Wow really?  This seems like a huge thing to miss...  What's the total pad network size before and after this change?

Chocobozzz:
Without this PR, these files are minified. But etherpad re minifies the files every time.

Please try with/without this patch by adding a `console.log` above https://github.com/ether/etherpad-lite/blob/develop/src/node/utils/Minify.js#L228

JohnMcLear:
Wow excellent spot!  Good job!  Will let @muxator review and merge.

muxator:
Nice catch, thanks! Testing right now.

Now that I look at that section of code, I think it could be reworked. WIP.

Chocobozzz:
> Now that I look at that section of code, I think it could be reworked. WIP.

Agreed.

Could we add the routes I mentioned (`/static/css/`, `/static/js`) in the [cache middleware](https://github.com/ether/etherpad-lite/blob/develop/src/node/hooks/express/static.js#L12)?

muxator:
Aaaand @chocobozzz wins the prize for the higest value/code ratio in history!

This is what happened:

By specification [0], the if-modified-since HTTP header sent by browsers does not include milliseconds.

Before this patch, let's say a file was generate at time:
```
    t_real-file         = 2020-03-22T02:15:53.548Z (note the fractional seconds)
```

When issuing a conditional request, the browser would truncate the fractional part, and only request an if-modified-since with this contents:
```
    t_if-modified-since = 2020-03-22T02:15:53.000Z
```

The `minify()` function would return `HTTP/304` only if `t_if-modified-since >= t_real-file`, but this would never be true unless, by chance, a file was generated at XX.000Z.

**This resulted in that file being minified/compressed again and resent to the client for no reason**. After this patch, the server correctly responds with HTTP/304 without doing any computation, and the browser uses the cached file.

Pulled in, many thanks.

[0] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since

JohnMcLear:
Well done @Chocobozzz !  Defo happy that you found this!  What were you looking for when you spotted it?

muxator:
Wow. This ended up landing in `master`. I will cherry pick in `develop`.

muxator:
The naming convention for the branches is a bit unfortunate, but that's another story.

JohnMcLear:
I think we should push for a release / master by the end of this week anyway?  How bout it? 💃

Chocobozzz:
Thanks @muxator & @JohnMcLear <3

> Well done @Chocobozzz ! Defo happy that you found this! What were you looking for when you spotted it?

With COVID-19, many people in Europe have to work from home and some of them came on our etherpad instances at Framasoft (https://framapad.org/) leading to CPU issues. I ran a CPU profiler on one of our instance and discovered the weird minify call. After some investigation, I found this bug :)

JohnMcLear:
Excellent, yeah @Chocobozzz btw we also put up https://video.etherpad.com to help with the strain if you guys wanna offload onto that we're cool with it  :)

PR #3752, opened 2020-03-24
Use a Set for padlist

Chocobozzz:
Avoid looping on the array, especially useful if you have many pads.

Also changed `findKeys` second argument and the `replace` regex.


src/node/db/PadManager.js
```diff
@@ -49,17 +49,18 @@ var globalPads = {
...
  */
 let padList = {
-  list: [],
+  list: new Set(),
+  cachedList: [],
   sorted : false,
   initiated: false,
   init: async function() {
-    let dbData = await db.findKeys("pad:*", "*:*:*");
+    let dbData = await db.findKeys("pad:*", "pad:*:*");
```
muxator: Nice potential **performance gain**.

If possible confirm my analysis, based on looking at the [mysql](https://github.com/ether/ueberDB/blob/5fbc4f3ac182956168860a6a27d11ca82e820836/mysql_db.js) driver (I did not look at other DB drivers).

Am I right that this results in going from this query:

```sql
SELECT key FROM store WHERE key LIKE 'pad:%' AND key NOT LIKE '%:%:%'
```

To this one?
```sql
SELECT key FROM store WHERE key LIKE 'pad:%' AND key NOT LIKE 'pad:%:%'
```

Looking at the [database documentation](https://etherpad.org/doc/v1.7.5/#index_pad_padid) the structure of DB keys is the following:
```
pad:$PADID
pad:$PADID:revs:$REVNUM
pad:$PADID:chat:$CHATNUM
```

So the purpose would be excluding any key that has a colon (`:`) after "pad:$PADID".

This change can be **split** in a dedicated commit and maybe we can add a comment before:

```js
/*
 * select only the keys of the form pad:$PADID, defensively excluding all the
 * others (for example pad:$PADID:revs:$REVNUM and pad:$PADID:chat:$CHATNUM)
 */
```


**Edit**: upon further analysis, I fear **this would break compatibility with some drivers**. See for example Redis:

https://github.com/ether/ueberDB/blob/5fbc4f3ac182956168860a6a27d11ca82e820836/redis_db.js#L81

```js
callback(new customError("redis db currently only supports *:*:* as notKey","apierror"), null);
```

We would break Redis compatibility.

If we want to include this, we have to fix https://github.com/ether/ueberDB/blob/5fbc4f3ac182956168860a6a27d11ca82e820836/redis_db.js. It's an old driver, maybe Redis has evolved in the meantime. I do not know.

We **also should check the effects with the other DBs**.

src/node/db/PadManager.js
```diff
@@ -49,17 +49,18 @@ var globalPads = {
...
 
     if (dbData != null) {
       this.initiated = true;
 
       for (let val of dbData) {
-        this.addPad(val.replace(/pad:/,""), false);
+        this.addPad(val.replace(/^pad:/,""),false);
```
muxator: This anchors the regexp to the start of the string. I see no dangers here, and a **performance gain**.
Nice catch.

This can be split in a dedicated commit.

src/node/db/PadManager.js
```diff
@@ -79,27 +80,25 @@ let padList = {
     await this.load();
 
     if (!this.sorted) {
-      this.list.sort();
+      this.cachedList = Array.from(this.list).sort();
```
muxator: In the previous version ```this.sorted``` was used to keep whether ```this.list``` (then a list, now a set) was sorted.

What is its meaning now?
I am asking because this could be a semantic change, and thus introduce subtle bugs down the road.

My proposal is to further analyze its meaning, and eventually keep the invariant (sorted means that something is sorted), or get rid of it altogether.

src/node/db/PadManager.js
Chocobozzz: I'll remove it. The performance gain is not very interesting, and we don't really want to break things.

src/node/db/PadManager.js
Chocobozzz: I'm not sure to understand. The `sorted` variable was only used in the pad list structure to know if `getPads` needs or not to sort the list before returning the array. I think it was just for performance, and not used anywhere else.

muxator: I asked because when reviewing I can only look from an high level, without digging deep as you have done for proposing this.

So, `sorted` is probably not used. Makes sense trying to get rid of it, isn't it?

Chocobozzz: I think I understood. I pushed another version that does not use the `sorted` variable.

JohnMcLear:
Can we ensure we have backend test coverage for this before merging please?  @Chocobozzz did you run the backend tests already?  Coverage for the potential issues raised by @muxator would be ideal.  If you need help writing/running backend tests lemme know.

Chocobozzz:
I added (and fixed tests)

JohnMcLear:
@muxator Good for review.

Good from me

src/node/db/PadManager.js
```diff
@@ -79,27 +80,25 @@ let padList = {
...
   },
   addPad: function(name) {
     if (!this.initiated) return;
 
-    if (this.list.indexOf(name) == -1) {
-      this.list.push(name);
+    if (!this.list.has(name)) {
+      this.list.add(name);
       this.sorted = false;
     }
```
muxator, line 95: We want to know if `add()` has actually inserted an element, because in that case we need assign `sorted = false` and invalidate `cachedList`. That's ok.

Proposal: instead of using `list.has()` (at cost O(log(n)) now that `list` is actually a `Set`), couldn't we always perform the `add()`, and compare the set size before and after the insertion? The cost of `Set.size()` is probably O(1), and in the case of successful insertion we would be doing half the work.

src/node/db/PadManager.js
```diff
@@ -79,27 +80,25 @@ let padList = {
...
     if (!this.initiated) return;
 
-    var index = this.list.indexOf(name);
-
-    if (index > -1) {
-      this.list.splice(index, 1);
+    if (this.list.has(name)) {
+      this.list.delete(name);
       this.sorted = false;
     }
```
muxator, line 103: same here, _mutatis mutandis_.

src/node/db/PadManager.js
Chocobozzz: `.has` and `.add` are `O(1)` because a `Set` is a hashtable

muxator: Thanks, perfect!

muxator:
Merge conflicts, I'm not at my workstation now & can't fix them.

muxator:
Created an equivalent PR #3851 to avoid the merge conflicts. Closing this.

I think this was an hiccup in the GitHub UI. I did not see any terrible conflicts.

PR #3755, opened 2020-03-24
Add template cache

Chocobozzz:
It's a bit of a shame to use sync functions when we render template: https://github.com/ether/etherpad-lite/blob/develop/src/node/eejs/index.js#L106

I don't think we can change this function to be async, so I added a `readFile` cache and an EJS compilation cache. This is just a proposal because I'm not sure we expect that templates never change after an etherpad startup.

See also https://github.com/ether/etherpad-lite/issues/642


JohnMcLear:
Can you confirm this renders the read only pads with the expected content?  IE it shouldn't show the stars for saved revision.

Chocobozzz:
Yes it works correctly
screen_2020-03-25-10:40:18

JohnMcLear:
Good for review/merge. @muxator

muxator:
We'll integrate this in **1.8.4**

Issue #2884, opened 2016-01-14
Keyboard Shortcuts and languages


kamilmowinski:
Hi,
i have a problem with shortcuts. When i set english language on my system i can not type "ć" (ALT + C) becouse it shortcut activate chat window. But when my language is polish everythink is ok. 

Its possible to disable this shortcut or can you show me when in code i can turn off it?

Regards,
Kamil Mówiński

JohnMcLear:
https://github.com/ether/etherpad-lite/blob/develop/src/static/js/ace2_inner.js#L3729

bhldev:
Please see #2891 for workaround, disable shortcut in settings.json

Please give [bounty](https://www.bountysource.com/issues/30676386-iceweasel-switching-tabs-with-pgup-down-moves-the-cursor-in-the-pad) for UI/plugin development "enable/disable shortcut keys", UI will look like this  
![image](https://cloud.githubusercontent.com/assets/4929336/13028377/5944fe28-d23b-11e5-936b-3d1783dfeb6c.png)

Halibutt:
In the closed thread referenced above I explained why this issue is important. The fix mentioned in this thread is not really usable for your average user, who has no access to server settings. 

Also, this issue is not really a duplicate since my system language is set to Polish, yet the chat hijacks the keyboard when I hit Right Alt + c anyway. I wonder if changing the shortcut to Left Alt + c is that complicated... Any ideas how to do that?

lpagliari:
@Halibutt as @JohnMcLear mentioned [above](https://github.com/ether/etherpad-lite/issues/2884#issuecomment-172845087), you need to change ace2_inner.js. You might check the event `location`(https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/location). Hope this helps.

Halibutt:
@lpagliari , thanks a lot for your comment above, but it seems I'm way dumber than I thought and I have no idea what you suggest. I'm an end user, not a programmer. 

From what I understand, the workaround might be cool for someone who hosts an instance of etherpad and has access to all the files (and could change ace2_inner.js, wherever it is). I'm but a guy who used to use Etherpad quite frequently, and now cannot. Because it doesn't work as expected, because of a conflict that most likely is easy to fix for some code ninja (say, fix it for ALL people out there instead for me only), but for an end user like me is pure magic.

lpagliari:
[...]

PR #3764, opened 2020-03-27
If AltGr is pressed it should not be a command key

JohnMcLear:
I'm not sure this is even still a bug, I need someone to test it who has the correct setup I guess!

Fixes https://github.com/ether/etherpad-lite/issues/2884

To test, create a pad (with this branch active) and press AltGr C, you should see a character, you shouldn't have contents copied to clipboard...


muxator:
I have tried this patch, but I see no changes wrt `Alt`/`AltGr` + 'c` behaviour. Does it fix for you?

Let's see what users answer to https://github.com/ether/etherpad-lite/issues/2884#issuecomment-604765606, because maybe that bug no longer exists.

muxator:
I am inclined to reject this if there is no benefit. Confirmed?

JohnMcLear:
I'm a pleb Brit so not the right person to validate this as I'm 100% uncultured.  🐙

JohnMcLear:
Closing because we think this was fixed upstream.

PR #3765, opened 2020-03-27
Add env var for defaultPadText

sapkra:
I had to escape the line breaks otherwise the default string will be converted by `JSON.parse()` into multiple strings over multiple lines so that the the `lookupEnvironmentVariables` logic will not work for this value.

### DIFF
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -69,6 +69,7 @@ Available options:
 * `DB_PASS`: the password for the database username
 * `DB_CHARSET`: the character set for the tables (only required for MySQL)
 * `DB_FILENAME`: in case `DB_TYPE` is `DirtyDB`, the database filename. Default: `var/dirty.db`
+* `DEFAULT_PAD_TEXT`: The default text of a pad
 * `ADMIN_PASSWORD`: the password for the `admin` user (leave unspecified if you do not want to create it)
 * `USER_PASSWORD`: the password for the first user `user` (leave unspecified if you do not want to create it)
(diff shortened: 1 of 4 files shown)


muxator:
Ouch. I am not comfortable with having to escape the line breaks in that way `\\n` -> `\n`.

What would have to be the value of `DEFAULT_PAD_TEXT` if I wanted to have this default text in Etherpad?

```
Default text
on two lines
```

If this escaping is necessary in the values of the environment variables, too, we have to document it somewhere.

Can we find a way to have it cleaner, maybe making [`lookupEnvironmentVariables()`](https://github.com/ether/etherpad-lite/blob/fa3e4b146a977d9f60ba315bb84f5246d5c2abf1/src/node/utils/Settings.js#L450) smarter?

muxator:
Quick hint for you to expand: we may re-escape newlines before returning in this snippet, for example:

https://github.com/ether/etherpad-lite/blob/fa3e4b146a977d9f60ba315bb84f5246d5c2abf1/src/node/utils/Settings.js#L472-L483

i.e.: from `ascii code 13` to the string "\n".

Probably you may have to do the same in `coerceValue()`:

https://github.com/ether/etherpad-lite/blob/fa3e4b146a977d9f60ba315bb84f5246d5c2abf1/src/node/utils/Settings.js#L410-L411


Please have a look. It may work.

sapkra:
I found an easier solution by just allowing line breaks in the regex.

muxator:
> I found an easier solution by just allowing line breaks in the regex.

That's elegant! And I also think that it solves a bug, after all.

Merged, danke.


**p.s.**: is Etherpad being used in [Jitsi Meet](https://jitsi.org/jitsi-meet/)?

sapkra:
Yes. It can be used together with Jitsi Meet. But it's nothing special. Every participant of a call has a link in the conference UI for the same document.

PR #3766, opened 2020-03-27
docker: add support for arbitrary user ids (OpenShift compatibility)

sapkra:
I'm using OpenShift and this is running the containers by arbitrary user ids for security reasons but the users are always a member of the root group. So the `USER` definition of the `Dockerfile` will be ignored and the container crashes because the user has no permissions. This PR will just adjust the permissions a bit.

More information you can find here:
https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#use-uid

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,13 +25,13 @@ ENV NODE_ENV=development
 # that do not allow images running as root.
 RUN useradd --uid 5001 --create-home etherpad
 
-RUN mkdir /opt/etherpad-lite && chown etherpad:etherpad /opt/etherpad-lite
+RUN mkdir /opt/etherpad-lite && chown etherpad:0 /opt/etherpad-lite
 
-USER etherpad:etherpad
+USER etherpad
 
 WORKDIR /opt/etherpad-lite
 
-COPY --chown=etherpad:etherpad ./ ./
+COPY --chown=etherpad:0 ./ ./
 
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Looks fine to me, but not a Docker guy..  Will this interfere with existing deployments IE change their permissions or so?

sapkra:
It shouldn't because the user `etherpad` still has all permissions for the necessary directories and files.

@pierreprinetti @muxator WDYT?

muxator:
@pierreprinetti, if you are there, can you give a hand? I have a gazillion of things to look at.

Schnitzel:
+1 to the changes, we're running OpenShift at amazee.io and these changes make total sense to run on OpenShift.

pierreprinetti:
@sapkra
 
I am not sure I follow.

> the container crashes because the user has no permissions

What permissions do you need the unprivileged user to have?

---

There's a case in which you might fail accessing dirtyDB files created with a previous version of the Dockerfile (as they are owned by a superuser). In that case, I would suggest to run a mass-chown. Is that your case?

pierreprinetti:
@sapkra Thank you for reporting your issues running OpenShift!

The change looks good to me, @muxator please feel free to merge

muxator:
Added the motivation behind this change and the link to the OpenShift documentation in the commit message.

Merged, thanks.

PR #3768, opened 2020-03-27
Fix save revision icon showing on left menu bar and on the timeslider

### LINKED ISSUE #3767: Save revision icon showing on left menu bar?! :D
[
awtf
](url)

JohnMcLear:
Fixes #3767 - A bug introduced by me :D


src/node/utils/toolbar.js
```diff
@@ -250,7 +250,7 @@ module.exports = {
   selectButton: function (attributes) {
     return new SelectButton(attributes);
   },
-  menu: function (buttons, isReadOnly) {
+  menu: function (buttons, isReadOnly, whichMenu) {
```
muxator: `whichMenu` is a new flag for `toolbar.menu()`.

src/templates/pad.html
```diff
@@ -64,12 +64,12 @@
 
             <ul class="menu_left" role="toolbar">
                 <% e.begin_block("editbarMenuLeft"); %>
-                <%- toolbar.menu(settings.toolbar.left, isReadOnly) %>
+                <%- toolbar.menu(settings.toolbar.left, isReadOnly, 'left') %>
```
muxator: In `pad.html` it gets the values `'left'` or `'right'` (two string literals).
Fine.

src/templates/timeslider.html
```diff
@@ -82,7 +82,7 @@
           <div class="editbarright toolbar" id="editbar">
             <% e.begin_block("timesliderEditbarRight"); %>
             <ul>
-              <%- toolbar.menu(settings.toolbar.timeslider, true) %>
+              <%- toolbar.menu(settings.toolbar.timeslider, true, false) %>
```
muxator: In timeslider.html it gets the `false` value (a boolean).
At this point it makes sense for it to be a string literal, too (in TypeScript it would be an `enum`).

Should we using one of `'left'`/`'right'` because it conveys a meaning of some sort or do we need a third one (`'whatever'`)?

JohnMcLear: 'right' would work but you might also go 'timesliderRight' , your call :)

muxator: timesliderRight

JohnMcLear:
Do you wanna patch this real quick as it's a one liner?



I'm working on other stuff for release



-----Original message-----

muxator:
Already done

gareth:
Just tracked down that this fix is stopping me from configuring `settings.toolbars` to remove the "Save revision" button completely. I get that this probably isn't a strong use case, but can you think of any simple way that I can avoid having this button automatically re-added?

Issue #3032, opened 2016-07-27
Importing document botches pad options


DanielBickley:
I've struggled a bit trying to figure out why importing a document pulls in a pad where my previously loaded pad options (i.e. page view plugin) seem non-existent. The problems comes about in pad.js where the the 'options' var is null within the "switchToPad" method called via pad-impexp.js. I've rewritten the encompassing "switchToPad" method locally so that it actually pulls in those options and replaces them like one would expect.

My branch is only 269 commits behind ;) I've tried to conduct a frontend test, but the security measures on my pad prevent this from taking place. I don't have the time to set up another branch, free of security measures, for testing, at this time. Hopefully this can help somebody until I can run a proper frontend test and request a pull. 

If it is not desired that I paste code here, please let me know, and I'll keep my bug fixes to myself until I can get them properly added to my local branch in git.

Version number: 1.5.6
Git sha: 8af8c37

pad.js -  switchToPad()

From:

```
switchToPad: function(padId)
  {
    var newHref = new RegExp(/.*\/p\/[^\/]+/).exec(document.location.pathname) || clientVars.padId;
    newHref = newHref[0];    

    if (typeof options != "undefined" && options != null){
      newHref = newHref + '?' + options;
    }

    if(window.history && window.history.pushState)
    {
      $('#chattext p').remove(); //clear the chat messages
      window.history.pushState("", "", newHref);      
      receivedClientVars = false;
      sendClientReady(false, 'SWITCH_TO_PAD');
    }
    else // fallback
    {
      window.location.href = newHref;
    }
  },
```

To:

```
switchToPad: function(padId)
  {
    var newHref = new RegExp(/.*\/p\/[^\/]+/).exec(document.location.pathname) || clientVars.padId;
    newHref = newHref[0];    

    var options = clientVars.padOptions;
    if (typeof options != "undefined" && options != null)
    {
        var option_str = [];
        $.each(options, function(k,v) {                    
          var str = k + "=" + v;
          option_str.push(str);
        });
        var option_str = option_str.join("&");

      newHref = newHref + '?' + option_str;
    }

    if(window.history && window.history.pushState)
    {
      $('#chattext p').remove(); //clear the chat messages
      window.history.pushState("", "", newHref);      
      receivedClientVars = false;
      sendClientReady(false, 'SWITCH_TO_PAD');
    }
    else // fallback
    {
      window.location.href = newHref;
    }
  },
```

PR #3769, opened 2020-03-27
Fix3032 - importing document botches pad options

JohnMcLear:
Someone please test this.

Fix from: https://github.com/ether/etherpad-lite/issues/3032


JohnMcLear:
My tests it seems to be okay, I create a pad w/ < 10 revs, import .etherpad file and it seems to work.

But *shrug...

PR #3712, opened 2020-03-15
Fixed typo in en i18n

nchaloult:
Ellipsis on the `pad.modals.reconnecting` string was missing a dot

Issue #2812, opened 2015-10-20
Removing the bullet point in memo doesn't de-indent paragraph - WAITING ON TESTING


g-borgulya:
I don't know if this is like that intentionally by design:
1. if you click the bullet list toolbar on a line in a pad, then it becomes indented and gets a bullet-point 
2. if you click that icon again, the bullet-point is removed but the indentation stays.

I'd expect the indentation to be removed too in the 2nd step.

bujatt:
I's expect de-indentation, too.

bhldev:
Fixed in #2887

JohnMcLear:
I love me a nearly 5 year old bug to look at it...  ;\  This isn't fixed.

PR #3771, opened 2020-03-27
Mega test thread for OL / UL #3712 - DO NOT MERGE - This is Test Coverage ONLY, these tests should fail.

JohnMcLear:
#2812 covered


JohnMcLear:
These tests will need merging at the same time as merging all the OL / UL fixes awaiting merge.

PR #3775, opened 2020-03-27
Fix clear cookies tests

orblivion:
I mentioned in #3730 that I thought that there needed to be a delay between resetting the cookies and returning from `newPad`, because the change wasn't taking effect otherwise. However, this test proves that incorrect, at least _almost all of the time_. Seemingly on rare occasion, it fails, which means either:

a) occasionally it does indeed need the delay
b) something else entirely is going on

When I have a bit more time, I can try to test it with and without the delay enough times to prove that the delay actually fixes it. But it's quite a rare occurrence so I'm not sure how to be sure about it. Alternately: it's so rare that it's fine, given that this is a test suite; this question is up to the reviewers.

### DIFF
--- a/tests/frontend/helper.js
+++ b/tests/frontend/helper.js
@@ -53,7 +53,10 @@ var helper = {};
   }
 
   helper.clearCookies = function(){
-    window.document.cookie = "";
+    // Expire cookies, so author and language are changed after reloading the pad.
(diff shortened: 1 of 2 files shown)


JohnMcLear:
+1 looks good

muxator:
This was not scheduled for **1.8.4** (which should only be a release for the regressions introduced in **1.8.3**), but is not a change to the core, and is long overdue.

Keeping & merging, thanks.

PR #3776, opened 2020-03-28
.git is not needed in docker image

ratopi:
The current docker images containing the complete .git-Repo.

Therefore I fixed content of .dockerignore

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,5 @@
+.git
+.gitignore
 .hg
 settings.json
 src/node_modules


JohnMcLear:
Looks fine to me.

muxator:
Thanks, @ratopi, no reason to keep the `.git` history, indeed.
I did not catch this because I do not use git in my workflow (mercurial fan).

I was almost merging this change, but then I realized that **Etherpad upon start performs version detection looking at the contents of the `.git` directory**:

https://github.com/ether/etherpad-lite/blob/7a999ee398595429f70df8e7533e82371af3f3a2/src/node/utils/Settings.js#L333-L351

We could get away statically generating a version file inside the container at build time, and hopefully reworking the snipped I linked (I find it ugly).

Would you be willing to try?

ratopi:
Hi @muxator,
I'll try that, but I need some time. ;-)

muxator:
FYI: we are going to release next version soon (~a week).

No worries, this can slide in the next if you do not have time.

muxator:
Moving to **1.8.4**, since I hope we'll release **1.8.3** soon.

@ratopi, drop a line if you need help.

ratopi:
Sorry, for not answering a while.

I tried to completely refactoring the Dockerfile, but not finished yet.

Therefore I made a workaround (removing all unused git-objects): See #3917

Issue #3770, opened 2020-03-27
Restarting Server Using the Admin panel returns Declaring the sessionKey in the settings.json is deprecated when sessionKey is not set.


alekso87:
<img width="1439" alt="1" src="https://user-images.githubusercontent.com/25965602/77760830-ad3e1900-703f-11ea-8384-b77879f7b342.png">
<img width="1439" alt="2" src="https://user-images.githubusercontent.com/25965602/77760852-b4fdbd80-703f-11ea-8ea9-a8ce6aa22df5.png">
The server started successfully, but when restarting the server through the admin panel, a warning appeared.

There is no sessionKey entry in settings.json itself.

JohnMcLear:
Unable to replicate on develop branch.  What branch are you on?

alekso87:
git clone --branch master https://github.com/ether/etherpad-lite.git I did the installation from the master

alekso87:
`git clone --branch develop https://github.com/ether/etherpad-lite.git` I installed the project from the develop branch, the warning did not go away.

JohnMcLear:
Can you paste your own entire settings.json (if required remove any passwords)

alekso87:
[settings.json.zip](https://github.com/ether/etherpad-lite/files/4395464/settings.json.zip)

JohnMcLear:
access to instance would be useful
email me details contact@etherpad.org

alekso87:
> access to instance would be useful
> email me details [contact@etherpad.org](mailto:contact@etherpad.org)

plz check your email

JohnMcLear:
And when I run it on your box..  
```

found 7 vulnerabilities (5 low, 2 high)
  run `npm audit fix` to fix them, or `npm audit` for details
Clearing minified cache...
Started Etherpad...
[2020-03-27 23:14:35.702] [DEBUG] console - Running on Node v12.16.1 (minimum required Node version: 8.9.0)
[2020-03-27 23:14:35.849] [INFO] console - All relative paths will be interpreted relative to the identified Etherpad base dir: /home/etherpad/etherpad-lite
[2020-03-27 23:14:35.850] [DEBUG] AbsolutePaths - Relative path "settings.json" can be rewritten to "/home/etherpad/etherpad-lite/settings.json"
[2020-03-27 23:14:35.850] [DEBUG] AbsolutePaths - Relative path "credentials.json" can be rewritten to "/home/etherpad/etherpad-lite/credentials.json"
[...]

PR #3777, opened 2020-03-28
Restart warning clearer msg to admins

JohnMcLear:
I toyed with forcing regen of sessionKey but then I figured it'd lessen the admins UI so I just figured we can tell the admin not to worry about this msg on restart.


muxator:
This fixes #3770, right?

JohnMcLear:
Yep

muxator:
Merged, thanks.

PR #3782, opened 2020-03-29
Fix #3421

### LINKED ISSUE #3421: plain text password in the database
I can see my admin password in plain text in the database.

```
sessionstorage:-FWbiguBVxp1_VWLERNNheogwak9aewi {"cookie":
{"path":"/","_expires":null,"originalMaxAge":null,"httpOnly":true,"secure":false},
"passport":{},"user":{"password":"Here is a plain text password!",
"is_admin":true,"username":"admin"}}
```
That must not happen!

v1.6.6

JohnMcLear:
Replace password with a fixed string.


muxator:
John, I do no understand this. :)

https://github.com/ether/etherpad-lite/blob/b082a4e2334ad2b7cc2b1a733a5b505af5087d9e/src/node/db/SessionStore.js#L41-L44

This results in writing a fake password in the DB.

1. does the authentication still work after that?
2. if is works, do we still need to store something at all in the DB?

JohnMcLear:
1. Yes.
2. No.  But I like storing a record of the auth happening.

muxator:
Merged, thanks.

Aki-7:
Hi, there!

I think the authentication doesn't work under the circumstances below due to the changes in this PR,

- login with a device
- try to login to the same account with another device

The authentication fails at the second login because `settings.users[username].password` is replaced with "PASSWORD_HIDDEN"

Shouldn't it be like this?

```js
SessionStore.prototype.set = function (sid, sess, fn) {
  messageLogger.debug("SET " + sid);

  // don't store passwords in DB
  var password = sess.user && sess.user.password;
  if (password !== undefined) {
    sess.user.password = "PASSWORD_HIDDEN";
  }

  db.set("sessionstorage:" + sid, sess);

  if (password !== undefined) {
    sess.user.password = password;
  }

  if (fn) {
    process.nextTick(fn);
  }
};
```

JohnMcLear:
Fwiw we have s proposal to remove internal authentication mechanisms and shift to plugins / external Auth providers.

The win here would be to use a plugin for your Auth and move away from these methods.  Imho this isn't official yet.

Aki-7:
Thank you for your reply!

My problem was resolved using a plugin as you said, thank you for your help.

Btw can I create a Issue or PR about this?

JohnMcLear:
Yeh always feel free to create issues btw.  No need to ask for permission.

muxator:
This PR broke the authentication to the `/admin` area.

See #4016.

PR #3783, opened 2020-03-29
[PRIORITY #2] Fixes #3019 - Better import from LO / OO / soffice - READY TO MERGE

### LINKED ISSUE #3019: import from OO document
When I import a document from OO into etherpad, I have some format issues : 
-I get some empty bullets for a reason I do not understand.
-at random places, 8 spaces are inserted.
-I loose font size.

In the opposite way, after importing a document into etherpad, and then exporting it to a doc, it creates a tab which was not in the initial document.

It is probably an issue since refactoring the imported document needs a lot of time.
Is there any trick to avoid this ? Does this depend on the used format? I actually tried with doc, docx, odt and html, all with the same behaviour.

JohnMcLear:
Discussion here: https://github.com/ether/etherpad-lite/issues/3019

I don't like the ``return`` in contentcollector.js -- I'm +1 a better method.

This fixes copy/paste and import too, in theory.  Not fully tested.

  - [ ] Once OO is working test abiword to see if it will take the same Filter for the html value.


muxator:
Crashes my clean instance when importing a test file. No logs.

See https://github.com/ether/etherpad-lite/issues/3019#issuecomment-605729725.

JohnMcLear:
Ugh, libreoffice get this wrong.

If libreoffice isn't fully installed IE you only install libreoffice-common it fails.

```sudo apt-get install libreoffice```

is required.

What should happen is instead of Returning ``Error: source file could not be loaded`` Libreoffice should explain why it's not loading and how to remedy.  Quite disappointed!

**Comment by JohnMcLear** [comment]
When using NodeJS spawn
```
Error in option: --convert-to html:"XHTML Writer File:UTF8"
```
When using CLI
```
soffice --headless --convert-to html:"XHTML Writer File:UTF8" --outdir . /home/jose/testOneLiner.odt --invisible --nologo --nolockcheck --writer
```
Works

**Comment by JohnMcLear** [comment]
I dunno tbh, this fix is fine as far how it's designed, the problem is with nodejs not properly passing the arguments to ``spawn``

I tried hardcoding and ``shell:true`` to no avail..

```
      var soffice = spawn(settings.soffice, [
        '--headless',
        '--invisible',
        '--nologo',
        '--nolockcheck',
        '--writer',
        '--convert-to', 'html:"XHTML Writer File:UTF8"',
        task.srcFile,
        '--outdir', tmpDir
      ],
        {
          shell : true
        }
      );


```

@muxator can you take a look and sanity check my work on this?

**Comment by JohnMcLear** [comment]
Bump @muxator for help :D

**Comment by muxator** [comment]
@JohnMcLear I confirm your problem.

When invoked directly on the shell this command works flawlessly:
```
libreoffice6.4 --headless --convert-to html:"XHTML Writer File:UTF8" --outdir . /tmp/oneline.odt --invisible --nologo --nolockcheck --writer
convert /tmp/oneline.odt -> /tmp/oneline.html using filter : XHTML Writer File:UTF8
```

The same command wrapped in a [`spawn()`](https://nodejs.org/docs/latest-v12.x/api/child_process.html#child_process_child_process_spawn_command_args_options) call in pure Javascript, without Etherpad being involved at all, fails:
```js
// file: test-libreoffice.js
const { spawn } = require('child_process');

//const ls = spawn('ls', ['-lh', '/usr']);
const ls = spawn('libreoffice6.4', [
  '--headless',
  '--convert-to',
  'html:"XHTML Writer File:UTF8"',
  '--outdir',
  '.',
  '--invisible',
  '--nologo',
  '--nolockcheck',
  '--writer',
  '/tmp/oneline.odt',
]);

ls.stdout.on('data', (data) => {
  console.log(`stdout: ${data}`);
});

ls.stderr.on('data', (data) => {
  console.log(`stderr: ${data}`);
});

ls.on('close', (code) => {
  console.log(`child process exited with code ${code}`);
});
```

Result:
```
$ node test-libreoffice.js 
convert /tmp/oneline.odt -> /tmp/oneline.html using filter : "XHTML Writer File:UTF8"
Error: Please verify input parameters... (SfxBaseModel::impl_store <file:///tmp/oneline.html> failed: 0x81a(Error Area:Io Class:Parameter Code:26))
child process exited with code 0
```

Tested on nodejs **12.13.1**.


I'll try to profile the LibreOffice execution.

JohnMcLear:
I'm glad you sanity checked this, glad it's not me being an idiot!  Thanks!  Please let me know if you solve and how, this one pissed me off good :D

muxator:
This is **not nodejs specific**.

The same program above, **converted in python, gives the exact same result**:

```python
#!/usr/bin/env python3
#
# file: /tmp/test.py

import subprocess

args = [
  'libreoffice6.4',
  '--headless',
  '--convert-to',
  'html:"XHTML Writer File:UTF8"',
  '--outdir',
  '.',
  '/tmp/oneline.odt',
  '--invisible',
  '--nologo',
  '--nolockcheck',
  '--writer',
]

result = subprocess.run(args, capture_output=True)

print('stdout:', result.stdout)
print('stderr:', result.stderr)
```

When executing:
```
$ ./test.py 
stdout: b'convert /tmp/oneline.odt -> /tmp/oneline.html using filter : "XHTML Writer File:UTF8"\n'
stderr: b'Warning: failed to read path from javaldx\nError: Please verify input parameters... (SfxBaseModel::impl_store <file:///tmp/oneline.html> failed: 0x81a(Error Area:Io Class:Parameter Code:26))\n'
```

muxator:
Aaaand this is the reason for which everyone _always_ needs to have a good diff tool at hand.

Here's the difference between the output of the succeeding (directly on the shell) and the failing (js via `child_process.spawn`) execution:

```diff
-convert /tmp/oneline.odt -> /tmp/oneline.html using filter : XHTML Writer File:UTF8
+convert /tmp/oneline.odt -> /tmp/oneline.html using filter : "XHTML Writer File:UTF8"
```

Do you see that in the js (& python) execution there are two double quotes more?


Now, **in hindsight**, it is easy to understand: the double quotes are needed only when executing via the shell, **because the shell needs escaping for spaces**.

```bash
# bash requires the double quotes BECAUSE of the spaces, to convey
# that "html:XHTML Writer File:UTF8" is really a single parameter
#
--convert-to "html:XHTML Writer File:UTF8"   
```

When we explicitly pass parameters to a subprocess via [`execve()`](https://linux.die.net/man/2/execve) (which is what all these library calls are doing in the end), there is no need to escape anything, because the parameters are already cleanly separated among the array elements.

**Leave out the double quotes and it should work**.

On my toy Javascript & python programs it worked.

JohnMcLear:
woop nice will add to my list of things to do 2mrw

muxator:
After looking at this snippet I think that, we might be having one more problem.

**LibreOffice exits with a 0 exit code even when the conversion fails**. It did when we were passing a bad parameter.

If Etherpad just checks for the process exit code it **may mistake a failed conversion for a successful one**.

**We should check for something else in addition**.

1. the process stdout/stderr? That's fragile
2. that the converted file was really created where it was supposed to be <-- I would trust this more.

JohnMcLear:
- [ ] Fix spawn command to not use "
- [ ] Ensure output file exists with ``fs.fileExists`` lookup.

JohnMcLear:
Lol I just looked and I already WAS leaving out the double quotes btw :D


```
  if(type === "html") type = "html:XHTML Writer File:UTF8" // note how I'm not escaping the "
```
```
    queue.push({"srcFile": srcFile, "destFile": destFile, "type": type, "...
```

```
 var soffice = spawn(settings.soffice, [
        '--headless',
        '--invisible',
        '--nologo',
        '--nolockcheck',
        '--writer',
        '--convert-to', task.type,
        task.srcFile,
```

JohnMcLear:
Yea, I'm fucking stumped.

Sanity check for me..

1. Checkout ``develop`` && specify ``soffice`` in ``settings.json``
1. Import a word file, it works.
1. Edit ``src/node/utils/LibreOffice.js`` Insert ``  if (type === 'html') type = "html:XHTML Writer File:UTF8";``  @ Line 43 
1. Restart Etherpad
1. Import a word file, it fails.

Yet running this isolated:

```
var soffice = spawn("/usr/bin/soffice", [
  '--headless',
  '--invisible',
  '--nologo',
  '--nolockcheck',
  '--writer',
  '--convert-to', 'html:XHTML Writer File:UTF8',
  '/home/jose/test.doc',
  '--outdir', '/home/jose/'
]);
```

Works fine...   

Can you confirm?  Am I losing my mind?

muxator:
I have a suspect.

**To see what happens, you'll have to run Etherpad under nodejs 10** (on nodejs 12 a deprecated feature was removed and you won't see anything useful in the logs). See #3834 and #3841.

The `type` parameter somehow ends up in the filename of a temporary file we generate, or we think we generated (a brittle choice in the first place :smile:)).

* When `type === 'html'`, the temporary file name is `/tmp/upload_XXX.html`.
* When `type === 'html:XHTML Writer File:UTF8'`, this is what is generated in the logs:

  ```
  [2020-04-06 22:39:31.270] [WARN] console - Converting Error: { [Error: ENOENT: no such file or directory, rename '/tmp/upload_f3354ec47b4d3989f15cd7bbaedd81de.html:XHTML Writer File:UTF8' -> '/tmp/etherpad_import_1137536363.html']
    errno: -2,
    code: 'ENOENT',
    syscall: 'rename',
    path:
   '/tmp/upload_f3354ec47b4d3989f15cd7bbaedd81de.html:XHTML Writer File:UTF8',
    dest: '/tmp/etherpad_import_1137536363.html' }
  Error: ENOENT: no such file or directory, rename '/tmp/upload_f3354ec47b4d3989f15cd7bbaedd81de.html:XHTML Writer File:UTF8' -> '/tmp/etherpad_import_1137536363.html'
```

Moreover, in the next line, you'll see why we need to integrate #3841 (or upgrade log4js), and why this is not visible under nodejs 12:
```
  [2020-04-06 22:39:31.274] [ERROR] console - (node:13895) [DEP0079] DeprecationWarning: Custom inspection function on Objects via .inspect() is deprecated
  ```

muxator:
I think that the line that needs to be changed is this one:

https://github.com/ether/etherpad-lite/blob/e6251687bf6f0d773ef5cd8883465d67e18f4816/src/node/utils/LibreOffice.js#L105

But I really would like to see this become more robust (without encoding information in the temporary file name).

Some more monkey patching on that file would be almost immoral. :sob:

JohnMcLear:
How the funk did I not see that?! 

I'm running v13.10.1 and that might explain why but wow, what a crazy situation tho..

Solution, pass the ``padId+somerandomnonce()`` as the output filename string?

JohnMcLear:
Yea, agreed.  This whole import/export pass to third party binaries is a mess.  Gonna fix this, make it stable then will do a good review for future versions.

muxator:
1. > Solution, pass the padId+somerandomnonce() as the output filename string?

   You mean, ask libreoffice to generate a converted file with a name we control? Something like that, if LibreOffice's command line allows it.

   I would prefer that.

2. or mess with `task.type` here:

   https://github.com/ether/etherpad-lite/blob/e6251687bf6f0d773ef5cd8883465d67e18f4816/src/node/utils/LibreOffice.js#L105

   Removing everything after the first ":", if exists.

   But it stinks.

JohnMcLear:
Right, let's get this moving.  Please merge this PR as it is now and then I can keep moving forward with changes.  I'm going to hit problems with contentcollector.js and my list work really soon!

muxator:
Merged with some reordering of the commits.

Thanks.

PR #3784, opened 2020-03-29
Fix for "CHROME: Pasting an OL inserts additional line break"

JohnMcLear:
I'm not exactly 100% sure why this fix works so well but it does, I have a feeling the commented out line was required for IE6 or so but I could be wrong!


JohnMcLear:
Ah bollocks I based it off a diff branch..  Oh well, you can see the fix and i the other stuff is merged this can go in.

muxator:
That internal `if` looks really weird now.

```js
    if (isBlock)
    {
      if (lines.length() - 1 == startLine)
      {
        // commented out to solve #2412 - https://github.com/ether/etherpad-lite/issues/2412
        // cc.startNewLine(state);
      }
      else
      {
        _ensureColumnZero(state);
      }
    }
```

Should be reversed (assuming `lenght()` and `startLine` are always integers.

```js
    if (isBlock)
    {
      if (lines.length() - 1 !== startLine)
        _ensureColumnZero(state);
      }
    }
```

But then, the code is full of these constructs.

Pulling in.

muxator:
BTW, I was not able to rework the patch because the branch name "2412" drove mercurial mad. It did not like a branch name being an integer.

Weird.

Issue #3731, opened 2020-03-18
Load more messages hangs on a wheel turning


elpraga:
1.8.0 does not show old chat messages. What can I do? I'm using mariadb

muxator:
Hi.

Could you give some logs?

muxator:
Also, please describe better what you are trying to do, because otherwise it is impossible to replicate your problem.

elpraga:
![Screenshot_2020-03-19-00-23-14-540_com android chrome~2](https://user-images.githubusercontent.com/13814570/77016503-18208d80-6978-11ea-8f60-47ff0d8f5ca6.jpg)

elpraga:
![Screenshot_2020-03-19-00-23-14-540_com android chrome~2](https://user-images.githubusercontent.com/13814570/77016518-25d61300-6978-11ea-9a56-91ea9c096fe4.jpg)

elpraga:
![Screenshot_2020-03-19-00-23-06-173_com android chrome~2](https://user-images.githubusercontent.com/13814570/77016530-34bcc580-6978-11ea-8858-9ad8c7169ae5.jpg)

elpraga:
I'll need to find logs first

elpraga:
Hello, @muxator !

I'm sorry for yesterday's not-ver-clear start I was just on my phone. 

I've enabled logging now:

```
[2020-03-19 10:38:58.063] [WARN] client - Uncaught TypeError: Cannot read property 'replace' of undefined -- { errorId: 'Z97RLLbGB2blnicWuMzy',
  msg:
   'Uncaught TypeError: Cannot read property \'replace\' of undefined',
  url: 'http://etherpad.tomaskaluza.net/p/madisonhomework',
  linenumber: 44,
  userAgent:
   'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.149 Safari/537.36' }
[2020-03-19 10:38:58.077] [INFO] ueberDB - Flushed 1 values
[2020-03-19 10:38:58.277] [INFO] ueberDB - Flushed 1 values
[2020-03-19 10:38:58.378] [INFO] ueberDB - Flushed 1 values
[2020-03-19 10:39:49.408] [WARN] message - Dropped message, GetChatMessages Message has no start!
[2020-03-19 10:39:49.452] [INFO] ueberDB - Flushed 1 values

```
That is what I get

[...]

PR #3785, opened 2020-03-29
chat: in addMessage(), be tolerant of missing userId

JohnMcLear:
Fixes a bug that isn't a bug but is some weird database issue #3731


muxator:
If this is something that should not happen, and we are just being defensive here, we may have to issue a warn log.

JohnMcLear:
Feel free to change to warn and merge

muxator:
Taken. Issues a `console.warn()` in the browser, so that the more technically inclined users can still help identifying when how the problem appears.

PR #3790, opened 2020-03-29
Document already existing support for Unix sockets

JohnMcLear:
- [x] Includes example in settings.json.template
  - [x] Checks at runtime if unix socket is set
  - [x] Includes console msg at startup for how to connect to unix socket

As per ask at https://github.com/ether/etherpad-lite/pull/3312


muxator:
Added the documentation `settings.json.docker`, too (it has to be in lockstep with `settings.json.template`).

muxator:
Taken, many thanks.

Issue #3681, opened 2019-12-09
Hash based authentication does not work anymore


c7hm4r:
## Introduction

I have `ep_hash_auth` installed.

The current `settings.json.template` explicitly states:
```
...
   * WARNING: passwords should not be stored in plaintext in this file.
   *          If you want to mitigate this, please install ep_hash_auth and
   *          follow the section "secure your installation" in README.md
   */

  /*
  "users": {
    "admin": {
      // 1) "password" can be replaced with "hash" if you install ep_hash_auth
      // 2) please note that if password is null, the user will not be created
      "password": "changeme1",
      "is_admin": true
    },
...
```

## Setup 1

`settings.json` looks like:

```
  "users": {
    "admin": {
      "hash": "{{ etherpad_admin_password_hash }}",
      "is_admin": true
    }
  },
```

Log output:

```
[2019-12-09 17:44:06.270] [WARN] console - Removing user "admin", because it has no "password" field.
...
[2019-12-09 17:44:10.252] [WARN] console - Admin username and password not set in settings.json.  To access admin please uncomment and edit 'users' in settings.json
```

Cannot login with the hashed password.

## Setup 2

`settings.json` looks like:

```
  "users": {
    "admin": {
      "hash": "{{ etherpad_admin_password_hash }}",
      "password": "plaintext_password",
      "is_admin": true
    }
  },
```

Log output contains nothing suspect.

Can login with the hashed password as well as with the plaintext password.

## Setup 3 (Workaround)

`settings.json` looks like:

```
  "users": {
    "admin": {
      "hash": "{{ etherpad_admin_password_hash }}",
      "password": 0,
      "is_admin": true
    }
  },
```

Log output contains nothing suspect.

Can login with the hashed password but not with a plaintext password.

## Conclusion

Password authentication code is buggy. The easiest solution would probably be to not remove users from the config if they have an attribute `hash`. However, is the behaviour under setup 2 desired?

heicrd:
this also seems to "break" every other authentication module that relies on an entry in the users field for configuration (ep_ldapauth, ep_oidc, ep_oauth2), although most of them don't seem to be maintained

JohnMcLear:
Can someone do a git blame on this?

JohnMcLear:
- [ ] ep_hash_auth does ``console.log`` for missing dep, it should be ``console.warn`` -- will let @LaKing handle?
 - [x] Modify settings to allow users access if pass or hash is set.
[...]

PR #3792, opened 2020-03-30
Hash fix

JohnMcLear:
Fixes https://github.com/ether/etherpad-lite/issues/3681


muxator:
Clever!

There is a small logic bug, will fix.

PR #3798, opened 2020-03-31
Correctly fix referer when opening a link

Chocobozzz:


ldidry:
Related to #3044

src/static/js/ace2_inner.js
```diff
@@ -3374,7 +3374,7 @@ function Ace2Inner(){
       {
         try
         {
-          var newWindow = window.open(n.href, '_blank');
+          var newWindow = window.open(n.href, '_blank', 'noopener, noreferrer');
```
muxator: According to the [MDN documentation about `window.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#Window_functionality_features) newWindow will be null after this call, because of `noopener`.

> ## noopener
> If this feature is set, the newly-opened window [...] will not have access back to the originating window (via `Window.opener` — it returns `null`). In addition, the `window.open()` call will also return `null`, so the originating window will not have access to the new one either. [...]

src/static/js/ace2_inner.js
```diff
@@ -3374,7 +3374,7 @@ function Ace2Inner(){
...
           newWindow.focus();
```
muxator: this will probaby fail because `newWindow` will be null.

Could you verify it please? You should see an error message in the browser's console.

Chocobozzz:
Nice catch! I did not see an error because of the empty `catch` below `.focus()`. I removed it since it does not seem useful (the new tab is automatically focused on chrome & firefox).

muxator:
Pulled in, thanks.

muxator:
@Chocobozzz, for your next PRs: please do them against the `develop` branch. I did not check (my fault) and merged this in `master`.  This is the second time this happens.

Thanks

PR #3799, opened 2020-03-31
Note for 2802 - Scheduled for merge 03/04/2020

JohnMcLear:
Don't merge yet, I'm gonna see if I can bring test coverage in somehow..  Probably can't but let's see.

#2802


muxator:
So this is good to go as-is?

muxator:
It's just a log change, after all.

Pulling in.

PR #3801, opened 2020-03-31
[PRIORITY] Disallow undo on "clear authorship colors" - READY FOR REVIEW / MERGE

### LINKED ISSUE #2802: Undoing causes disconnect

JohnMcLear:
Don't allow user to undo clear authorship colors.

This fixes #2802 in a not so graceful way.

I also think it will fix a lot of broken Etherpad related issues, however, it doesn't stop the potential pad breakage issue.  That needs to be fixed somehow, for example, if someone did manage to get the server in a bad state **this is only remedied here but not fixed**.

  - [x] Notifies user prior to clear that this cannot be undone
  - [x] Test coverage (will do once accepted as a method [but prior to merge])
  - [x] Spoken to some admins to see if they would accept this change in behavior and they agreed it was fine.  The reason being is that if user X clicks clear authorship colors then it's not as if user Y can undo that change anyway.
  - [ ] All translations will need to be updated :|

The best thing to remedy this would be a plugin (such as ep_author_hover) which shows the author of a span on hover.


JohnMcLear:
@muxator wanna give your opinion on this?  I'd like to merge it.

muxator:
I agree removing support for undoing `clearauthorship` is the most practical approach right now.

As usual, there is an underlying bug laying somewhere that needs to be tackled (I suspect the usual joys of dynamic languages & lack of runtime validation).

Testing it now.

muxator:
Just a not to be on the same page: provided #3838 is related (as per @Chocobozzz hypotesis), this change on disallowing undoing of clear authorship (that I agree with) is not going to fix that problem.

muxator:
Since this is a change of behaviour, I [added a note in the changelog](https://github.com/JohnMcLear/etherpad-lite/blob/301e65b9a8ef1dce129f86c82cd90786351b7467/CHANGELOG.md#183).

Waiting for final comments at #2802, or merging if nothings surfaces.

JohnMcLear:
NOTE: You will need to merge this into develop then give TW / @Nikerabbit 's teams some time to do translations before releasing..

- 1. [ ] Test Coverage
- 1. [ ] Merge to develop
- 1. [ ] Get Translations

JohnMcLear:
Due to time constraints here I will get test coverage done within an hour.

JohnMcLear:
Tests in.

JohnMcLear:
One thing worth noting is that the change in behavior (lack of undo event in the stack) might confuse some people.

For example; 

1. user types hello
1. clears authorship
1. expects undo stack to clear authorship but it actually removes hello

I could write a "blank event" to the undo stack OR we just leave it as it currently stands in this PR.  For me I'm inclined to leave it as it is because it's bad practice to pollute the undo stack with "null" events and also we do say "it can't be undone in the dialog (note that this is not present on clearing just a selection of text".

I'm no UX expert and I'm mindful we need to get this in quickly.

Another option could be to allow undo events of clear authorship to be written to the stack but on execution of the undo event to pop up a gritter msg saying "can't undo clear authorship events"..  

Either way, we need translations stat :)

muxator:
> expects undo stack to clear authorship but it actually removes hello

Non nice, but acceptable for me. The other options seem even more confusing.

Waiting for translations.

JohnMcLear:
Don't wait for translations, see above.  Afaik TW can't do work until it's in develop! :D

PR #3807, opened 2020-04-01
Add getStats API

Chocobozzz:


JohnMcLear:
I like this! 👯

muxator:
Nice!

Don't make me be the guy who always asks for docs, please :(

Chocobozzz:
Oup's sorry

muxator:
Pulled in with the following modifications, and a **TODO** (see point 4):

1. added a missing bit in the docs ("_The latest version is ~~1.2.13~~ -> 1.2.14_").
   Folded in your commit.
2. added another commit on top of this that rewrites `activePads` computation.
   Instead of starting from an empty `Set`, and then mutating it via iteration over `sessionKeys`:

   ```js
   const activePads = new Set();

   sessionKeys
     .map(k => padMessageHandler.sessioninfos[k])
     .forEach(o => activePads.add(o.padId));
   ```

   Compute it directly via a **single expression**:

   ```js
   const activePads = new Set(Object.entries(sessionInfos).map(k => k[1].padId));
   ```

   But it's really the same, after all.
3. added *test coverage*, which basically checks that the body of the api response contains these three fields, and that they are of the correct type:
   `totalPads`: `int`
   `totalSessions`: `int`
   `totalActivePads`: `int`
4. **TODO**: the development version is now migrated to OpenAPI, and includes an endpoint `/api/openali.json` that exposes the schema of the API requests and responses. A modification needs to be done there, as well.
   I could not do it because this PR is based on a development version that is not recent enough. It needs to be done now.


Thanks!

PR #3813, opened 2020-04-01
Backend test runner - Ready for review / feedback / squash / merge - Scheduled for merge 03/04/2020

JohnMcLear:
- [x] Checks for SAUCE environment variables.
    - [ ] Stop the test if env vars aren't set
  - [x] Runs the Backend tests first ( I think they are sequential? ).
  - [ ] Is it normal for projects to only run tests if the user has the sauce keys available?  This feels like a shortcoming to me that I have to give out sauce keys to each dev?


tests/frontend/travis/runner.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+if [ -z ${SAUCE_USERNAME} ]; then echo "SAUCE_USERNAME is unset - exiting"; exit 1; fi
```
muxator, line 2: Use double quotes around expanded bash variables: `"{SAUCE_USERNAME}"`

JohnMcLear: Ack

.travis.yml
```diff
@@ -24,20 +24,28 @@ env:
...
-        - "tests/frontend/travis/runner.sh"
-    - name: "Test the Dockerfile"
+    - name: "Run the Backend"
       install:
         - "cd src && npm install && cd -"
       script:
         - "docker build -t etherpad:test ."
         - "docker run -d -p 9001:9001 etherpad:test && sleep 3"
         - "cd src && npm run test-container"
+        - "npm run test"
```
muxator, line 34: why don' we put this in a dedicated section, so we have separated tests for Docker build (& its minimal testing), and the backend testing itself?

JohnMcLear: could do, I'm just playing with something that should be stable for now.

tests/frontend/travis/runner.sh
JohnMcLear: looking at other references it' ``"${SAUCE_USERNAME}"``

You sure on ``"{SAUCE_USERNAME}"``?

JohnMcLear:
No need for early review btw, I'm mostly just playing right now :)

muxator:
Ok, I'll slow down in the next ~24 h

JohnMcLear:
I mean just on this PR btw, feel free to review other stuff :D  Merge away!

JohnMcLear:
Works at this point, but jeez that was a lock of back and forth -_-

PR #3817, opened 2020-04-01
Fix2877 - Ready for review / Merge- Bump to 0.4.5 scheduled 03/04/20202

### LINKED ISSUE #2877: Two inconsistent versions with %20 at the end
One can access the same etherpad 
https://beta.etherpad.org/p/test20percent
with the following link as well
https://beta.etherpad.org/p/test20percent%20.
Now these version will split: 
If you start editing one the other will not change.  
Can be seen here:
https://beta.etherpad.org/p/20percenttest
https://beta.etherpad.org/p/20percenttest%20.

Maybe this a feature as one can steal/copy pads.

JohnMcLear:
Update UeberDB2 to fix #2877

Explanation: we now include BINARY in the MySQL WHERE select clause.

Test coverage might be desirable?


JohnMcLear:
package-lock.json got made a mess, just need the ueberdb part..

src/package.json
```diff
@@ -59,7 +59,7 @@
     "socket.io": "2.1.1",
     "swagger-node-express": "2.1.3",
     "tinycon": "0.0.1",
-    "ueberdb2": "0.4.3",
+    "ueberdb2": "0.4.4",
```
muxator, line 62: This has to be `0.4.5`, right?

JohnMcLear:
Yes :)



-----Original message-----

muxator:
John, _please_, delete your development clone and start from a fresh one ASAP!

JohnMcLear:
Is it even possible to delete my development close without destroying a bunch of pull requests / history that we need to merge in?  Sorry, I haven't used Git much in like 5 years and I'm re-learning a lot ;\

muxator:
Committed directly on `develop` via 3a46e010cea637c4b51f35def04a290091044d4c.
The change was easy, and untangle the merge conflicts would have taken hours for no reason.

Closing.


> Is it even possible to delete my development close without destroying a bunch of pull requests / history that we need to merge in?

Probably not at this time.

> Sorry, I haven't used Git much in like 5 years and I'm re-learning a lot ;\

The world would have been an easier place had [Mercurial](https://www.mercurial-scm.org) won instead of git :smile:

akosiaris:
Unfortunately, this has caused a significant regression on our instance, https://etherpad.wikimedia.org.

By the nature of our service we have a pretty large dataset, some 233M rows in the `store` table.

Upon upgrading we had an outage and reverted to the previous version. Fetching e.g.  https://etherpad.wikimedia.org/p/foobar (the pad name is actually utterly unimportant) via curl took a really long time, in the order of ~1m 20s.

After debugging a bit, we looked in the database and we noticed that SELECTs would now take >1m to complete. An EXPLAIN points out why

```
MariaDB [etherpadlite]> explain SELECT `value` FROM `store` WHERE `key` = 'pad:foobar';
+------+-------------+-------+-------+---------------+---------+---------+-------+------+-------+
| id   | select_type | table | type  | possible_keys | key     | key_len | ref   | rows | Extra |
+------+-------------+-------+-------+---------------+---------+---------+-------+------+-------+
|    1 | SIMPLE      | store | const | PRIMARY       | PRIMARY | 402     | const |    1 |       |
+------+-------------+-------+-------+---------------+---------+---------+-------+------+-------+
1 row in set (0.00 sec)

vs 

MariaDB [etherpadlite]> explain SELECT `value` FROM `store` WHERE BINARY `key` = 'pad:foobar';
+------+-------------+-------+------+---------------+------+---------+------+-----------+-------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref  | rows      | Extra       |
+------+-------------+-------+------+---------------+------+---------+------+-----------+-------------+
|    1 | SIMPLE      | store | ALL  | NULL          | NULL | NULL    | NULL | 233303599 | Using where |
+------+-------------+-------+------+---------------+------+---------+------+-----------+-------------+
```

So, casting to BINARY there makes the SELECT not use the Primary Key and instead scan the entire table (which in our case takes >1m). 

My guess is that this is going to cause a pretty big performance regression, perhaps up to the point of it appearing not functioning, for any sufficiently large installation.

JohnMcLear:
Really good spot, it's on my radar.

akosiaris:
For what is worth, it might be worth looking into https://dev.mysql.com/doc/refman/8.0/en/charset-binary-collations.html#charset-binary-collations-trailing-space-comparisons instead of casting to BINARY. But `utf8mb4_0900_bin` seems to be pretty new per https://dev.mysql.com/doc/relnotes/connector-net/en/news-8-0-17.html

Pita:
Yeah maybe to get this performant the column needs to be altered. jeez, full table scan :/

JohnMcLear:
I think I have a fix.

Source: https://forums.mysql.com/read.php?25,114183,114335#msg-114335

`` mysql> SELECT * FROM members WHERE username = 'aaa' AND binary username = 'aaa'; ``

JohnMcLear:
https://github.com/ether/ueberDB/commit/7762f0096458bcb1ada3ad634d1a43f99a1354f0

Ueberdb 0.4.6.

## Before
```
MariaDB [etherpad_lite_db]> explain SELECT `value` FROM `store` WHERE BINARY `key` = 'MYSQL_MIGRATION_LEVEL';
+------+-------------+-------+------+---------------+------+---------+------+--------+-------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref  | rows   | Extra       |
+------+-------------+-------+------+---------------+------+---------+------+--------+-------------+
|    1 | SIMPLE      | store | ALL  | NULL          | NULL | NULL    | NULL | 105469 | Using where |
+------+-------------+-------+------+---------------+------+---------+------+--------+-------------+
1 row in set (0.00 sec)
```

## After patch is applied
```
MariaDB [etherpad_lite_db]> explain SELECT `value` FROM `store` WHERE `key` = 'MYSQL_MIGRATION_LEVEL' AND BINARY `key` = 'MYSQL_MIGRATION_LEVEL';
+------+-------------+-------+-------+---------------+---------+---------+-------+------+-------+
| id   | select_type | table | type  | possible_keys | key     | key_len | ref   | rows | Extra |
+------+-------------+-------+-------+---------------+---------+---------+-------+------+-------+
|    1 | SIMPLE      | store | const | PRIMARY       | PRIMARY | 402     | const |    1 |       |
+------+-------------+-------+-------+---------------+---------+---------+-------+------+-------+
1 row in set (0.00 sec)
```

@Pita fyi the UeberDB tests are broken.  IDK if I have the energy to fix your code, it might be quicker for me just to rewrite using some modern test framework.  If you wanted to fix tho I'd be cool to have it off my radar ;)

akosiaris:
Tested in production db really quickly and indeed explain is happy

```
MariaDB [etherpadlite]> explain SELECT `value` FROM `store` WHERE `key` = 'pad:Scrum-of-Scrums' AND BINARY `key` = 'pad:foobar`;
+------+-------------+-------+-------+---------------+---------+---------+-------+------+-------+
| id   | select_type | table | type  | possible_keys | key     | key_len | ref   | rows | Extra |
+------+-------------+-------+-------+---------------+---------+---------+-------+------+-------+
|    1 | SIMPLE      | store | const | PRIMARY       | PRIMARY | 402     | const |    1 |       |
+------+-------------+-------+-------+---------------+---------+---------+-------+------+-------+
```

JohnMcLear:
Wicked tnx @akosiaris - sending PR into Etherpad-lite now.

muxator:
Hi @akosiaris and everyone in this thread.

Before going in with the fix, I would like an opinion on https://github.com/ether/ueberDB/commit/7762f0096458bcb1ada3ad634d1a43f99a1354f0#r38780442, since I want to make sure we are not introducing another bug.

If everything is ok, we can issue a patch release only with this fix.

JohnMcLear:
Just to loop back and close this off from WMF perspective:  I added CI to Ueber which adds 1M rows and replicates WMF problem then does some get/set/findKeys performance benchmarking.  This means this problem (at least w/ MySQL) wont rear it's head again and test coverage is in.  Thanks for reporting and thanks for being patient for the fix.  We should be merging a patch early next week and merging the CI over the coming weeks.

akosiaris:
That's nice! Thanks very much @JohnMcLear. It's appreciated.

PR #3818, opened 2020-04-01
Set author info error - Ready for Review / Merge [urgent]

JohnMcLear:
On a production instance after pulling develop I was getting setAuthorInfo errors thrown on visiting pads.

Turns out it was trying to init with a blank user, tihs patch fixes that but doesn't remedy why it was firing with a null value?  @muxator ?


muxator:
Was the production instance on 1.8 already, or did it start as soon as you migrated to 1.8?

I suspect this is a bug from the `async` code in **1.8**.

BTW, this is the reason for which I'd like the developers cared about writing meaningful commits & messages, otherwise bisecting when a bug is introduced takes ages, particularly for complex code.

JohnMcLear:
It started as soon as I migrated BUT it might be because we have more debug / error handling now?

PR #3820, opened 2020-04-02
Fix #3819 Strip cookie value from optional quotes

### LINKED ISSUE #3819: sessionID cookie enveloped in quotes is not parsed properly
Some servers, like WildFly 11, surround cookie value in double quotes.  It depends on cookie value characters and cookie version. Here is some background
https://issues.redhat.com/browse/UNDERTOW-1096
Etherpad SecurityManager expects sessionID cookie value without quotes. If quotes are present, they are considered a part of session ID, resulting in "session not found" error.

MarcinCieslak:
A simple fix to make different cookie versions work

### DIFF
--- a/src/node/db/SecurityManager.js
+++ b/src/node/db/SecurityManager.js
@@ -96,7 +96,30 @@ exports.checkAccess = async function(padID, sessionCookie, token, password)
   // get information about all sessions contained in this cookie
   if (sessionCookie) {
     let groupID = padID.split("$")[0];
-    let sessionIDs = sessionCookie.split(',');
+
+    /*
+     * Sometimes, RFC 6265-compliant web servers may send back a cookie whose
+     * value is enclosed in double quotes, such as:
+     *
+     *   Set-Cookie: sessionCookie="s.37cf5299fbf981e14121fba3a588c02b,s.2b21517bf50729d8130ab85736a11346"; Version=1; Path=/; Domain=localhost; Discard
+     *
+     * Where the double quotes at the start and the end of the header value are
+     * just delimiters. This is perfectly legal: Etherpad parsing logic should
+     * cope with that, and remove the quotes early in the request phase.
+     *
+     * Somehow, this does not happen, and in such cases the actual value that
+     * sessionCookie ends up having is:
+     *
+     *     sessionCookie = '"s.37cf5299fbf981e14121fba3a588c02b,s.2b21517bf50729d8130ab85736a11346"'
+     *
+     * As quick measure, let's strip the double quotes (when present).
+     * Note that here we are being minimal, limiting ourselves to just removing
+     * quotes at the start and the end of the string.
+     *
+     * Fixes #3819.
+     * Also, see #3820.
+     */
+    let sessionIDs = sessionCookie.replace(/^"|"$/g, '').split(',');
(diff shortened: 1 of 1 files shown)


src/node/db/SecurityManager.js
```diff
@@ -96,7 +96,7 @@ exports.checkAccess = async function(padID, sessionCookie, token, password)
   // get information about all sessions contained in this cookie
   if (sessionCookie) {
     let groupID = padID.split("$")[0];
-    let sessionIDs = sessionCookie.split(',');
+    let sessionIDs = sessionCookie.replace(/"/g, '').split(',');
```
muxator, line 99: If this is meant to always remove the double quotes (`"`), we could achieve it without firing a regexp:

```js
replace('"', '')
```

MarcinCieslak: This removes only first quote. We need to remove both quotes, at the beginning and at the end.

MarcinCieslak:
@muxator 
Thank you for the advice.
I described the problem in the issue ticket I referenced
https://github.com/ether/etherpad-lite/issues/3819
I will just paste it here
---
Some servers, like WildFly 11, surround cookie value in double quotes. It depends on cookie value characters and cookie version. Here is some background
https://issues.redhat.com/browse/UNDERTOW-1096
Etherpad SecurityManager expects sessionID cookie value without quotes. If quotes are present, they are considered a part of session ID, resulting in "session not found" error.
---
I will keep in mind to comment in commit message and in code.

src/node/db/SecurityManager.js
muxator: Yup

muxator:
Thanks, @MarcinCieslak.

From [the Undertow issue](https://issues.redhat.com/browse/UNDERTOW-1096) you linked, and skimming [RFC 6265](https://tools.ietf.org/html/rfc6265) it seems that the real error is Etherpad not understanding cookie quoting.

Can you help me understand? If it is really like so, we could maybe accept this as quick fix, but open another issue tracking that problem.

For opening the issue, it would be useful an example of the raw `set-cookie` HTTP header. Could you provide that?

MarcinCieslak:
@muxator 
When when want to give an user access to multiple Etherpad sessions, we put their IDs in a single cookie sessionID. WildFly 11 generates the following header:

`Set-Cookie: sessionID="s.37cf5299fbf981e14121fba3a588c02b,s.2b21517bf50729d8130ab85736a11346"; Version=1; Path=/; Domain=localhost; Discard`

It is then sent back to Etherpad like this
`sessionID="s.37cf5299fbf981e14121fba3a588c02b,s.2b21517bf50729d8130ab85736a11346"`

Current, non-patched version of Etherpad extracts following 2 session IDs
"s.37cf5299fbf981e14121fba3a588c02b
s.2b21517bf50729d8130ab85736a11346"
i.e. it includes quotes as part of session ID.

muxator:
Modified the logic to only remove quotes at the start & the end of the string (should also do unescaping, but I did not bother to see if RFC defines escaping rules), added a comment to contextualize (which I'll report in #3819).

Thanks.

MarcinCieslak:
@muxator 
Thank you as well

PR #3822, opened 2020-04-02
i18n order of items in dropbox

JohnMcLear:
1. Isn't an actual issue, just something I thought was worthwhile.
1. Puts top 10 world used languages at top of select
1. Puts rest below (alphabetically sorted but with a caveat of it sorts by key (see comment in source)).

I wouldn't merge this unless someone thinks it's worthwhile, I'm not a UX guy.


JohnMcLear:
@seballot any thoughts on this?  Good UX / Bad UX?

seballot:
Hi !

I think a common practice is when we use "top choice" in a select, add a divider between the top choice and the rest of the choice. And also here as there are a lot of options, I would still leave the top choice language in the second part, so it's still ordered alphabetically. For exemple if top choice is only "en" and "es"

English
Spanish
_----------_
Arabic
English
French
Spanish

But having top choice set from the most used language (vs configurable per instance) does not seems to have a lot of benefits to me, but it has no harm either, so as you want !

JohnMcLear:
Hey @seballot that's a good idea but sadly not easy to implement.  The problem is that we push one object IE

{
en: ...,
de:...,
es:...
}

so I can't add es*x without losing the kye and having to rewrite loads of other logic.

So I think maybe this isnt worth the effort so I will close.

Issue #3487, opened 2018-09-26
Spaces are lost around wrapped line - Ready for Review #2 and merge


rupran:
On my hosted etherpad-lite instance, I have the issue that spaces typed at the beginning of a wrapped line are being ignored, leading to the last word of the previous line and the new word wrapping without a space between them.

I made a recording of this problem. This is on a newly created pad. Both times, I type `<space><s><e><e><s><space>`. With the default text in place, this works, but after erasing everything I typed and typing the same sequence again, the leading space gets dropped.

![screen recording 2018-09-26 at 12 08 32](https://user-images.githubusercontent.com/6428272/46073397-1318be80-c185-11e8-8266-f193f64f4f07.gif)

**Relevant system information:**
- Etherpad-Lite: from git at https://github.com/ether/etherpad-lite/commit/2a5e87cc7da748da94d1169df2a470b6406b65b9 (most recent commit in `develop`)
- Client OS: macOS 10.14 (18A391) Mojave
- Browser: Google Chome Version 69.0.3497.100 (Official Build) (64-bit)
- Host OS: CentOS 7, shared hosting
  - npm: 5.6.0
  - node: v8.11.3

**Steps to reproduce:**
Open a new or existing pad, start typing a space followed by other characters with the cursor at the beginning of a wrapped line.

**What did you expect to happen?**
The last word of the previous line stays in the previous line, the new word starts at the current cursor position.

**What actually happened?**
The line break is removed, merging the last word of the previous line and the new word without a space even though it was entered.

JohnMcLear:
Confirmed.   This is a horrible bug, probably due to pre-wrap setting.

JohnMcLear:
@rupran I was just testing this and I'm 99% sure I can't replicate this bug on Firefox, can you confirm?

JohnMcLear:
Also if so..
commenting out the below seems to fix it...
  /*
  word-wrap: break-word; /* fix for issue #1648 - firefox not wrapping long $
  */
Can anyone confirm?

JohnMcLear:
Does removing the comments around

```
   white-space: pre-wrap;
``` 
in ``iframe_editor.css``

JohnMcLear:
Great spot @seballot - yeah it's a complex problem compounded by another bug for caret positioning on wraps.

I removed the JS hack and CSS "fix" for the JS hack.

[...]

PR #3831, opened 2020-04-03
3487 wrap - Ready for review / merge

JohnMcLear:
Chrome specific CSS rule.  

Fixes https://github.com/ether/etherpad-lite/issues/3487


src/static/css/iframe_editor.css
```diff
@@ -38,6 +38,17 @@ body.grayedout { background-color: #eee !important }
   color: black;
 }
 
+@media screen and (-webkit-min-device-pixel-ratio:0){
```
muxator, line 41: I'm no frontend person.

If the bug is chrome specific, should'n this `-webkit-min-device-pixel-ratio` only apply to Chrome?

Then again, from my distant past I think to remember that it is better to do capability detection instead of hand-patching for specific browsers.

@seballot, can you have a look? This is related to issue #3487.

seballot: This wrapping thing seems not easy ! From ace2_inner.js

```javascript
    // Chrome can't handle the truth..  If CSS rule white-space:pre-wrap
    // is true then any paste event will insert two lines..
    // Sadly this will mean you get a walking Caret in Chrome when clicking on a URL
    // So this has to be set to pre-wrap ;(
    // We need to file a bug w/ the Chromium team.
    if(browser.chrome){
      $("#innerdocbody").addClass("noprewrap");
    }
```

And from ifram_editor.css
```css
body {
  white-space: nowrap;
  word-wrap: normal;
}
body.doesWrap {
  display: block !important;
  /* white-space: pre-wrap; */
  /*
    Must be pre-wrap to keep trailing spaces. Otherwise you get a zombie caret,
    walking around your screen (see #1766).
    WARNING: Enabling this causes Paste as plain text in Chrome to remove line breaks
    this is probably undesirable
    WARNING: This causes copy & paste events to lose bold etc. attributes
    NOTE: The walking-zombie caret issue seems to have been fixed in FF upstream
    so let's try diabling pre-wrap and see how we get on now.
    For more details see: https://github.com/ether/etherpad-lite/issues/2574
  */
  word-wrap: break-word; /* fix for issue #1648 - firefox not wrapping long lines (without spaces) correctly */
}
.noprewrap{
  white-space: normal;
}
body.doesWrap:not(.noprewrap) > div {
  /* Related to #1766 */
  white-space: pre-wrap;
}
```

What I think is that all those problems have been fixed in recent browsers, and we could simplify it and use `white-space: pre-wrap;` everywhere. But it would need a bit more investigation and testing

JohnMcLear:
Great spot @seballot - yeah it's a complex problem compounded by another bug for caret positioning on wraps.

I removed the JS hack and CSS "fix" for the JS hack.

Full disclaimer: Older versions of Chrome may have a bug. But we can make the assumption that most peoples chrome version is kept up to date so I'm not going to worry about this. What we could do is target older chrome versions with a fix.

I think we can merge and close this and I will just have to do more test coverage for caret position and character string entry when I submit my fix for caret positioning (which is WIP and due within 48 hours).

muxator:
So this change now only amounts to having _less_ code in `ace2_inner.js`?

That's a net win!

PR #3833, opened 2020-04-04
URGENT!  Max file size, Rate limiting & can only import with author present on pad - Ready for review.

JohnMcLear:
- [x] Rate limits
  - [x] Settings
  - [ ] Tests <-- can't test without an authorID.
- [x] Restricts import only if user is on pad
  - [x] Settings
  - [ ] Tests <-- can't test without an authorID.
- [x] Max file size on import
  - [x] Settings
  - [ ] Tests <-- can't test without an authorID

Also

 - [x] loadSettings was not getting the correct settings, it was only parsing settings.json and ignoring defaults not present in the settings.json file.
Tests without creating an authorID is not really feasible, I tried but I can't find a decent way.

This is good to merge as is, knowing that a future job is to write test coverage which includes a method to create an authorID that is present on a specific padID.


tests/backend/loadSettings.js
```diff
@@ -7,20 +7,10 @@
 
 var jsonminify = require(__dirname+"/../../src/node_modules/jsonminify");
 const fs = require('fs');
+var settings = require("../../src/node/utils/Settings");
 
 function loadSettings(){
-  var settingsStr = fs.readFileSync(__dirname+"/../../settings.json").toString();
```
muxator, line 12: So this code was bugged since forever, isnt'it?

If you read at the top of this file

https://github.com/ether/etherpad-lite/blob/c71706fd86c8e91323ff731f2b472179cb01a486/tests/backend/loadSettings.js#L1-L6

There is a "cloned" version in `tests/container/loadSettings.js`. Until the code is unified, we have to manually port the same modifications there. But: the previous (bugged) code "apparently" gave the possibility to explicitly load a settings file, and in `tests/container/loadSettings.js`, we were using it:

https://github.com/ether/etherpad-lite/blob/83d72d27a405f63782803e0168b1e9121f3cd650/tests/container/loadSettings.js#L12

See that we are loading `settings.json.docker`?

What happens when we port this modification to that file? What was the settings bug about?

JohnMcLear: The settings bug was about if you ran ``loadSettings`` it only got the settings from the ``settings.json`` file, not the default settings IE after being processed by ``Settings.js``.

And yes this code was bugged since forever, the reason being is that if you require("Settings"); it has to init DB etc.  For tests it's probably fine to use ``settings.json`` but I wouldn't say it's ideal because your test settings might !== settings being loaded by ``Settings.js``

src/node/hooks/express/importexport.js
```diff
@@ -33,13 +38,38 @@ exports.expressCreateServer = function (hook_name, args, cb) {
   });
 
   // handle import requests
+  args.app.use('/p/:pad/import', limiter);
   args.app.post('/p/:pad/import', async function(req, res, next) {
     if (await hasPadAccess(req, res)) {
       let exists = await padManager.doesPadExists(req.params.pad);
       if (!exists) {
+        console.warn("Someone tried to import to a pad that doesn't exist");
```
muxator, line 46: I'll always try to add infos about the actual parameter values in the log messages. Rule of thumb is: a log message should be a `` `template string with some ${variable} value inside` `` instead of `"a simple fixed string"`.

This is more a operator rather than end-user UX, but I think it can improve ergonomics.
For example in this case we could add the non-existend pad name.

I'll do it here and in similar places, no problem.

JohnMcLear: yeah good point, will do so in future :)

JohnMcLear:
Am I blocking further review of this?

settings.json.template
```diff
@@ -409,6 +409,25 @@
...
+   * 
+   * importRateLimiting defaults at 10 requests per IP per 90 seconds.  After this it's rejected, value is in mS.  See express-rate-limit for more options
+   * requireAuthorSessionToImport defaults to true meaning that an author must be on the pad to import.
+   * importMaxFileSize defaults to 50Mb meaning that a single imported file max size is 50Mb.  File size is specified in bytes.
+   * 
+   * Uncomment to customize.
+   */
+
+  /*
+  "importRateLimiting": {
```
muxator, line 423: Looking at the code, the rate limiting is applied (correctly, IMHO), on both export and import routes:

https://github.com/ether/etherpad-lite/blob/bb409df1d82e78aba7499e12ff6ffd7db011ed60/src/node/hooks/express/importexport.js#L12

https://github.com/ether/etherpad-lite/blob/bb409df1d82e78aba7499e12ff6ffd7db011ed60/src/node/hooks/express/importexport.js#L12

The setting name should reflect it.

I'd use `importExportRateLimiting`, unless we find something less verbose.

settings.json.template
```diff
@@ -409,6 +409,25 @@
...
+    "windowMs": 90000,
```
muxator, line 424: I'd go with a seconds granularity. Too many digits for small benefit using ms.

src/node/utils/Settings.js
```diff
@@ -298,6 +298,24 @@ exports.scrollWhenFocusLineIsOutOfViewport = {
  */
 exports.exposeVersion = false;
 
+
+/*
+ * Import Rate limiting, File size and If a session is required to import
+ *
+ * importRateLimiting defaults at 10 requests per IP per 90 seconds.  After this it's rejected, value is in mS.  See express-rate-limit for more options
```
muxator, line 305: Question: did you test this feature behind a reverse proxy?

Is the middleware smart enough to honor the [`X-Forwarded-For`](https://en.wikipedia.org/wiki/X-Forwarded-For) Header? Otherwise the source IP will always be the reverse proxy's.

settings.json.template
JohnMcLear: Ack feel free to change setting name, that's a good shout!

settings.json.template
```diff
@@ -409,6 +409,25 @@
   "indentationOnNewLine": false,
   */
 
+  /*
```
muxator, line 412: These modifications will have to be ported to settings.json.docker, too (they have to be the same file). I can do it.

Do we want to expose the rate limiting config knobs via an environment variable under docker? I think it may make sense. Agreed?

settings.json.template
JohnMcLear: I use whatever the middleware uses.  https://www.npmjs.com/package/express-rate-limit  -- And I think we should stick with that so we can refer to their docs.

src/node/utils/Settings.js
JohnMcLear: It can be included in the json blob no?  

https://github.com/ded/express-limiter/blob/master/README.md#examples

settings.json.template
```diff
@@ -409,6 +409,25 @@
...
+   * Import Rate limiting, File size and If a session is required to import
+   * 
+   * importRateLimiting defaults at 10 requests per IP per 90 seconds.  After this it's rejected, value is in mS.  See express-rate-limit for more options
+   * requireAuthorSessionToImport defaults to true meaning that an author must be on the pad to import.
+   * importMaxFileSize defaults to 50Mb meaning that a single imported file max size is 50Mb.  File size is specified in bytes.
+   * 
+   * Uncomment to customize.
```
muxator, line 419: I think it's better leaving this section uncommented: after all, now import/export is always going to be rate limited.

New installas will inherit this file, and have the config knob ready. Old installs will use the default mechanism you implemented in Settings.js.

settings.json.template
JohnMcLear: I literally know nothing about ANY container platforms so I have no idea!  The docker guys would know :)

src/node/hooks/express/importexport.js
```diff
@@ -33,13 +38,45 @@ exports.expressCreateServer = function (hook_name, args, cb) {
   });
 
   // handle import requests
+  args.app.use('/p/:pad/import', limiter);
```
muxator, line 41: From now on, Etherpad will always have its import/export rate limited (eventually configurable).

We should mention this in the changelog (I'll do it), and in the documentation.

@JohnMcLear, what is the best place to document the fact that import/export is rate limited?

settings.json.template
JohnMcLear: Sounds good to me!  Happy for that change to be made :)  Sorry if I'm being defective in making these easy to make behavioral changes btw, I'm super deep into the whole list issues and it's reaching the limit of my mental capacity.

settings.json.template
muxator: Uhg, I still do not like it (end users/administrators are not required to know the internals of the library we are using, we do. And a human thinks in terms of seconds at best).

But I can also live with ms for sure. :)

src/node/utils/Settings.js
muxator: Don't we already have a setting that has to be set when running behind a reverse proxy (or am I wrong)?

- if not, let's add the configuration in `config.json.template`.
- if yes, let's verify if the middleware confarms to it (it should be an express-level configuration). If it does not, I say it's on us to further configure the rate limiter, without asking anything to the users.

settings.json.template
muxator: Ok I'll do it.

src/node/utils/Settings.js
muxator: Found the setting. Its name is `trustProxy`:

https://github.com/ether/etherpad-lite/blob/83d72d27a405f63782803e0168b1e9121f3cd650/settings.json.template#L302-L311

settings.json.template
muxator: Done. I'll introduce environment variables if we decide to uncomment the rate limiting configurations in the next thread.

muxator:
Quick questions: in this series, two final changes "bubbled up": they introduce and subsequently remove tests for the functionality (`tests/backend/specs/api/rateLimitFileSizeAndAuthorSessionExists.js`), leaving no trace of them.

I understand you said they cannot be implemented right now. **Can I drop those two commits** altogether?

src/node/hooks/express/importexport.js
JohnMcLear: +1 ``settings.json.template`` is good I guess?  Perhaps when a session IS rate limited we should drop something to logs to help admins debug? :)

JohnMcLear:
> Quick questions: in this series, two final changes "bubbled up": they introduce and subsequently remove tests for the functionality (`tests/backend/specs/api/rateLimitFileSizeAndAuthorSessionExists.js`), leaving no trace of them.
> 
> I understand you said they cannot be implemented right now. **Can I drop those two commits** altogether?

I can't remember but if they are commented out they can be left in.  The idea is that we have a reference for tests for when I can figure out how to create a session and put that session present in a pad.  That or provide a method to bypass that with ``loadtest: true``;

Either way it would be nice to have a reference / reminder knowing how infrequent I contribute :D

src/node/hooks/express/importexport.js
JohnMcLear: btw +1 good logging, I'm not sure I did this, if you don't mind it should be easy to add as a console.warn with details..  

Ref: onRateLimited: Function called when a request exceeds the configured rate limit.

Perhaps we should include that by default in the blob?   I'm +1 that!

src/node/utils/Settings.js
muxator: > It can be included in the json blob no?
> https://github.com/ded/express-limiter/blob/master/README.md#examples
>
> this is why I think we should keep the original ms approach instead of seconds :)

Ugh, that's another middleware :)

The rate limiter you used is this one: https://github.com/nfriedly/express-rate-limit

muxator:
> I can't remember but if they are commented out they can be left in.

No they are not commented out. The test file is introduced in a commit and wiped altogether in the next one. This is why I asked.

src/node/hooks/express/importexport.js
muxator: Please take care of the documentation part, or at least drop two lines here.

src/node/utils/Settings.js
JohnMcLear: Lol yeah sorry :|  It's been a while!

src/node/hooks/express/importexport.js
JohnMcLear: See: https://github.com/nfriedly/express-rate-limit#handler for handler controls.

src/node/utils/Settings.js
JohnMcLear: https://github.com/nfriedly/express-rate-limit/issues/165

JohnMcLear: tldr; we'd need to do ``app.set('trust proxy', 1);``, that would need to be a separate setting in itself?

src/node/hooks/express/importexport.js
JohnMcLear: Ack, will try sort 2mrw to get this unblocked.  Will take a few days off the list stuff because it's hurting my brain.

JohnMcLear:
In heinsight, I should have split rate limiting and author requirements up into two PRs.  My apologies @muxator.

muxator:
> In heinsight, I should have split rate limiting and author requirements up into two PRs. My apologies @muxator.

I was thinking about that just now. I'm going to split them.

1. max file size
2. author requirements
3. rate limiting

settings.json.template
muxator: Kept `windowMs`.

settings.json.template
muxator: Changed the name of the config section: `importRateLimiting` -> `importExportRateLimiting`.

Reworded the description.

settings.json.template
muxator: Ok agreed.

The new settings for import/export checks on size and author existence + rate limiting **will be already uncommented** in `config.json.template`. **Their values will be equal to the default ones**.

If old installs won't upgrade the configuration file, will inherit the default settings via Settings.js.

This is the backwards-compatibility mechanism that is already in place, btw.

src/node/hooks/express/importexport.js
muxator: > it's hurting my brain

You are not alone :)

src/node/hooks/express/importexport.js
muxator: Done by me for this PR.

src/node/utils/Settings.js
```diff
@@ -298,6 +298,12 @@ exports.scrollWhenFocusLineIsOutOfViewport = {
  */
 exports.exposeVersion = false;
 
+/*
+ * If true, importing to a pad is allowed only if an author has a session
+ * estabilished and has already contributed to that specific pad.
+ */
+exports.requireAuthorSessionToImport = true;
```
muxator, line 322: I like this being flipped on for future releases. But it is certainly a breaking change wrt old versions.

Are we going to keep it? If so, let's mention the change in `CHANGELOG.md`.

settings.json.template
```diff
@@ -409,6 +409,12 @@
   "indentationOnNewLine": false,
   */
 
+  /*
+   * If true, importing to a pad is allowed only if an author has a session
+   * estabilished and has already contributed to that specific pad.
+   */
```
muxator, line 432: I'd really like an help to better describe what happens when this setting is true (use case, only applies to imports via the web interface? Has an effect on API calls?).

Depending on this "brainstorming", we may also want to:
- how to describe this change in the docs (we can include this in #3858)?
- do we want to flip it on when an old installation upgrades to **1.8.3** (see my next comment)?

tests/backend/loadSettings.js
muxator: After removing all the unused imports (1b300242400e) this is what remains of `tests/backend/loadSettings.js`:

https://github.com/ether/etherpad-lite/blob/1b300242400e9c7b664a5b84985a230f5736a4ec/tests/backend/loadSettings.js#L8-L14

I say there is no point in keeping it. **The backend test specs can directly require `src/node/utils/Settings`**.

There is a problem, though: the `loadSettings` mechanism was put in place to be able to live patch a configuration object, we discovered it never worked, so we are discarding it, but it is needed for testing the docker build.

We'll have to re-implement it, and all the Settings class is ging to need to be reworked. :sob: Basically we'll have to get ride of that huge side effect performed at require() time (parses the cli, reads files, dynamically modifies its exports... all implicitly).

src/node/utils/Settings.js
JohnMcLear: I mean, IMHO this is a security patch.  I'd say it's required by default.  Also sessions here is confusing..  An author session is not the same as an Etherpad session.

Author session = Socket IO author session IE user is on pad.
Etherpad session = required for logging into "group pads"

Don't get confused!

tests/backend/loadSettings.js
JohnMcLear: Yep!

JohnMcLear: What's the conclusion here?  Anything required by me?

muxator: > What's the conclusion here?

- I am going to drop `etherpad-lite/tests/backend/loadSettings.js`. All the backend tests will directly require Settings, since that's the intended behaviour anyhow;
- I am letting `tests/container/loadSettings.js` live as it was, because it served its purpose, and it is going to be cahnged once Settings is reimplemented as a module that does not do heavy side effects on load

> Anything required by me?
Not here. See my next point on UI when the rate limiter triggers.

src/node/utils/Settings.js
muxator: > tldr; we'd need to do app.set('trust proxy', 1);, that would need to be a separate setting in itself?

I think the issue you mentioned (nfriedly/express-rate-limit#165) solves the question. `express-rate-limit` will always look at the `ip` field but - when told so via `trust proxy` setting, Express will automatically fill that value with `x-forwarded-for`.

Our `trustProxy` settings already does exactly that. I'll document it in the source.

Resolving the conversation.

src/node/hooks/express/importexport.js
```diff
@@ -3,8 +3,22 @@ var settings = require('../../utils/Settings');
 var exportHandler = require('../../handler/ExportHandler');
 var importHandler = require('../../handler/ImportHandler');
 var padManager = require("../../db/PadManager");
+var authorManager = require("../../db/AuthorManager");
+const rateLimit = require("express-rate-limit");
+
+settings.importExportRateLimiting.onLimitReached = function(req, res, options) {
+  // when the rate limiter triggers, write a warning in the logs
+  const ip = req.headers["x-forwarded-for"] || req.connection.remoteAddress;
```
muxator, line 11: This is not necessary (to be verified...): `req.ip` should already have the correct value once `settings.trustProxy` is set.

See: https://expressjs.com/en/guide/behind-proxies.html

settings.json.template
muxator: Done

src/node/hooks/express/importexport.js
muxator: Ok verified.

Issue #3450, opened 2018-08-11
txt export: ordered lists number == indentation level


muxator:
[edit]: **please see below for a more precise characterization of this bug**.

Manually insert this list in a pad:
![txt-export-bug-on-ol](https://user-images.githubusercontent.com/4067621/43986344-e1ea71bc-9d0f-11e8-8458-baeeb64fb017.png)

Export it as txt. The result will be:
```
1. a
1. b
1. c
1. d
```
This bug was discovered while replicating #3447: that bug inserts an html via API, and the text gets both displayed and exported wrongly.

This case is _simpler_: **the same list** is inserted manually via the GUI (and thus displayed correctly), yet the exported txt has **exactly** the same errors as in #3447.

zorrikon:
Btw, exported TXT lists start with with whatever level indent they are. So all items indented once start with "2." and those indented twice start with "3." etc.

JohnMcLear:
Test coverage in for this

JohnMcLear:
Fix here: https://github.com/ether/etherpad-lite/pull/3839

JohnMcLear:
Final part of fix landed here: https://github.com/ether/etherpad-lite/pull/4086

PR #3839, opened 2020-04-06
A different approach to txt export - WIP -- I DONT LIKE THIS

JohnMcLear:
In this PR we use ``html-to-text`` instead of our own logic to create the txt export.

Advantages of using a third party module:
  1. We only have to care about rendering out HTML.
  1. Improvements made upstream will be absorbed.
  1. Can test things on CLI.
  1. Consistent experience.

``html-to-text`` can do things like style heading(h1 et al) exports
``html-to-text`` is poorly maintained.  https://github.com/werk85/node-html-to-text/issues/198.  It looks like it's going to be landing in new hands soon so hopefully the bugs will be resolved.  

Currently I'm using a fork.  https://www.npmjs.com/package/@mxiii/html-to-text which you will need to install to test.

Assuming html-to-txt gets properly maintained we could look at using this for future releases.  I will leave this open until then..

Also this solves: https://github.com/ether/etherpad-lite/issues/3450

Todo
  - [ ] Check html-to-txt is being properly maintained
  - [ ] Tweak below code so tests pass (IE look at how urls are exports, what's expected by the test runner and what is outputted are different things but probably fixable.
  - [x] Make sure getText is returning the right content, I'm sure it's delivering original pad content?!
  - [ ] Move ``html-to-text`` settings into settings.json.template ??if possible?? so site admins can tweak their own export settings.
  - [ ] Remove duplicates of this code
```
    // get the text of this revision
    let text = htmlToText.fromString(html, {
      format: {
        unorderedList: function(elem, fn, options){
          var h = fn(elem.children, options);
          if(elem.attribs && (elem.attribs.class === 'indent')){
            return '    ' + h + '\n';
          }else{
            return h;
          }
        }
      },
      wordwrap: 130
    }
    return { text };
  });
```


JohnMcLear:
I think html to text is just not reliable enough of a module.  We will need an internal fix for the issues.

KillyMXI:
Hey, just fyi:
New version of `html-to-text` is finally live and I hope it is in a better shape now.

PR #3852, opened 2020-04-09
scroll: replace absolute import with relative one

muxator:
This was introduced by f1fcd16894e5 ("Add settings to scroll on edition out of viewport") in 2018-01-03.

@joassouza: was there a specific reason to use `require('/caretPosition')` instead of `require('./caretPosition')`, or was it just an oversight?


joassouza:
No reason at all. Just a oversight

PR #3855, opened 2020-04-10
travis: Add test for DB with docker-compose

### LINKED ISSUE #3806: Add database support tests in Travis-CI and DockerHub
I've built a docker container for etherpad with a few plugins and production mode enabled by default (see https://github.com/Monogramm/docker-etherpad), but more importantly it comes with a very simple test for several databases.
The overall idea is:

1. start etherpad and a database through docker-compose with etherpad configured to use the database container
2. download the etherpad home page and check that it is valid (check the title for instance)

The tests are done in a specific container, allowing both Travis-CI and Dockerhub to test the container. Executing the tests in Dockerhub could allow to NOT publish unstable containers (if that's wanted).
See https://docs.docker.com/docker-hub/builds/automated-testing/ for more details on this.

I think having this tests run in Travis might help to detect earlier issues like #3805.
I might try to send a PR for this if that's interesting for you and if I find the time 😅

madmath03:
travis: Add test for DB with docker-compose to fix #3806

I made this based on how we handled database support in our fork [Monogramm/docker-etherpad](https://github.com/Monogramm/docker-etherpad) but it might not be the best here.
@pierreprinetti let me know if you have any advice

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ script:
 
 env:
   global:
+    - IMAGE_NAME=etherpad:test
     - secure: "WMGxFkOeTTlhWB+ChMucRtIqVmMbwzYdNHuHQjKCcj8HBEPdZLfCuK/kf4rG\nVLcLQiIsyllqzNhBGVHG1nyqWr0/LTm8JRqSCDDVIhpyzp9KpCJQQJG2Uwjk\n6/HIJJh/wbxsEdLNV2crYU/EiVO3A4Bq0YTHUlbhUqG3mSCr5Ec="
     - secure: "gejXUAHYscbR6Bodw35XexpToqWkv2ifeECsbeEmjaLkYzXmUUNWJGknKSu7\nEUsSfQV8w+hxApr1Z+jNqk9aX3K1I4btL3cwk2trnNI8XRAvu1c1Iv60eerI\nkE82Rsd5lwUaMEh+/HoL8ztFCZamVndoNgX7HWp5J/NRZZMmh4g="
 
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -40,8 +41,22 @@ jobs:
...
-        - "docker run -d -p 9001:9001 etherpad:test && sleep 3"
+        - "docker build -t ${IMAGE_NAME} ."
+        - "docker run -d -p 9001:9001 ${IMAGE_NAME} && sleep 3"
+        - "cd src && npm run test-container"
+    - name: "Test a docker-compose with MySQL"
+      install:
+        - "cd src && npm install && cd -"
+      script:
+        - "docker-compose -f docker-compose_mysql.yml build"
+        - "docker-compose -f docker-compose_mysql.yml up -d && sleep 60"
```
muxator, line 52: Are we waiting for the DB to come up here? If so, is there a way to wait for some event, like - for example - a TCP port that starts accepting connections?

Hardcoded timeouts always make my scars aching again :smile:

muxator:
Added an humble comment on a `sleep` statement, but for all the serious stuff I totally rely on @madmath03 & @pierreprinetti here, since I am focusing on integrating the new functionalities for **1.8.3**.

.travis.yml
madmath03: > Are we waiting for the DB to come up here?

Yes, that's what is done here.

> If so, is there a way to wait for some event, like - for example - a TCP port that starts accepting connections?

Made it the same way the "_Test the Dockerfile_" job waits for Etherpad to be ready, but I guess it should be possible using custom netcat commands.

The best though would be to add the wait directly into the Dockerfile, so that Etherpad waits for the DB to be ready before starting.
This is something I've done in our fork through a custom simple script [Monogramm/docker-etherpad wait-for-service.sh](https://github.com/Monogramm/docker-etherpad/blob/master/template/wait-for-service.sh) but I was not sure if you were interested.
Also wondering if using the standard [wait-for-it](https://github.com/vishnubob/wait-for-it) might be better or not here.

muxator: > Made it the same way the "Test the Dockerfile" job waits for Etherpad to be ready

And it was a wise decision. Sticking to whatever a project is doing is a sign of maturity, thanks. In this case, a better way exists and we are deciding to use it.

> but I guess it should be possible using custom netcat commands.

Wow, I think I've done something similar in commit 4a17443a2e6a:
https://github.com/ether/etherpad-lite/blob/5503ebdb949eeb9e4058ee5e3e072436a14bc255/tests/frontend/travis/runner.sh#L23-L31

But I did not know of the existence of more engineered solutions.

About what to use (something in the Dockerfile, or a custom command invoked from `.travis.yml`, I'll only offer two remarks:
- a dockerised Etherpad can use various DB backends. Waiting for an external process to be ready may not be necesary in every case (think about dirtydb or sqlite).
- if possible, let's choose a single mechanism to wait for an open port where applicable (it may be a custom one or [wait-for-it](https://github.com/vishnubob/wait-for-it). Can we backport that mechanism to `ests/frontend/travis/runner.sh`, too?

muxator:
Scheduling for **1.8.4** (right after 1.8.3 is released, hopefully ~days).

Niwla23:
I think there should also be a user firendly version of this, meaning:
* preset Image name
* Forwarded ports
If you tweak these things it works as a production system.
#4095

madmath03:
Tweaks mentioned by @Niwla23 added to branch.

>   If you tweak these things it works as a production system.
>   #4095

I would rather say that it works as "_development_" system, not a production one since `NODE_ENV=development`

madmath03:
@muxator & @pierreprinetti Added a simple entrypoint with custom function to check if the external database is up.

Issue #3861, opened 2020-04-13
Broken Pad: TypeError: r.dropdowns is undefined - Changeset related. - Waiting on review.


JohnMcLear:
I finally have a clear bug that causes the r.dropdowns.

1. Minify is on w/ high maxage
1. Database is MySQL
1. Instance is medium traffic/volume but nothing concerning (loads super low).
1. Instance is runing in production mode.

## Steps to replicate
When user visits Etherpad the pad doesn't fully render.

## Things I tested.
1. .etherpad file can be exported from /export/etherpad
1. checkPad.js and extradPadData.js are no use.
1. Restart changes nothing.
1. /export/txt returns a blank pad with no warnings or errors in console.

## Error in console:
```
Error: mismatch error setting raw text in setDocAText testBrokenPad:798:13
    setDocAText https://hidden.com/p/testBrokenPad line 24 > scriptElement line 232 > Function:798
    setDocumentAttributedText https://hidden.com/p/testBrokenPad line 24 > scriptElement line 232 > Function:206
    setBaseAttributedText https://hidden.com/p/testBrokenPad line 24 > scriptElement line 232 > Function:10300
    withCallbacks https://hidden.com/p/testBrokenPad line 24 > scriptElement line 232 > Function:202
    inCallStack https://hidden.com/p/testBrokenPad line 24 > scriptElement line 232 > Function:492
    inCallStackIfNecessary https://hidden.com/p/testBrokenPad line 24 > scriptElement line 232 > Function:561
    withCallbacks https://hidden.com/p/testBrokenPad line 24 > scriptElement line 232 > Function:199
    setBaseAttributedText https://hidden.com/p/testBrokenPad line 24 > scriptElement line 232 > Function:10286
    ace_setBaseAttributedText https://hidden.com/p/testBrokenPad line 24 > scriptElement line 232 > Function:1017
    Ace2Editor/</editor[fnName]< https://hidden.com/javascripts/lib/ep_etherpad-lite/static/js/pad.js?callback=require.define:2857
    action https://hidden.com/javascripts/lib/ep_etherpad-lite/static/js/pad.js?callback=require.define:2810
    doActionsPendingInit https://hidden.com/javascripts/lib/ep_etherpad-lite/static/js/pad.js?callback=require.define:2830
    forEach https://hidden.com/javascripts/lib/ep_etherpad-lite/static/js/ace2_common.js?callback=require.define:12709
[...]

PR #3863, opened 2020-04-13
Hack fix for 3861

JohnMcLear:
I haven't tested much, this is just a patch fix to show a solution to the problem.  If someone wants to investigate further IE on the server then this would be a good reference.  This can probably be merged as it wont cause any damage / harm, or at least, shouldn't...

Solves: https://github.com/ether/etherpad-lite/issues/3861

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -737,6 +737,22 @@ function Ace2Inner(){
 
   function setDocAText(atext)
   {
+    if (atext.text === "") {
+      /*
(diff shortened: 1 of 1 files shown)


muxator:
Wow, conflicts. This PR was based on 1.6.5, from two years ago.

Please, for the future contributions, start again from a clean clone!

muxator:
I somehow managed to solve the conflicts.
Please, really, do your future contributions from a clean clone.

Merging.

Issue #3864, opened 2020-04-14
Latest develop import


JohnMcLear:
I'm gonna do some polish on this today.

- [x] import errors should bubble up to user.  
  - [x]  Max file size
  - [x] User has to contribute to pad before uploading -- Ideally just being present on a pad should be enough.

- [x] using lenient-import branch - PDF imports are not being properly decoded <-- I didn't have ``soffice or abiword`` set.
- [x] .Doc imports are not being properly decoded <-- I didn't have ``soffice or abiword`` set.
- [x] This page needs updating: https://github.com/ether/etherpad-lite/wiki/How-to-enable-importing-and-exporting-different-file-formats-with-AbiWord-or-LibreOffice
- [x] Need to test the whole UTF8 logic / issue, https://github.com/ether/etherpad-lite/issues/3019

muxator:
Fine, I was thinking, indeed, about opening an issue about these UI polishings that need to be done now that the import has been reworked.

The black of feedback on failed import was indeed evident yesterday while testing. Thanks!

JohnMcLear:
```
  "pad.importExport.importUnsuccessful.maxFileSize": "Import Failed, file size too big, contact your site admin to increase file size",
  "pad.importExport.importUnsuccessful.permission": "Import Failed, please contribute to the pad and try again!",
```

These should do it.

JohnMcLear:
https://github.com/ether/etherpad-lite/pull/3865

JohnMcLear:
I am just going to hide import functionality until their has been a contribution from that user.  Hide and say "you need to contribute to the pad prior to importing."  That seems logical.

JohnMcLear:
Okay PR is in, please review.

JohnMcLear:
Closing as this is all merged.

PR #3865, opened 2020-04-14
Notify the user on failed import for permission denied (and max file size, already integrated in develop)

JohnMcLear:
Details are on the github issue.

**Edit**: this is a follow up to #3864.

### DIFF
--- a/src/static/js/collab_client.js
+++ b/src/static/js/collab_client.js
@@ -313,6 +313,16 @@ function getCollabClient(ace2editor, serverVars, initialUserInfo, options, _pad)
     }
     else if (msg.type == "ACCEPT_COMMIT")
     {
+      /*
+       * this is the first time this user contributed to this pad. Let's record
+       * it, because it will be used for allowing import.
+       *
+       * TODO: here, we are changing this variable on the client side only. The
+       *       server has all the informations to make the same deduction, and
+       *       broadcast to the client.
+       */
+      clientVars.thisUserHasEditedThisPad = true;
+
       var newRev = msg.newRev;
       if (msgQueue.length > 0)
       {
--- a/src/static/js/pad_editbar.js
+++ b/src/static/js/pad_editbar.js
@@ -394,6 +394,17 @@ var padeditbar = (function()
 
     toolbar.registerCommand("import_export", function () {
       toolbar.toggleDropDown("import_export", function(){
+
+        if (clientVars.thisUserHasEditedThisPad) {
+          // the user has edited this pad historically or in this session
+          $('#importform').show();
+          $('#importmessagepermission').hide();
+        } else {
+          // this is the first time this user visits this pad
+          $('#importform').hide();
+          $('#importmessagepermission').show();
+        }
+
         // If Import file input exists then focus on it..
         if($('#importfileinput').length !== 0){
           setTimeout(function(){
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -144,5 +144,6 @@
   "pad.impexp.importfailed": "Import failed",
   "pad.impexp.copypaste": "Please copy paste",
   "pad.impexp.exportdisabled": "Exporting as {{type}} format is disabled. Please contact your system administrator for details.",
-  "pad.impexp.maxFileSize": "File too big. Contact your site administrator to increase the allowed file size for import"
+  "pad.impexp.maxFileSize": "File too big. Contact your site administrator to increase the allowed file size for import",
+  "pad.impexp.permission": "Import is disabled because you never contributed to this pad. Please contribute at least once before importing"
 }
--- a/src/node/handler/PadMessageHandler.js
+++ b/src/node/handler/PadMessageHandler.js
@@ -939,6 +939,16 @@ async function handleClientReady(client, message)
     });
   }));
 
+  let thisUserHasEditedThisPad = false;
+  if (historicalAuthorData[statusObject.authorID]) {
+    /*
+     * This flag is set to true when a user contributes to a specific pad for
+     * the first time. It is used for deciding if importing to that pad is
+     * allowed or not.
+     */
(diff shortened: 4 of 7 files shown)


muxator:
This should go in **1.8.3**, but there are conflicts.

muxator:
Conflicts solved.

muxator:
I'll continue in the next days.

JohnMcLear:
@muxator bump :)

src/static/js/collab_client.js
```diff
@@ -159,6 +159,9 @@ function getCollabClient(ace2editor, serverVars, initialUserInfo, options, _pad)
         rev=newRev;
         if (msg.type == "ACCEPT_COMMIT")
         {
+          // useful for knowing if a user has committed content to pad or not
+          clientVars.thisUserHasEditedThisPad = true;
```
muxator, line 163: Please initialize this variable to `false` somewhere earlier (perhaps at first client connection?)

src/node/handler/ImportHandler.js
```diff
@@ -82,6 +82,12 @@ async function doImport(req, res, padId)
         if (err) {
           console.warn("Uploading Error: " + err.stack);
         }
+
+        // I hate doing indexOf here but I can't see anything to use...
+        if (err.stack.indexOf("maxFileSize") !== -1) {
+          reject("maxFileSize");
+        }
```
muxator, line 89: Doesn't exist a way of properly bubbling the error?

The server logs are not so good looking when an import is rejected when a file is too big:

```
When the 
[2020-04-21 22:02:53.499] [WARN] console - Uploading Error: Error: maxFileSize exceeded, received 11353 bytes of file data
    at Stream.<anonymous> (<BASE>/src/node_modules/formidable/lib/incoming_form.js:222:19)
    at Stream.emit (events.js:210:5)
    at Stream.EventEmitter.emit (domain.js:475:20)
    at MultipartParser.parser.onPartData (<BASE>/src/node_modules/formidable/lib/incoming_form.js:387:14)
    at callback (<BASE>/src/node_modules/formidable/lib/multipart_parser.js:102:31)
    at dataCallback (<BASE>/src/node_modules/formidable/lib/multipart_parser.js:115:11)
    at MultipartParser.write (<BASE>/src/node_modules/formidable/lib/multipart_parser.js:237:15)
    at IncomingForm.write (<BASE>/src/node_modules/formidable/lib/incoming_form.js:159:34)
    at IncomingMessage.<anonymous> (<BASE>/src/node_modules/formidable/lib/incoming_form.js:125:12)
    at IncomingMessage.emit (events.js:210:5)
```

JohnMcLear: No, this error is coming from formidable.

JohnMcLear:
1 & 2 are handled.  3 Does bubble up afaik but we don't suppress the server side msg.

src/node/handler/ImportHandler.js
muxator: I see. But a way to catch it must exist.

Let's skim on it for now (== forever). Closing.

muxator:
> 1 & 2 are handled

Indeed. Perfect, thanks!

muxator:
There is still one last thing I want to see & I am having a (probably unrelated) nasty crash on pad loading that is blocking me.

src/static/js/pad_editbar.js
```diff
@@ -392,6 +392,38 @@ var padeditbar = (function()
...
+
+                /*
+                 * I'm setting this because it might be useful in the future.
+                 * Also, we don't want to iterate authors every time we click
+                 * import/export
+                 */
+                hasEdits = true;
+              }
+            });
+          }
```
muxator, line 417: This code covers the case in which we open a pad (or hit F5) where we have already contributed.

Shoudn't be the server to already send us the correct value for `thisUserHasEditedThisPad`? He is the one setting `historicalAuthorData`, after all.

muxator:
I have split the functionalities in this PR. Now they are independent.

1. the first one implements the error message when the imported file is too big. This works very well and does not need any modifications;
2. the second one disables/enables the import UI when the user contributes to the pad at least once. This requires a modification moving some logic from the client to the server side.

I am going to import 1, which is ready, directly on the `develop` branch.

This PR now only deals with 2 (import permission), which is the only one needing fixes.

JohnMcLear:
RE: 

> the second one disables/enables the import UI when the user contributes to the pad at least once. This requires a modification moving some logic from the client to the server side.

The server does check to ensure the client exists..  The client side code is just to provide nice UX.

I don't see much benefit moving the logic to the server?

muxator:
> I don't see much benefit moving the logic to the server?

It's the server that sends the client variables to the client on first connection.

`thisUserHasEditedThisPad` is a new variable we are introducing in this PR, whose value on first connection was `undefined` (before my comment) and now `false` (after my request of initializing it).

I'd say it's safe to expect that something that is called `thisUserHasEditedThisPad` lives up to its name. It's the server that creates it, he knows what to do.

Moving the logic on the server and removing the client side code guarantees an invariant, that otherwise we are introducing and breaking at the same time.


If we do not do this, we are re-enacting one of the main ways of giving life to bugs.

src/static/js/pad_editbar.js
JohnMcLear: I don't see the gain.  Either the server sends a boolean or we iterate the object and set the boolean.  Either way the boolean is set.

JohnMcLear:
Okay noted RE shifting code to server.  Will look to see if this is workable within time constraints.

JohnMcLear:
Fix is in, give that a go :)

muxator:
> Fix is in, give that a go :)

It is **perfect** now! Kudos!

I've just squashed the commits and moved a comment around.

JohnMcLear:
Tnx, sorry for push back you were right to make me rework it

Issue #3830, opened 2020-04-03
pads silently disconnect when browser suspends a background tab


ryanpitts:
My team uses etherpad-lite for a lot of internal planning work and to document open community meetings, so we often have multiple pads open in different tabs for extended periods of time. Many times when a pad loses connection (because someone shut their laptop and their machine went to sleep, or because the browser offloads tabs that have been in the background for a while), the pad will show a "disconnected" prompt. But not always.

We regularly experience silent disconnections—if you happen to notice and reload, the pad reconnects just fine. But if you return to a tab and _don't_ notice that it disconnected, changes that you make aren't stored and it's easy to lose a lot of work. We're totally aware of _why_ a socket might lose connection, but it's unclear why sometimes there's a warning and sometimes there's not.

Is this a known issue? Are we strange in the way we use etherpads? I feel a little awkward about filing this as a bug, but I did a ton of searching and I can't find much description of this problem anywhere.

JohnMcLear:
No, you are using Etherpad right and this is not a known issue.  You should have a disconnected msg and an overlay.  Can you provide some exact steps to replicate?

Thanks for letting us know.

This is the connection status logic: https://github.com/ether/etherpad-lite/blob/develop/src/static/js/pad_connectionstatus.js

And the auto reconnect logic: https://github.com/ether/etherpad-lite/blob/develop/src/static/js/pad_automatic_reconnect.js

joassouza:
We had the some problem with a user with unstable internet connection. This bug is quite difficult to replicate and really serious. We've tried to simulate an unstable connection without success

JohnMcLear:
Looking into this now.

JohnMcLear:
Wait..  In settings.json

```

  /*
   * Time (in seconds) to automatically reconnect pad when a "Force reconnect"
   * message is shown to user.
   *
   * Set to 0 to disable automatic reconnection.
   */
  "automaticReconnectionTimeout": 0,

```

This is the default, what's the logic here?  Surely that's not a good default setting?! :D  cc @muxator any idea what the deal is here?

My method for initial testing is..

1. connect to pad.  Type stuff.
1. minimize browser.
1. killall node (makes server disappear)
1. wait a few mins (I'd love to know what the timeout is here)
1. tab back to browser, see if I have any msg.

luixxiul:
````
  "automaticReconnectionTimeout": 0,
````

[...]

PR #3868, opened 2020-04-14
Timeout logic - Ready for review

JohnMcLear:
Solution for https://github.com/ether/etherpad-lite/issues/3830

- [x] Doesn't allow edit when in a disconnected state (dunno how that happened).
- [ ] Doesn't make for badchangeset upon reconnection
OR
- [x] Temp fix in place which is not as great a UX but works fine.

### DIFF
--- a/src/static/js/pad.js
+++ b/src/static/js/pad.js
@@ -173,8 +173,20 @@ function sendClientReady(isReconnect, messageType)
   //this is a reconnect, lets tell the server our revisionnumber
   if(isReconnect == true)
   {
+    // Hammer approach for now.  This is obviously wrong and needs a proper fix
+    // TODO: See https://github.com/ether/etherpad-lite/issues/3830
+    document.location=document.location;
+
+    // Switching to pad should work but doesn't...
+    // return pad.switchToPad(padId); // hacky but whatever.
+    // It might be related to Auth because failure logs...
+    //   [ERROR] console - Auth was never applied to a session.
+    //   If you are using the stress-test tool then restart Etherpad
+    //   and the Stress test tool.
+
     msg.client_rev=pad.collabClient.getCurrentRevisionNumber();
     msg.reconnect=true;
+
   }
 
   socket.json.send(msg);
--- a/src/static/js/pad.js
+++ b/src/static/js/pad.js
@@ -210,6 +210,7 @@ function handshake()
   });
 
   socket.on('reconnecting', function() {
+    padeditor.disable();
     pad.collabClient.setStateIdle();
     pad.collabClient.setIsPendingRevision(true);
     pad.collabClient.setChannelState("RECONNECTING");


JohnMcLear:
Okay I only have time for a temp fix that reloads the pad upon reconnect, it's not ideal but it works.

**It is a vastly better UX that what's currently in develop** but really the pad contents should reload without an entire page refresh.  At least an entire page refresh in Etherpad is cheap but still, it's hacky.

muxator:
Scheduled for **1.8.4**.

muxator:
> At least an entire page refresh in Etherpad is cheap

It is when Etherpad runs in production mode. Otherwise is super slow. I'm thinking about changing the default running mode to production for this reason. Developers can switch it back if they want.

JohnMcLear:
I'm inclined to agree RE auto deploy as production.

JohnMcLear:
Feels like this can be merged?

rhansen:
Would you mind rebasing this onto current develop?

```
git checkout timeout-logic
git rebase origin/develop
git push -f
```

JohnMcLear:
done

JohnMcLear:
@rhansen if you can do reviews/lemme know if things work it might speed things up with the backlog of PRs?  I rebased a bunch of other PRs.

rhansen:
Thanks for rebasing.

I'm going to test this PR to see if it reduces the number of spontaneous mid-edit disconnects I'm seeing. Unless this PR breaks something horribly, I probably won't be able to report back until the next round of heavy usage on my Etherpad a week from now (for a weekly meeting).

rhansen:
With this PR merged in, I have a large pad that keeps reloading itself over and over on its own.

The pad is 967 lines long (many of of those lines are actually long paragraphs), about a dozen authors, and months of edit history. It takes about 55 seconds to load on my machine (`NODE_ENV=production`, sqlite3 database).

Could the long load times be triggering a connection timeout which then triggers the reload?

JohnMcLear:
Can you hit me up with a ``.etherpad`` export file.  I get the whole "weekly meeting" thing but if you can be if you want me to help you (essentially for free) I kinda need you to reciprocate...

Fwiw I dropped sqlite3 on the next version of ueber so is that a deal breaker for you?

rhansen:
> Can you hit me up with a `.etherpad` export file.

Sent you an email with the export attached.

More details about my Etherpad setup:
  * Current tip of `develop` branch (f79b50d3118401b4ff5f457a1188e861a1ec0c54)
  * Run in a Docker container
  * `NODE_ENV=production`
  * Plugins:
    * ep_adminpads2
    * ep_align
    * rhansen/ep_author_neat
    * ep_delete_empty_pads
    * ep_font_color
    * ep_font_family
    * ep_font_size
    * ep_headings2
    * ep_padlist2
    * ep_prompt_for_name

> Fwiw I dropped sqlite3 on the next version of ueber so is that a deal breaker for you?

I can spin up a PostgreSQL server and feed it a dump of the sqlite3 database if necessary, but I'd rather stick with sqlite3. I like sqlite3 because it's low maintenance: (a) It's trivial to back up the database, (b) it's trivial to move the database to another server, (c) I don't have to worry about username+password management, and (d) I don't have to worry about accidentally exposing a TCP port to the world.

rhansen:
Without this PR, I get a silent disconnect (no pop-up) after the pad finishes loading and all edits are silently discarded. So I guess the behavior with this PR is better than without—at least I know it's broken. :smile:

rhansen:
Another thing to note: I tried Chromium (I usually use Firefox) and it loaded much faster and did not disconnect when it finished loading.

JohnMcLear:
Pad loads in about 1 second for me. 

Initial load is slow because it's creating minified JS/CSS.

Can you see if disabling plugins makes a difference?  

I'm using dirtydb, no plugins, firefox.

rhansen:
I'll do some more troubleshooting to see if I can identify the source of slowness.

Ignoring my particular setup, I think that merging this PR is better than not. Users with slow servers, slow clients, or massive pads might start complaining about reload loops, but without this PR I think those reload loops would instead be silent discards, and that's worse.

Thoughts for a future PR (or this one if you feel up to it):
  * Is it feasible to replace the automatic reload with a "you have been disconnected" pop-up? Would that be better?
  * I'm curious to know what's causing the infinite reload loop, and whether the initial pad load can be restructured to avoid it. I don't know much about Etherpad's internals, but I suspect there's some sort of heartbeat (WebSockets ping/pong?) that continually tests the connection health, and if a heartbeat times out it triggers a reconnect. If that's true, I wonder if it's possible to postpone starting the heartbeats until the pad has finished its initial load.

rhansen:
The slowdown I'm experiencing seems to be caused by the ep_author_neat plugin (clkao/ep_author_neat#17).

Please merge this. The lost edits bug is severe enough that this should be merged, even if improvements could be made. (Those improvements can be made in a future pull request.)

PR #3870, opened 2020-04-14
PadMessageHandler: fix for scoping error hiding original error

squiddle:
`r` is undefined outside of the for loop, but used in the catch block of the try
statement

This fix is a quickfix.

This PR is based on devlop


muxator:
Good catch, thanks!

PR #3873, opened 2020-04-16
docker: make settings fully configurable via env vars

sapkra:
* Refactored docs for env vars using tables
* Added new env vars for nearly all configuration options
* Added possibility to set env setting to `null` using env vars

Please note: I was not able to test everything. Theoretically everything should work.

### DIFF
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -58,11 +58,11 @@ The `settings.json.docker` available by default enables some configuration to be
 
 | Variable           | Description                                                                                | Default                                                                                                                                                                                                                             |
 | ------------------ | ------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `TITLE`            | The name of the instance                                                                   |                                                                                                                                                                                                                                     |
-| `FAVICON`          | favicon default name, or a fully specified URL to your own favicon                         |                                                                                                                                                                                                                                     |
-| `DEFAULT_PAD_TEXT` | The default text of a pad                                                                  |                                                                                                                                                                                                                                     |
-| `IP`               | IP which etherpad should bind at. Change to `::` for IPv6                                  |                                                                                                                                                                                                                                     |
-| `PORT`             | port which etherpad should bind at                                                         |                                                                                                                                                                                                                                     |
+| `TITLE`            | The name of the instance                                                                   | `Etherpad`                                                                                                                                                                                                                          |
+| `FAVICON`          | favicon default name, or a fully specified URL to your own favicon                         | `favicon.ico`                                                                                                                                                                                                                       |
+| `DEFAULT_PAD_TEXT` | The default text of a pad                                                                  | `Welcome to Etherpad! This pad text is synchronized as you type, so that everyone viewing this page sees the same text. This allows you to collaborate seamlessly on documents! Get involved with Etherpad at https://etherpad.org` |
+| `IP`               | IP which etherpad should bind at. Change to `::` for IPv6                                  | `0.0.0.0`                                                                                                                                                                                                                           |
+| `PORT`             | port which etherpad should bind at                                                         | `9001`                                                                                                                                                                                                                              |
 | `ADMIN_PASSWORD`   | the password for the `admin` user (leave unspecified if you do not want to create it)      |                                                                                                                                                                                                                                     |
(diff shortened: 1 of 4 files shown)


muxator:
Wow, nice work, thanks!

In the weekend I'll test it and merge ASAP.

flecno:
Yeah! Very nice! Thank you so much :tada: :raised_hands:

flecno:
What is wrong with the CI tests?

muxator:
Now that #3744 is merged I'll have to do a small rework.

sapkra:
@muxator I've updated the PR to include the skin variants.

muxator:
> @muxator I've updated the PR to include the skin variants.

Thanks, it's not common seeing such attention. Really appreciated.

doc/docker.md
```diff
@@ -54,30 +54,140 @@ And point your browser to `http://<YOUR_IP>:<DESIRED_PORT>`
...
+| --------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------- |
+| `SHOW_SETTINGS_IN_ADMIN_PAGE`     | hide/show the settings.json in admin page                                                                                                                                                              | true                                                                                                              |
+| `TRUST_PROXY`                     | set to `true` if you are using a reverse proxy in front of Etherpad (for example                                                                                                                       | Traefik for SSL termination via Let's Encrypt). This will affect security and correctness of the logs if not done |
```
muxator, line 166: I guess the text ended up being separated by mistake. I'll fix it.

muxator: Done

muxator:
> Added possibility to set env setting to null using env vars

I am going to split this in a dedicated commit immediately before your change.

src/node/utils/Settings.js
```diff
@@ -441,6 +442,15 @@ function coerceValue(stringValue) {
...
     }
 
+    if (stringValue === "null") {
+      /*
+       * There is already a way of setting a default value to null, and it is
+       * by using the syntax ${VAR_NAME}. This allows to use ${VAR_NAME:null} as
+       * well.
+       */
+      return null;
+    }
```
muxator, line 452: I have split this modification in a dedicated commit (55dd44cc8048) and tried to document it.

However, I am not sure if it's better to include this or not, because it introduces an edge case (the string "null" would become special, and a crazy admin would no longer be able to call his user "null").

In theory, there is already complete syntactic support for "null" values without introducing special strings. See for example:

https://github.com/ether/etherpad-lite/blob/70990afd6680045b9887559b18e8a380aabc814c/settings.json.docker#L59-L66

On the other side, I understand that `${VAR_NAME:null}` is more explicit than `${VAR_NAME}`.

My questions for you are:
1. did you introduce this snippet because the documentation was not evident enough? In this case, I would prefer to drop this change & improve the documentation;
2. or did you do it because you were aware of the existing support but found it uncomfortable? In this case, let's discuss. I do not like the edge case, but user ergonomics is important.

muxator:
I am going without introducing the coercition `"null"` -> `null`.

I am adding this snippet in `settings.json.template`, hoping it clarifies what is supported and what is not.

```
If you want to use an empty value (null) as default value for a variable,
simply do not set it, without putting any colons: "${ABIWORD}".
```

The `settings.json.docker` file in this PR is going to be updated accordingly.

muxator:
Mentioned in `DB_TYPE` that the default is "not set, thus will fall back to `DirtyDB` (please choose one instead)"

muxator:
This PR is really useful.

Many thanks for taking the time to prepare and update it.

Pulled in.

muxator:
@sapkra: I have discovered that the documentation generator that Etherpad uses (https://github.com/markedjs/marked) is an ancient version, and **does not render Markdown tables** at all.

See what happens to your wonderful docker documentation:

image

Updating the library helps, but it's still horrible:

image

I stumbled on this issue while working on another documentation PR (#3921). If you are willing to help, you are welcome to drop a line there.

sapkra:
Sorry but I have to say that I'm not even using etherpad yet. But I may will in the future and will create a few PRs then.

I just created these PRs to cleanup the docker-jitsi-meet repository a bit.

muxator:
Ok no problem, thanks!

Consider this just a follow-up while we are moving towards the release.

sapkra:
Maybe you can use something like docusaurus 2.0 to prettify the docs and/or website.

PR #3877, opened 2020-04-17
Clean HTML prior to import & Content Collector test coverage - READY FOR REVIEW

JohnMcLear:
Done:
1. Introduce ``contentcollector.js`` tests
1. Fix issue with OL LI items not being properly numbered after import
1. Fix issue with nested OL LI items being improperly numbered on export
1. Fix issue with new lines not being introduced after lists in on import #3961
1. Sanitize HTML on the way in (import) <-- conflicts with another PR.
1. Fix ExportHTML CSS because it needs to support ``OL > LI > OL`` not ``OL > OL`` [The latter being the correct format]
1. Fix backend tests.

Solves: https://github.com/ether/etherpad-lite/issues/3620 and others from the UL/LI mega issue thread..

I get that this is a big PR and it's complex touching lots of files / endpoints but it's touching a lot of legacy code and really bringing in a lot of resilience so please be patient and caring with merging it.  I expect there are edge cases, let's do our best to find them and get them in before merging.

**Still work in progress, but ready for testing..  Please be gentle and patient with me**

ETA for completion is mid may.  I expect a few weeks of back and forth getting as close to perfect as is possible considering this is modifying a HTML parser which is no small task.

Also I promise this is the last big PR from my fork then I will nuke and only send clean PRs.  But after this I will probably tone down my open source contributions. <3

## Run tests with

```
cd src && npm run test-content-collector
```


src/static/js/linestylefilter.js
```diff
@@ -100,8 +102,8 @@ linestylefilter.getLineStyleFilter = function(lineLength, aline, textAndClassFun
             {
               classes += ' list:' + value;
             }
-            else if (key == 'start')
-            {
+            else if (key == 'start'){
```
JohnMcLear, line 105: No salt plz :P

Issue #3711, opened 2020-03-14
Startup is really slow due to a bunch of npm ops, it is a lot slower than it used to be.  Perhaps provide mechanism to skip checking npm deps?


JohnMcLear:

muxator:
I have been using the following script (`bin/fastrun.sh`) since a long time when developing. It just runs Etherpad, assuming all the dependencies are in place. I never felt like worth commiting.

```bash
#!/bin/bash

set -eu

# source: https://stackoverflow.com/questions/59895/how-to-get-the-source-directory-of-a-bash-script-from-within-the-script-itself#246128
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"

echo "Running directly, without checking/installing dependencies"
node "${DIR}/node_modules/ep_etherpad-lite/node/server.js" "${@}"
```

Also, see the startup statement in the Dockerfile:

https://github.com/ether/etherpad-lite/blob/467fc11b723723275f1ebd0fed556a0525ea5f0b/Dockerfile#L50

Yes, Etherpad is slow to start because `run.sh` ends up running `npm install` every time, in time `npm` itself became more strict with its resolution dependency, and the dependencies got bigger. Try uploading [package.json](https://github.com/ether/etherpad-lite/blob/467fc11b723723275f1ebd0fed556a0525ea5f0b/src/package.json) to https://bundlephobia.com, for example.

JohnMcLear:
fastRun.sh is an awesome idea.  

Can you get that committed as a PR please and include a modification to the README and Developer docs?  +1 thanks!

Also, should we maybe have a message for developers when they run Etherpad that they can use fastRun if they don't want to check for deps each time?

Feels like a low hanger but defo worth it!

muxator:
[fastRun.sh](https://github.com/ether/etherpad-lite/issues/3711#issuecomment-599135831) was committed in 8956efc4ae020bb75663be32b73de6fc8d36597b.

The final nail in the coffin for this will be when we will start using `npm ci` instead of `npm install` in `installDeps.sh`. The RFC is #3778, hopefully scheduled for **1.8.4**.

PR #3878, opened 2020-04-17
Adds fastRun script for developers

seballot:


bin/fastRun.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -eu
+
+# source: https://stackoverflow.com/questions/59895/how-to-get-the-source-directory-of-a-bash-script-from-within-the-script-itself#246128
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+
+echo "Running directly, without checking/installing dependencies"
+node "${DIR}/node_modules/ep_etherpad-lite/node/server.js" "${@}"
```
muxator, line 9: This won't work, because in this PR `fastRun.sh` is inside `bin`.

But it's really my fault, because in https://github.com/ether/etherpad-lite/issues/3711#issuecomment-599135831 I've indicated that I used `fastRun.sh` from `<base>/bin`, while really I use it from `<base>`.

I'll push the necessary corrections :)

muxator: Done

muxator:
Merged, thanks

muxator:
Yesterday I made a mistake and pushed the not working version on `develop`. I had to force push the branch just now.

The correct commit is this one: 8956efc4ae02. Sorry.

seballot:
ah ah yes I checked this mroning without seeing any change with your new version :)

But for me, the version I pushed works fine. From base folder I run `sh bin/fastRun.sh` and it's working nicely... Maybe behavior is different from docker? (I'm not using it for etherpad)

PR #3879, opened 2020-04-17
Using hype WIP - Ready for review/merge but be mindful of the contentcollector PR as that one has better support than this one.  Keeping owner if you want granual increments

JohnMcLear:
Part of my huge experiment in lists.  This is actually good to merge but please don't, I wanna keep working.


muxator:
I am confident that this is ok, but it is out of scope for **1.8.4**.

Moving to the next release. Hopefully its intent can be described a bit more clearly in the meantime. :smile:

JohnMcLear:
It's an easy one..

```
  rehype()
    .use(format, opts)
    .process(html, function(err, output){
      if(err) console.error("Error making HTML nice");
      html = String(output);
      html = html.replace(/(\r\n|\n|\r)/gm,"");
  });
``` 
is pretty much all the logic.  rehype basically takes crappy html and sanitizes it and if it's bad it makes it less bad.  It fixes a load of import issues we currently **have that are relatively severe bugs.**  Lots of clients are complaining about these so I wanted to prioritize.

PR #3899, opened 2020-04-19
Fix timeslider rev tests and temp disable attrib tests

JohnMcLear:
Things to note.

1. Had to disable all scroll tests.
2. Had to disable one of @lpagliari tests, and modify her other.
3. Only testing against Firefox until Saucelabs is fixed.


JohnMcLear:
Mheh just merge this and see what happens, I have lost the will to live.

muxator:
Would you be ok with moving this to **1.8.4** or would this cause an unrecoverable loss of happiness?

muxator:
Since these changes are in test code and not in Etherpad's core, I was willing to merge them for the next release. But there are unsolved conflicts, so it will have to wait. :(

PR #3905, opened 2020-04-19
Socketio bump

JohnMcLear:
You asked for it....  📦


muxator:
Danke schön.

We'll bump the major version immediately after releasing **1.8.3**, so that we have time to stabilize them.

PR #3909, opened 2020-04-20
close all gritters on escape key

JohnMcLear:
This PR closes gritter pop ups when your focus is in the editor.  It closes ALL gritter notifications.  

This solves the problem of gritters taking too much focus from the editor, it is an accessibility requirement.


muxator:
Merged, thanks.

Issue #3907, opened 2020-04-20
New FlexLayout / Browser compatibility


seballot:
We should test it seriously on most possible browser. I guess there are some tools to do that online, if you know some please inform me :)

Also, I think we should add a prefixer in our assets pipeline.
Prefixer convert css rule like `display: flex` to something like
```
display: -webkit-box;
display: -ms-flexbox;
display: flex
```

How the assets pipeline works for now? Where does the css minify takes place? If you could point me the right direction would be nice :)

Please assign to me

luixxiul:
Maybe it is too late but how about implementing [`normalize.css`](https://github.com/csstools/normalize.css)? It would definitely break the existing plugins layout so I am not quite if it is right time to do so...

seballot:
Hi ! thanks but that is different, and I already added it along with refactoring ;)

https://github.com/ether/etherpad-lite/blob/develop/src/static/css/pad/normalize.css

JohnMcLear:
![image](https://user-images.githubusercontent.com/220864/79734065-03744400-82ee-11ea-8a57-213a3a8542fd.png)

Note the white bar at the bottom clipping the pad text

Firefox on Windows

luixxiul:
by the way @JohnMcLear 幸せの肉 = Meat of happiness / Happy Meat 😄

muxator:
> How the assets pipeline works for now? Where does the css minify takes place? If you could point me the right direction would be nice :)

Have a look at [src/node/utils/Minify.js](https://github.com/ether/etherpad-lite/blob/4ceb42603e2c9d4051df1a9140728ea30f2f833f/src/node/utils/Minify.js).

That specific part (along with the "peculiar" startup sequence and cache handling) is one of the parts of thecodebase that really show their age.

I won't spend more than a couple of lines on that file. Anything more complex deserves taking the time to tear apart what exists and rebuild on more modern foundations (hopefully without bringing in too much complexity).

It can be a fruitful line of work, indeed.

JohnMcLear:
> by the way @JohnMcLear 幸せの肉 = Meat of happiness / Happy Meat 😄 🍖

Hahah no way I worked in Nagoya and Tokyo for a bit and I never was told this!  Japanese politeness I guess!  Awesome translation!  :P

JohnMcLear:
[...]

PR #3913, opened 2020-04-21
css: some fix for browser compatibility refs #3907

seballot:


muxator:
> css: some fix for browser compatibility refs #3907 (#3913)

If possible give a hint of what you have done in the commit messages. Having only pointers to github issues ties us to GitHub and is one more level of indirection when reading the code.

PR #3915, opened 2020-04-21
Import bug fixes

JohnMcLear:
* ace2_inner.js needs top for any ``console warn``
* Close modal after succesfull import
* Proper evaluation of ``directDatabaseAccess`` as it's a string not a bool.

- [ ] Still todo.  Fix duplicate page issue.

cc @seballot


muxator:
Does `ace2_inner.js` need `top` for only for `console.warn()` or for `console.whatever()` too?

src/static/js/pad_impexp.js
```diff
@@ -243,8 +243,19 @@ var padimpexp = (function()
...
+      // directDatabaseAccess is a string not a boolean
+      if(directDatabaseAccess !== 'undefined'){
+        // commented out for now.  This needs a proper fix after 1.8.3
+        // TODO: pad.switchToPad() should not require a reload of the page.
+        // pad.switchToPad(clientVars.padId);
+        location.reload();
+        // Switches to the pad without redrawing the page
+        $('#import_export').removeClass('popup-show');
       }
-      if(directDatabaseAccess) pad.switchToPad(clientVars.padId);
```
muxator, line 247: If `directDatabaseAccess` was a literal `"undefined"` string, this means that this instruction should had never fired.

Is this the case?

src/static/js/pad_impexp.js
```diff
@@ -243,8 +243,19 @@ var padimpexp = (function()
       if (status !== "ok")
       {
         importFailed(status);
+      }else{
+        $('#import_export').removeClass('popup-show');
+      }
+
+      // directDatabaseAccess is a string not a boolean
+      if(directDatabaseAccess !== 'undefined'){
```
muxator, line 251: Do we really want to do this instead of finding the server side bug?

src/static/js/pad_impexp.js
```diff
@@ -243,8 +243,19 @@ var padimpexp = (function()
...
+        // commented out for now.  This needs a proper fix after 1.8.3
+        // TODO: pad.switchToPad() should not require a reload of the page.
+        // pad.switchToPad(clientVars.padId);
+        location.reload();
```
muxator, line 255: This sidesteps the "two columns" bug.

How is it possible that `pad.switchToPad()` causes a duplication of the DOM nodes? @seballot work may have helped to uncover it, but I doubt this is just due to his recent work.

Can we investigate?

JohnMcLear:
> 
> 
> Does `ace2_inner.js` need `top` for only for `console.warn()` or for `console.whatever()` too?

``.whatever`` too

src/static/js/pad_impexp.js
JohnMcLear: I wrote the switchToPad logic maybe 8 years ago, I can look more into why it's duplicating but I'm worried about release Vs amount of time in day to get all these things done!  So this is a patch fix to ship 1.8.3 which causes very _little negative_ user impact as a trade off.

src/static/js/pad_impexp.js
JohnMcLear: It was doing this before....  This isn't new code, it's just fixing it up.

src/static/js/pad_impexp.js
JohnMcLear: It worked at one point, something must have changed it to break it.  I am kinda surprised by it too..

muxator:
> .whatever too

ok, thanks. I'll commit a change for the whole `ace2_inner.js`. There are mixed `console.whatever()` statements there, and people may get confused.

src/static/js/pad_impexp.js
muxator: I am ok with keeping the refresh _for now_. I want to get this out.

Can you assure me that the day after releasing **1.8.3** we can try to fix it?

src/static/js/pad_impexp.js
muxator: Look at this. May it be here?

https://github.com/ether/etherpad-lite/blob/789dd7acd6a92d4b750ba2bf290690090052ecb3/src/node/handler/ImportHandler.js#L283

src/static/js/pad_impexp.js
JohnMcLear: I can fix it before you release but don't merge it because I'd rather release something stable/guaranteed to work than something optimized.  Assurance made.

JohnMcLear:
```

jose@server:~/etheretherpad-lite$ git bisect bad
5fd6aeeea62674cecf997421546a675d91cf45ef is the first bad commit
commit 5fd6aeeea62674cecf997421546a675d91cf45ef
Author: Sebastian Castro <sebastian.castro@protonmail.com>
Date:   Thu Apr 2 14:36:49 2020 +0200

    css: refactor element positioning

    No more javascript to change css properties
    Remove a number of useless tables
    Try to stop positioning elements with absolute, but use flex-boxes instead

    Adds comment to pad template, and move popups and chatbox inside editorcontainerbox (so absolute positioning is straightforward)

    Make the design more consistent: always use base color, font-family and font-size. USe relative font size if necessary (.9rem instead of 11px for example)

    Remove two columns in the popups, just use one column

    Remove css meant to support old browser (like -webkit-box-shadow, -moz-box-shadow). Those css rules are quite common now, and If we want to support very old browser, we should use clean-css or other tools to add them automatically

:040000 040000 978d82573baabc336d70aa1ddf2f7de12eba090e 5d24f199c525d3cab707d13f1a89d2b736a8f4d1 M   src

```

5fd6aeeea62674cecf997421546a675d91cf45ef

JohnMcLear:
Fault applies on both ``no-skin`` and ``colibiris``.  It was introduced in https://github.com/ether/etherpad-lite/commit/5fd6aeeea62674cecf997421546a675d91cf45ef by @seballot - I expect it was the change to pad.html

seballot:
What is the problem here I don't understand?

JohnMcLear:
> 
> 
> What is the problem here I don't understand?

For some reason your commit (above) is causing the duplicate frame issue on ``.etherpad`` import.  I used git bisect to discover the cause.

JohnMcLear:
If you do ``git checkout 0603bf8097b6d3366d4c65435916a4e6f4c5536a`` which is the parent, then re-run Etherpad you will notice ``.etherpad`` file import works as expected (doesn't create two editors).

JohnMcLear:
Ignore my notes re pad.html / ace.js not being the fault, I was being an idiot.  They still may be..

JohnMcLear:
yeah not ``pad.html`` or ``ace.js`` -- I'm at a bit of a loss..  It's defo this conmit but I can't narrow it down any further

JohnMcLear:
@seballot I tried reverting the changes to ``pad.html`` and ``ace.js`` thinking that they surely were the cause but that didn't fix it.  I'm at a bit of a loss..  I hope I have pointed you in the right direction though if you can take a look at your code to see if it's something obvious?  

This is what I used to figure out this was the offending commit: https://mclear.co.uk/2013/04/02/using-git-bisect-to-debug-etherpad-issues/

seballot:
I had a look, but cannot understand either

JohnMcLear:
@seballot okay well I will take another look and see if I can figure it out!

JohnMcLear:
Figured it out.  Jesus what a nasty bug..  Good job @seballot this was 100% NOT your fault.  This is my fault and your "good" code highlighted my sloppy job :)  

image

Note the two ``ace_outer`` elements...  This was in the older versions....  *sighs..  I need to nuke one before bringing the other one in..

JohnMcLear:
Fix https://github.com/ether/etherpad-lite/pull/3925

seballot:
ohh so you mean the two iframe have always been there after import, but as they were positioned in absolute the last created was just on top of the old one ! Funny one ah ah

JohnMcLear:
Yes exactly :D  *laughs in shame.

muxator:
Kudos! Great job!

I am going to prepare the release during the weekend, since almost everything is ready (I have still a small problem with the doc generation, fix underway).

Just to be clear: #3925 and this PR go together, right?

JohnMcLear:
3925 does the job afaik, this isn't needed?  Feel free to close.

muxator:
Everything that was in this PR was integrated (e.g. closing of the modal after import, changes on logging in ace2_inner), or replaced with a better solution (e.g.: removing the duplicate iframe instead of reloading the page).

The only non merged change that is still contained in this PR is the following:
```diff
diff --git a/src/static/js/pad_impexp.js b/src/static/js/pad_impexp.js
--- a/src/static/js/pad_impexp.js
+++ b/src/static/js/pad_impexp.js
@@ -253,7 +253,8 @@ var padimpexp = (function()
         $('#import_export').removeClass('popup-show');
       }
 
-      if (directDatabaseAccess) {
+      // directDatabaseAccess is a string not a boolean
+      if (directDatabaseAccess !== 'undefined') {
         // Switch to the pad without redrawing the page
         pad.switchToPad(clientVars.padId);
         $('#import_export').removeClass('popup-show');
```

What problem did it fix?

If it is important, a proper fix should be done, maybe looking at
https://github.com/ether/etherpad-lite/blob/789dd7acd6a92d4b750ba2bf290690090052ecb3/src/node/handler/ImportHandler.js#L283

Closing, let's eventually open a new PR to define & fix this last issue.

PR #3918, opened 2020-04-22
chore(docker): removing git-objects from final docker image

ratopi:
Removing git-objects, which are not needed, from final docker image, to make it smaller.

@muxator : This follows wrong fix in issue #3776

Code base on branch "develop"

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,22 @@
 .hg
+
+# Remove the git objects, logs, etc. to make final image smaller.
+# Some files still need to be in the .git directory, because Etherpad at
+# startup uses them to discover its version number.
+.git/branches
+.git/COMMIT_EDITMSG
+.git/config
+.git/description
+.git/FETCH_HEAD
+.git/hooks
+.git/index
+.git/info
+.git/logs
+.git/objects
+.git/ORIG_HEAD
+.git/packed-refs
+.git/refs/remotes/
+.gitignore
+
 settings.json
 src/node_modules


ratopi:
BTW: I don't understand, why one should fix a thing on a unstable branch!?

The failed tests are not broken by my change. ;-)

You should consider to accept changes against a stable branch (master?).  In case you do, I can change my pull request to be based on another branch.

muxator:
> Removing git-objects, which are not needed, from final docker image, to make it smaller.

This is a great idea, actually better & simpler than what I suggested at https://github.com/ether/etherpad-lite/pull/3776#issuecomment-605492321. No code changes needed, and the best code is the one that you do not write :smile: 

You gave me an idea, mixing your first proposal (putting the whole `.git` inside `.dockerignore`) and this PR: let's put these entries:

https://github.com/ether/etherpad-lite/blob/d682983c8de966660f65ea2d1b24e72a9cc27a55/Dockerfile#L37-L38

Directly inside the `.dockerignore`.

In this way we won't even send the contents of the .git folder to the docker daemon via the context mechanism, and building the container will be faster!

muxator:
I have tried the `.dockerignore` trick: no new bugs introduced, and the container is 23 MB smaller :)

I'll push the changes and merge them.

ratopi:
@muxator :-D  ... sometimes one is blind ... could not see the simple thing.

.dockerignore : Good job! 👍

PR #3925, opened 2020-04-23
Fix duplicated iframes after importing text in a pad

JohnMcLear:


muxator:
This becomes part of #3915, correct?

(I'll do in the weekend)

JohnMcLear:
Yea but I kept it separate because it was something that was historically broken and someone might want to cherry pick it into a previous version...   Also it keeps things clean.

muxator:
Ok

muxator:
For documentation's sake: this is an old bug in `pad.js`: after an import there were always two iframes with duplicated content, and the last one was simply on top of the old one.

It was not evident with the old absolutely positioned layout, but after moving to flexbox it stood out.

Issue #3932, opened 2020-04-24
Timeslider errors when content is deleted and content is no longer consistant. - AWAITING REVIEW/MERGE


JohnMcLear:
I'm placing this as an early catch, this needs working on urgently.

https://youtu.be/Wf0TJa3CucQ shows the timeslider content being incorrect.

Error is: 

```
Uncaught Error: Can't copy unknown attrib (reference attrib string to non-existant pool entry). Inconsistent attrib state!
    at Object.error (timeslider.js?callback=require.define:2540)
    at timeslider.js?callback=require.define:4033
    at String.replace (<anonymous>)
    at Object.exports.moveOpsToNewPool (timeslider.js?callback=require.define:4030)
    at Object.handleMessageFromServer (timeslider.js?callback=require.define:6157)
    at r.<anonymous> (timeslider.js?callback=require.define:93)
    at r.emit (index.js:83)
    at r.onevent (index.js:83)
    at r.onpacket (index.js:83)
    at r.<anonymous> (index.js:83)
```
**And it is reproducable by deleting all pad content**.

This error was present in 1.6.2..  Jeez

## Steps to replicate
1. Create pad
1. Open timeslider in another window
1. Delete pad contents.
1. Type new pad contents.

## who is impacted
Anyone who reviews pad content[with the timeslider when a pad is being edited.  Most people don't use this functionality but it's a really key part of reviewing pad edits and IMHO a decent feature worth promoting/fixing.

JohnMcLear:
## timeslider
```          console.warn("obj.apool", obj.apool)
          console.warn("padcontents.apool", padContents.apool)
          var changesetBack = Changeset.moveOpsToNewPool(
            changesetBack, (new AttribPool()).fromJsonable(obj.apool), padContents.apool);
```
Returns

![image](https://user-images.githubusercontent.com/220864/80241635-d7ddba80-865b-11ea-9f66-ebf3f49845bf.png)

JohnMcLear:
Surgical gloves were required for this one...  *wipes sweat from brow..

muxator:
Nice catch. This can go in **1.8.3** for sure.

PR #3934, opened 2020-04-24
Fix inconsistencies when deleting content while timeslider is open

JohnMcLear:
It's a simple fix but it had to be nuanced enough to solve it properly.

Bug Ref: #3932

### DIFF
--- a/src/static/js/Changeset.js
+++ b/src/static/js/Changeset.js
@@ -1529,7 +1529,17 @@ exports.moveOpsToNewPool = function (cs, oldPool, newPool) {
   return upToDollar.replace(/\*([0-9a-z]+)/g, function (_, a) {
     var oldNum = exports.parseNum(a);
     var pair = oldPool.getAttrib(oldNum);
-    if(!pair) exports.error('Can\'t copy unknown attrib (reference attrib string to non-existant pool entry). Inconsistent attrib state!');
+
+    /*
+     * Setting an empty pair. Required for when delete pad contents / attributes
+     * while another user has the timeslider open.
+     *
+     * Fixes https://github.com/ether/etherpad-lite/issues/3932
+     */
+    if (!pair) {
+      pair = [];
+    }
+
     var newNum = newPool.putAttrib(pair);
     return '*' + exports.numToString(newNum);
   }) + fromDollar;


muxator:
Moving to **1.8.4** since still needs some love.

JohnMcLear:
What love does it need?  Imho this is urgent...

muxator:
Hi,

I moved the issue to **1.8.4** because that was the explicit indication given in the ticket. Until 2 hours ago, the starting comment of this PR https://github.com/ether/etherpad-lite/pull/3934#issue-408695006 read:

> It's a simple fix but it had to be nuanced enough to solve it properly..
>
> **Ugh, don't merge, I just realized it doesn't copy attributes that are set..** <-- this was deleted 2 hours ago
>
> Bug Ref: #3932

I will be glad to merge the PR if it's useful and it's mature.

Please explain clearly what problem it's fixing and what are the remaining drawbacks, if any.

Then we can decide if it's material for **1.8.3** or has to slip to **1.8.4**.

JohnMcLear:
Yea my bad RE that comment I thought I removed it.



What it's fixing.

The problem is that if you are streaming(Play button active) the timeslider and another editor removes all pad attributes (by deleting content) then the logic said "redo the attribute pool from X to Y" but failed if Y was [].  This allows Y to be [] IE the attribute pool to be "empty".  This failing logic caused pad content to be incorrect!  My fix is to allow the attribute pool to be changable from X to [] without erroring.



Drawbacks

I'm not sure, I ran all tests after and tried to replicate other weird edge cases.  It seems the logic only affects timeslider..  I was pretty surgical in the fix going as far deep into the Changeset as I could to ensure I wasn't affecting anything else negatively.  We really need better "collaborative in real time" test coverage I guess to test these Changeset changes.



Thanks!



-----Original message-----

muxator:
Ok, thanks.

I tried to replicate it, but I did not succeed.

What's the way for removing all pad attributes? I've tried removing all the text (CTRL+A, DELETE) while streaming in another window, but I was not able to crash anything.

Is some plugin needed? What's the expected (buggy) outcome? A crash?



**p.s.**: when trying to replicate this issue, I spotted **another** bug in the timeslider: if I delete some text  (for example via backspace), while the play button is running, the deleted text does not disappear from the timeslider window. Worth another issue for a later time, imho.

JohnMcLear:
Afaik that's the bug, if you apply the patch do you still not see the content deleting?  Look in console you should see the error in timeslider when you hit backspace..  *without the patch.  With the patch the pad content should be correct...

muxator:
Confirmed the bug, confirmed the fix.
Scheduling for **1.8.3**, thanks!

JohnMcLear:
You are welcome, and yeah, this is probably my best fix to date for 1.8.3.   Thanks!

PR #3943, opened 2020-04-26
release: prepare for 1.8.3

muxator:
This release is _huge_.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,28 @@
 # 1.8.3
+* FEATURE: colibris is now the default skin for new installs
+* FEATURE: improved colibris visuals, and migrated to Flexbox layout
+* FEATURE: skin variants: colibris skin colors can be easily customized. Visit http://127.0.0.1:9001/p/test#skinvariantsbuilder
 * REQUIREMENTS: minimum required Node version is **10.13.0 LTS**.
+* MINOR: stability fixes for the async migration in 1.8.0 (fixed many UnhandledPromiseRejectionWarning and the few remaining crashes)
+* MINOR: improved stability of import/export functionality
+* MINOR: fixed many small UI quirks (timeslider, import/export, chat)
+* MINOR: Docker images are now built & run in production mode by default
+* MINOR: reduced the size of the Docker images
+* MINOR: better documented cookies and configuration parameters of the Docker image
+* MINOR: better database support (especially MySQL)
+* MINOR: additional test coverage
+* MINOR: restored compatibility with ep_hash_auth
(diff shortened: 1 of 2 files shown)


Gared:
Because of the breaking changes this should be the 1.9 release

muxator:
> Because of the breaking changes this should be the 1.9 release

Maybe my wording in the changelog is too strong.

* There is no API change for plugins at all, only the CSS layout changed a lot. The real breaking change happened when we migrated to async in **1.8**.
* Requiring node 10 is something that was widely anticipated, and the program will stop n startup with a very explicit message.

If these are the only changes, I'd be inclined to stay at 1.8.x. But I may be forgetting something.

Anyway: I do not have strong feelings towards the version number. Right now it is a strange hybrid. Not time based a-la Mozilla, not semantic. We'd need a policy on that.

Gared:
Okay, I see. As you mentioned there is no policy which version ist next, but in my opinion after 1.8.0 (current Version) should follow 1.8.1 and not 1.8.3. What happened with the other versions?

muxator:
Expect a less than robust answer here :cry: 

When **1.8.0-beta1** was released on November 2019 the [changelog](https://github.com/ether/etherpad-lite/blob/503b11730c4ee6fe0fff50209b5126b73268eccd/CHANGELOG.md) stated:

> REQUIREMENTS: minimum required Node version is 8.9.0 LTS. Release 1.8.3 will require at least Node 10.13.0 LTS

I am trying to stick to that statement, because I think it would be less surprising than going back to 1.8.1 requiring Node 10.

Since this release is big enough, I am comfortable with skipping two minors.


It's a compromise. I really would be happier to put a clear plan in place. This depends on how much community engagement we are going to gain back.

Would you be willing to contribute to a proposal for the next version? I would be happy to have a wide agreement.

tiblu:
Congrats on the 1.8.3! The new layout looks good!

> There is no API change for plugins at all, only the CSS layout changed a lot. The real breaking change happened when we migrated to async in 1.8.

I'm not sure what's your take on it, but MAYBE the big layout changes should have been at least minor version update, that is 1.9.
The trick is that some deployments MAY use custom CSS overrides directly or using plugins like ep_themes_ext (https://www.npmjs.com/package/ep_themes_ext) or ep_custom_styles (https://www.npmjs.com/package/ep_custom_styles) to modify the layout. 
IF the CSS only does overrides to the default bundled with EP and the default changes, there is high chance for layouts to break, something that you would probably not expect from patch version?

That said, from our perspective, there is no need to revert or change anything now, but brought this to attention in case you're discussing versioning and releases.

Again, I'm not sure what is the correct approach, just thinking out loud.

PR #3944, opened 2020-04-26
startup: use "npm ci" instead of "npm install" in startup scripts

### LINKED ISSUE #3778: installDeps.sh: use "npm ci" instead of "npm install"?
In #3750 a user implicitly suggested we could use `npm ci` to install dependencies. At the moment, `bin/installDeps.sh` uses `npm install`:

https://github.com/ether/etherpad-lite/blob/1ad6e9c2880ed074ba7f6185bdb399d84b298310/bin/installDeps.sh#L103

Reading from the [`npm ci` docs](https://docs.npmjs.com/cli/ci.html):

> it’s meant to be used in [...] any situation where you want to make sure you’re doing a clean install of your dependencies.
> It can be **significantly faster** than a regular npm install by skipping certain user-oriented features.
> It is also more strict than a regular install, which **can help catch errors or inconsistencies caused by the incrementally-installed local environments** of most npm users.

## Possible advantages
1. **increased reproducibility** of the installations (both direct and via Docker);
2. once installed, **checking dependencies would be very fast**. This would solve the underlying issue below #3711.

   ~~Afterwards, we can still commit [`fastRun.sh`](https://github.com/ether/etherpad-lite/issues/3711#issuecomment-599135831) if we want.~~ **Edit**: `fastRun.sh` was commited in 8956efc4ae020bb75663be32b73de6fc8d36597b and will be part of **1.8.3**);
3. **often, user reported problems are due to poorly installed environments**. This could help reduce the confusion.

## Possible problems
1. **compatiblity with installing plugins** has to be checked.

   Plugins are internally installed via `npm install`, which would render inconsistent the contents of `node_modules` and `package-lock.json`.
   On the other side, plugins seem to end up in a different directory.

   This really has to be checked.

2. among other things, being more strict means that:

   > **If dependencies** in the package lock **do not match** those in package.json, **npm ci will exit with an error**, instead of updating the package lock.

   This **may confuse our users**, but it's better to have a clear error which a clear solution (== reinstall) than having half-backed installs laying around, or random bug report coming in;

3. something different I may be overlooking

muxator:
Advantages:
- reproducible install: every user will have the same, exact install, instead of
  a slightly different one
- speed: installation of dependencies is measurably faster
- explicit: if a user setup is broken, from now on he'll have a clear error
  message

Closes #3778

**TODO**:
Check impact on plugins (both via npm install and the admin interface)

### DIFF
--- a/bin/installDeps.sh
+++ b/bin/installDeps.sh
@@ -100,7 +100,7 @@ echo "Ensure that all dependencies are up to date...  If this is the first time 
   cd node_modules
   [ -e ep_etherpad-lite ] || ln -s ../src ep_etherpad-lite
   cd ep_etherpad-lite
-  npm install --save --loglevel warn
+  npm ci
 ) || {
   rm -rf src/node_modules
   exit 1
--- a/bin/installOnWindows.bat
+++ b/bin/installOnWindows.bat
@@ -14,7 +14,7 @@ cd /D node_modules
 mklink /D "ep_etherpad-lite" "..\src"
 
 cd /D "ep_etherpad-lite"
-cmd /C npm install --loglevel warn || exit /B 1
+cmd /C npm ci || exit /B 1
 
 cd /D "%~dp0\.."
 


muxator:
Compatibility with plugin installation was validated here: https://github.com/ether/etherpad-lite/issues/3778#issuecomment-618587303

Merging now.

This will end up in **1.8.5** (or whatever comes after **1.8.4**).

Issue #3094, opened 2016-12-09
Question: What CSP headers to add to secure for etherpad


aswen:
Hi there,
I don't really understand CSP so I fail to compose the right CSP config for my etherpad instance. Can anyone here give me an example of what it should be to keep Etherpad working but improve security on it?
I think it could improve documentation when such examples are added.
Thanks a lot.

JohnMcLear:
Afaik we did CSP years ago, pretty sure you don't need to do anything.

hex-m:
I just installed the latest version and use nginx as reverse proxy, configured like in the [documentation](https://github.com/ether/etherpad-lite/wiki/How-to-put-Etherpad-Lite-behind-a-reverse-Proxy). Mozilla Observatory tells me

> Content Security Policy (CSP) header not implemented

So I think "you don't need to do anything" is not true when you use a proxy.

JohnMcLear:
Yea, I did the scan on the video site and get a similar result.

JohnMcLear:
Ah whoops I scanned the wrong hostname, doing again for video.etherpad.com...  Fwiw I did .org 📦

JohnMcLear:
- [ ] Look to see which parts of https://observatory.mozilla.org/analyze/video.etherpad.com we can do in Etherpad code base w/ out damaging user experience.  This will probably mean putting some stuff in settings.
 - [ ] Bits we can't do we should document

Note on Video.etherpad.com the results are

![image](https://user-images.githubusercontent.com/220864/78786771-368b0f00-79a1-11ea-9210-9b62bd15aeea.png)

JohnMcLear:
@hex-m fwiw in nginx I did
```
add_header Content-Security-Policy "default-src ws://video.etherpad.com https://video.etherpad.com 'unsafe-inline'; script-...
add_header Strict-Transport-Security "max-age=31536000; includeSubDomains" always;
add_header X-Frame-Options "SAMEORIGIN";
add_header X-Content-Type-Options nosniff;
```

Just as a quick play around I got it from F to B, or something..

hex-m:
@JohnMcLear Thanks a lot!

In your config example the first line is truncated.

What I came up with (researching and trying stuff - not really knowing what I do):

```
add_header Strict-Transport-Security "max-age=63072000; includeSubDomains" always;
add_header X-Content-Type-Options "nosniff" always;
add_header Content-Security-Policy "frame-ancestors https://my-jitsi-server.example.com;";
[...]

PR #3953, opened 2020-04-27
Csp nonce - Security! - Ready for Review/Merge

JohnMcLear:
This provides CSP functionality inside of Etherpad

Etherpad has a lot of inline JS, instead of removing it (which would be nice but not required) CSP provides the functionality to generate a nonce (generated using the ep_helmet plugin) and providing that to the page.

Pages tested and fixed:
- [x] Index
- [x] Pad
- [x] Timeslider
- [x] Pad Import
- [x] Admin pages
- [x] Test pages (known broken but WIP on another branch so don't want to work on it)
- [x] A bunch of plugins.
- [ ] Ace has a placeholder Nonce in for when it renders the outerHTML to the page.  I can't find a good way for this but I'm not actually convinced it's exploitable...

Required for https://github.com/ether/etherpad-lite/issues/3094

### DIFF
--- a/src/static/js/ace.js
+++ b/src/static/js/ace.js
@@ -35,7 +35,7 @@ var _ = require('./underscore');
 
 function scriptTag(source) {
   return (
-    '<script type="text/javascript">\n'
+    '<script type="text/javascript" nonce="aceTemporaryNonce">\n'
     + source.replace(/<\//g, '<\\/') +
     '</script>'
   )
--- a/src/templates/index.html
+++ b/src/templates/index.html
@@ -5,7 +5,7 @@
 <html>
 
         <title><%=settings.title%></title>
-    <script>
+    <script nonce="<%=settings.nonce && settings.nonce.padLicense || ''%>">
       /*
       |@licstart  The following is the entire license notice for the
       JavaScript code in this page.|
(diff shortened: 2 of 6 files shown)


JohnMcLear:
To get the actual nonce values ep_helmet is required, else no nonce is provided.

src/static/js/ace.js
```diff
@@ -35,7 +35,7 @@ var _ = require('./underscore');
 
 function scriptTag(source) {
   return (
-    '<script type="text/javascript">\n'
+    '<script type="text/javascript" nonce="aceTemporaryNonce">\n'
```
rhansen, line 38: Shouldn't this nonce come from `settings.nonce`?

src/templates/index.html
```diff
@@ -5,7 +5,7 @@
 <html>
 
         <title><%=settings.title%></title>
-    <script>
+    <script nonce="<%=settings.nonce && settings.nonce.padLicense || ''%>">
```
rhansen, line 8: If `settings.nonce.padLicense` is nullish/empty, wouldn't it be better to not add the `nonce` attribute? Like this:

```suggestion
    <script <%=(settings.nonce && settings.nonce.padLicense) ? `nonce="${settings.nonce.padLicense}"` : ''%>>
```

rhansen, line 8: A nonce should only be used once: refreshing should get a different nonce, and two different users should never see the same nonce. If the same nonce is used every time, an attacker could visit the page, look at the nonce, then use it in the injected code.

JohnMcLear:
Closing as stale.

PR #3957, opened 2020-04-28
 Some ui fix after 1.8.3 closes #3954

### LINKED ISSUE #3954: UI issue after 1.8.3 release
- [x] in safari, the select dropdowns inside the toolbar (ep_headings, ep_font_color) do not work (because of the overflow of the toolbar. I already noticed this bug, but on chrome and firefox using a `position: fixed` fixed it. Not on safari :) )

- [x] A strange vertical line is displayed on middle of the screen on load : https://youtu.be/MiJ5AcS6S-g

- [x] Index page need to use new font
- [x] ep_set_title_on_pad is not colored by skin variants
Others small stuff I already fixed, not yet pushed

Please assign to me !

seballot:
see #3954

### DIFF
--- a/src/static/js/pad_editbar.js
+++ b/src/static/js/pad_editbar.js
@@ -312,7 +312,11 @@ var padeditbar = (function()
       // reset style
       $('.toolbar').removeClass('cropped')
       var menu_left = $('.toolbar .menu_left')[0];
-      if (menu_left && menu_left.scrollWidth > $('.toolbar').width()) {
+
+      // on mobile the menu_right get displayed at the bottom of the screen
+      var isMobileLayout = $('.toolbar .menu_right').css('position') === 'fixed';
+
+      if (menu_left && menu_left.scrollWidth > $('.toolbar').width() && isMobileLayout) {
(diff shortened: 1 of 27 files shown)


src/static/js/pad_editbar.js
```diff
@@ -303,7 +307,9 @@ var padeditbar = (function()
       // reset style
       $('.toolbar').removeClass('cropped')
       var menu_left = $('.toolbar .menu_left')[0];
-      if (menu_left && menu_left.scrollWidth > $('.toolbar').width()) {
+      // on mobile the menu_right get displayed at the bottom of the screen
+      var isMobileLayout = $('.toolbar .menu_right').css('position') == 'fixed';
```
muxator, line 311: If possible always use "===" instead of "==". I'll fix it.

muxator: Done

muxator:
I like when a PR solves a problem, and actually results in _less_ code than it started with!

Pulled in, thanks.

seballot:
thanks !

Issue #3849, opened 2020-04-08
[Docker] Setting Static API Key?


jkaosw:
I'm attempting to set up an instance of Etherpad in Kubernetes to be embedded by one of our applications.

However, upon doing a quick test for reliability/availability on Kubernetes, when destroying and redeploying the Etherpad pod, it respawns with a new API key in the APIKEY.txt file.  We don't intend to have high availability in terms of having multiple replicas, but having the API key change upon restart seems not so great.

Is there any way to force an Etherpad container to spawn up with a set API key currently?

JohnMcLear:
Afaik you just set it in ``APIKEY.txt`` in the root folder before running it for the first time?

I could be wrong...  It's been a long time since I touched that logic.

jkaosw:
Interesting.  I'm just curious how that logic works.  Does Etherpad check that file to verify the APIKEY? If so, mounting a ConfigMap over it would be fine and we could just run for that.

I will test tomorrow and put findings here.

JohnMcLear:
Afaik yes.  But this is code I wrote X years ago, it may have changed and I'm too lazy to look :)

muxator:
@jkaosw, when Etherpad starts, it looks for `APIKEY.txt` at the root of its installation (the basedir of the repo).

You can see this in the logs at startup time:

```
[2020-04-09 01:35:42.366] [INFO] APIHandler - Api key file read from: "<base>/APIKEY.txt"
```

If that file does not exist, it is created with random contents:
```
[2020-04-09 01:36:02.122] [INFO] APIHandler - Api key file "<base>/APIKEY.txt" not found. Creating with random contents.
```

If you do not like the directory where `APIKEY.txt` lives, you can pass the `--apikey` parameter at startup with the full path to an arbitrary text file:

```
node <BASE>/node_modules/ep_etherpad-lite/node/server.js --apikey /somewhere/NAME_OF_APIKEY.EXT
```

The default location will be overridden and not looked for. All the rest will stay the same.


The code that implements this functionality is here:
https://github.com/ether/etherpad-lite/blob/df08883a009aeaebfbe59a53c592ac25cd3aff06/src/node/utils/Cli.js#L48

**However**, in Docker:

Currently, there is no support for passing command line arguments in the official Dockerfile:
https://github.com/ether/etherpad-lite/blob/df08883a009aeaebfbe59a53c592ac25cd3aff06/Dockerfile#L53

[...]

PR #3958, opened 2020-04-28
assets: add versioning to override the browser cache on deploys closes #3955

### LINKED ISSUE #3955: Versioning assets for clearing browser cache
On the server I run with etherpad, I noticed after moving to 1.8.3 that I needed to refresh my browser cache to see the changes. That is not very nice because most of the user will not do that.

We could easily change that by adding a version when loading assets

`<link href="path/to/css?v=1.8.3">`

WDYT? please assign to me if you agree

seballot:


Gared:
This exposes the installed etherpad Version. Maybe we can generate a local Hash instead?

muxator:
Good observation, @Gared. We can't do expose the version in that way.

We may generate a random string, like we do for APIKEY.TXT, and use hash(RND.TXT + version) as hash.

seballot:
Is that okay if I use APIKEY.txt instead of generating a new key?

Gared:
I think this should work if you hash APIKEY.txt + version

seballot:
ok done !

muxator:
Thanks, @seballot.

I am thinking of transforming the next version **1.8.4** in a stabilizing release for **1.8.3**, just including your recent fixes on Colibris, the fix for the performance regression on MySql, and this very PR (which solves an annoying bug).

Since I am not an expert, I do not feel confident on evaluating the consequences of using `APIKEY.txt` + a not cryptographically secure hash algorithm for generating the version hash.

I am thinking of reworking this PR and make it use a new, independent file instead of `APIKEY`. No action needed on your part, thanks!

seballot:
Hi ! 
> of transforming the next version 1.8.4 in a stabilizing release for 1.8.3,
Yes I agree, I was about to propose you the same thing !

> reworking this PR and make it use a new, independent file
Yes you right, probably better. I didn't found in the code where is was generated that's why I used a already generated one :)
Actually I am thinking: maybe instead of relying on the version of etherpad (combined with a generated key), we could use a random string that we generate again each time etherpad is restarted? So it will clear cache when people update etherpad, but also if they do a hotfix or whatever change and they restart etherpad. WDYT?

muxator:
> Actually I am thinking: maybe instead of relying on the version of etherpad (combined with a generated key), we could use a random string that we generate again each time etherpad is restarted? So it will clear cache when people update etherpad, but also if they do a hotfix or whatever change and they restart etherpad. WDYT?

I was thinking about it, too.

Advantages of using a random string:
* code simplicity;
* this code will be thrown away with joy once we have a proper build pipeline. Throwing away a cheaper thing generates less suffering;
* no new file is introduced. Each new file may require special attention by administrators. See for example what one of our users is doing with `APIKEY.txt` when running under Kubernetes: https://github.com/ether/etherpad-lite/issues/3849#issuecomment-611724196
* no security risks whatsoever.

Drawbacks of using a random string:
* after a server restart each client will eventually hit the server requesting the assets, storming it

Overall, the advantages seem to be more that the drawbacks.

Let's go with a short random string generated at server start and kept in memory.

muxator:
This is the last remaining issue for **1.8.4**. I'll try to work on it now or tomorrow so the release can be unblocked.

muxator:
This change (using query strings for cache busting) has the potential to break caching proxies, and is not a recommended practice (https://stackoverflow.com/questions/23603023/file-caching-query-string-vs-last-modified#23604412).

On the other side, it is hard to find sufficiently updated information. Most articles point to articles dating back to a decade ago.

Once an asset pipeline will be in place, it will be important to move to using hashed filenames instead of query strings.

seballot:
Hi @muxator ! I think we missed versioning the plugins assets. But I'm not sure we can do something right?

muxator:
Where are they located?

Anyway, we'll have a look in **1.8.5**.

I really hope to get rid of query strings ASAP. We only need to fix a 12 year frontend codebase, after all :smile:

ber8749:
Hello @seballot and @muxator,

First, thank you for all your hard work on the etherpad, it's a really fantastic app.

Second, I was wondering if plugin asset versioning is still on the docket for **1.8.5**? We're using an older version of etherpad (**1.6.1**) and would like to update to the latest, however, doing so breaks two of our plugins (`ep_align` and `ep_font_family`). It appears this issue is caused by references to outdated plugin files. Hard reloading the page fixes the issue, but this is a less than ideal solution for our end users.

Issue #3964, opened 2020-05-02
Setting to disable the new import security checks


JohnMcLear:
If you have pre-authed your user IE put them through an auth module you probably don't want the additional import security checks so they should be disablable by a plugin.

Plugin sohuld do       ``clientVars.thisUserHasEditedThisPad = true;``

But server side ``hooks/express/importexport.js`` needs some logic.

JohnMcLear:
Waiting on review/merge of https://github.com/ether/etherpad-lite/pull/3965

PR #3965, opened 2020-05-03
Setting for allowing import - Ready for Review / Merge

JohnMcLear:
https://github.com/ether/etherpad-lite/issues/3964

Reasoning:  If you have already authed a user you can assume they are "trusted".  So a setting to override this is fine for instances which use an authentication plugin.


muxator:
Moving to **1.8.5**. Not planned for **1.8.4**.

PR #3969, opened 2020-05-05
timeslider: important fix pad content was cropped 

seballot:
If the pad is bigger than height of screen, is was cropped (impossible to access the bottom). Fixed now


muxator:
Metged, thanks

Issue #3956, opened 2020-04-28
Toolbar Icons not showing after update (1.6.2 -> 1.8.3) when exposing in a subdir via a reverse proxy and minify = true


skramsa:
![image](https://user-images.githubusercontent.com/16946810/80459501-be29c500-8932-11ea-92eb-8788ceb3c96c.png)

The toolbar Icons are not showing and I get missing font family errors, but they're all in their folders, nothing was changed. Can you help me fix this issue? 
 
GET <server>/static/font/Roboto-Regular.ttf net::ERR_ABORTED 404 (Not Found)

I get this for all font families.

JohnMcLear:
Please provide details.

Are you running behind reverse proxy?
Are you running in development or production mode.
Minified?
Secured copy or your settings.json
Etc.

https://mclear.co.uk/2015/01/15/how-to-write-a-bug-report/

skramsa:
It's running in development mode, although the systemd service has "Environment=NODE_ENV=production" in it, no reverse proxy, the css is minified, settings.json is barely touched (only mysql and admin pw is changed).

It looks like etherpad can't load all fonts because it looks at <server>/static/font/*.ttf instead of <server>/etherpad/src/static/font/

JohnMcLear:
Try turn minify to false and maxage to 0 in settings.json

skramsa:
I did, nothing changed

JohnMcLear:
At my laptop now, might be able to help more.  Most useful thing is to get your settings.json (but remove any passwords).  Can you provide that please?

FWIW I tested 1.6.* > 1.8.3 and didn't experience this but it could be related to your settings.json having some legacy cruft in :!

JohnMcLear:
If you don't want to provide settings one thing to do is:

``cp settings.json settings.backup.json``
``cp settings.json.template settings.json``
Then copy your database host/user/pass into ``settings.json``

Restart and see if it's fixed.

skramsa:
/*
  This file must be valid JSON. But comments are allowed

  Please edit settings.json, not settings.json.template

  To still commit settings without credentials you can
  store any credential settings in credentials.json
*/
{
  // Name your instance!
  "title": "HelpMe-Etherpad",

  // favicon default name
  // alternatively, set up a fully specified Url to your own favicon
  "favicon": "favicon.ico",

  //IP and port which etherpad should bind at
  "ip": "0.0.0.0",
  "port" : 9001,

[...]

PR #3984, opened 2020-05-08
migrateDirtyDBtoRealDB: fix the script, which did not work

JohnMcLear:
Fixes https://github.com/ether/etherpad-lite/issues/2214

node dirty returns wrong length on ``.load``

I fix by doing forEach twice, once to get the length then again to do the process.  Then I terminate on end of process..


muxator:
This was not scheduled for **1.8.4**. However, it is out of core, and if it is useful we can include it anyway.

@JohnMcLear, could you briefly describe what is the problem you are fixing so I can annotate the change and we can merge it right away?

JohnMcLear:
The syntax used for dirty didn't work.

1. Fixed issue with .length being wrong on .load
2. used working syntax for dirty
3. potentially introduced a bug where I wait 5 seconds before killing

made the script work.

muxator:
Do we have a issue on ueberDB for tracking the defect on dirtyDB?

Seems serious enough. How old is the dependency on `dirty` library in ueberDB? Maybe it is time to upgrade?

JohnMcLear:
@muxator  no RE issue on ueber, I figured it was a quirk and easy enough to work around.

Afaik dirty is up to date, I did it recently.

I didn't wanna go too far into fixing or working around deps, I figured for a quick one time use script it doesn't matter if the op is a little slower if it means it's reliable.

muxator:
> it doesn't matter if the op is a little slower

I did not object on the double iteration, indeed. I am just saying that the info about the `length` parameter in ueberDB being broken when using dirtyDB should be written somewhere in ueberDB's code (a comment may be ok, a `warn()` on invocation may be even more evident).

It's a bug and took a lot of time to be discovered, let's not lose it again.

JohnMcLear:
Sorry I was not clear...

The bug appears to be in dirty, or at least the understanding by the initial author of how dirty provides the length value?

muxator:
That's clear, thanks.

Let's add a comment somewhere in https://github.com/ether/ueberDB/blob/master/dirty_db.js to explain that `length` is broken upstream. You spent a lot of time to figure this out, it has to be written somewhere better than in github comments.

I am working on #3956 now.

JohnMcLear:
I put it on the new database reference docs.   dirtyref is branch, I'm stopping my commits to master as dirtyref is ready for review.

Issue #3250, opened 2017-08-20
Reliable backups for high-activity pad


njoyard:
Hi,

We have a production instance of etherpad-lite 1.6.1 for a nonprofit, which apart from being used normally with many pads, has a specific pad that has a lot of activity and history, as it is used kind of as a board of things to do/in progress/recently done and is updated many times every day by several members of the org, and it is hardly ever deleted and recreated.

This makes it a pad that comes to have tens of thousands of revisions after a few months. I'm quite sure this is not really what etherpad-lite was designed for, but "unfortunately" the org members like this way of working very much and are very used to it and we've still not found a better tool.

We've already had several catastrophes with this specific pad due to kernel panics, unclean shutdowns, mysql restarts (mostly for sec upgrades) without stopping etherpad first, corrupt changesets that lead to the high-activity pad being unreadable (fails with `Failed assertion: Invalid changeset (checkRep failed)` client-side). Other lower activity pads on the instance seem to cope with those events quite nicely though. Additionally, some members sometimes have a faulty connection that causes their browser to reconnect very often and I wonder if that doesn't generate even more revisions to fade their author color each time. That's a secondary problem, but if it's the case then it also makes the pad history grow even faster and increases the chances of failures as I perceive it.

Restoring attempts for this pad usually includes:
- calling for a member to not close his pad browser tab and copy-paste the html somewhere
- running checkPad.js / repairPad.js which takes ages due to the huge history and does not change the situation
- trying to get a proper backup using getText/getHTML API methods (calls always fail)
- trying to restore a working version using restoreRevision/copyPad API methods (both calls succeed but copyPad takes ages and just creates a 2nd nonworking pad)
- trying to call deletePad via the API (takes ages *and fails*)
- remove pad rows from database manually and recreate the pad from a copy-paste of the HTML, losing both history and authorship

That situation led me to try and setup frequent backups of the whole instance. I'm actually doing hourly mysql dumps right now (at least for the last 24h). Unfortunately I discovered that restoring those backups also lead to a nonworking `checkRep failed` pad. Which led me to believe that doing mysql dumps actually produces a faulty database image unless etherpad is stopped.

I would have used the API to make backups but after a few weeks/months of activity the API calls just take longer than the backup interval. And stopping the instance every hour to run mysqldump would be quite disruptive.

So here are my questions:
- is there a way to tell a running etherpad to somehow flush everything to database so that a mysqldump has a higher chance of being usable ?
[...]

PR #3991, opened 2020-05-11
Include pool every 100 revs - Ready for Testing / Merge

JohnMcLear:
This is part of https://github.com/ether/etherpad-lite/issues/3250

I realized when doing pad rebuilds that the Attribute pool has to be rebuilt from Revision 0, and if you have broken changesets(due to database curruption) then your attribute pool is unable to be built.  

Including the pool (``attribToNum`` and ``numToAttrib``) every 100 revisions increases database storage (by ~0.1%) but it gives us the ability to restore a pad from any 100th revision.   Currently we store the attribute values IE line1 has attribute 0 but we don't store that attribute 0 is a list item(variables previously mentioned)..

Disclaimer:  I need this validated, I bumped @pita because I want to be sanity checked.


JohnMcLear:
We(Me and @pita) discussed and we think I'm correct and agree it's worth storing the data for integrity even if I'm wrong.

We can't see a downside.

PR #3995, opened 2020-05-12
Various fix for 1.8.4

seballot:
One of them, re wrapping, is critical because some functionalities have been broken because of this

### DIFF
--- a/src/static/css/iframe_editor.css
+++ b/src/static/css/iframe_editor.css
@@ -28,6 +28,14 @@ html.inner-editor {
   padding: 10px;
   overflow: hidden;
   background-color: white;
+
+  /* Be careful editing following rules. Longs words should not overflow, ep_align justify should work,
+     Test on chrome, firefox and safari... Copy / Paste a word inside a sentence should not add line-breaks
+     and preserve the style */
+  display: block; /* for safari and firefox, otherwise the break-word does not work */
+  white-space: normal;
+  word-wrap: break-word;
+  overflow-wrap: break-word;
 }
 #innerdocbody, #sidediv {
   padding-top: 10px; /* Both must have same top padding, so line number are aligned with the rows */
@@ -66,38 +74,6 @@ body.mozilla, body.safari {
 }
 .safari div {
   padding-right: 1px; /* prevents the caret from disappearing on the longest line of the doc */
-}
-
-
-/* -------------- */
-/* -- WRAPPING -- */
-/* -------------- */
-
-body {
-  white-space: nowrap;
-  word-wrap: normal;
-}
-body.doesWrap {
-  display: block !important;
-  /* white-space: pre-wrap; */
-  /*
-    Must be pre-wrap to keep trailing spaces. Otherwise you get a zombie caret,
-    walking around your screen (see #1766).
-    WARNING: Enabling this causes Paste as plain text in Chrome to remove line breaks
-    this is probably undesirable
-    WARNING: This causes copy & paste events to lose bold etc. attributes
-    NOTE: The walking-zombie caret issue seems to have been fixed in FF upstream
-    so let's try diabling pre-wrap and see how we get on now.
-    For more details see: https://github.com/ether/etherpad-lite/issues/2574
-  */
-  word-wrap: break-word; /* fix for issue #1648 - firefox not wrapping long lines (without spaces) correctly */
-}
-.noprewrap{
-  white-space: normal;
-}
-body.doesWrap:not(.noprewrap) > div {
-  /* Related to #1766 */
-  white-space: pre-wrap;
 }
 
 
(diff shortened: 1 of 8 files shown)


src/static/css/iframe_editor.css
```diff
@@ -28,6 +28,13 @@ html.inner-editor {
   padding: 10px;
   overflow: hidden;
   background-color: white;
+
+  /* Be careful editing following rules. Longs words should not overflow, ep_align justify should work,
+     Test on chrome, firefox and safari... Copy / Paste a word inside a sentence should not add line-breaks
+     and preserve the style */
+  white-space: normal;
```
JohnMcLear, line 35: I think this will break caret positioning.  Check the comments in the iframe_editor.css around these properties.

seballot: Yes I read the comments about zombie caret, but actually the fix for zombie caret (white-space: pre-wrap) was already commented since a while. And I cannot reproduce the zombie caret. Or are you talking about another caret issue?

muxator:
@seballot, please check this branch before continuing to work. I force pushed here by mistake. I **think** I reverted the original situation, but compare your local version with this one if possible. Sorry.

seballot:
Hi @muxator, I cannot see any difference with my local seballot:ui, I think it's fine !

And yes, same as #3957, would need to deploy this in 1.8.4

seballot:
but would be good before deploying to dedicate some effort on testing, especeially on safari which introduced a lot of bugs, so please tell me when you think 1.8.4 is ready, so I can test the final version

muxator:
The only remaining PRs for **1.8.4** is this one and #4002.

#4002 involves the impact of minification on css fonts paths. If this PR does not modify anything in `font-face` or does not change location to css stylesheets, you are green light on testing it on safari, for sure. :smile: 

Thanks!

seballot:
I deployed the changes here if you want to do some testing : https://pad.colibris-outilslibres.org/p/test

For now it seems to work pretty well

JohnMcLear:
Last minute suggestion, not related to this:

image

Perhaps if text is RTL we should shift Chat to left side of screen? :D

JohnMcLear:
I noticed changing from open sans to lato doesn't change the font?

JohnMcLear:
image

Weird topbar in timeslider bug :)

seballot:
We do not provide Lato font, I think it for mac users, I kept some legacy font in this dropdown, but I'm okay to remove them and only propose the one we really handle, do you agree?

What do you mean by "weird toolbar" ? I can see

- [ ] authors should not be in white, but in black or white depending on the color
- [ ] return to pad button should not shrink so we can always read it

- [ ] with rtl content put chat on left

JohnMcLear:
RE fonts, I'm easy :)  You decide.

Yea sorry I was vague RE timeslider, I just assume your eyes get drawn to things that aren't perfect because everything else is so great!

muxator:
My 2 cents about the legacy fonts in the dropdown: reading _Helvetica_ or _Times New Roman_ and remembering when they were a thing on Windows 3.1 makes me feel ancient :fearful:

Back to the point: this PR seems to be mature, pulling in.

Thanks.

src/static/js/pad_utils.js
```diff
@@ -526,12 +526,27 @@ function setupGlobalExceptionHandler() {
...
-      if ($("#editorloadingbox").attr("display") != "none"){
-        //show javascript errors to the user
-        $("#editorloadingbox").css("padding", "10px");
-        $("#editorloadingbox").css("padding-top", "45px");
-        $("#editorloadingbox").html("<div style='text-align:left;color:red;font-size:16px;'><b>An error occurred</b><br>The error was reported with the following id: '" + errorId + "'<br><br><span style='color:black;font-weight:bold;font-size:16px'>Please press and hold Ctrl and press F5 to reload this page, if the problem persists please send this error message to your webmaster: </span><div style='color:black;font-size:14px'>'"
-          + "ErrorId: " + errorId + "<br>URL: " + padutils.escapeHtml(window.location.href) + "<br>UserAgent: " + userAgent + "<br>" + msg + " in " + url + " at line " + linenumber + "'</div></div>");
+
+      var msgAlreadyVisible = false;
+      $('.gritter-item .error-msg').each(function() {
+        if ($(this).text() == msg) msgAlreadyVisible = true;
```
muxator, line 532: will replace "==" with "==="

muxator: In these cases, you could also do an early exit once you found a condition that makes the variable true.

In pure JS it would be a `break`. In jQuery there was a way to exit early a `.each()`. I _think_ it was returning `false` from the lambda, but I'd have to look ad the documentation again.

seballot: Yes I have been lazy :) thanks !

seballot:
oh thanks, but I wanted to fix the small things listed previous comment before release !
That's fine, will be for next one !

Issue #3978, opened 2020-05-07
Facilitate skin customization/extending 


brunob:
Hi, at https://pad.infini.fr/ we use latest 1.8.3 with the new default skin customized with a plugin available here https://framagit.org/infini/ep_infini

First, thx for the work on this new fronted and for the new skin system :)

As you can see in our plugin, we are trying to achieve some small tweaks on "default" skin : 

- add a script in order to get our topemnu on all page of etherpad
- add a small css to customize the index page

FTR, before upgrading to 1.8.3 we were using a customized theme bases on `custom` dir cf https://framagit.org/infini/themes-outils-libres/-/commit/6c71dc24b7619a07d6d27cb37b1c5da232148bd1 but i'd like to stick to the new skin as much as possible in order to facilitate our upgrades in the future :)

I'm not really satisfied of what i've done to achieve this, since i have to use `eejsBlock_indexWrapper` hook in order to insert the js and the css on home page, maybe the index page lack a new hook in order to add content in head of the page.

We also have to customize the index page in order to change `go2Name()` & `go2Random()` functions, but for now i haven't find a way to provide a template from a plugin.

I've read @muxator comment https://github.com/ether/etherpad-lite/issues/3441#issuecomment-422155860 saying that for now it's not possible to add another layer of customization on top of skins, and from what i've read here https://github.com/ether/etherpad-lite/issues/3471 "in case of errors, fall back to no-skin" it doesn't seem possible to "extend" colibris skin, since any file not provided by a personal skin will fall back to no-skin. Am i right ?

As i was discussing about this with @mrflos here https://framapiaf.org/@infini/104120325892102445 i think i'll have to end up wiht a plugin in order to insert our topmenu script on pads page + a skin for the index customization. 

Some questions : 

- it is possible to provide a skin as a plugin ?
- it is possible to provide a peronal index template from a plugin ?
- if i want to make a personal skin based on colibris, i think i have to copy entirely the original, am i right ?

Anyway, i know that you are in a transitional state, since v2 will change things around this subject.

JohnMcLear:
cc @seballot

seballot:
Hi @brunob !

I guess what you can do is create a new skin folder, and import colibris css with 

`@import url("../colibris/pad.css");`

?

For js... Cannot see an easy solution, but the js code in colibris skin is quite small so I guess it should be ok to copy it...

[...]

PR #3996, opened 2020-05-12
add hooks for index.html template

brunob:
ref #3978

Tested with this branch of our infini plugin and it works :)

https://framagit.org/infini/ep_infini/-/tree/new_hooks

Still have to test the `indexCustomInlineScripts` but it should work also from what i see.

Open to discussion on naming of these new hooks.


JohnMcLear:
Looks good to me, putting into 1.8.5 as

1) Low hanging
2) Very unlikely to cause problems
3) Not breaking

brunob:
thx @JohnMcLear do you want a separate PR in order to add these new hooks to documentation ?

JohnMcLear:
There is docs for EJS hooks?  I didn't think there was...  That's a lot of docs if there is!

brunob:
Yes, it seem https://etherpad.org/doc/v1.8.4/#index_eejsblock_name (see my precedent PR about it).

JohnMcLear:
Yeah docs pr would be required then

PR #4000, opened 2020-05-13
feat(i18n) Custom i18n strings

orblivion:
Allow locale string overrides in `settings.json`, an described in https://github.com/ether/ep_webrtc/issues/31. See documentation change in this PR for example.

### DIFF
--- a/doc/localization.md
+++ b/doc/localization.md
@@ -92,3 +92,23 @@ For example, if you want to replace `Chat` with `Notes`, simply add...
 , "pad.chat": "Notes"
 }
 ```
+
+## Customization for Administrators
+
+As an Etherpad administrator, it is possible to overwrite core mesages as well as messages in plugins. These include error messages, labels, and user instructions. Whereas the localization in the source code is in separate files separated by locale, an administrator's custom localizations are in `settings.json` under the `customLocaleStrigns` key, with each locale separated by a sub-key underneath.
+
+For example, let's say you want to change the text on the "New Pad" button on Etherpad's home page. If you look in `locales/en.json` (or `locales/en-gb.json`) you'll see the key for this text is `"index.newPad"`. You could add the following to `settings.json`:
+
+```
+  "customLocaleStrings": {
+    "fr": {
+      "index.newPad": "Créer un document"
+    },
+    "en-gb": {
+      "index.newPad": "Create a document"
+    },
+    "en": {
+      "index.newPad": "Create a document"
+    }
+  }
+```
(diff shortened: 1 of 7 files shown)


src/node/hooks/express/adminsettings.js
```diff
@@ -29,9 +29,9 @@ exports.socketio = function (hook_name, args, cb) {
           return console.log(err);
         }
 
-        //if showSettingsInAdminPage is set to false, then return NOT_ALLOWED in the result
+        // if showSettingsInAdminPage is set to false, then return NOT_ALLOWED in the result
         if(settings.showSettingsInAdminPage === false) {
-          socket.emit("settings", {results:'NOT_ALLOWED'});
+          socket.emit("settings", {results: 'NOT_ALLOWED'});
```
orblivion, line 34: Just a couple spacing issues I found randomly before.

orblivion:
Actually I'm not certain what the best place would be to put the documentation for this thing. `README.md` is way too small. The Wiki has links to Docs, but those docs are somewhat redundant with the rest of the Wiki. Etc.

orblivion:
Also for consideration: I'm putting this in `settings.json` because it started as a suggestion to customize a few error messages. However perhaps the expectation is that people will use this extensively, in which case we could just have a separate "user locales" directory that is formatted like the other ones.

I'm actually unfamiliar with the need for this feature altogether, so I can't say which way to go.

doc/localization.md
```diff
@@ -92,3 +92,23 @@ For example, if you want to replace `Chat` with `Notes`, simply add...
...
+## Customization for Administrators
+
+As an Etherpad administrator, it is possible to overwrite core mesages as well as messages in plugins. These include error messages, labels, and user instructions. Whereas the localization in the source code is in separate files separated by locale, an administrator's custom localizations are in `settings.json` under the `customLocaleStrigns` key, with each locale separated by a sub-key underneath.
+
+For example, let's say you want to change the text on the "New Pad" button on Etherpad's home page. If you look in `locales/en.json` (or `locales/en-gb.json`) you'll see the key for this text is `"index.newPad"`. You could add the following to `settings.json`:
+
+```
+  "customLocaleStrings": {
+    "fr": {
+      "index.newPad": "Créer un document"
```
orblivion, line 105: Chances are Google Translate gave me the infinitive instead of the imperative, so I probably am embarrassing myself here. I welcome any francophones to correct me.

orblivion:
Oh was this done with a squash-merge, github-side?

JohnMcLear:
Yep

PR #4005, opened 2020-05-14
like #2998 : fix ueberdb module path

brunob:
In order to repair bin/rebuildPad.js for example.


muxator:
Self-contained change that fixes long standing bugs.
Accepting right away, even if not strictly part of what was planned for **1.8.4**.

Thanks.

PR #4012, opened 2020-05-15
Script to create session and store token <> author more throughly (dont create ghosts)

### LINKED ISSUE #4006: Import always fails: Author X exists but he never contributed to this pad
Using 4816785aef457f0 I receive:

```
[2020-05-14 21:37:24.764] [WARN] console - Unable to import file into "g.YYYY$TTTTT". Author "a.XXXXXX" exists but he never contributed to this pad
```

```
$ node -v
v10.20.1
$ npm -v
6.14.4
$ uname -a
Linux 6f0add8352b7 4.15.0-99-generic #100-Ubuntu SMP Wed Apr 22 20:32:56 UTC 2020 x86_64 GNU/Linux
```

- Build based on the official Docker Image.
- Client: Firefox

Expected behavior:
- Import works when I contributed to a pad.

Actual behavior:
- Import does not work
- Request is terminated with `HTTP Status Code 405 Method not allowed ` Body: `405: Only the HEAD or GET methods are allowed.` - however the above message (`Author X exists but he never contributed to this pad`) is logged.

JohnMcLear:
Fixes #4006

This PR is:
1. Is a script for debug
1. Is a fix for #4006 
1. Is possibly a fix for countless other group session related problems. 

My findings:
* Token wasn't set for an author on first visit of a group pad from a new session, this meant a new author was made (even though the user had an author / session assigned!)

My solution:
Now when a user comes in and it has an author the users ``token`` value is ``.set`` in the database. 

My concerns:
1. This could mean a user could send a package saying "Hey I'm author a.xxxx (public) and this is my token (whatevz)" then pretend to be that user?  I'm not sure if this is how it works.  This is what I hope @muxator can figure out.

1. I feel like I'm missing something, surely this worked at some point?  Surely this hasn't been broken for all this time without anyone noticing that they were getting the wrong authorID for a group pad?!  That would be insane.

Possible ways to mitigate my concerns:
If author already exists, don't create the ``token2author`` assignment?  I'm not sure if this will have negative impact and we don't have thorough test coverage to make me confident.

Things this has brought up:
We should have better test coverage for session creation/timings/multiple editors/authors etc.  It's all good having unit and integration testing but if you need to replicate time/edits/browser life cycles etc it's hard.  On top of that we have countless settings and at the moment we only test the default settings.  I'm not sure what the solution is, just pointing out we lack test coverage.

Also I hate our sessions / token / authorId model.  It works but it's confusing, leaving 4006 as a record for understanding it might be useful.  I also hate our use of language/variables as ``token`` is ambiguous within Etherpad, as is ``session``.


JohnMcLear:
@muxator this one is one for your attention, it seems like such an obvious fix that I want a second opinion..

JohnMcLear:
I'm going to merge this but without any warranty and mindful we might need to revert it.  My time is running out on waiting on review so it will need to be reviewed by people using develop.  Sorry!

rhansen:
I believe this PR is a no-op because the call to `securityManager.checkAccess` already saves the author to the database:
  * `securityManager.checkAccess` calls `authorManager.getAuthor4Token`
  * `authorManager.getAuthor4Token` calls `mapAuthorWithDBKey('token2author', token)`
  * If there is not already an entry in the DB, `mapAuthorWithDBKey` calls `authorManager.createAuthor` to create an author object, then saves the author ID in `token2author`

Issue #4016, opened 2020-05-15
Security regression on password  (PASSWORD_HIDDEN) from commit 5c5b99fc9ad33054fb0291b92084e00ae1e634ef


alasserr:
On Etherpad 1.8.4 with Docker + postgresql, admin password can be used only once and is then replaced by "PASSWORD_HIDDEN" (on login form, password that is indeed checked, not only in cache).
This seems to be a regression from 5c5b99fc9ad33054fb0291b92084e00ae1e634ef => https://github.com/anttiviljami/etherpad-lite/commit/5c5b99fc9ad33054fb0291b92084e00ae1e634ef

See : #3421

JohnMcLear:
Even reading the code it is only setting the password in the database, how that string lands at the client baffles me...

muxator:
**Confirmed**.
Many thanks @alasserr for reporting this.

This bug impacts the users defined in `settings.json`. At least for the users that have access to the `/admin` area, **after a single login their password is implicitly changed to `PASSWORD_HIDDEN`**.

The bug may impact non-administrative users, too, but I did not test them.

# Minimal test case
0. supposing `settings.json` contains the following snippet:

   ```json
     "users": {
       "admin": {
         "password": "pwd",
         "is_admin": true
       },
     }
   ```

1. successful login to the admin console with the correct password:
   ```
   $ curl -u "admin:pwd" http://localhost:9001/admin/
   <!doctype html>
   <html>
     <head>
       <title>Admin Dashboard - Etherpad</title>
       [...]
   </html>
   ```

2. subsequent logins with the correct password fail (to do this in a browser, reset everything via `CTRL + SHIFT + DEL`):
   ```
   $ curl -u "admin:pwd" http://localhost:9001/admin/
   Authentication required
   ```

3. the password of `admin` is now `PASSWORD_HIDDEN`:
   ```
   $ curl -u "admin:PASSWORD_HIDDEN" http://localhost:9001/admin/
   <!doctype html>
   <html>
     <head>
       <title>Admin Dashboard - Etherpad</title>
       [...]
   </html>
   ```

4. the good news is that **the breakage is not stored in the database**: **after a server restart** users are able to authenticate again (only once):
   ```
   $ curl -u "admin:pwd" http://localhost:9001/admin/
   <!doctype html>
   <html>
     <head>
       <title>Admin Dashboard - Etherpad</title>
       [...]
   </html>
   ```

JohnMcLear:
[...]

PR #4019, opened 2020-05-16
Dont store authd users in database

JohnMcLear:
To try to resolve: https://github.com/ether/etherpad-lite/issues/4016

I can't see why you would want to have auth'd sessions persist in the database.  They are in the settings file anyway..  Thoughts?  Will this break something?  Tests pass...  

Opinions welcome.

PR #4023, opened 2020-05-16
Revert "SessionStore: replace password with PASSWORD_HIDDEN when storing in db"

rhansen:
This reverts commit 53f126082a8b3d094e48b159f0f0bc8a5db4b2f4, which
broke user authentication.

Fixes issue #4016.
Reopens issue #3421.

### DIFF
--- a/src/node/db/SessionStore.js
+++ b/src/node/db/SessionStore.js
@@ -38,11 +38,6 @@ SessionStore.prototype.get = function(sid, fn) {
 SessionStore.prototype.set = function(sid, sess, fn) {
   messageLogger.debug('SET ' + sid);
 
-  // don't store passwords in DB
-  if (sess.user && sess.user.password) {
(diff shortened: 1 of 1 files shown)


rhansen:
@JohnMcLear

JohnMcLear:
Hopefully https://github.com/ether/etherpad-lite/pull/4019 will require this PR defunct, but let's see! :D

rhansen:
My preference would be to merge this then merge PR #4019. Rationale: PR #4019 has a non-zero chance of breaking something. If this PR is not merged and PR #4019 needs to be reverted, then we're back to having broken authentication.

muxator:
> My preference would be to merge this then...

Definitely, thanks.
Let's revert the problematic commit and start over.
I will have a look at the proposed remediation tomorrow.

This time we'll have to be stricter.

Issue #4031, opened 2020-05-18
iframe embedding a pad


joshr:
Hi all, new here and a huge fan of etherpad for a very long time.

i'm building a site that is made up of etherpad pages in iframes but i'm having problems. i'm essentially using the embed code from the embed code button in the edit bar, but i'm making the iframe full page sized.

i've been chatting to John, who's gone above and beyond helping me get this far (i'm not a coder really!)

in safari and brave browsers users aren't able to load the pads in iframes at all due to (i think) cross domain cookies being blocked by the browser, etherpad is blocking being embedded in iframes altogether, or combo of those two things. some reports of this happening in incognito mode in other browsers too.

here's an example page https://reorganise.org/etherpad-page/ to try in safari (this is a wordpress generated page)
my etherpad install is at https://etherpad.xyz (digital ocean, ubuntu, nginx)

i'm on the latest etherpad-lite, and i'm running the latest versions of: comments_page, embedded_hyperlinks2, headings2, set_title_on_pad, spellcheck, table_of_contents

john and i have been thinking to solve this with helmet, but it occurred to me this is an out-of-the-box feature that doesn't/shouldn't rely on a plugin, since iframe embedding is in the vanilla install.

many thanks in advance for advice, fixes, and diagnosis!

aspiers:
Don't suppose https://github.com/vitr/safari-cookie-in-iframe is any use here?

JohnMcLear:
I wonder if cors might do the trick.  https://www.npmjs.com/package/cors & https://github.com/helmetjs/helmet/issues/79

JohnMcLear:
I created a test scenario for this just to show it's not Etherpad that's broken here: 

Test @ https://admissivefalcon.htmlpasta.com/

All I'm doing is putting cookie set and get from mozilla MDN docs into the iframe HTML.

```
<!DOCTYPE html>
<html>
<body>
<button onclick="alertCookie()">Show cookies</button>
<script type="text/javascript">
document.cookie = "name=oeschger";
document.cookie = "favorite_food=tripe";
function alertCookie() {
  alert(document.cookie);
}
</script>
</body>
</html>
```

And this fails in Brave/Safari

JohnMcLear:
https://gist.github.com/iansltx/18caf551baaa60b79206  I think this would need a hack for Safari, which frankly I'm not willing to do.  

[...]

PR #4034, opened 2020-05-19
Better ux for no cookies

JohnMcLear:
Here is a PR with a much better UX.  This is a proposed fix.

Safari / Brave users CAN use a pad in an iframe BUT if they refresh/rejoin Etherpad is not aware of who they are.

For #4031

- [ ] Needs testing against different security setups / sessions etc. @muxator 
- [ ] ?i18n the warning? and make the gritter a warning and persist? @seballot 
- [ ] Style the Cookie error @ seballot

image

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -24,7 +24,7 @@
   "pad.colorpicker.cancel": "Cancel",
 
   "pad.loading": "Loading...",
-  "pad.noCookie": "Cookie could not be found. Please allow cookies in your browser!",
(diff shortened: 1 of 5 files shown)


seballot:
Hi, there is already a way to display error gritter

```javascript
$.gritter.add({
      title: "Error",
      text: "Ep_author_neat: Please upgrade to etherpad 1.8.4 for this plugin to work correctly",
      sticky: true,
      class_name: "error"
    });
```

you can reuse this code, and remove the message in the background ("cookie could not be load, please allow cookies in your browser"), the gritter is enough I guess !

JohnMcLear:
@seballot I pushed a change, it now looks like this. 

image

 That okay?

JohnMcLear:
Okay I slightly modified the language too, over to @muxator for review.

PR #4036, opened 2020-05-19
Rate limit Socket IO communication - WIP

JohnMcLear:
1. Includes settings
1. Includes i18n
1. Includes a nice notification
1. Disconnects on rate limit
1. Includes feeding into metrics/stats
1. Include console warn to server console.

- [x] Needs testing.
- [ ] Is applied to all messages from user, should it only be on a specific type of message?  IE user change.
- [x] Basic value (of 10 per second) needs reviewing.
- [x] Discuss if Rate limiting based on IP is a good idea (considering Edu use etc.)

> It may be issue if there are many users behind one IP address. If there is some login procedure or uniqueUserId, use it to limit on per user basis. Otherwise, you may try to limit by socket.id and limit number of allowed re-connections from the same IP.  <-- we don't have this luxury unless ``requireSession`` is ``true``

To test set rateLimiting value per second to 1 and write in a pad.  You should get a popup modal and disconnect.


JohnMcLear:
I did a bunch of thinking/research about this and I **also** think a global setting of ``maxPendingEdits`` is a good idea..  https://github.com/JohnMcLear/etherpad-load-test/issues/1

``maxPendingEdits`` would stop new users connecting to a pad once the load is at ``50`` (setting.maxPendingEdits).  Until this value is below 50 no NEW users can join.  Existing users can connect to a pad.

This wouldn't stop a client generating 1 new lurker every second until it reached several thousand videos then sending a bunch of edits.  I would need logic for this too :\  I think the best thing here is to restrict number of sessions that one IP address can have with ``maxSessionsPerIP``.  I'd suggest a default value of ``100`` as that would mean a single IP address would hardly put a dent on Etherpad usage (remembering that commits are rateLimited by the PR).

Remembering.  Authors are "expensive", Lurkers are "cheap", the biggest gain here is in this PR.

Also remembering we will need a white list system, are it's possible that business' running Etherpad internally might want to whitelist an entire subnet IE ``192.168.1.*``.

Finally also remembering that authenticated sessions might also want to be exempt from limiting / max sessions.

JohnMcLear:
See: https://gypseouspondskater.htmlpasta.com/ and https://github.com/ether/etherpad-load-test/issues/1 for more thoughts and todos

JohnMcLear:
Also I just realized timeslider may be a good attack vector due to being able to request multiple changesets

PR #4039, opened 2020-05-21
Remove top console logs

JohnMcLear:
If Etherpad is embedded is an iframe it will throw an error and block the error if a pad tries to throw an error.

A cleaner way might be to check to see if console is available first and use if it is..


src/static/js/ace2_inner.js
```diff
@@ -518,7 +518,7 @@ function Ace2Inner(){
         }
       }
       currentCallStack = null;
-      if (profiling) top.console.profileEnd();
+      if (profiling) // top.console.profileEnd();
```
squiddle, line 521: commenting out the only expression of the if statement creates a Syntax Error. 

```suggestion
      if (profiling) { /* top.console.profileEnd(); */ }
```

JohnMcLear: Good spot tnx :D <3

rhansen:
Please delete the code rather than comment it out. Commented code makes the code harder to maintain: Do future changes have to worry about the commented code coming back? Was the code commented because there's a separate problem that remains unresolved? Also, commented code rots over time (at some point the commented code will stop working when uncommented).

PR #4041, opened 2020-05-22
Update mocha version on front-end tests

### LINKED ISSUE #3434: frontend tests: stale mocha.js version
File `tests/frontend/lib/mocha.js` is quite old: first committed in 2012 with ba4ebbba3ba6, slightly modified the day after in 7aee98bce82e ("made test helpers more cross browser compatible").

```diff
-    var query = Mocha.utils.parseQuery(window.location.search || '');
-    if (query.grep) mocha.grep(query.grep);
+    //var query = Mocha.utils.parseQuery(window.location.search || '');
+    //if (query.grep) mocha.grep(query.grep);
 ```

It would be probably useful to update the frontend library like we did for the backend one.

### PR DESCRIPTION
**Pull request by alexanmtz**
A new version of Mocha 7.1.2 was added, downloaded from Mocha official source, and integrated into the front-end tests infrastructure and running on the Browser, and now with this Pull request I will figure out how this will behave on Travis since I did changes on the remote runner.

The current version of Mocha was used with a workaround to extend the HTML reporter to add a custom reporter as a hidden div when runs on the browser, so the remote runner was using this element to extract the result to connect on Sauce Labs and provide the test results.

To update this version I had to remove this custom runner `WebdriverAndHtmlReporter`, used to generate a text output in a hidden div to be retrieved by the remote runner with the test results.

So I had to add a jQuery logic on the remote runner to obtain test results from HTML. The code is not completed yet, as we would need with this current approach some changes on the CSS to fit with the current version. But I would like to know if we can proceed using this approach, otherwise, we would need to explore some other approaches. I tried several ways to avoid bigger changes in the way the tests run today, but I faced several issues when I updated the Mocha version into the current code structure. 

If this is not enough, I would suggest and introduce alternative approaches:

1. Use Webpack in order to have module imports and then we could extend Mocha to create custom reports like the one we have today with the most recent features from ES6, but we would need some way to build from the source before running the build, and we would need further investigation to figure out the best way to do it for this project.

2. Use a headless browser, like Phantom JS or other frameworks, or a browser runner from some of these frameworks, that works well with SauceLabs, but I would need to change completely from what we have today, and I'm not sure if we would have the best results since I could see that we test in different browsers.

I'm creating this Pull request in an earlier stage cause I'm not able to test on Sauce Labs locally, or please let me know if there's a way to do it.

Closes #3434

> Screenshots/GIFs

pageshot of 'Frontend tests' @ 2020-05-22-2314'02

### CODE REVIEW

File: tests/frontend/travis/remote_runner.js [hunk 1]
```diff
@@ -52,7 +52,8 @@ var sauceTestWorker = async.queue(function (testSettings, callback) {
 
     var knownConsoleText = "";
     var getStatusInterval = setInterval(function(){
-      browser.eval("$('#console').text()", function(err, consoleText){
+      var testResultFormatedScript = "function customReportFromHTML(){var globalResult='';var report=$('#report').clone();report.find('pre').remove();report.find('> li').each(function(index,element){var titleText=$(element).find('> h1').text();var testResultTitle=$(element).find('> ul > li > h1').text();var result='';$(element).find('> ul > li > ul > li').each(function(index,element){if($(element).hasClass('pass'))result+='[green]PASSED[clear] : ';if($(element).hasClass('fail'))result+='[red]FAILED[clear] : ';if($(element).hasClass('pending'))result+='[yellow]PENDING[clear] : ';result+=$(element).text()});globalResult+=titleText+''+testResultTitle+'  -> '+result+'  '});return globalResult;};customReportFromHTML();";
```
JohnMcLear, line 55: 🐙  ``\`` is a thing for line breaks to make this readable, worth using?

tests/frontend/runner.js
```diff
@@ -1,166 +1,4 @@
 $(function(){
```
JohnMcLear, line 1: Less code!  Awesome :)

JohnMcLear:
Testing

- [x] npm test still works
- [x] Frontend test runner works
- [x] Individual tests are still greppable ``tests/frontend/index.html?grep=helper.js``
- [ ] Travis-CI is being properly reported to <-- still needs testing.  I'm not sure the best way to test tihs.  Usually me or @muxator need to submit the PR then it has the keys required.

tests/frontend/travis/remote_runner.js
alexanmtz: @JohnMcLear, I have tried to use the line breaks as the original implementation, cause the runner handle this input to split in line breaks, but somehow using inside `eval` is throwing an error, I will try to figure out a way to add the line breaks properly in this case.

tests/frontend/runner.js
alexanmtz: Yes, the runner now is running Mocha pretty straightforward, but the responsibility is now with Selenium on `remote_runner` to parse the results :p

tests/frontend/travis/remote_runner.js
JohnMcLear: Don't worry too much about it :)  Just a readability thing and this is not the worst none-line-broken piece of code in Etherpad ;)

alexanmtz: @JohnMcLear but the remote runner split the tests by these line breaks:
https://github.com/ether/etherpad-lite/blob/develop/tests/frontend/travis/remote_runner.js#L38

alexanmtz: If I can't find a solution I will figure out other separators...

JohnMcLear: Ahhhhhh...  Okay that makes sense..  Did you try to throw the lines at remote runne rto see what happens yet?   I assume you implemented the sauce keys I sent you?

JohnMcLear: If you are working on this now I will pause the debug I'm doing as they will interfere..

alexanmtz: Hey @JohnMcLear, yeah, I'm working on this and I got the sauce API keys and I'm able to run the tests on CI 🙏

JohnMcLear:
In case this catches you out...

https://docs.travis-ci.com/user/caching/#npm-cache

To disable npm caching, use:
```
cache:
  npm: false
```

alexanmtz:
great @JohnMcLear, I was having cache issues and this will help 🙏

JohnMcLear:
Make sure to rebase, I just pushed some fixes for testing.

JohnMcLear:
Tests should at least be more stable for you man..  That was a lot of pain to get here... ;\

alexanmtz:
@JohnMcLear , after a lot of tries by extracting the report from the HTML using a parser, I figure out there's no way to do this on the remote runner side, cause even extracting the content of the test report on HTML source there's no way to know that the test ended to run, and the way is implemented now, with a time interval it's not gonna work. I will send a new pull request right now from Mocha side using events to keep the raw report hidden to be used on the runner and I hope it works now.

alexanmtz:
So now we should proceed with this PR: https://github.com/ether/etherpad-lite/pull/4074

PR #4042, opened 2020-05-25
Testing: Add and test padPrefs for helper.newPad

orblivion:
This allows for setting "prefs" in tests, which is going to be very useful if not vital for testing ep_webrtc. (I use it here: https://github.com/ether/ep_webrtc/pull/56).

Also, I fixed a couple unrelated front end tests that were failing locally. All pass now on Firefox and Chromium!

### DIFF
--- a/tests/frontend/helper.js
+++ b/tests/frontend/helper.js
@@ -52,11 +52,26 @@ var helper = {};
     return win.$;
   }
 
-  helper.clearCookies = function(){
+  helper.clearSessionCookies = function(){
     // Expire cookies, so author and language are changed after reloading the pad.
     // See https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#Example_4_Reset_the_previous_cookie
     window.document.cookie = 'token=;expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/';
     window.document.cookie = 'language=;expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/';
+  }
+
+  // Can only happen when the iframe exists, so we're doing it separately from other cookies
+  helper.clearPadPrefCookie = function(){
+    helper.padChrome$.document.cookie = 'prefsHttp=;expires=Thu, 01 Jan 1970 00:00:00 GMT';
+  }
+
+  // Overwrite all prefs in pad cookie. Assumes http, not https.
+  //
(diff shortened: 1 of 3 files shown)


tests/frontend/helper.js
```diff
@@ -100,10 +115,16 @@ var helper = {};
...
+        if (opts.clearCookies) {
+          helper.clearPadPrefCookie();
+        }
+        if (opts.padPrefs) {
+          helper.setPadPrefCookie(opts.padPrefs);
+        }
         helper.waitFor(function(){
           return !$iframe.contents().find("#editorloadingbox").is(":visible");
         }, 50000).done(function(){
-          helper.padChrome$ = getFrameJQuery(                $('#iframe-container iframe'));
```
orblivion, line 106: Whereas the other two seem to require the page to have loaded, the iframe itself doesn't. And that's great, because now we can grab it ahead of time and set the prefs cookie way sooner. Soon enough for my ep_webrtc test anyway.

tests/frontend/specs/embed_value.js
```diff
@@ -19,8 +19,8 @@ describe("embed links", function(){
     var width = $embediFrame.attr("width");
     var height = $embediFrame.attr("height");
     var name = $embediFrame.attr("name");
-    expect(width).to.be('600');
-    expect(height).to.be('400');
+    expect(width).to.be('100%');
+    expect(height).to.be('600');
```
orblivion, line 23: 23307d14d57c462b66a4bb15d3acde9e9dcf4c90

tests/frontend/specs/helper.js
```diff
@@ -82,21 +92,39 @@ describe("the test helper", function(){
...
+            expect(window.document.cookie).to.not.be(firstCookie);
 
             // click on the settings button to make settings visible
             var $userButton = chrome$(".buttonicon-showusers");
             $userButton.click();
 
             // confirm that the session was actually cleared
             var $usernameInput = chrome$("#myusernameedit");
-            expect($usernameInput.val()).to.be('')
+            expect($usernameInput.val()).to.be('Enter your name');
```
orblivion, line 108: Not sure how this was passing before. This text looks like it was introduced long ago. Maybe we recently fixed a bug that was keeping it from appearing?

JohnMcLear: Probably!

Issue #3883, opened 2020-04-18
Wrong url address when I use read only mode


alekso87:
Hi everyone! I use this config https://github.com/ether/etherpad-lite/wiki/How-to-put-Etherpad-Lite-behind-a-reverse-Proxy#hosted-at--with-friendly-urls-for-pads, a little modified for myself.
![33](https://user-images.githubusercontent.com/25965602/79641711-e3376e80-81a1-11ea-9172-62076eed092a.gif)

JohnMcLear:
Works fine here: https://video.etherpad.com/p/testreadonlyURI

Possibly a setting?

Video.etherpad.com is running nginx reverse proxy too..

alekso87:
@JohnMcLear I see, but i use settings without /p/

JohnMcLear:
Afaik the read only URI is computed based on the location of ``/p/``, I think you would need to look at the logic around that and compute it from a setting or so?  Are you hoping to hack on some of this stuff or just making issues for now?

I just don't want them to go into a black hole because I wouldn't consider this a bug but a quirk of your specific deployment so for core devs it's gonna be low down on the priority list -_-  We don't like having open bugs though!

alekso87:
> Are you hoping to hack on some of this stuff or just making issues for now?
I followed the instructions from your wiki. I like this setting, where there are no extra parameters in the url.

I showed that with this configuration and the inclusion of "read-only", the URL breaks and there is no mention of a domain.

> We don't like having open bugs though!
And who loves bugs?

muxator:
@alekso87, I am sure there is some quirk somewhere in the code.

If possible, try to find it, and send a PR.

If it takes too much time, or happens to be too difficult, please [update the wiki page](https://github.com/ether/etherpad-lite/wiki/How-to-put-Etherpad-Lite-behind-a-reverse-Proxy#hosted-at--with-friendly-urls-for-pads) mentioning the problem you just found. At least this would not be surprising anymore.

That would be helpful and really appreciated. Thank you.

alekso87:
@muxator what PR can be used for this fix?

JohnMcLear:
I think what @muxator id saying is that you will need to create a pr to address this bug.

JohnMcLear:
Nice job on getting a PR in!  We will review ASAP.  We're currently in release mode so give us a few days to get this release out please <3

alekso87:
[...]

PR #4070, opened 2020-06-01
setEmbedLinks: fix the wrong URL when we use etherpad without /p/ in the URL

alekso87:
This solution fixes the problem #3883


alekso87:
@JohnMcLear done

PR #4072, opened 2020-06-02
Fix bad merge conflict resolution

rhansen:
The series of commits I made for PR #4008 were squashed into a single commit and rebased. Somewhere along the way a mistake was made in a merge conflict resolution, resulting in some bad code in `bin/buildForWindows.sh`. This PR reverts the squashed commit and re-does the original series of commits.

**PLEASE DO NOT SQUASH AND MERGE. CHOOSE "Create a merge commit" INSTEAD.**

### DIFF
--- a/bin/buildForWindows.sh
+++ b/bin/buildForWindows.sh
@@ -1,26 +1,26 @@
 #!/bin/sh
 
-#Move to the folder where ep-lite is installed
+# Move to the folder where ep-lite is installed
 cd $(dirname $0)
 
-#Was this script started in the bin folder? if yes move out
+# Was this script started in the bin folder? if yes move out
 if [ -d "../bin" ]; then
   cd "../"
 fi
 
-#Is wget installed?
+# Is wget installed?
 hash wget > /dev/null 2>&1 || {
   echo "Please install wget" >&2
   exit 1
 }
 
-#Is zip installed?
+# Is zip installed?
 hash zip > /dev/null 2>&1 || {
   echo "Please install zip" >&2
   exit 1
 }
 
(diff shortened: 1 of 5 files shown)


JohnMcLear:
This repo doesn't have create a merge commit enabled.  Can you ELI5 why that approach is better?

rhansen:
> This repo doesn't have create a merge commit enabled.

Hmm, that's unfortunate. While that "Create a merge commit" option can result in a tangled history that is hard to review, squashing is often worse. The third option ("Rebase and merge") is not great as it doesn't create a merge commit (it fast-forwards). Ideally GitHub would add a fourth choice to rebase and create a merge commit (see isaacs/github#1143 and isaacs/github#1017).

Even with that option disabled you should be able to manually create a merge commit:

```shell
# Clone the etherpad repo from scratch so you don't have to worry
# about local divergence, the current state of your upstream tracking
# branches, etc.
git clone git@github.com:ether/etherpad-lite.git
cd etherpad-lite
# Add a remote for my fork
git remote add rhansen https://github.com/rhansen/etherpad-lite.git
# Fetch the commits from my fork
git remote update -p
# Merge no-ff my PR branch into your local develop branch.
# THE --no-ff ARGUMENT IS CRITICAL TO PREVENT A FAST-FORWARD MERGE.
git merge --no-ff rhansen/fix-bad-conflict-resolution
# Push the result
git push
```

> Can you ELI5 why that approach is better?

See https://stackoverflow.com/a/15721436.

In this specific case, squashing the commits will squash the revert with the re-do, which is generally bad practice.

PR #4073, opened 2020-06-02
Timeslider export links numeric pads

JohnMcLear:
Uses basic splits instead of complex regular expression.

All we're doing is rewriting a url not landing a rocket ship.


JohnMcLear:
Ehh? ?  We have loads of ``const`` and ``let``'s in the code base...  Lemme try that out..

JohnMcLear:
Okay looks like most JS isn't minified that uses const/let..  will fix on this branch and send another PR to use https://www.npmjs.com/package/terser

JohnMcLear:
pushed, give that a tr y:)

rittmeier:
>  ./src/node_modules/.bin/uglifyjs src/static/js/timeslider.js
Parse error at src/static/js/timeslider.js:157,12
        let type = this.href.split('export/')[1];
            ^
ERROR: Unexpected token: name «type», expected: punc «;»

It does not like the let in line 157 and 158....

JohnMcLear:
Sorry was focusing on getting the uglifyJS stuff migrated to terser, that's done now and this should now work :)

JohnMcLear:
Okay will do some test coverage shortly if I can, either way this is good for merge.  Will wait for @muxator to merge though as it's my PR.

PR #4098, opened 2020-06-08
pluginfw: Dont cache if max age is 0

JohnMcLear:
Fixes a bug where the cache would cache plugin templates even if maxAge is 0.  This sucks if you are actively developing something..


Chocobozzz:
Oup's thanks!

PR #4116, opened 2020-06-14
Timesliding slidebar visual representation of pad

JohnMcLear:
image

This is my PR working on improving the UX of the timeslider.

- [ ] Script to generate a graph of activity over time.  Mostly done, see  ``bin/createPadTimesliderVizData.js``
- [x] Script to generate report of pad activity over time, line by line.
- [x] Script to generate report of pad activity day by day.
- [ ] Documentation
- [x] Setting to disable some Log4js noise when running script

I'm putting the logic etc. here but it's likely this will need to go into a plugin.


JohnMcLear:
Fwiw I published

ep_what_have_i_missed
and
ep_who_did_what

plugins which begin to pluginify this code.  Admin scripts may be useful but not necessary as the user is empowered to handle themselves.

PR #4133, opened 2020-06-26
Feature: Timeslider follow

JohnMcLear:
Isolated PR with the timeslider follow which I think is worthwhile functionality to merge in.

- [x] Not dependent on jQ
- [x] Test coverage
- [x] Setting to turn it on/off

Test coverage and settings should be easy to do, so I would say this can be handled as an easy to solve :)


tests/frontend/specs/timeslider_follow.js
```diff
@@ -0,0 +1,52 @@
...
+
+        setTimeout(function() {
+          //make sure the text has changed
+          var newTop = timeslider$('#innerdocbody').offset();
+          expect( originalTop ).not.to.eql( newTop );
+          done();
+        }, 1000);
+
+      }, 2000);
+    }, revs*timePerRev);
```
webzwo0i, line 48: Maybe I misunderstand, but does this work and test what you wanted to test? I think this waits until all edits have been made before entering timeslider, so it won't really follow new edits?

We probably need special test equipment for this to work, eg two browsers (one in timeslider, one doing the edits) like in https://webdriver.io/docs/multiremote.html

I think we can hack this together with wd, too: we need access to known pad ids, but we can set them via existing code (like we do with specs_list) during test setup. Then, at the beginning or end of the tests we wait in those "special" pads and run our tests as soon as other browser join. Of course this needs more thinking - should the browsers roll the dice to find out who edits the pad and who watches the edits in the timeslider? Should they switch roles so that we can test if the code works across browsers? We probably also want to handle local testing with multiple browsers..

JohnMcLear:
Afaik it plays from rev0

PR #4134, opened 2020-06-26
Fix backendRunner's exitcode

webzwo0i:
Changes the behaviour of backendRunner.sh:
It will not terminate immediately on error, but instead run all the tests. I think it's useful to see the results of all test suites even when a single one fails (although there are only two atm - but this might change in the future). There was also a "kill and sleep" at the end, which had no effect. (Maybe it was meant to kill the node-process?)
Also removes the minification step for backend tests, because they don't need them.

### DIFF
--- a/tests/frontend/travis/runnerBackend.sh
+++ b/tests/frontend/travis/runnerBackend.sh
@@ -31,12 +31,6 @@ echo "Successfully connected to Etherpad on http://localhost:9001"
 # a copy of settings.json is necessary for the backend tests to work
 cp settings.json.template settings.json
 
-# Build the minified files?
-curl http://localhost:9001/p/minifyme -f -s > /dev/null
-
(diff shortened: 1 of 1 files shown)


tests/frontend/travis/runnerBackend.sh
```diff
@@ -31,20 +31,12 @@ echo "Successfully connected to Etherpad on http://localhost:9001"
 # a copy of settings.json is necessary for the backend tests to work
 cp settings.json.template settings.json
 
-# Build the minified files?
-curl http://localhost:9001/p/minifyme -f -s > /dev/null
```
JohnMcLear, line 35: why remove this?

webzwo0i: shouldn't be necessary because api- and contentcollector-tests both include files directly. minification is only relevant for frontend tests, no?

PR #4161, opened 2020-07-10
frontend test improvements

webzwo0i:
1. prints duration of every test case to travis console
2. saves browser's console output in saucelabs (better way would be to print it too, but I don't know how to do that right now) - this will help in the future to detect warnings/errors printed to browser console during test runs - maybe we should also include ep server logs in travis?
3. sends sauce connect proxy tunnelIdentifier during test runs
4. sets parallel runs to ~~4~~ 6 again - this seems to work now (because of tunnelIdentifier?)
5. removes hard coded 10 seconds wait time for minification
6. update sauce connect proxy to 4.6.2
7. fixes the behaviour how we check, if a test run has succeeded. When the [killTimeout in runner.js](https://github.com/ether/etherpad-lite/blob/develop/tests/frontend/runner.js#L72) stops our tests like in https://travis-ci.org/github/ether/etherpad-lite/jobs/706938845#L715 our tests should fail. In the past this was not the case
8. includes MacOS 10.14-safari and Win7-firefox
9. pinned browser versions

### DIFF
--- a/tests/frontend/runner.js
+++ b/tests/frontend/runner.js
@@ -60,7 +60,7 @@ $(function(){
       stats.tests = stats.tests || 0;
       stats.tests++;
       if ('passed' == test.state) {
-        append("->","[green]PASSED[clear] :", test.title);
+        append("->","[green]PASSED[clear] :", test.title," ",test.speed,"ms");
       } else if (test.pending) {
         append("->","[yellow]PENDING[clear]:", test.title);
       } else {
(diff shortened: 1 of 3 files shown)


webzwo0i:
With parallelism we are at ~8 minutes for the frontend tests now. 2 min preparation and safari with 6 minutes.

JohnMcLear:
Awesome progress.  This good to merge?

webzwo0i:
Yes it's safe. I updated parallel sessions to 6 and added OSX 10.14/safari 12.0 and Win7/Firefox 78.0, too and pinned the other versions.
We are at 8.5 minutes for those 6, safari being the slowest.

RE frontend tests speed I will come up with something soon.

PR #4173, opened 2020-07-17
Async bump 2

JohnMcLear:
I don't know where these commits went..  They weren't reverted..  Ref: https://github.com/ether/etherpad-lite/pull/3994/files


JohnMcLear:
``async`` and ``await`` are not supported in IE or Chrome 48 whihc is why the tests are failing.

PR #4175, opened 2020-07-17
Large test file timeslider

JohnMcLear:
Just a large .etherpad file that can be used for testing when you need 10k revisions in a single pad.  I am not merging because it's too large but leaving here in case it's useful for others.


JohnMcLear:
cc @webzwo0i will be useful for your tests.

PR #4177, opened 2020-07-17
Bugfix / editor: Attempt to fix #3037

### LINKED ISSUE #3037: Text input not allowing non-breaking space
Hello,

It seems that etherpad lite's content edit area does not allow for non-breaking spaces. 

Instead, any non-breaking space inserted seems to automatically be converted into a normal (breaking) space.

This behaviour breaks typographic rules in French and probably other languages too. Also, I don't see any convicing reason for automatically converting non-breaking spaces into breaking spaces.

(Litewrite had the same problem, so this might help: https://github.com/litewrite/litewrite/issues/199 )

Thanks,

JohnMcLear:
WIP.

### DIFF
--- a/src/static/js/domline.js
+++ b/src/static/js/domline.js
@@ -317,7 +317,7 @@ domline.processSpaces = function(s, doesWrap)
     for (var i = 0; i < parts.length; i++)
     {
       var p = parts[i];
-      if (p == " ")
+      if (p == "&nbsp;")
       {
         parts[i] = '&nbsp;';
       }


JohnMcLear:
Revisiting this PR while investigating #3037. The one-line change here (`domline.processSpaces`) sits in the *display* path — it runs after the editor's text model is already built, and transforms ordinary spaces into `&nbsp;` for rendering. By the time this code runs, any U+00A0 the user typed has already been replaced with U+0020 by the ingestion pipeline, so there is nothing for this patch to preserve.

The actual stripping happens in four ingestion points that all normalize `\xa0` → ` `:

- `src/static/js/contentcollector.ts:82` — `textify()` replaces `\xa0` in text read from the DOM
- `src/static/js/ace2_inner.ts:2045` — a second `textify()` does the same
- `src/static/js/ace2_inner.ts:479` — `importText(..., dontProcess=true)` throws if raw text contains nbsp
- `src/node/db/Pad.ts:55` — server-side `cleanText()` used by `spliceText`, default pad content, and `setHTML`

To fix #3037 the nbsp needs to survive into the changeset text bank (which it can — the changeset wire format only regex-parses the op list before `$`; the text bank is read by numeric offset/length via `StringIterator`, and `' '.length === 1`). Then `processSpaces` in `domline.ts` and `ExportHtml.ts` need to treat U+00A0 as its own class — emit `&nbsp;` verbatim and exclude it from the "last space in a run stays plain" bookkeeping so that adjacent regular spaces don't get miscounted. Plain-text export should preserve ` ` too.

In short: this diff can be discarded — the fix has to land on the input side first, and the display tweak is then a different (and larger) change than the one here. Happy to open a follow-up PR against `develop` with the full pipeline change if that's useful.

PR #4189, opened 2020-07-22
Fix(locales): fix wiki links in several locales

nodeg:
This fixes the AbiWord wiki link in several locales files.


JohnMcLear:
Please use translate wiki to submit language patches

Issue #4191, opened 2020-07-23
Connection diagnostics don't work


webzwo0i:
Since https://github.com/ether/etherpad-lite/commit/d26df864902510ca7f325b786183dc9031e04e42 diagnostics from disconnects are POSTed to the wrong url. We listen to /ep/pad/connection-diagnostic-info in https://github.com/ether/etherpad-lite/blob/develop/src/node/hooks/express/apicalls.js#L8 while that change made requests go to ep/pad/connection-diagnostic-info relative to path, so when you are in a pad at /p/abc diagnostics are sent to /p/ep/pad/connection-diagnostic-info 

https://github.com/ether/etherpad-lite/pull/2912 was the original pull request that introduced this behaviour, so I asked there if they can post their configuration.

We must decide, if we leave it up to ops to write configuration in their reverse proxy to handle *all* etherpad-specific paths (maybe include a section in the wiki for subdir-installations, ie https://yourhost.org/specific/path/for/etherpad) or if we include a setting in settings.json, eg "subdirectory", that makes same of the reverse proxy-path configs obsolete. socket.io and all javascripts/css files are relative already, but /jserror, /locales.json etc are served at a static path and we'd need to serve them at subdirectory+/jserror, subdirectory+/locales.json etc.

webzwo0i:
Note I was wrong about locales.json, it's is relative already.

webzwo0i:
And jserror seems to be relative too. Need to investigate if it is really just diagnostic-info that makes problems

AbhishekG-Codes:
PR opened: #7051 —> updates the diagnostics POST URL to work correctly in subdirectory setups (e.g., /p/ep/pad/...). This should resolve the path mismatch issue. Let me know if any adjustments are needed.

PR #4192, opened 2020-07-23
WIP: Revert "made url relative"

webzwo0i:
This reverts commit d26df864902510ca7f325b786183dc9031e04e42.
See https://github.com/ether/etherpad-lite/issues/4191 for the problem. In case you think it's better to include logic for subdirectory in settings.json and make express endpoints listen to a subdirectory-prefix (e.g. https://yourhost.org/specific/path/for/etherpad/jserror, https://yourhost.org/specific/path/for/etherpad/locales.json when subdirectory="/specific/path/for/etherpad") then just close this and I will probably issue a PR for that

### DIFF
--- a/src/static/js/pad.js
+++ b/src/static/js/pad.js
@@ -855,7 +855,7 @@ var pad = {
       $.ajax(
       {
         type: 'post',
-        url: 'ep/pad/connection-diagnostic-info',
+        url: '/ep/pad/connection-diagnostic-info',
         data: {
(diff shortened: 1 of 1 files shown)


webzwo0i:
Don't merge yet, needs more investigation. First I thought that more paths are prone to not work when etherpad is served behind a reverse proxy with subdirectory, but probably everything works except diagnostics.

PR #4194, opened 2020-07-23
fix: fix async code inside of loop

### LINKED ISSUE #4193: Error with clientVars hooks
Since we've updated to the 1.8.4 we've got a bug where the variables set on 
clientVars is not working. A deep look in to the code, I've realised there's one
part of code that's causing the error. That was added on https://github.com/ether/etherpad-lite/commit/d1330a1e1c0fc536244c77605f0612cf4412e9fe. 

The misbehaviour happens because it has async code inside a forEach loop
https://github.com/ether/etherpad-lite/commit/d1330a1e1c0fc536244c77605f0612cf4412e9fe#diff-e4533b536e57db4e3299a5214e5c2e24R88

**To Reproduce**
I've tried to reproduce on a controlled scenario (only one plugin) without success. Maybe, the error
is easily reproducible on instances with lots of plugins and hooks.  

**Expected behavior**
It should allow set variables on clientVars hook

**Screenshots**

 - OS: [windows, linux, mac]
 - Browser [chrome]

joassouza:
This PR fixes the error of not waiting the async code to finish.
As the forEach did not wait until the async code finish we may get a
hook set up incorrectly. To fix it, we use an "Array.map" to iterate and
wait the promises to be resolved and then return the values.
Fixes #4193
A better explanation why this should avoided can found here
https://codeburst.io/javascript-async-await-with-foreach-b6ba62bbf404

### DIFF
--- a/src/static/js/pluginfw/hooks.js
+++ b/src/static/js/pluginfw/hooks.js
@@ -82,17 +82,16 @@ async function aCallAll(hook_name, args, cb) {
   if (!cb) cb = function () {};
   if (exports.plugins.hooks[hook_name] === undefined) return cb(null, []);
 
-  var newArray = [];
-  // This should be a map.
-  await exports.plugins.hooks[hook_name].forEach(async function(hook, index){
-    let test = await hookCallWrapper(hook, hook_name, args, function (res) {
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Thanks :D <3

Issue #4188, opened 2020-07-21
Frontend tests problems with Saucelabs


webzwo0i:
There are still some problems with frontend tests on Saucelabs. I am working on this, tracking this here.

1. after the async update, queue.drain must be called with a callback (fix is https://github.com/ether/etherpad-lite/commit/ccfbe22c06eb735b6dd853fd9106a8e1ffb8e4c7)
2. travis cancels the SL-job when no output was received for 10 minutes, but in that case mocha results won't be logged (fix is https://github.com/ether/etherpad-lite/commit/a48079588d90174de01a0f3aa5023edcdf1091b2 setting a max limit for all SL-tests to 9.5 minutes)
~~3. for unknown reasons the frontend test output in travis is broken in some cases, might be a problem in https://github.com/ether/etherpad-lite/blob/develop/tests/frontend/runner.js#L107~~ travis output log is not 100% in time and sometimes changes for me on reload - the problems I've seen (in PASSED/FAILED only the first five letters are green/red or sometimes there was a line missing) are gone on reload of the page
4. failed hooks are not printed as FAILED, so one can get the wrong impression that all tests pass (fix is to move FAILED out of "test end"-handler, because "test end" is not called when a hook failed)
5. related to 4., browsers seem to not be able to POST to /jserror in some cases (this makes sense when in the middle of loading a pad some error occurs, but in other cases I have the feeling it should work but doesn't during frontend tests) - eval if it's possible to setup this error-logging earlier on loading

JohnMcLear:
Are things stable if you just target one browser?  I often found that the more devices you target the less reliable saucelabs was

webzwo0i:
Yes and no. The problems described above are bugs in runner.js/remote_runner.js, fixes with explanations will come today! Without those it will look like everything is fine, while there are actually failing tests due to newPad (or something else) in a before(Each)-hook throwing. (You can see failing before-hooks on the left side during tests, but they are not logged because mocha's "test end"-hook, where we log the pass/fail/pending, will not be called - fix is to log in the pass/fail/pending events)

[...]

PR #4200, opened 2020-07-28
Fix frontend tests (#4188), ugly work around for "Pad never loaded"

webzwo0i:
sorry for the delay, but I had to run the tests a lot in travis and try out different things.

This fixes problems reported in #4188:
- no output in Travis
Sometimes the output in Travis does not contain anything but that the job is canceled because there are no messages for 10 minutes (Travis cancels sessions, if it does not receive data for ten minutes). We can configure this timeout in .travis.yml but for the moment I anticipate this timeout and stop the tests after 9.5 minutes on Saucelabs. This ensures that mocha results, at least for the tests up to this point in time, are logged. When we have a lot more tests we can increase this value. That there are no messages for 10 minutes is just a symptom, see below.

- missing FAILED messages in Travis
Errors in `before`/`beforeEach`-hooks are not printed in Travis, so it looks like everything works, but in fact it doesn't. The errors are printed in the left side of the browser (this is done by mocha), but we have a special element (`#console`) where we log pass/fail/pending, using that log to check if the tests have finished and to print them in Travis. Mocha has the design decision to not call `test end` events on `before`/`beforeEach` failure. The solution is to move our `#console` logging out of `test end` into the respective `pass`/`fail`/`pending` events.

- The tests never finish
In `runner.js` all events from mocha are handled and for every test a counter `stats.tests` is increased. This is done on `test end`.  Mocha has its own stats on the number of tests in `runner.total` (accounting also for `grep`-filter). When the `end` event occurs (all suites have finished), there is a check, if `stats.tests >= runner.total`. When there was a failure in `before`/`beforeEach` those check will always be false (see above) and "FINISHED" is never printed to `#console`. However, in our webdriver-script, we check for the existence of "FINISHED" to decide if all tests were run. The result is, the tests will never end and no message is printed in Travis. The solution is to always finish the tests on `end` event, no matter if some of the tests were "lost" in between. Mocha skips the remaining suite when there was an error in `before`/`beforeEach`, so the number in `stats.tests` varies a lot depending where the error happens.

- "Pad never loaded"
There are a lot of this messages, in different browsers and different suites. I also see them occasionally on my local machine. I think they caused the impression that SL is unstable with multiple parallel sessions. I tried reverting the jQuery-update, but that does not seem to be the reason. Sometimes (when looking at the browser sessions on SL) there is an error screen right before it happens ("ReferenceError - padutils not defined", "SyntaxError" - they also differ!). I systematically parse Travis output and I see them happen for a very long time. Because they also happen on my machine - sometimes - it's not an SL problem. This needs investigation, but I put a retry-logic in `newPad`. Now it tries up to 4 times to create a pad. Now all the tests run again! Note that we cannot use mocha's retry logic, because it does not work for hooks. Also note that it really just retries the pad loading-stuff and nothing else, no test and no hook is run twice.

Additionally:
- async#drain semantics changed with the last update, the exit code is now working again
- includes the server log (WARN level) into the output of the frontend tests - this really helps with problems the tests unveil
- removes some unnecessary sleeps
- removes some cruft (unused variables, globals etc)
- changes the maximum time of the tests from 200 minutes to 9.5 minutes
- dropped jquery.iframe.js and delete the iframe in `newPad` directly (because <IE9 is not supported anymore)
- the Travis log is sent in the callback of `browser.quit`. This should guard against cases like https://app.saucelabs.com/tests/cb8225375d274cbcbb091309f5466cfd when the log in Travis appears and remote_runner.js exits but webdriver on SL does not receive the DELETE command to end the session.


JohnMcLear:
Amazing work and excellent job!!

JohnMcLear:
[firefox Windows 7 52.0] -> FAILED : "before each" hook for "does not show Cancel button nor timer anymore" Error

Failed on that.

PR #4205, opened 2020-07-29
WIP: Fix check pad deltas.js

JohnMcLear:

> Screenshots/GIFs


webzwo0i:
Oh I did not see this PR. I think checkPad and checkPadDeltas were largely the same in the past?!? I took the logic of applying the changesets starting with revision 0 up to the head revision and comparing it to the key revisions in https://gist.github.com/webzwo0i/c6d75c09c4db0599ded42e7988cb39aa#file-checkpaddeltas-js
I could not get `diff` to work so I left it out, but replaced async-npm with async/await and Promise.all.

webzwo0i:
Also in my version it's probably not good for large pads to cache all changesets

PR #4222, opened 2020-08-20
Modifying instruction of installation

hs29:
There was a confusion when I tried to install etherpad-lite on Windows. The second instruction of "With a "Run as administrator" command prompt execute bin\installOnWindows.bat" may make misunderstand to run their Command Prompt as 'Run as administrator'.

The instruction can be replaced with "Open the folder of etherpad-lite, select bin, then run installOnWindows.bat.


JohnMcLear:
What is going on here???  This doesn't look like a good pull request.

JohnMcLear:
I closed because this PR made no sense ;\  If you want to change the instructions please modify the instruction files

PR #4232, opened 2020-08-27
Document the authenticate hook

rhansen:


JohnMcLear:
Awesome thanks @rhansen

PR #4235, opened 2020-08-30
bugfix/import: doc import bugfix

JohnMcLear:
.doc import was broken.

also includes uncommenting of test coverage


src/node/utils/LibreOffice.js
```diff
@@ -43,7 +43,9 @@ exports.convertFile = function(srcFile, destFile, type, callback) {
 
   if (type === "html") {
     // "html:XHTML Writer File:UTF8" does a better job than normal html exports
-    type = "html:XHTML Writer File:UTF8";
+    if (path.extname(srcFile).toLowerCase() === ".doc") {
+      type = "html";
+    }
```
rhansen, line 48: `type` is already set to `'html'` so these three lines can be deleted.

Issue #1997, opened 2013-11-18
Import from MS Word into Etherpad is problematic with regards to formatting Edit


bertpareyn:
I have tried to import fairly basic MS Word documents into Etherpad using the import option but very little of the formatting is retained and the content become somewhat unusable without reformatting.  Below are screen shots of the word document and the result after importing.

I've tried to simply copy and past from Word into Etherpad and the formatting actually seems to be retained better then using the import.  It still need to be reformatted but with less effort.

The MS Word document is a MS Word 2010 file (.docx) and I was using Firefox.
![ms word version](https://f.cloud.github.com/assets/5169198/1423820/1827891e-4012-11e3-8c41-d98bc7715c01.JPG)
![etherpad version after import](https://f.cloud.github.com/assets/5169198/1423822/1aaa8006-4012-11e3-986a-695930d074f2.JPG)
![copyandpasteintoehterpad](https://f.cloud.github.com/assets/5169198/1423823/1c681d2c-4012-11e3-9977-5166e84aa80b.JPG)

Originally reported at https://github.com/oaeproject/3akai-ux/issues/3274

PR #4240, opened 2020-09-02
.doc with multiple line breaks test file -- need help with fix

JohnMcLear:
To see the problem run
``soffice --convert-to "html:XHTML Writer File:UTF8" tests/backend/specs/api/bio.doc``
Note the line break after ``Richard``

``soffice --convert-to "html:XHTML Writer File:UTF8" tests/backend/specs/api/bio.doc``
Has the correct formatting.  We should be using this filter on ``.doc`` files.. 

Note that ``.docx`` is fine.

Note that for some reason we pass .doc > .odt > .html - I don't know if this is the cause.

I'm 99% sure LibreOffice.js is to blame but I can't see where to add logic to ensure .doc gets the correct filter and everytime I try to make it the correct filter I get empty pad contents.

I also included a temporary test file that can be run with: ``npx mocha ../tests/backend/specs/api/docImport.js`` -- it might be wise to merge this into the import API test coverage file with a copy/paste.

@webzwo0i any insight on this?


webzwo0i:
Spaces are probably missing due to https://github.com/ether/etherpad-lite/blob/develop/src/node/utils/ImportHtml.js#L37 because the filter creates a document that has newlines instead of spaces.

I don't see .doc -> .odt -> .html but only .doc -> .html.

See my commit in this branch, but it is not good:
[rehype does not support XHTML/XML](https://github.com/rehypejs/rehype/issues/42) so when we switch filter, we get a problem. The test passes (because with XHTML-filter no unwanted newlines are in the body and so the text has spaces) but the HTML is garbage.

webzwo0i:
Also not sure if we should use rehype-sanitize or if we filter the html somewhere else, so that no bad data can be imported into the editor

webzwo0i:
the reason you saw a blank page: this title replacements did not work (not closing the comment -->)

webzwo0i:
another thing: soffice has spans and when closing&immediately open another span then also a space is missing

JohnMcLear:
@webzwo0i think you can get a fix together for it please?  I'm gonna sort the security notifications stuff.

JohnMcLear:
Breaks export: https://travis-ci.org/github/ether/etherpad-lite/jobs/723684566#L951

webzwo0i:
hey, just an update:
With xhtml writer lists/font-styles are more broken than with Starwriter (default).
(cf https://github.com/ether/etherpad-lite/issues/1997).

webzwo0i:
maybe not "more broken". starwriter import has problems:
- no strikethrough
- https://github.com/ether/etherpad-lite/issues/1997 first numbered list has only "1."s instead of "1.", "2." etc
- second list starts with "2." instead of "1."
- nested list with letters (1.a, 1.b, 1.c...) become 1.1, 1.2, 1.3...

And xhtml writer problems:
- no font style (bold/italic/underline/strikethrough)
- wrong numbering (no use of ol-type attribute, has an additional span-element that includes the list number, so in the end there are two list numbers for every list item)

However, a positive news is that it don't seem to make a difference if it's odt, docx or doc anymore. (I only tested libreoffice 6.1.5-3)

JohnMcLear:
hrm, but this is failing the basic tests we have for export ;\  I haven't tested it locally but Travis isn't happy about the output for doc, docx and odt... ??

webzwo0i:
Yeah, I think because of this parameter-stuff with quotation marks that you also stumbled upon some time ago. This will work soon. :-D

I noticed you put some effort into getting ordered lists to work recently. I will check if the space-thing is a regression after those changes. My first tests seem to indicate that ordered lists are still broken, but I see your contentcollector-tests work, so I need to get to the bottom of it. (I am pretty sure I tested this with xhtml and html converter from soffice, but maybe it's an edge case that is not covered in the contentcollector tests)

One question though: What was your intention when adding rehype? Did you feel the export from soffice/abiword is broken? I tested some more complicated files with w3m and they seem to render correctly.

In case I am not finished by tomorrow noon, I will just put a workaround for the space-thing in, ignoring bugs with lists completely.

JohnMcLear:
> One question though: What was your intention when adding rehype? Did you feel the export from soffice/abiword is broken? I tested some more complicated files with w3m and they seem to render correctly.

rehype took bad HTML input and fixed it so it was machine readable.  Basically it meant we could import bad HTML which is obviously useful.

JohnMcLear:
Is this ready to go or does it still have list caveats?? :)

webzwo0i:
not ready :-(

JohnMcLear:
Gentle bump :)  Please wrap this up for 1.8.7 :)

webzwo0i:
fixed in develop

PR #4245, opened 2020-09-04
PadMessageHandler: Pass socket.io Socket object to clientVars hook

rhansen:
Also revise the clientVars hook documentation.


JohnMcLear:
Have you tested this with other plugins that use clientvars?

rhansen:
This is adding a new item to the context so existing plugins should be unaffected.

rhansen:
I tested a plugin that uses clientVars and it still works OK.

PR #4249, opened 2020-09-05
webaccess: Always sleep for 1s before returning HTTP 401

rhansen:
Not all authentication plugins require the Authorization header, so it might not be present in subsequent attempts. (In particular, a reverse proxy might strip it.)


JohnMcLear:
Dudes on fire today :1st_place_medal:

PR #4254, opened 2020-09-05
webaccess: Return 401 for authn failure, 403 for authz failure

rhansen:
This makes it possible for reverse proxies to transform 403 errors into something like "upgrade to a premium account to access this pad".

### DIFF
--- a/doc/api/hooks_server-side.md
+++ b/doc/api/hooks_server-side.md
@@ -370,7 +370,8 @@ A plugin's authFailure function is only called if all of the following are true:
 Calling the provided callback with `[true]` tells Etherpad that the failure was
 handled and no further error handling is required. Calling the callback with
 `[]` or `undefined` defers error handling to the next authFailure plugin (if
-any, otherwise fall back to HTTP basic authentication).
+any, otherwise fall back to HTTP basic authentication for an authentication
+failure or a generic 403 page for an authorization failure).
 
 Example:
 
--- a/src/node/hooks/express/webaccess.js
+++ b/src/node/hooks/express/webaccess.js
@@ -58,19 +58,6 @@ exports.checkAccess = (req, res, next) => {
     hooks.aCallFirst('authorize', {req, res, next, resource: req.path}, hookResultMangle(grant));
   };
 
-  /* Authentication OR authorization failed. */
-  const failure = () => {
-    return hooks.aCallFirst('authFailure', {req, res, next}, hookResultMangle((ok) => {
-      if (ok) return;
-      // No plugin handled the authn/authz failure. Fall back to basic authentication.
-      res.header('WWW-Authenticate', 'Basic realm="Protected Area"');
-      // Delay the error response for 1s to slow down brute force attacks.
-      setTimeout(() => {
-        res.status(401).send('Authentication Required');
-      }, 1000);
-    }));
-  };
-
   // Access checking is done in three steps:
   //
   // 1) Try to just access the thing. If access fails (perhaps authentication has not yet completed,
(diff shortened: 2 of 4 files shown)


JohnMcLear:
Please include test coverage with this one!  I think it's an area we might be missing test coverage in.  

https://github.com/ether/etherpad-load-test/blob/master/.travis.yml has an example of modifying settings.json in test but you might also want to just use a plugin that's included to test auth?  Your call but I think we need to make sure we have test coverage prior to merge.  Sound fair?

rhansen:
> Please include test coverage with this one!

Any tips on how to get started? I don't see any existing plugin tests that I can emulate, nor do I see any backend tests that inject custom settings (they all seem to use `settings.json` as-is).

rhansen:
I think I figured it out; stay tuned.

rhansen:
I added tests, but I need PR #4259 merged before this can be merged. Once that PR is merged I'll mark this as ready for review.

JohnMcLear:
merged 4259

rhansen:
> merged 4259

Thanks! I rebased this onto current develop.

rhansen:
friendly ping @JohnMcLear

rhansen:
Now that the authn/authz bypass tests are in place, I can now see that mocha appears to load all of the `tests/backend/specs/**/*.js` files simultaneously in a single node process. I was expecting it to run the tests in each `*.js` independently (for each `*.js` file start a new node process).  Unexpectedly, the `before` and `after` functions in `socketio.js` and `webaccess.js` don't appear to conflict with each other.

If my understanding of how mocha works is correct, what is the best way to organize the tests? Should we have a single `*.js` file that is responsible for the overall test setup and teardown, and have it `require()` the other `*.js` files?

JohnMcLear:
I don't know enough about mocha to be able to answer this question.  @webzwo0i has been pretty active on it so might be able to share some insights.

webzwo0i:
Don't think in terms of files, but in "suites" (although the files define the order of suites). Whenever you write describe or context, it defines a new "suite". Suites contain tests (it). Tests have hooks, like after/before. There is a special suite, the root-suite, that contains all the suites that you define. That's why you can define after/before outside describe in webaccess/socketio, but those will actually belong to the root-suite, which might result in unexpected behaviour.

Mocha will build up the file list and go through every file. For every file, it will execute code that's outside a suite (it's actually in the root-suite), code that's inside describe/context but outside any tests and it will do the same for any nested suites.

All of this will happen *before* the first test is executed. In your case, you define before-hooks that go to the root-suite in webaccess.js and also in socketio.js. Both will be executed before the first test is executed! This is probably not what you want.

I think it's best to avoid code inside describe, unless it's inside it or any of the available hooks. Otherwise it might be difficult to know for sure, if the tests can be run independently from one another. I am not sure if I understand you correctly, but I see no need for global setup&teardown, "file-local" should be enough (start express, stop express etc.)

P.S. Your tests are very inspiring!

rhansen:
Thank you for the information, that was very helpful.

> I think it's best to avoid code inside describe, unless it's inside it or any of the available hooks. Otherwise it might be difficult to know for sure, if the tests can be run independently from one another.

I'm not quite sure what you mean. Take a look at the tests I'm adding in PR #4312; all but the last three tests (`it()` invocations) must run in their declared order because the next test depends on the one before it. The tests have shared state, which means there must be variables declared at `describe()` scope. Is that a non-idiomatic way to write mocha tests?

webzwo0i:
I think sharing state like this is okay (e.g. declaring variables in describe callback that you need to access from different tests or storing a function inside a variable in describe callback that you need to call from multiple tests). There is an alternative described at https://github.com/mochajs/mocha/wiki/Shared-Behaviours but I would not recommend using it, because there does not seem to be a defined interface, so in the future mocha could add methods that would conflict with test code. Also describe callbacks don't have access to the same `this` context, so better use lexical scope like you already do or closures.

Don't think to long about it, but tests that depend on each other can be problematic, so it's better to keep them independent as much as possible. They can share state, but if test 1 fails for some reason you probably don't want to fail all subsequent tests. Sometimes when bisecting, to speed things up, I use `mocha --grep test-description`, so only tests that match test-description are run. This probably isn't possible if tests need to be run in a specific order.

One side note: an async function as describe callback seems to work, but could be misleading. It implies that the function returns a Promise but afaik the return value in describe is ignored.
In your case it's okay, but just in case you stumble upon it: Use arrow functions as hooks/tests callback and you loose the test context (this.retries or this.timeout won't work any longer)

rhansen:
I tweaked the tests a bit:

  * Thanks to PR #4329 these tests now play nicely with the tests in socketio.js.
  * I pass non-`async` functions to `describe()`.
  * I switched from arrow functions to `function`.
  * All of the tests are independent.

PR #4255, opened 2020-09-05
hooks: New mechanism to deprecate hooks

rhansen:
I plan on splitting authFailure into authnFailure and authzFailure so that separate authentication and authentication plugins can coexist peacefully. This change will make it possible to mark the authFailure hook as deprecated (which simply logs a warning).

### DIFF
--- a/src/static/js/pluginfw/hooks.js
+++ b/src/static/js/pluginfw/hooks.js
@@ -1,9 +1,32 @@
 var _ = require("underscore");
 
+// Maps the name of a server-side hook to a string explaining the deprecation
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Needs lots of docs with it :)

rhansen:
> Needs lots of docs with it :)

By "it" do you mean the authnFailure and authzFailure hooks that I will be adding in a future PR? If so, yes, I've already written the docs for them.

If by "it" you mean this deprecation mechanism, I expanded the comments.

PR #4259, opened 2020-09-06
plugins: Move plugin definitions to avoid monkey patching

rhansen:
Also document the plugin data structures.


JohnMcLear (commented):
Should this tie in more to the web ui installation logic?  Ie using these states to update web ui

What if a plugin depends on another plugin and that's installed by npm install?

I think I might require plugins in some of my plugins but I'd need to get to a pc to confirm.

rhansen:
FYI: I renamed `plugin_state` to `plugin_defs` so that people don't get confused and think the module is for holding state that is generated/consumed by plugins.

> Should this tie in more to the web ui installation logic? Ie using these states to update web ui

All client-side plugins are loaded during editor initialization, so I'm not sure how they could be more tied in than they already are.

Here's my understanding of client-side plugin loading:
  1. The client calls `client_plugins.js:ensure` during editor initialization.
  2. `client_plugins.js:ensure` calls `client_plugins.js:update` which GETs `/pluginfw/plugin-definitions.json` from the server.
  3. The server serializes `parts` and `plugins` from `plugin_state.js` into JSON and replies to the client.
  4. The client parses the JSON and saves `parts` and `plugins` in its own `plugin_state`.
  5. The client iterates over `parts` and calls `require` for each `.js` file mentioned in the client-side hook definitions.
  6. Each `require` GETs the plugin's `.js` from the server and evaluates it.
  7. The hook functions are saved in `hooks`.

> 
> What if a plugin depends on another plugin and that's installed by npm install?

This PR doesn't change any behavior, it only moves things around a bit. So if things worked before, they should continue to work.

I don't have in-depth knowledge about the plugin load code, but if I'm reading [`plugins.js` line 113](https://github.com/ether/etherpad-lite/blob/efa4ba3cf15772aaeae47282a91822a92c13358b/src/static/js/pluginfw/plugins.js#L113) correctly, I think Etherpad only loads top-level `ep_*` plugins, not their `ep_*` dependencies.

PR #4264, opened 2020-09-08
import: Use the correct author ID when using sessions

rhansen:
There are two different ways an author ID becomes associated with a user: either bound to a token or bound to a session ID. (The token and session ID come from the `token` and `sessionID` cookies, or, in the case of socket.io messages, from the `token` and `sessionID` message properties.) When `settings.requireSession` is true or the user is accessing a group pad, the session ID should be used. Otherwise the token should be used.

Before this change, the `/p/:pad/import` handler was always using the token, even when `settings.requireSession` was true. This caused the following error because a different author ID was bound to the token versus the session ID:

> Unable to import file into ${pad}. Author ${authorID} exists but he never contributed to this pad

This bug was reported in issue #4006. PR #4012 worked around the problem by binding the same author ID to the token as well as the session ID.

This change does the following:
  * Modifies the import handler to use the session ID to obtain the author ID (when appropriate).
  * Expands the documentation for the SecurityManager checkAccess function.
  * Removes the workaround from PR #4012.
  * Cleans up the `bin/createUserSession.js` test script.


bin/createUserSession.js
```diff
@@ -2,77 +2,46 @@
  * A tool for generating a test user session which can be used for debugging configs
  * that require sessions.
  */
+const m = (f) => __dirname + '/../' + f;
```
JohnMcLear, line 5: wtf? :D

JohnMcLear:
Where the fsk is the travis test? :D

JohnMcLear:
tnx for fixing my sloppy code dude :P  Nice to see it done properly ;)

PR #4270, opened 2020-09-08
webaccess: Enforce creation of `req.session.user` by authn plugins

rhansen:
The authorization logic determines whether the user has already successfully authenticated by looking to see if `req.session.user` exists. If an authentication plugin says that it successfully authenticated the user but it did not create `req.session.user` then authentication will re-run for every access, and authorization plugins will be unable to determine whether the user has been authenticated. Return a 500 internal server error to prevent these problems.


rhansen:
I will add unit tests to this PR if you merge PR #4254 first (so that I can add to the unit tests that are added in PR #4254).

JohnMcLear:
Can this wait till post release?

rhansen:
> Can this wait till post release?

Yup, that's fine.

Issue #4217, opened 2020-08-12
"Restart Etherpad" in settings.json editor in admin does not remove deleted fields


orblivion:
**Describe the bug**
If I remove a field in the `settings.json` editing Admin panel and click "Save" followed by "Restart Etherpad", the removed field will not be reflected in the effective settings as seen in the site's behavior. New fields and changes to existing fields are reflected, however. Once the server is manually restarted (ctrl-c, re-run), the removed fields are reflected in the effective settings as expected.

**To Reproduce**

1. Open a pad and watch your console. You'll see log items, including INFO.
1. Go to /admin/settings
2. Change "loglevel" to "ERROR"
3. Click "Save" and "Restart Etherpad"
4. Reload pad and watch your console. You'll see no log items (assuming no errors).
5. Go back to /admin/settings
6. Comment out "loglevel"
7. Click "Save" and "Restart Etherpad"
8. Reload pad and watch your console. You'll *still* see no log items (assuming no errors).
9. Ctrl-C to quit Etherpad, and run it again.
10. Reload pad and watch your console. You'll again see log items, including INFO.

**Expected behavior**

When you uncomment "loginfo" and save and restart the server, it should behave according to its default value of "INFO".

**Desktop (please complete the following information):**
 - Debian on QubesOS
 - Firefox
 - 68.11.0esr (64-bit)

jinbullsushil:
Hi,
Try to open the setting.json into notepad and save the file. After that restart the etherpad.

orblivion:
To be clear, the core of the error is in what is meant by "restart etherpad". There are two options:

`#1` Kill the etherpad server, and start it again
`#2` Click the "Restart Etherpad" button in the admin

For `#1`, there is no error. The error comes when I do `#2`.

As you suggested, I tried editing `settings.json` using vim (I'm on Linux), but then clicked "restart etherpad" in the admin. (I didn't change anything in the admin or click "Save Settings". I found the same error. So, I'm pretty sure that the error is to do with the manner of "restarting" rather than the manner of editing or saving `settings.json`.

JohnMcLear:
@orblivion yeah this sounds about right, I haven't looked at it yet but I imagine logic is something like this

```
var foo = "bar";
```
then you remove ``foo`` from settings.json and when it reloads ``foo`` is still ``bar`` because we don't nullify values that aren't present.

The fix here would be to nuke ``settings`` with  ``settings = {}``; then repopulate.  I imagine we just populate.

[...]

PR #4273, opened 2020-09-08
WIP fix(settings): Reloading after removing settings keys

orblivion:
Addressing issue in #4217 

# The Problem

If you have a field in settings with a default value (hardcoded in `src/node/utils/Settings.js`), you can override it in a custom setting in `settings.json`. The settings admin (`/admin/settings`) lets you edit that custom setting, and click "save" and "reload" and it takes effect as expected. However if in the settings admin you _remove_ that setting, it doesn't go back to the default value; the value just stays the same. However if you do a hard restart of the server, the value does go back to the default.

# The Explanation

When you call `reloadSettings()` from the settings.json editor admin, it will take the your new settings and merge it, more or less, with the _currently loaded_ settings, not the _default_ settings. When you remove a key and reload within the settings admin, it overrides that now-missing key with the existing value, which is the custom value you had there before you removed it. If you do a hard reset, the previous custom value is flushed from memory, and it goes through the usual process of loading the default before loading any custom values.

This phenomenon happens at different levels of the `settings.json` object hierarchy, but I'm not sure if it happens at all of them, because of the merging method used. This may also depend on whether it's a plugin setting or not.

# The Proposed Fix

For my fix, I make a new function called `setDefaults()` which overwrites all of the settings with the defaults. I call it within `reloadSettings()` before `storeSettings(settings)` and `storeSettings(credentials)`. This also involves putting the defaults into its own easily identified object.

Assuming this is the desired path (per core developers' opinion(s)), I'll see if I can write some tests, and then clean up some confusing comments.

### DIFF
--- a/src/node/utils/Settings.js
+++ b/src/node/utils/Settings.js
@@ -39,22 +39,24 @@ const suppressDisableMsg = ' -- To suppress these warning messages change ' +
     'suppressErrorsInPadText to true in your settings.json\n';
 const _ = require('underscore');
 
+const defaultSettings = {}
+
 /* Root path of the installation */
-exports.root = absolutePaths.findEtherpadRoot();
+defaultSettings.root = absolutePaths.findEtherpadRoot();
 console.log('All relative paths will be interpreted relative to the identified ' +
-            `Etherpad base dir: ${exports.root}`);
+            `Etherpad base dir: ${defaultSettings.root}`);
 
 /**
  * The app title, visible e.g. in the browser window
  */
-exports.title = 'Etherpad';
+defaultSettings.title = 'Etherpad';
 
 /**
  * The app favicon fully specified url, visible e.g. in the browser window
  */
-exports.favicon = 'favicon.ico';
-exports.faviconPad = `../${exports.favicon}`;
-exports.faviconTimeslider = `../../${exports.favicon}`;
+defaultSettings.favicon = 'favicon.ico';
+defaultSettings.faviconPad = `../${defaultSettings.favicon}`;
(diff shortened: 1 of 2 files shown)


JohnMcLear:
@rhansen are you happy with this?

rhansen:
I think this is better than what we have now, so it's worth merging after tests are added.

Some easy (I think) ways to improve this PR, if you feel up to it:
* Do a deep copy of `defaultSettings` instead of a shallow copy.
* Re-run the `loadSettings` hook on reload.

If you want to go the extra mile, you could use [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to track dynamic modifications so that they can be reapplied after a reload. `Proxy` could also be used to provide an updatable view onto the current settings, so that a module that does `const foo = settings.foo;` automatically sees the new `foo` settings object.

JohnMcLear:
@orblivion can you get the tests in and conflicts resolved please? :)  Thanks!

JohnMcLear:
Still waiting on @orblivion for testing?

JohnMcLear:
@rhansen I wonder if your latest round of work makes this PR defunct?

rhansen:
> @rhansen I wonder if your latest round of work makes this PR defunct?

No, this PR is still valid.

orblivion:
Got probably most of the tests. Tried to catch lots of scenarios. Have a few more. Still need to consider other comments here.

JohnMcLear:
Bump @orblivion  :)

PR #4277, opened 2020-09-10
docs: Revise documentation for handleMessage and handleMessageSecurity

rhansen:


doc/api/hooks_server-side.md
```diff
@@ -398,22 +412,36 @@ Called from: src/node/handler/PadMessageHandler.js
...
 ```
-function handleMessageSecurity ( hook, context, callback ) {
-  if ( context.message.boomerang == 'hipster' ) {
-    // If the message boomer is hipster, allow the request
-    callback(true);
-  }else{
-    callback();
-  }
+// Using an async function:
+exports.handleMessageSecurity = async (hookName, {message, client}) => {
```
JohnMcLear, line 436: wouldn't this break every plugin that used this? :)

rhansen: I'm just documenting the current behavior. Currently plugin authors can either write their `handleMessageSecurity` function as a regular function with a callback or as an async function. Either way works.

PR #4278, opened 2020-09-10
Pad: Disable toolbar and import/export when reconnecting

rhansen:
I noticed this while troubleshooting a difficult-to-reproduce reconnect problem. If you know how to simulate a disconnect please let me know.

### DIFF
--- a/src/static/js/pad.js
+++ b/src/static/js/pad.js
@@ -222,7 +222,6 @@ function handshake()
   });
 
   socket.on('reconnecting', function() {
-    padeditor.disable();
     pad.collabClient.setStateIdle();
     pad.collabClient.setIsPendingRevision(true);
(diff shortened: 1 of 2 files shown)


JohnMcLear:
What happens if you are disconnected and need to export your pad?  Let's say the pad(editor) is broken but you still need the export.

rhansen:
> What happens if you are disconnected and need to export your pad? Let's say the pad(editor) is broken but you still need the export.

This PR only affects the `RECONNECTING` state. If you are in the `DISCONNECTED` state, then even without this change you already can't export (see [lines 785-787](https://github.com/ether/etherpad-lite/blob/80639fdc6a3be4f76b0747e611a1b1d54c51e68a/src/static/js/pad.js#L785-L787)).

PR #4260, opened 2020-09-07
PadMessageHandler: Various cleanups

rhansen:
This PR contains a few commits that each do some minor cleanups to PadMessageHandler.js.

**PLEASE DO NOT SQUASH MERGE THIS PR—do a regular merge instead.**

  * Simplify `handleClientReady` a bit
  * Invert logic to improve readability
  * Use `===` instead of `==` for comparison
  * Rename createSessionInfo to createSessionInfoAuth

PR #4281, opened 2020-09-11
PadMessageHandler: Use `await` instead of `p.then()`

rhansen:


rhansen:
Actually I'm going to move this to PR #4260.

Issue #4242, opened 2020-09-04
Feature Request/RFC: Shallow Copy Feature


joassouza:
name: Feature request
about: Implement a way to copy a pad that is less resource demanding
title: 'Shallow Pad Copy'
labels: enhancement
assignees: me

* * *
**Is your feature request related to a problem? Please describe.**
We use the duplication of pads (copyPad API) a lot in our use case. We provide it as a way to the users to create versions of their pads. With that users can keep track of the different versions of the pad and the work of their team as well. It's really good because they use this feature in the same way as we use with code branches.
The main problem with it is because Etherpad does not provide a simple copy of pads. Every copy it's like a clone of the source pad. The whole content: revisions, chat, etc is copied. This by itself it brings some problems with pads with lots of revisions. It's not difficult to find pads with over 100.000 revisions in our instance. So, every copy it needs to loop over more than 100.000 revisions. This causes problems in our Etherpad instance because the Ueberdb has a problem to handle this amount of data, please check https://github.com/ether/ueberDB#dobulk-operations-on-set-out-of-memory . On the contrary what it's said in the documentation that is easily reproducible.

**Describe the solution you'd like**
The shallow copy would only copy the content of the pad and set the head revision to 0. It would be quite similar to what we have today but only the pad content, users who contributed to this pad would be copied. We could provide a flag to warn the plugins that the operation triggered is a shallow copy, so the plugins could act properly depending on the type of the copy. Something like:
hooks.callAll('padCopy', { 'originalPad': this, 'destinationID': destinationID, fullCopy: fullCopy });

**Describe alternatives you've considered**
I could implement that feature on a plugin, but I think other people could face a similar problem like us, so being in the Etherpad core would be easier to use and benefit the community more.
I've found a plugin that does something similar to that https://github.com/redhog/ep_copypad. I didn't try this plugin and two things concerns me:
This plugin does not have any test
The last update was 7 years ago

**Additional context** 
We had to upgrade the Etherpad machine we use on AWS to one that costs **10x** more to solve this temporally. That solution does not scale well.
Usually our pads have more than 100 pages (over 5000 lines).

**Plugin?**
As I mentioned before, implementing this feature on the core has some advantages.

**Where the Copy is made**
https://github.com/ether/etherpad-lite/blob/develop/src/node/db/Pad.js#L359

[...]

PR #4295, opened 2020-09-14
Feature: copy without history

joassouza:
This PR is an implementation of https://github.com/ether/etherpad-lite/issues/4242
With this change we provide a way to copy pads without bringing all the history information (revisions, chats).

### Motivation
Etherpad is such a great tool, but in our instance we were having some problems with the current mechanism of pads duplication. A pad with 100k revisions has to perform 200k operations at least - get 100k information on the source pad and save 100k into the new pad. Although Etherpad provides a good mechanism of caching, that amount of operation occupies the CPU time too much. 
In our case we could use small instances of AWS machines t2.micro* to host Etherpad without a hitch if it wasn't that problem. As we have lots of users duplicating their pads, we had to increase our server capacity to a t2.large* instance. That means a cost of 10x. One of the greatest things about Etherpad is how lightweight it is! With that in mind this PR provides a lightweight alternative to duplicate pads.

### Observations
This method should **NOT** be preferable over `copyPad`. The `copyPad` API brings all the information so we have an exact copy of the pad. We can think the "copyWithoutHistory" as a copy/paste from one pad to another one.

### Tests
We implemented some missing tests on the copyPad on this PR as well. Due to limitations on the test suite we couldn't provide an automated way to check if the Timeslider works. I did a manual smoke test with a large pad, 11.000 lines, and everything worked properly.

### CPU usage
example

#### AWS t2 types
* https://aws.amazon.com/ec2/instance-types/t2/

closes https://github.com/ether/etherpad-lite/issues/4242

### DIFF
--- a/doc/api/http_api.md
+++ b/doc/api/http_api.md
@@ -530,7 +530,7 @@ copies a pad with full history and chat. If force is true and the destination pa
 * API >= 1.2.15
 
 copies a pad without copying the history and chat. If force is true and the destination pad exists, it will be overwritten.
-Note, all the revisions will be lost! In most of the cases one should user `copyPad` instead.
+Note that all the revisions will be lost! In most of the cases one should use `copyPad` API instead.
 
 *Example returns:*
 * `{code: 0, message:"ok", data: null}`
--- a/doc/api/http_api.md
+++ b/doc/api/http_api.md
@@ -526,6 +526,16 @@ copies a pad with full history and chat. If force is true and the destination pa
   * `{code: 0, message:"ok", data: null}`
   * `{code: 1, message:"padID does not exist", data: null}`
 
+#### copyPadWithoutHistory(sourceID, destinationID[, force=false])
+* API >= 1.2.15
+
+copies a pad without copying the history and chat. If force is true and the destination pad exists, it will be overwritten.
+Note, all the revisions will be lost! In most of the cases one should user `copyPad` instead.
+
+*Example returns:*
+* `{code: 0, message:"ok", data: null}`
+* `{code: 1, message:"padID does not exist", data: null}`
+
 #### movePad(sourceID, destinationID[, force=false])
  * API >= 1.2.8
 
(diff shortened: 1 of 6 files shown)


JohnMcLear:
How do you prefer copyPadWithoutHistory so it matches other API name?

JohnMcLear:
Can you show the cpu usage comparison between copypad and this?  I'm just curious.

JohnMcLear:
Testing needs to keep being worked on, make sure it doesn't error on group if it doesn't exist and if permissions / security is wrong make sure it fails.  Test for failure, I guess :+1:

joassouza:
@JohnMcLear I "think" the develop branch is broken actually
https://travis-ci.org/github/ether/etherpad-lite/builds/727447357
I'm working on new tests :)

JohnMcLear:
I'm aware of the breakage in the tests, I'm working on it atm, see Travis :x:

joassouza:
@JohnMcLear as you requested here's the comparison between copyPad and copyPadWithoutHistory.
It uses the same pad, a really simple one. Just one line! I've created **13956 revisions**. 
```json
{
  _id: ObjectId("5f5fd9fdde1f46ddea6244f8"),
  key: "pad:g.GroupId$padId", // changed the value
  val: "{\"atext\":{\"text\":\"line 1200242299996808\\n\",\"attribs\":\"*0+l|1+1\"},\"pool\":{\"numToAttrib\":{\"0\":[\"author\",\"a.bVnYTJfxU2HC4SB6\"]},\"nextNum\":1},\"head\":13956,\"chatHead\":-1,\"publicStatus\":false,\"passwordHash\":null,\"savedRevisions\":[]}"
}
```
_Every revision it was the replacing of one char, no new lines were generated. The idea was to create lots of revisions in the most simple pad possible._ 


### 1. Duplicating pad using `copyPad API`
copyPad


### 2. Duplicating pad using `copyPadWithoutHistory API`
copy pad without history


##### The second time we run `copyPadWithoutHistory` we trigger it **5 times!**

JohnMcLear:
That's really cool to show the change in CPU usage between the two methods :)

PR #4298, opened 2020-09-15
css: Line up line numbers with their rows

### LINKED ISSUE #4297: Line numbers don't line up with `no-skin` skin
As of PR #4267 (I think), line numbers no longer line up with the lines when using the `no-skin` skin:

Screenshot_20200914_231906

cc @seballot

rhansen:
Tested with both `no-skin` and `colibris`.

Fixes #4297.

cc @seballot


seballot:
oops, thank for the fix ! this double skin is a bit a pain to maintain !!

PR #4304, opened 2020-09-15
tests: Use wtfnode to determine why mocha isn't exiting

rhansen:
If mocha hangs after running the tests, hit Ctrl-C and wtfnode will print open files, open sockets, running timers, and running intervals. Adding an `after` function that closes/stops all of those things will ensure that mocha exits when it finishes running the tests.

### DIFF
--- a/src/package-lock.json
+++ b/src/package-lock.json
@@ -8835,6 +8835,12 @@
       "resolved": "https://registry.npmjs.org/ws/-/ws-7.3.0.tgz",
       "integrity": "sha512-iFtXzngZVXPGgpTlP1rBqsUK82p9tKqsWRPg5L56egiljujJT3vGAYnHANvFxBieXrTFavhzhxW52jnaWV+w2w=="
     },
+    "wtfnode": {
+      "version": "0.8.3",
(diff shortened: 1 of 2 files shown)


webzwo0i:
Doesn't https://mochajs.org/#-exit help?

rhansen:
> Doesn't https://mochajs.org/#-exit help? (Not with finding the not-cleaned-up-yet state but with definitely exiting the tests)

Yes, that would work too. I decided not to take that approach because (1) I wanted to understand why it wasn't exiting, (2) I like doing it "the right way", and (3) I was a bit worried that `--exit` might hide real bugs.

webzwo0i:
wtfnode seems to always exit with 0 which breaks the usefulness of the backend tests, so better don't run it in ci?

If the tests don't finish:
- disable all suites (describe -> xdescribe) or run a specific mocha file
- if it does not exit properly the bug is somewhere in setup code outside of `describe` (better move it into an "before"-hook and do the teardown in an "after"-hook)
- if it does than enable the suites one by one until you have the offending area

PR #4312, opened 2020-09-17
utils: Fix promise creation accounting bug in promises.timesLimit

rhansen:
Before this change, `promises.timesLimit()` created `concurrency - 1` too many promises. The only users of this function use a concurrency of 500, so this meant that 499 extra promises were created each time it was used. The bug didn't affect correctness, but it did result in a large number of unnecessary database operations whenever a pad was deleted. This change fixes that bug.

Also:
  * Convert the function to async and have it resolve after all of the created promises are resolved.
  * Reject concurrency of 0 (unless total is 0).
  * Document the function.
  * Add tests.

### DIFF
--- a/src/node/utils/promises.js
+++ b/src/node/utils/promises.js
@@ -35,27 +35,21 @@ exports.firstSatisfies = (promises, predicate) => {
   return Promise.race(newPromises);
 };
 
-exports.timesLimit = function(ltMax, concurrency, promiseCreator) {
-  var done = 0
-  var current = 0
-
-  function addAnother () {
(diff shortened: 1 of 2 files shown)


rhansen:
@webzwo0i would you mind giving your opinion about the tests I'm adding in this PR? I'm wondering if this is a non-idiomatic way of writing mocha tests. (Also see https://github.com/ether/etherpad-lite/pull/4254#issuecomment-693732671.)

JohnMcLear:
Nice job!

PR #4313, opened 2020-09-17
async-ify more functions, and await completion

rhansen:
Where feasible I put the `await` at the end of the function to minimize the impact on latency.

My motivation for this change: Eliminate a race condition in tests I am writing.

### DIFF
--- a/src/node/utils/ImportHtml.js
+++ b/src/node/utils/ImportHtml.js
@@ -22,8 +22,7 @@ const rehype            = require("rehype")
 const format            = require("rehype-format")
 
 
-exports.setPadHTML = function(pad, html)
-{
+exports.setPadHTML = async (pad, html) => {
   var apiLogger = log4js.getLogger("ImportHtml");
 
   var opts = {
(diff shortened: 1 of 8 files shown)


JohnMcLear:
My only concern here is if there are some scripts in bin/ that call some of these files directly that won't be using async/await?  I don't think they have test coverage.

rhansen:
> My only concern here is if there are some scripts in bin/ that call some of these files directly that won't be using async/await? I don't think they have test coverage.

This PR does not cause any code that was previously synchronous to run asynchronously. All non-`async` callers of these newly-`async` functions were already continuing execution without waiting for the asynchronous code to complete—the only difference is they are now ignoring a Promise instead of ignoring an inaccessible background operation.

rhansen:
Also: I used `git grep` to audit all callers of these functions to make sure they were either updated or already correct.

src/node/utils/ImportHtml.js
```diff
@@ -103,6 +102,8 @@ exports.setPadHTML = function(pad, html)
   var theChangeset = builder.toString();
 
   apiLogger.debug('The changeset: ' + theChangeset);
-  pad.setText("\n");
-  pad.appendRevision(theChangeset);
+  await Promise.all([
+    pad.setText('\n'),
```
webzwo0i, line 106: Is it possible that this can lead to a race when `pad.appendRevision` completes before `pad.setText`?

rhansen: Looking at the code I don't think there's currently a race condition, but that could easily change if the implementation of `Pad.setText()` or `Pad.appendRevision()` is changed. Either way, this code is not very readable; waiting for `setText()` to complete before calling `appendRevision()` would be an improvement.

(Note that this code was problematic before—this PR didn't introduce the problem.)

PR #4322, opened 2020-09-19
Improve toolbar responsiveness for small screen Fixes #4321

### LINKED ISSUE #4321: Etherpad on mobile devices
iPads (and other mobile devices with relatively large screens) do not use the mobile version of etherpad-lite. This results in broken icons in the toolbar. I think the [media queries](http://stephen.io/mediaqueries/#iPad) have to be adapted.

Screenshot_20200919_125849

seballot:
Until now, the "mobile layout" (with right toolbar on bottom of the screen) was displayed only when screen was smaller than 800px. It made the toolbar break for screen about 1000px when a lot of plugins are in the toolbar.

Now instead, we detect with javascript when the toolbar icons overflow the natural space available, and we switch in "mobile layout" in such case

toolbar

### DIFF
--- a/src/static/css/pad/layout.css
+++ b/src/static/css/pad/layout.css
@@ -47,8 +47,6 @@ body {
   width: 0; /* hide when the container is empty */
 }
 
-@media only screen and (max-width: 800px) {
-  #editorcontainerbox {
-    margin-bottom: 39px; /* Leave space for the bottom toolbar on mobile */
-  }
+.mobile-layout #editorcontainerbox {
+  margin-bottom: 39px; /* Leave space for the bottom toolbar on mobile */
 }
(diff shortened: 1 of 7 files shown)


AliKarpuzoglu:
Hey guys, hi @JohnMcLear 
When do you think a new release would be ready? This feature would warrant a release in my opinion since it fixes the Etherpad for the budget (standard) iPad which is presumably the most used resolution (50% of tablet share according to https://gs.statcounter.com/screen-resolution-stats/tablet/worldwide)

JohnMcLear:
@AliKarpuzoglu afaik we provide milestones that need solving to get a release shipped.  Please refer to them :+1:

PR #4324, opened 2020-09-19
SessionManager: Fix session expiration check

### LINKED ISSUE #4323: Wrong condition on access check to group pad
https://github.com/ether/etherpad-lite/blob/12bd617f511de1da4f4511617ffed6e2804ac9d7/src/node/db/SessionManager.js#L75

The last condition is wrong: si.validUntil <= now, which means that only expired sessions will be matched. Instead we should match only si.validUntil > now sessions.

rhansen:
This bug was introduced in 8b0baa96797718985b0557d25d4696c19220c309.

Fixes #4323

### DIFF
--- a/src/node/db/SessionManager.js
+++ b/src/node/db/SessionManager.js
@@ -72,7 +72,7 @@ exports.findAuthorID = async (groupID, sessionCookie) => {
     return undefined;
   });
   const now = Math.floor(Date.now() / 1000);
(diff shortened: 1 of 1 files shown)


rhansen:
Should this go to `master` instead of `develop`?

JohnMcLear:
It can be cherry picked into master if @Gared wants :)

Gared:
Probably best way to do a new release...

PR #4329, opened 2020-09-21
Refactor startup/shutdown for tests

rhansen:
* `src/node/server.js` can now be run as a script (for normal operation) or imported as a module (for tests).
  * Move shutdown actions to `src/node/server.js` to be close to the startup actions.
  * Put startup and shutdown in functions so that tests can call them.
  * Use `await` instead of callbacks.
  * Block until the HTTP server is listening to avoid races during test startup.
  * Add a new `shutdown` hook.
  * Use the `shutdown` hook to:
      * close the HTTP server
      * call `end()` on the stats collection to cancel its timers
      * call `terminate()` on the Threads.Pool to stop the workers
  * Exit with exit code 0 (instead of 1) on SIGTERM.
  * Export the HTTP server so that tests can get the HTTP server's port via `server.address().port` when `settings.port` is 0.

### DIFF
--- a/doc/api/hooks_server-side.md
+++ b/doc/api/hooks_server-side.md
@@ -10,6 +10,28 @@ Things in context:
 
 Use this hook to receive the global settings in your plugin.
 
+## shutdown
+Called from: src/node/server.js
+
+Things in context: None
+
+This hook runs before shutdown. Use it to stop timers, close sockets and files,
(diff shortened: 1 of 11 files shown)


rhansen:
Looks like I need to increase the 5s test timeout.

rhansen:
> Looks like I need to increase the 5s test timeout.

Done.

PR #4339, opened 2020-09-23
Include plugins in tests

JohnMcLear:
Includes all ether/ plugins when we test Etherpad.

The idea is that we can spot any breaking (to plugin) changes by having sufficient coverage here.


webzwo0i:
Any chance running both with and without plugins? We only have 5 parallel travis runner, but maybe we can reuse them (without discarding the always-clear-setup-environment requirement)? I have concerns that only running with plugins can shadow bugs.

I think for backend tests it's already possible to have them run with and without plugins, because they are so fast (~2min). For frontend I am confident, that at one point we can parallelize the tests (i.e. having 10+ tabs that run the individual test files). So if you think running frontend tests two times takes too long, I am okay with that and probably come back when the frontend tests are faster :-)

JohnMcLear:
> Any chance running both with and without plugins? We only have 5 parallel travis runner, but maybe we can reuse them (without discarding the always-clear-setup-environment requirement)? I have concerns that only running with plugins can shadow bugs.
> 
> I think for backend tests it's already possible to have them run with and without plugins, because they are so fast (~2min). For frontend I am confident, that at one point we can parallelize the tests (i.e. having 10+ tabs that run the individual test files). So if you think running frontend tests two times takes too long, I am okay with that and probably come back when the frontend tests are faster :-)

testing running both :)

PR #4353, opened 2020-09-26
PadMessageHandler: Always save the author ID in the session info

rhansen:
Before, the author ID was only saved in the session info during the initial `CLIENT_READY`, not when the client sent a `CLIENT_READY` due to a reconnect. This caused the handling of subsequent messages to use an undefined author ID.


JohnMcLear:
wohoo :)  The illusive error is found!  Good job!

PR #4358, opened 2020-09-26
webaccess: Log all authentication successes/failures

rhansen:
This loses some of the granularity of the default HTTP basic auth (unknown username vs. bad password), but there is considerable value in having logging that is consistent no matter what authentication plugins are installed.


JohnMcLear:
Can you just confirm we have test coverage for this logic?  I don't recall writing anything that tests auth but it might already exist?  If not, can we include it please? :)

rhansen:
Yes, we already have test coverage: https://github.com/ether/etherpad-lite/blob/develop/tests/backend/specs/webaccess.js

PR #4362, opened 2020-09-26
webaccess: Move pre-authn authz check to a separate hook

rhansen:
Before this change, the authorize hook was invoked twice: once before authentication and again after (if settings.requireAuthorization is true). Now pre-authentication authorization is instead handled by a new preAuthorize hook, and the authorize hook is only invoked after the user has authenticated.

Rationale: Without this change it is too easy to write an authorization plugin that is too permissive. Specifically:

  * If the plugin does not check the path for /admin then a non-admin user might be able to access /admin pages.
  * If the plugin assumes that the user has already been authenticated by the time the authorize function is called then unauthenticated users might be able to gain access to restricted resources.

This change also avoids calling the plugin's authorize function twice per access, which makes it easier for plugin authors to write an authorization plugin that is easy to understand.

This change may break existing authorization plugins: After this change, the authorize hook will no longer be able to authorize non-admin access to /admin pages. This is intentional. Access to admin pages should instead be controlled via the `is_admin` user setting, which can be set in the config file or by an authentication plugin.

Also:
  * Add tests for the authenticate and authorize hooks.
  * Disable the authentication failure delay when testing.

### DIFF
--- a/doc/api/hooks_server-side.md
+++ b/doc/api/hooks_server-side.md
@@ -212,6 +212,50 @@ Things in context:
 
 I have no idea what this is useful for, someone else will have to add this description.
 
+## preAuthorize
+Called from: src/node/hooks/express/webaccess.js
+
+Things in context:
+
+1. req - the request object
+2. res - the response object
+3. next - bypass callback. If this is called instead of the normal callback then
+   all remaining access checks are skipped.
+
+This hook is called for each HTTP request before any authentication checks are
+performed. Example uses:
+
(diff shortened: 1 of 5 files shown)


rhansen:
While working on adding tests for this I noticed some minor problems, so I changed this to draft. I'll mark it as ready to review once those issues are fixed.

JohnMcLear:
Looks good to me, just need to do a second read then will merge commit

PR #4367, opened 2020-09-27
webaccess: Log hook errors

rhansen:


JohnMcLear:
could this ever post credentials into a log file?

rhansen:
> could this ever post credentials into a log file?

Only if someone includes the credentials in the error. This could happen if a plugin logs `Unknown username ${username}` and users accidentally enter their password in the username box. This is a common enough problem that any authentication plugin author should know better. :smile:

PR #4373, opened 2020-09-29
security: Support proxy with rate limiting and include CI test coverage for nginx rev proxy

webzwo0i:
It's ugly but works

### DIFF
--- a/src/node/handler/PadMessageHandler.js
+++ b/src/node/handler/PadMessageHandler.js
@@ -181,10 +181,11 @@ exports.handleMessage = async function(client, message)
   var env = process.env.NODE_ENV || 'development';
 
   if (env === 'production') {
+    const clientIPAddress = (settings.trustProxy && client.handshake.headers['x-forwarded-for']) ? client.handshake.headers['x-forwarded-for'] : client.handshake.address;
     try {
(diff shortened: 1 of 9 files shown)


src/node/handler/PadMessageHandler.js
```diff
@@ -181,10 +181,11 @@ exports.handleMessage = async function(client, message)
   var env = process.env.NODE_ENV || 'development';
 
   if (env === 'production') {
+    const clientIPAddress = (settings.trustProxy && client.handshake.headers['x-forwarded-for']) ? client.handshake.headers['x-forwarded-for'] : client.handshake.address;
```
rhansen, line 184:
```suggestion
    const clientIPAddress = remoteAddress[client.id];
```

JohnMcLear:
Sorry dude I just pushed package-lock, can you please resolve that conflict?

src/node/handler/PadMessageHandler.js
webzwo0i: Perfect. I saw this, but did not realize it's set in the connection setup phase and is available to later socket.io messages.

rhansen:
I rebased the `fix-rate-limit` branch onto `develop`.

PR #4380, opened 2020-10-01
Localize `/admin` pages

rhansen:
Not every string was localized:

  * `/admin/plugins` has some CSS magic to draw the tables of plugins differently on narrow (mobile) screens, and the l10n library we use does not support that particular magic. The strings that were not localized are "Name", "Description", "Version", and "Time". These strings are only stuck in English when the page is viewed on a narrow screen; normal desktop users will see translated strings. The CSS magic ought to be replaced with something more robust (lots of nested `div`s); those remaining strings can be localized whenever that happens.

  * Strings from external sources such as plugin descriptions, error messages, and `settings.json` comments are not localized.

cc @ovari


ovari:
If we fork this once this PR is merged, then we can localize  src/locales/en.json into into `hu`. Is that okay? Thank you

rhansen:
Yes, that should work. Or you can wait for translatewiki.net to automatically pick up the new strings and use that to translate to `hu`.

ovari:
Does translatewiki.net send us an email when there are new strings? Thank you

rhansen:
> Does translatewiki.net send us an email when there are new strings? Thank you

I don't know how translatewiki works. @JohnMcLear might be able to answer your question.

JohnMcLear:
We get commits from TW usually it's daily.

Issue #1922, opened 2013-10-06
Log4js is stale due to upstream change


JohnMcLear:
Log4js is stale

JohnMcLear:
It's still creating a new file btw..

JohnMcLear:
Mheh it turns out log4js made loads of breaking changes over the years and now I gotta support old settings/configs to pass to their now broken API.  FML

JohnMcLear:
- [ ] Write mechanism to see if this is pre v2 config (appenders.out !== 'undefined')
  - [ ] Write mechanism to see if this is pre v2 install of log4js (version) 
  - [ ] Write mechanism to port v1 config to v2.
  - [ ] Update settings.json.template with new config style
  - [ ] Make changes to logic handlers IE reload of config [ a8ec242d0047e501b410d44108d7a70a1d7cb705 ]

JohnMcLear:
I'm not gonna work on this, just don't feel like it :)

rhansen:
> I'm not gonna work on this, just don't feel like it :)

I'd be surprised if anyone feels like working on this. :smile:

Unless there is an existing library that can convert a v0.6.35 config to something compatible with v3.0 and later, I think the only feasible option is to break backward compatibility and force users to re-write their logging settings. (Also: log4js seems to have gone out of favor, so we may want to switch to [`winston`](https://www.npmjs.com/package/winston) or something else.)

To make it possible for us to change the logging library again in the future without breaking user configs, I suggest we simplify the logging settings by replacing them with the following:

* `settings.loglevel`: Unchanged.
* `settings.logfile`: If set, also write log messages to this file. Defaults to unset.

And if we switch to a logging library that supports syslog, add the following

* `settings.syslog`: If true, also write log messages to syslog (`LOG_USER` facility). Defaults to false.

Most users aren't going to want anything more sophisticated than the above. For exotic configurations, we can either tell users to patch their local copy of Etherpad or we can have a (hidden?) `advancedLoggingConfig` object in `settings.json` that is not guaranteed to remain compatible across Etherpad versions.

rhansen:
> This is a configuration breaking change

I don't understand what you mean by "this".

Note that PR #5178 deprecated the `logconfig` setting, which will make it possible for us to upgrade log4js once users have been given enough time to remove `logconfig` from their `settings.json`.

squiddle:
> > This is a configuration breaking change
> 
> I don't understand what you mean by "this".


[...]

PR #4383, opened 2020-10-02
import: Replace the `allowAnyoneToImport` check with `userCanModify`

### LINKED ISSUE #4381: Revisit allowAnyoneToImport
PR #3833 added a requirement to edit a pad before import, unless `settings.allowAnyoneToImport` is true (see issue #3964 and PR #3965). I don't know all of the background that went into adding that requirement, but with recent authentication and authorization changes (see issue #4325) it might now be safe to rely solely on [`securityManager.checkAccess`](https://github.com/ether/etherpad-lite/blob/ceb09ce99ad080ae804d322804b22e3e8f0be005/src/node/db/SecurityManager.js#L54) and [`webaccess.userCanModify`](https://github.com/ether/etherpad-lite/blob/ceb09ce99ad080ae804d322804b22e3e8f0be005/src/node/hooks/express/webaccess.js#L38-L47) to check if a user is allowed to import. Removing the need to edit a pad before import will simplify certain use cases such as programmatic imports (see issue #4379).

rhansen:
This PR removes the `allowAnyoneToImport` setting and replaces it with a call to the `webaccess.canUserModify` function. **Please do not squash merge this PR** because it has a few commits that are intentionally separate:

  * tests: Factor out common server setup/teardown
  * tests: Switch import/export tests to self-contained server
  * tests: Always run the import unsupported file type test
  * webaccess: Check for read-only pad ID in `userCanModify`
  * webaccess: Fix pad ID extraction for import and export paths
  * import: Allow import if pad does not yet exist
  * import: Replace the `allowAnyoneToImport` check with `userCanModify`

Fixes #4381.

### DIFF
--- a/tests/frontend/travis/runnerBackend.sh
+++ b/tests/frontend/travis/runnerBackend.sh
@@ -12,11 +12,8 @@ cd "${MY_DIR}/../../../"
 # Set soffice to /usr/bin/soffice
 sed 's#\"soffice\": null,#\"soffice\":\"/usr/bin/soffice\",#g' settings.json.template > settings.json.soffice
 
-# Set allowAnyoneToImport to true
-sed 's/\"allowAnyoneToImport\": false,/\"allowAnyoneToImport\": true,/g' settings.json.soffice > settings.json.allowImport
-
 # Set "max": 10 to 100 to not agressively rate limit
-sed 's/\"max\": 10/\"max\": 100/g' settings.json.allowImport > settings.json.rateLimit
+sed 's/\"max\": 10/\"max\": 100/g' settings.json.soffice > settings.json.rateLimit
 
 # Set "points": 10 to 1000 to not agressively rate limit commits
 sed 's/\"points\": 10/\"points\": 1000/g' settings.json.rateLimit > settings.json
--- a/settings.json.template
+++ b/settings.json.template
@@ -497,19 +497,6 @@
    */
   "importMaxFileSize": 52428800, // 50 * 1024 * 1024
 
-
-  /*
-   * From Etherpad 1.8.3 onwards import was restricted to authors who had
-   * content within the pad.
-   *
-   * This setting will override that restriction and allow any user to import
-   * without the requirement to add content to a pad.
-   *
-   * This setting is useful for when you use a plugin for authentication so you
-   * can already trust each user.
-   */
-  "allowAnyoneToImport": false, 
-
   /*
    * From Etherpad 1.9.0 onwards, when Etherpad is in production mode commits from individual users are rate limited
    *
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -188,6 +188,5 @@
   "pad.impexp.importfailed": "Import failed",
   "pad.impexp.copypaste": "Please copy paste",
   "pad.impexp.exportdisabled": "Exporting as {{type}} format is disabled. Please contact your system administrator for details.",
-  "pad.impexp.maxFileSize": "File too big. Contact your site administrator to increase the allowed file size for import",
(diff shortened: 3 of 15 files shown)


tests/frontend/travis/runnerBackend.sh
```diff
@@ -12,11 +12,8 @@ cd "${MY_DIR}/../../../"
 # Set soffice to /usr/bin/soffice
 sed 's#\"soffice\": null,#\"soffice\":\"/usr/bin/soffice\",#g' settings.json.template > settings.json.soffice
 
-# Set allowAnyoneToImport to true
-sed 's/\"allowAnyoneToImport\": false,/\"allowAnyoneToImport\": true,/g' settings.json.soffice > settings.json.allowImport
-
 # Set "max": 10 to 100 to not agressively rate limit
-sed 's/\"max\": 10/\"max\": 100/g' settings.json.allowImport > settings.json.rateLimit
+sed 's/\"max\": 10/\"max\": 100/g' settings.json.soffice > settings.json.rateLimit
```
JohnMcLear, line 16: This looks wrong...  :??

JohnMcLear:
1. Translations need to be worked on separately.

1. We need to support the "old" setting as some people will have set that.

1. Comment RE soffice

1. Let's be mindful that log4js is a problem and stale, it doesn't support the old settings method.  See https://github.com/ether/etherpad-lite/issues/1922

tests/frontend/travis/runnerBackend.sh
rhansen: It's correct. The pattern for this file is:

  * `sed a > b`
  * `sed b > c`
  * `sed c > d`
  * etc.

I'm removing something from the middle of the chain:

  * `sed a > b`
  * `sed b > d`
  * etc.

rhansen:
> 1. Translations need to be worked on separately.

How should I handle the removal of the strings? Just edit `en.json` and let the other languages be updated automatically?

> 
> 2. We need to support the "old" setting as some people will have set that.

If a user has that setting they'll see the following warning in their log but otherwise it is ignored:

```
[WARN] console - Unknown Setting: 'allowAnyoneToImport'. This setting doesn't exist or it was removed
```

> 
> 3. Comment RE soffice

Replied to that comment.

> 
> 4. Let's be mindful that log4js is a problem and stale, it doesn't support the old settings method.  See #1922

The logging changes here should be forward compatible with newer log4js versions.

JohnMcLear:
> How should I handle the removal of the strings? Just edit `en.json` and let the other languages be updated automatically?

Yes.  Exactly :)


> If a user has that setting they'll see the following warning in their log but otherwise it is ignored:

Ugh, that's not ideal ;\  Peoples experience will change and that's not great..  I can live with it but for a few lines of code to catch it seems worthwhile not to upset users when they upgrade...



> The logging changes here should be forward compatible with newer log4js versions.

They are, it's our settings.json object that isn't...   So having this abstraction could be good for if we did want to use a newer version of log4js we could mutate the object here.

rhansen:
> > How should I handle the removal of the strings? Just edit `en.json` and let the other languages be updated automatically?
> 
> Yes. Exactly :)

Done. (I also rebased onto latest `develop`.)

> 
> > If a user has that setting they'll see the following warning in their log but otherwise it is ignored:
> 
> Ugh, that's not ideal ;\ Peoples experience will change and that's not great.. I can live with it but for a few lines of code to catch it seems worthwhile not to upset users when they upgrade...

Users shouldn't see a behavior change, except they are no longer required to create and edit the pad before importing. (It's still OK if they do create and edit the pad before importing, of course.)

Issue #4389, opened 2020-10-05
timeslider follow content: Cannot read property 'offsetTop' of undefined


webzwo0i:
**Describe the bug**
Introduced with bf24063234b29f34a3674fcedd6a2da281b04d4b
The timeslider cannot display the contents in certain situations and errors, so that timeslider play or backwards/forwards buttons do not work anymore
**To Reproduce**
1. Create pad 
2. Delete everything (ctrl+a, delete)
3. Press enter
4. goto timeslider, activate "follow pad contents"
5. it errors when going from revision 1 to 2, and when going from 1 to 0.

Error is: 
`timeslider.js?callback=require.define&v=33c0661d:6342 Uncaught TypeError: Cannot read property 'offsetTop' of undefined
    at goToLineNumber (timeslider.js?callback=require.define&v=33c0661d:6342)
    at applyChangeset (timeslider.js?callback=require.define&v=33c0661d:6004)
    at Function.goToRevision (timeslider.js?callback=require.define&v=33c0661d:6099)
    at Array.goToRevisionIfEnabled (timeslider.js?callback=require.define&v=33c0661d:6309)
    at _callSliderCallbacks (timeslider.js?callback=require.define&v=33c0661d:6401)
    at setSliderPosition (timeslider.js?callback=require.define&v=33c0661d:6470)
    at HTMLButtonElement.<anonymous> (timeslider.js?callback=require.define&v=33c0661d:6693)
    at HTMLButtonElement.dispatch (ace2_common.js?callback=require.define&v=33c0661d:5492)
    at HTMLButtonElement.elemData.handle (ace2_common.js?callback=require.define&v=33c0661d:5300)`

In the case of revision 1 to 2: in rev 1 the pad contains one line ("\n") and the changeset in rev 2 inserts another one and goToLineNumber will try to go to line 2 and fail.
In the case of revision 1 to 0: in rev 1 the pad contains one line ("\n") and the changeset for 0 will insert 7 lines for the default pad text. goToLineNumber will try to go to line 8 and fail.

JohnMcLear:
@webzwo0i closed this as the 4390 was merged.

PR #4390, opened 2020-10-05
timeslider follow pad contents: only goToLineNumber if it exists

webzwo0i:
fix for https://github.com/ether/etherpad-lite/issues/4389

### DIFF
--- a/src/static/js/broadcast.js
+++ b/src/static/js/broadcast.js
@@ -255,9 +255,15 @@ function loadBroadcastJS(socket, sendSocketMsg, fireWhenAllScriptsAreLoaded, Bro
       debugLog(e);
     }
 
-    var lineNumber = Changeset.opIterator(Changeset.unpack(changeset).ops).next().lines;
+    // we want to scroll to the area that is changed before the lines are mutated and we can't go
+    // to a line that is not there yet
+    let lineNumber = Changeset.opIterator(Changeset.unpack(changeset).ops).next().lines;
     if($('#options-followContents').is(":checked") || $('#options-followContents').prop("checked")){
-      goToLineNumber(lineNumber);
+      if(padContents.currentLines.length <= lineNumber){
+        goToLineNumber(padContents.currentLines.length-1)
+      } else {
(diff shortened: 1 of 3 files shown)


JohnMcLear:
Worth building test coverage into this?

webzwo0i:
tests are in and I checked the exception is raised without the fix in this PR. I used the rightstep/leftstep buttons and compared the pad contents, so there is now test coverage for all three timeslider buttons.

JohnMcLear:
https://travis-ci.org/github/ether/etherpad-lite/jobs/733045689#L1796

Fails FF 78 [firefox Windows 7 78.0]     -> FAILED : follows only to lines that exist in the current pad view Error: Timeout of 2

webzwo0i:
Ha! Thanks!
Testing with larger timeout now

JohnMcLear:
@webzwo0i our testing framework can be so annoying :+1:

webzwo0i:
:-D 
I migrated the test to use some of the helper methods I was working on two month ago but did not finish yet and it works (https://travis-ci.org/github/ether/etherpad-lite/jobs/733305037)
https://github.com/ether/etherpad-lite/blob/workaround-dom-latency/tests/frontend/specs/timeslider_follow.js#L32
imo more readable, but the main advantage is that you can use await in a lot of places to avoid setTimeout and problems related to mixing setTimeout and waitFor.

Disadvantage is that it's adding a lot of helper methods, which can be cumbersome to work with. I did this to ensure you don't have to explicitly write waitFor or setTimeouts when using click events, navigation, etc. (Like https://github.com/ether/etherpad-lite/blob/workaround-dom-latency/tests/frontend/helper/methods.js#L85

Probably better to invest time into getting PRs ready for the new frontend helpers than to spend it on try-and-error here :-)

JohnMcLear:
Okay, so do you want to remove the testing stuff/resolve conflicts and I can merge this as a patch then we can add tests later?

webzwo0i:
Yes. You can squash and merge

PR #4391, opened 2020-10-05
tests: Disable non-test logging unless level <= DEBUG

rhansen:
This makes it easier to see the test results, and it hides some scary-looking but intentional error messages.

This code will likely have to be updated if/when we change the logging library (see issue #1922).


JohnMcLear:
Do I need to do anything in ep_comments for the tests to pass or?  Will run them anyway just to see.

JohnMcLear:
Looks to have worked, tnx!  Will fix up ep_comments tests ASAP

Issue #4335, opened 2020-09-22
Node 10 is broken


JohnMcLear:
Updating fixes..  I'm wondering if they hot patched an old release and now we're seeing issues with it...

JohnMcLear:
```
Ensure that all dependencies are up to date... If this is the first time you have run Etherpad please be patient.
wd@1.12.1 install /home/pi/etherpad-lite/src/node_modules/wd
node scripts/build-browser-scripts
added 1127 packages in 41.911s
Clearing minified cache...
Starting Etherpad...
All relative paths will be interpreted relative to the identified Etherpad base dir: /home/pi/etherpad-lite
Random string used for versioning assets: 650daa0b
[2020-09-22 18:08:58.773] [DEBUG] AbsolutePaths - Relative path "settings.json" can be rewritten to "/home/pi/etherpad-lite/settings.json"
[2020-09-22 18:08:58.777] [DEBUG] AbsolutePaths - Relative path "credentials.json" can be rewritten to "/home/pi/etherpad-lite/credentials.json"
settings loaded from: /home/pi/etherpad-lite/settings.json
No credentials file found in /home/pi/etherpad-lite/credentials.json. Ignoring.
[2020-09-22 18:08:58.856] [INFO] console - Using skin "colibris" in dir: /home/pi/etherpad-lite/src/static/skins/colibris
[2020-09-22 18:08:58.857] [INFO] console - Session key file "/home/pi/etherpad-lite/SESSIONKEY.txt" not found. Creating with random contents.
[2020-09-22 18:08:58.860] [WARN] console - DirtyDB is used. This is fine for testing but not recommended for production. File location: /home/pi/etherpad-lite/var/dirty.db
[2020-09-22 18:09:04.523] [INFO] APIHandler - Api key file "/home/pi/etherpad-lite/APIKEY.txt" not found. Creating with random contents.
[2020-09-22 18:09:05.183] [INFO] console - Installed plugins:
[2020-09-22 18:09:05.195] [INFO] console - Report bugs at https://github.com/ether/etherpad-lite/issues
[2020-09-22 18:09:05.197] [INFO] console - Your Etherpad version is 1.8.6 (299bd96)
[2020-09-22 18:09:06.047] [INFO] console - You can access your Etherpad instance at http://0.0.0.0:9001/
[2020-09-22 18:09:06.048] [WARN] console - Admin username and password not set in settings.json. To access admin please uncomment and edit 'users' in settings.json
[...]

PR #4392, opened 2020-10-05
WIP: Run node 10 with '--experimental_worker' flags

jeanfabrice:

Fixes: #4335 for both Docker and Linux setup:
- add `--experimental-worker` to Dockerfile
- move common constants and functions to `functions.sh` adding a new `compute_node_args` that may compute additionals node arguments based on several tests. Here, use this function to add `--experimental-worker` when node 10 is detected.
- source `functions.sh` in all startup scripts, and use `compute_node_args` as a node argument

I did not check Windows environment since I don't have any

### DIFF
--- a/bin/functions.sh
+++ b/bin/functions.sh
@@ -12,45 +12,63 @@ error() { log "ERROR: $@" >&2; }
 fatal() { error "$@"; exit 1; }
 is_cmd() { command -v "$@" >/dev/null 2>&1; }
 
-compute_node_args() {
-  NODE_VERSION_STRING=$(node --version)
-  NODE_VERSION_STRING=${NODE_VERSION_STRING#"v"}
-  DETECTED_MAJOR=$(pecho "$NODE_VERSION_STRING" | cut -s -d "." -f 1)
-  ARGS=""
 
-  [ -n "$DETECTED_MAJOR" ] || fatal "Cannot extract $PROGRAM_LABEL major version from version string \"$VERSION_STRING\""
+get_program_version() {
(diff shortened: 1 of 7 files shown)


bin/functions.sh
```diff
@@ -0,0 +1,56 @@
...
+fatal() { error "$@"; exit 1; }
+is_cmd() { command -v "$@" >/dev/null 2>&1; }
+
+compute_node_args() {
+  NODE_VERSION_STRING=$(node --version)
+  NODE_VERSION_STRING=${NODE_VERSION_STRING#"v"}
+  DETECTED_MAJOR=$(pecho "$NODE_VERSION_STRING" | cut -s -d "." -f 1)
+  ARGS=""
+
+  [ -n "$DETECTED_MAJOR" ] || fatal "Cannot extract $PROGRAM_LABEL major version from version string \"$VERSION_STRING\""
```
rhansen, line 21: This logic is repeated in `require_minimal_version`. Please move it to a separate function that is called by both `compute_node_args` and `require_minimal_version`.

rhansen (commented):
Thank you for this! I have one suggested change, otherwise LGTM.

bin/functions.sh
jeanfabrice: pushed 5aaeddf to improve this logic

JohnMcLear:
Ready for review?  It's still flagged as WIP:

jeanfabrice:
> Ready for review? It's still flagged as WIP:

'cause I'm not able to address the Windows part :/

Issue #4401, opened 2020-10-07
Move all inline code to `.js` files


rhansen:
I would like to move all of the inline code into `.js` files so that users can set a [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) (CSP) that prohibits inline code. Prohibiting inline code reduces the potential number of XSS vulnerabilities.

rhansen:
As a first step, I would like to delete the indexCustomInlineScripts hook that was added in PR #3996 for issue #3957.

cc @brunob

JohnMcLear:
Makes sense to me.  I used ep_helmet to handle CSP by adding nonces btw but moving it out from being inline is way better.  Peep ep_helmet first and the nonce open PR as reference points :+1:

brunob:
Why not, but what alternative would we have to modify the index inline script after that ?

JohnMcLear:
@brunob isn't there a hook to include JavaScript in the head?

brunob:
@JohnMcLear i don't think so, here is what we do with that hook https://framagit.org/infini/ep_infini/-/blob/master/infini.js#L10

JohnMcLear:
Okay so @rhansen why remove it?  Just ensure the docs say to people to only reference scripts here IE how the template does it..

```<script src="static/skins/<%=encodeURI(settings.skinName)%>/index.js?v=<%=settings.randomVersionString%>"></script>
```

An example of how not to do it could be

```
function(){alert("pow")};
```
An example of how to do it could be..
```
<script src="myawesomepluginscript.js"></script>
```
We could also just in the HTML put a comment saying not to include inline JS...

I'm lazy too which is why I prefer nonce approach over removing inline completely because this catches all potential plugin issues...

rhansen:
I moved the indexCustomInlineScripts-specific discussion to PR #4402.

> I prefer nonce approach over removing inline completely because this catches all potential plugin issues...

Adding a nonce might be a good temporary fix, but I don't like it as a long-term fix because:

  * It is easy to do it incorrectly (insecurely).
  * It breaks caching. By definition, a nonce can only be used once.
  * We would have to set the CSP header in Etherpad itself rather than let users set the CSP header in their reverse proxy. (This is because only Etherpad knows the nonce value.)
  * We should migrate away from inline JavaScript anyway because it's harder to maintain.

SamTV12345:
Already solved with the latest 2.2.x versions

PR #4402, opened 2020-10-07
CSP: Disable the indexCustomInlineScripts hook

rhansen:
Addresses #4401.

cc @brunob

### DIFF
--- a/doc/api/hooks_server-side.md
+++ b/doc/api/hooks_server-side.md
@@ -118,7 +118,6 @@ Available blocks in `pad.html` are:
  * `indexCustomStyles` - contains the `index.css` `<link>` tag, allows you to add your own or to customize the one provided by the active skin
  * `indexWrapper` - contains the form for creating new pads
  * `indexCustomScripts` - contains the `index.js` `<script>` tag, allows you to add your own or to customize the one provided by the active skin
- * `indexCustomInlineScripts` - contains the inline `<script>` of home page, allows you to customize `go2Name()`, `go2Random()` or `randomPadName()` functions
 
 ## padInitToolbar
 Called from: src/node/hooks/express/specialpages.js
--- a/src/static/js/pluginfw/shared.js
+++ b/src/static/js/pluginfw/shared.js
@@ -1,6 +1,13 @@
 var _ = require("underscore");
 var defs = require('./plugin_defs');
 
+const disabledHookReasons = {
+  hooks: {
+    indexCustomInlineScripts: 'The hook makes it impossible to use a Content Security Policy ' +
+        'that prohibits inline code. Permitting inline code makes XSS vulnerabilities more likely',
+  },
+};
+
 function loadFn(path, hookName) {
   var functionName
     , parts = path.split(":");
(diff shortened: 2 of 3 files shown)


JohnMcLear:
I am not opposed to this but we need to look at impact (# of plugins / usage) prior to merging.  It is inevitable though so best we notify devs ASAP.  That's a me job unless you want to do it?

JohnMcLear:
Some changes required.

1. Don't remove hook.
1. issue log warning IF inline JS is detected.  References to other JS files here should be fine too.
1. include some comment in HTML saying don't use inline here!  Also include same warning in hook docs.
1. Promote usage of including JS files IE ``<script src=...`` in this hook.

That way it wont be a breaking change but will gently move people away from doing inline JS.  Win-win

rhansen:
(moving discussion in issue #4401 here)

> but what alternative would we have to modify the index inline script after that?

A future commit will move index's inline script to a `.js` file. Perhaps I should squash that into this so that plugins can start migrating from `indexCustomInlineScripts` to `indexCustomScripts`.

> Okay so why remove it? Just ensure the docs say to people to only reference scripts here

The `indexCustomInlineScripts` EJS block is inside `<script>...</script>`, so users can't do `<script src="..."></script>` with this hook—they would have to use the `indexCustomScripts` hook instead.

If we leave the `<script>` block in `index.html` unchanged, then a CSP that prohibits inline code will break the index page. We could move the existing inline code to a separate `.js` to empty the inline `<script>` block except for the `indexCustomInlineScripts` hook, but that would break plugins that want to replace (not augment) the existing code.

> we need to look at impact (# of plugins / usage) prior to merging

The plugin is new (added in May), which I hope means there aren't many users. Do we have a good way to audit existing uses?

rhansen:
I just had an idea that would allow us to remain backward compatible: Instead of moving `index.html`'s inline script to `static/js/index.js`, I could move it to `template/index.js` and turn it into an EJS template that still has the `indexCustomInlineScripts` block. I would have to set up an Express router for that template, but that's not difficult. What do you think?

brunob:
> A future commit will move index's inline script to a .js file. Perhaps I should squash that into this so that plugins can start migrating from indexCustomInlineScripts to indexCustomScripts.

:+1: 

> The plugin is new (added in May), which I hope means there aren't many users. Do we have a good way to audit existing uses?

:+1: too since this hook became public only after 1.8.5 release 29days ago i may be the only one using it :p

JohnMcLear:
+1.  This is a new hook and an easy one for us to habdle first.  A lot of other inline js is added by plugins that interact with the pad.  This isn't a huge task but it's a good day or so of modifying plugins.

rhansen:
@brunob I sent you a merge request to update your plugin: https://framagit.org/infini/ep_infini/-/merge_requests/1

brunob:
@rhansen woot & thx for that and sorry for the delay, i'll test that asap !

rhansen:
Friendly ping @brunob

brunob:
@rhansen tested and i confirm it works well, thx for that i'll merge the PR from our side on next release.

JohnMcLear:
@rhansen as only this plugin used the hook I am +1 merging this.  Did you want to do any final polish before I hit go?

rhansen:
>  Did you want to do any final polish before I hit go?

Feel free to merge any time.

brunob:
@rhansen ftr your PR on our plugin/theme doesn't work with latest etherpad release, maybe my first test was a fail...

If i had an unbind on original script events, our index.js file override the original script well, see :

```js
$(() => {
  $('#go2Name').unbind('submit');
  $('#button').unbind('click');
});
```

I'll may try to make it with `removeEventListener` instead.

Issue #4344, opened 2020-09-24
Discard wd in favour of webdriver.io or similar


webzwo0i:
I probably did not open a ticket for this, yet. We use https://github.com/admc/wd that only supports old Selenium (legacy version 2).

This has impact on our Saucelabs tests, among others:
- we cannot get console output
- we cannot get performance metrics

It must be guaranteed that we can still launch the frontend tests in a local browser after migration.

JohnMcLear:
+1 yes please! :+1:

webzwo0i:
I had some luck today with https://www.npmjs.com/package/selenium-webdriver 
so it will definitely happen in the near future... :-)

I am not sure if this will help with test stability, but at least we'll know, that it's not related to selenium 2 and we'll have console.log

Issue #4274, opened 2020-09-09
responsiveness.js breaks FF52 


JohnMcLear:
Test needs uncommenting then debugging on FF52.

PR #4408, opened 2020-10-10
refactor some frontend tests

webzwo0i:
- include test for #4389 
- don't include sendkeys in index.html as it's included in helper.init
- don't force the usage of callbacks in `it`, so we can use async
- mocha opts: add default timeout and replace ignoreLeaks with checkLeaks as the former is deprecated

refactored chat.js, timeslider_labels, timeslider_numeric_padID and timeslider_follow

Should I issue single PRs only with one test file at a time?

### DIFF
--- a/tests/frontend/helper.js
+++ b/tests/frontend/helper.js
@@ -145,6 +145,24 @@ var helper = {};
         helper.padOuter$.fx.off = true;
         helper.padInner$.fx.off = true;
 
+        /*
+         * chat messages received
+         * @type {Array}
+         */
+        helper.chatMessages = [];
+        
+        /*
+         * changeset commits from the server
+         * @type {Array}
+         */
+        helper.commits = [];
+        
+        /*
+         * userInfo messages from the server
+         * @type {Array}
+         */
+        helper.userInfos = [];
+
         // listen for server messages
         helper.spyOnSocketIO();
         opts.cb();
--- a/tests/frontend/helper.js
+++ b/tests/frontend/helper.js
@@ -145,6 +145,8 @@ var helper = {};
         helper.padOuter$.fx.off = true;
         helper.padInner$.fx.off = true;
 
+        // listen for server messages
+        helper.spyOnSocketIO();
         opts.cb();
       }).fail(function(){
         if (helper.retry > 3) {
--- a/tests/frontend/helper.js
+++ b/tests/frontend/helper.js
@@ -220,6 +220,8 @@ var helper = {};
   /**
    * Same as `waitFor` but using Promises
    *
+   * @returns {Promise}
+   *
    */
   helper.waitForPromise = async function(...args) {
     // Note: waitFor() has a strange API: On timeout it rejects, but it also throws an uncatchable
--- a/tests/frontend/helper/methods.js
+++ b/tests/frontend/helper/methods.js
@@ -60,9 +60,7 @@ helper.linesDiv = function(){
  * @returns {Array.<string>} lines of text
  */
 helper.textLines = function(){
-  return helper.padInner$('.ace-line').map(function(){
-    return $(this).text()
-  }).get()
+  return helper.linesDiv().map((div) => div.text());
 }
 
 /**
--- a/tests/frontend/helper/methods.js
+++ b/tests/frontend/helper/methods.js
@@ -51,7 +51,7 @@ helper.edit = async function(message, line){
 helper.linesDiv = function(){
   return helper.padInner$('.ace-line').map(function(){
     return $(this)
-  })
+  }).get()
 }
 
 /**
--- a/tests/frontend/helper/methods.js
+++ b/tests/frontend/helper/methods.js
@@ -74,6 +74,12 @@ helper.defaultText = function(){
 
 /**
  * Sends a chat `message` via `sendKeys`
+ * You *must* include `{enter}` at the end of the string or it will
+ * just fill the input field but not send the message.
+ *
+ * @example
+ *
+ * `helper.sendChatMessage('hi{enter}')`
  *
  * @param {string} message the chat message to be sent
  * @returns {Promise}
--- a/tests/frontend/helper/methods.js
+++ b/tests/frontend/helper/methods.js
@@ -91,3 +91,99 @@ helper.textLines = function(){
 helper.defaultText = function(){
   return helper.padChrome$.window.clientVars.collab_client_vars.initialAttributedText.text;
 }
+
+/**
+ * Sends a chat `message` via `sendKeys`
+ *
+ * @param {string} message the chat message to be sent
+ * @returns {Promise}
+ */
+helper.sendChatMessage = function(message){
+  let noOfChatMessages = helper.chatMessages.length;
+  helper.padChrome$("#chatinput").sendkeys(message)
+  return helper.waitForPromise(function(){
+    return noOfChatMessages + 1 === helper.chatMessages.length;
+  })
+}
+
+/**
+ * Opens the settings menu if its hidden via button
+ *
+ * @returns {Promise}
+ */
+helper.showSettings = function() {
+  if(!helper.isSettingsShown()){
+    helper.settingsButton().click()
+    return helper.waitForPromise(function(){return helper.isSettingsShown(); },2000);
+  }
+}
+
+/**
+ * Hide the settings menu if its open via button
+ *
+ * @returns {Promise}
+ * @todo untested
+ */
+helper.hideSettings = function() {
+  if(helper.isSettingsShown()){
+    helper.settingsButton().click()
+    return helper.waitForPromise(function(){return !helper.isSettingsShown(); },2000);
+  }
+}
+
+/**
+ * Makes the chat window sticky via settings menu if the settings menu is
+ * open and sticky button is not checked
+ *
+ * @returns {Promise}
+ */
+helper.enableStickyChatviaSettings = function() {
+  var stickyChat = helper.padChrome$('#options-stickychat');
+  if(helper.isSettingsShown() && !stickyChat.is(':checked')) {
+    stickyChat.click();
+    return helper.waitForPromise(function(){
+      return helper.isChatboxSticky();
+    },2000);
+  }
+}
+
+/**
+ * Unsticks the chat window via settings menu if the settings menu is open
+ * and sticky button is checked
+ *
+ * @returns {Promise}
+ */
+helper.disableStickyChatviaSettings = function() {
+  var stickyChat = helper.padChrome$('#options-stickychat');
+  if(helper.isSettingsShown() && stickyChat.is(':checked')) {
+    stickyChat.click();
+    return helper.waitForPromise(function(){return !helper.isChatboxSticky()},2000);
+  }
+}
+
+/**
+ * Makes the chat window sticky via an icon on the top right of the chat
+ * window
+ *
+ * @returns {Promise}
+ */
+helper.enableStickyChatviaIcon = function() {
+  var stickyChat = helper.padChrome$('#titlesticky');
+  if(helper.isChatboxShown() && !helper.isChatboxSticky()) {
+    stickyChat.click();
+    return helper.waitForPromise(function(){return helper.isChatboxSticky()},2000);
+  }
+}
+
+/**
+ * Disables the stickyness of the chat window via an icon on the
+ * upper right
+ *
+ * @returns {Promise}
+ */
+helper.disableStickyChatviaIcon = function() {
+  if(helper.isChatboxShown() && helper.isChatboxSticky()) {
+    helper.titlecross().click()
+    return helper.waitForPromise(function(){return !helper.isChatboxSticky()},2000);
+  }
+}
--- a/tests/frontend/helper/methods.js
+++ b/tests/frontend/helper/methods.js
@@ -187,3 +187,41 @@ helper.disableStickyChatviaIcon = function() {
     return helper.waitForPromise(function(){return !helper.isChatboxSticky()},2000);
   }
 }
+
+/**
+ * Sets the src-attribute of the main iframe to the timeslider
+ * In case a revision is given, sets the timeslider to this specific revision
+ * It waits until the timer is filled with date and time
+ *
+ * @param {number} [revision] the optional revision
+ * @returns {Promise}
+ * @todo for some reason this does only work the first time, you cannot
+ * goto rev 0 and then via the same method to rev 5. Use buttons instead
+ */
+helper.gotoTimeslider = function(revision){
+  revision = Number.isInteger(revision) ? '#'+revision : '';
+  var iframe = $('#iframe-container iframe');
+  iframe.attr('src', iframe.attr('src')+'/timeslider' + revision);
+
+  return helper.waitForPromise(function(){
+    return helper.timesliderTimerTime()
+      && !Number.isNaN(new Date(helper.timesliderTimerTime()).getTime()) },5000);
+}
+
+/**
+ * Clicks in the timeslider at a specific offset
+ * It's used to navigate the timeslider
+ *
+ * @todo no mousemove test
+ * @param {number} X coordinate
+ */
+helper.sliderClick = function(X){
+  let sliderBar = helper.sliderBar()
+  let edown = new jQuery.Event('mousedown');
+  let eup = new jQuery.Event('mouseup');
+  edown.clientX = eup.clientX = X;
+  edown.clientY = eup.clientY = sliderBar.offset().top;
+
+  sliderBar.trigger(edown);
+  sliderBar.trigger(eup);
+}
--- a/tests/frontend/helper/methods.js
+++ b/tests/frontend/helper/methods.js
@@ -1,0 +1,93 @@
+/*
+ * chat messages received
+ * @type {Array}
+ */
+helper.chatMessages = [];
+
+/*
+ * changeset commits from the server
+ * @type {Array}
+ */
+helper.commits = [];
+
+/*
+ * userInfo messages from the server
+ * @type {Array}
+ */
+helper.userInfos = [];
+
+/**
+ * Spys on socket.io messages and saves them into several arrays
+ * that are visible in tests
+ */
+helper.spyOnSocketIO = function (){
+  helper.contentWindow().pad.socket.on('message', function(msg){
+    if (msg.type == "COLLABROOM") {
+
+      if (msg.data.type == 'ACCEPT_COMMIT') {
+        helper.commits.push(msg);
+      }
+      else if (msg.data.type == 'USER_NEWINFO') {
+        helper.userInfos.push(msg)
+      }
+      else if (msg.data.type == 'CHAT_MESSAGE') {
+        helper.chatMessages.push(msg)
+      }
+    }
+  })
+}
+
+/**
+ * Makes an edit via `sendkeys` to the position of the caret and ensures ACCEPT_COMMIT
+ * is returned by the server
+ * It does not check if the ACCEPT_COMMIT is the edit sent, though
+ * If `line` is not given, the edit goes to line no. 1
+ *
+ * @param {string} message The edit to make - can be anything supported by `sendkeys`
+ * @param {number} [line] the optional line to make the edit on starting from 1
+ * @returns {Promise}
+ * @todo needs to support writing to a specified position
+ *
+ */
+helper.edit = async function(message, line){
+  let editsNr = helper.commits.length;
+  line = line ? line - 1 : 0;
+  helper.linesDiv()[line].sendkeys(message);
+  return helper.waitForPromise(function(){
+    return editsNr + 1 === helper.commits.length;
+  })
+}
+
+/**
+ * The pad text as an array of divs
+ *
+ * @example
+ * helper.linesDiv()[2].sendkeys('abc') // sends abc to the third line
+ *
+ * @returns {Array.<HTMLElement>} array of divs
+ */
+helper.linesDiv = function(){
+  return helper.padInner$('.ace-line').map(function(){
+    return $(this)
+  })
+}
+
+/**
+ * The pad text as an array of lines
+ *
+ * @returns {Array.<string>} lines of text
+ */
+helper.textLines = function(){
+  return helper.padInner$('.ace-line').map(function(){
+    return $(this).text()
+  }).get()
+}
+
+/**
+ * The default pad text transmitted via `clientVars`
+ *
+ * @returns {string}
+ */
+helper.defaultText = function(){
+  return helper.padChrome$.window.clientVars.collab_client_vars.initialAttributedText.text;
+}
--- a/tests/frontend/helper/methods.js
+++ b/tests/frontend/helper/methods.js
@@ -28,15 +28,15 @@ helper.spyOnSocketIO = function (){
  * @param {string} message The edit to make - can be anything supported by `sendkeys`
  * @param {number} [line] the optional line to make the edit on starting from 1
  * @returns {Promise}
- * @todo needs to support writing to a specified position
+ * @todo needs to support writing to a specified caret position
  *
  */
 helper.edit = async function(message, line){
-  let editsNr = helper.commits.length;
+  let editsNum = helper.commits.length;
   line = line ? line - 1 : 0;
   helper.linesDiv()[line].sendkeys(message);
   return helper.waitForPromise(function(){
-    return editsNr + 1 === helper.commits.length;
+    return editsNum + 1 === helper.commits.length;
   })
 }
 
@@ -56,6 +56,7 @@ helper.linesDiv = function(){
 
 /**
  * The pad text as an array of lines
+ * For lines in timeslider use `helper.timesliderTextLines()`
  *
  * @returns {Array.<string>} lines of text
  */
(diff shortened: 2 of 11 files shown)


JohnMcLear:
Looks good to me, @rhansen any thoughts?

rhansen:
I'll take a look. I'm not familiar with this part of the Etherpad codebase so it will take a bit of studying before I can provide useful feedback.

tests/frontend/helper.js
```diff
@@ -197,6 +217,38 @@ var helper = {};
     return deferred;
   }
 
+  /**
+   * Same as `waitFor` but using Promises
+   *
+   */
+  helper.waitForPromise = function(conditionFunc, _timeoutTime, _intervalTime){
```
rhansen, line 224: `helper.waitFor` already returns a "thenable" (a [Deferred](https://api.jquery.com/category/deferred-object/) object), so callers can already do `await helper.waitFor(...)`, `helper.waitFor(...).then(...)`, etc.

If you need something that doesn't allow the caller to control the Deferred object's state, you can simplify this function:

```javascript
  helper.waitForPromise = function(conditionFunc, _timeoutTime, _intervalTime) {
    return this.waitFor(conditionFunc, _timeoutTime, _intervalTime)).promise();
  }
```

The above returns a jQuery Promise object. If you need a standard Promise object for some reason, or if you want to improve readability, you can do this:

```javascript
  helper.waitForPromise = async function(conditionFunc, _timeoutTime, _intervalTime) {
    return await this.waitFor(conditionFunc, _timeoutTime, _intervalTime);
  }
```

tests/frontend/specs/chat.js
```diff
@@ -2,134 +2,86 @@ describe("Chat messages and UI", function(){
...
-    }
+    expect(failed).to.be(true);
 
-    // due to animation, we need to make some timeout...
-    setTimeout(function() {
-      //check if chat changed to get the stickychat Class
-      var $chatbox = chrome$("#chatbox");
-      var hasStickyChatClass = $chatbox.hasClass("stickyChat");
-      expect(hasStickyChatClass).to.be(true);
+    await helper.sendChatMessage(`${chatValue}{enter}`); // simulate a keypress of typing mluto and enter (to send 'mluto')
```
rhansen, line 41: This and later lines are now unnecessary due to the way you restructured the test.

tests/frontend/specs/timeslider_numeric_padID.js
```diff
@@ -4,64 +4,26 @@ describe("timeslider", function(){
...
+    await helper.waitForPromise(function(){return helper.contentWindow().location.hash === '#1'})
 
-      setTimeout(function() {
-        var timeslider$ = $('#iframe-container iframe')[0].contentWindow.$;
-        var $sliderBar = timeslider$('#ui-slider-bar');
+    // expect URI to be similar to
+    // http://192.168.1.48:9001/p/2/1/export/html
+    // http://192.168.1.48:9001/p/735773577399/1/export/html
+    let rev1ExportLink = helper.contentWindow().$('#exporthtmla').attr('href');
+    expect(rev1ExportLink.indexOf('/1/export/html')).to.not.be(-1);
```
rhansen, line 21:
```suggestion
    expect(rev1ExportLink).to.have.string('/1/export/html');
```

tests/frontend/specs/timeslider_numeric_padID.js
```diff
@@ -4,64 +4,26 @@ describe("timeslider", function(){
...
-          var exportLink = timeslider$('#exporthtmla').attr('href');
-          var checkVal = padId + "/0/export/html";
-          var includesCorrectURI = exportLink.indexOf(checkVal);
-          expect(includesCorrectURI).to.not.be(-1);
-          done();
-        }, 400);
-      }, 2000);
-    }, 2000);
+    let rev0ExportLink = helper.contentWindow().$('#exporthtmla').attr('href');
+    expect(rev0ExportLink.indexOf('/0/export/html')).to.not.be(-1);
```
rhansen, line 27:
```suggestion
    expect(rev0ExportLink).to.have.string('/0/export/html');
```

tests/frontend/specs/timeslider_labels.js
```diff
@@ -2,63 +2,61 @@ describe("timeslider", function(){
...
+    let timerTimeLast = new Date(helper.timesliderTimerTime()).getTime();
+
+    // the day of this revision, e.g. August 12, 2020 (stripped the string "Saved")
+    let dateLast = new Date(helper.revisionDateElem().substr(6)).getTime();
+
+    // the label/revision, ie Version 3
+    let labelLast = helper.revisionLabelElem().text();
+
+    // the datetime should be a date
+    expect( Number.isNaN(timerTimeLast)).to.eql(false);
```
rhansen, line 31:
```suggestion
    expect(timerTimeLast).to.be.finite;
```

tests/frontend/specs/timeslider_labels.js
```diff
@@ -2,63 +2,61 @@ describe("timeslider", function(){
...
+
+    // the Date object of the day should not be NaN
+    expect( Number.isNaN(dateLast) ).to.eql(false)
```
rhansen, line 34:
```suggestion
    expect(dateLast).to.be.finite;
```

tests/frontend/specs/timeslider_labels.js
```diff
@@ -2,63 +2,61 @@ describe("timeslider", function(){
...
+
+    // the label should match Version `Number`
+    expect( labelLast.indexOf(`Version ${revs}`) ).to.not.be(-1);
```
rhansen, line 37:
```suggestion
    expect(labelLast).to.have.string(`Version ${revs}`);
```

tests/frontend/specs/timeslider_labels.js
```diff
@@ -2,63 +2,61 @@ describe("timeslider", function(){
...
+
+    // Click somewhere left on the timeslider to go to revision 0
+    helper.sliderClick(30);
```
rhansen, line 40: Because there are three revisions, I would expect that clicking at 30 would go to revision 1. To be safe, maybe click more to the left:

```suggestion
    helper.sliderClick(1);
```

tests/frontend/specs/timeslider_labels.js
```diff
@@ -2,63 +2,61 @@ describe("timeslider", function(){
...
+    let timerTime = new Date(helper.timesliderTimerTime()).getTime();
+
+    // the day of this revision, e.g. August 12, 2020
+    let date = new Date(helper.revisionDateElem().substr(6)).getTime();
+
+    // the label/revision, e.g. Version 0
+    let label = helper.revisionLabelElem().text();
+
+    // the datetime should be a date
+    expect( Number.isNaN(timerTime)).to.eql(false);
```
rhansen, line 52:
```suggestion
    expect(timerTime).to.be.finite;
```

tests/frontend/specs/timeslider_labels.js
```diff
@@ -2,63 +2,61 @@ describe("timeslider", function(){
...
+    // the last revision should be newer or have the same time
+    expect(timerTimeLast - timerTime >= 0);
```
rhansen, line 54:
```suggestion
    expect(timerTimeLast).to.be.at.least(timerTime);
```

tests/frontend/specs/timeslider_labels.js
```diff
@@ -2,63 +2,61 @@ describe("timeslider", function(){
...
+
+    // the Date object of the day should not be NaN
+    expect( Number.isNaN(date) ).to.eql(false)
```
rhansen, line 57:
```suggestion
    expect(date).to.be.finite;
```

tests/frontend/specs/timeslider_labels.js
```diff
@@ -2,63 +2,61 @@ describe("timeslider", function(){
...
+
+    // the label should match Version 0
+    expect( label ).to.match( /Version 0/);
```
rhansen, line 60:
```suggestion
    expect(label).to.have.string('Version 0');
```

tests/frontend/specs/timeslider_follow.js
```diff
@@ -1,55 +1,69 @@
...
+    // set to follow contents as it arrives
+    helper.contentWindow().$('#options-followContents').prop("checked", true);
+    helper.contentWindow().$('#playpause_button_icon').click();
 
-        // set to follow contents as it arrives
-        timeslider$('#options-followContents').prop("checked", true);
+    let newTop;
+    return helper.waitForPromise(function(){
+      newTop = helper.contentWindow().$('#innerdocbody').offset();
+      return newTop.top < originalTop.top;
```
rhansen, line 28: Won't this only wait for a single detected advancement?

Also, if playback is considerably faster than the poll period then this won't detect the intermediate advancements.

tests/frontend/helper.js
webzwo0i: Thanks for the input! Trying your last two options the tests in 
https://github.com/ether/etherpad-lite/pull/4408/files#diff-4c39b0646494534de0cbe94479400dabR190
https://github.com/ether/etherpad-lite/pull/4408/files#diff-4c39b0646494534de0cbe94479400dabR201
https://github.com/ether/etherpad-lite/pull/4408/files#diff-4c39b0646494534de0cbe94479400dabR226
fail.

I will try again later, but in case you see a problem in the helper.js tests, it would definitely help :) I don't want to have a massive amount of try-catch

tests/frontend/specs/timeslider_follow.js
```diff
@@ -1,55 +1,69 @@
...
+    let rev1text = "\xa0"; // timeslider will add &nbsp;
+    let rev2text = 'Test line'
 
-        setTimeout(function() {
-          //make sure the text has changed
-          var newTop = timeslider$('#innerdocbody').offset();
-          expect( originalTop ).not.to.eql( newTop );
-          done();
-        }, 1000);
+    helper.padInner$('body').empty();
```
rhansen, line 37: Why empty it?

tests/frontend/specs/timeslider_follow.js
```diff
@@ -1,55 +1,69 @@
...
+    await helper.waitForPromise(function(){
+      return helper.commits.length === 1; 
+    })
```
rhansen, line 40: I'm a bit confused by this. What is creating the single commit? (Or does a new pad start with a single commit?)

tests/frontend/specs/timeslider_follow.js
```diff
@@ -1,55 +1,69 @@
...
+    await helper.edit(rev2text)
```
rhansen, line 41: It's unclear to me what this does. Does it:

  * add the text to the beginning of the existing line 1?
  * add the text to the end of the existing line 1?
  * replace the existing line 1?
  * insert a new line before the existing line 1?

tests/frontend/specs/timeslider_follow.js
```diff
@@ -1,55 +1,69 @@
...
-    }, revs*timePerRev);
-  });
+    await helper.gotoTimeslider();
+
+    // set to follow contents as it arrives
+    helper.contentWindow().$('#options-followContents').prop("checked", true);
+
+    helper.contentWindow().$('#leftstep').click();
+    await helper.waitForPromise(function(){
+      return helper.timesliderTextLines()[0] === rev1text;
```
rhansen, line 50: How does this test whether the browser is following the edits? It looks like it's only checking that the contents match.

tests/frontend/specs/timeslider_follow.js
```diff
@@ -1,55 +1,69 @@
...
+    let newTop;
+    return helper.waitForPromise(function(){
+      newTop = helper.contentWindow().$('#innerdocbody').offset();
+      return newTop.top < originalTop.top;
+    })
+  });
 
-        var originalTop = timeslider$('#innerdocbody').offset();
-        timeslider$('#playpause_button_icon').click();
+  it("only to lines that exist in the current pad view #4389", async function(){
```
rhansen, line 32: What do you mean by "lines that exist"? How is it testing that?

rhansen (approved):
Looks good! I left a bunch of comments, most of which are for minor nits. Don't feel obligated to address them all.

tests/frontend/helper.js
rhansen: Hmm, interesting. I'll take a look.

rhansen:
Oops, I didn't realize we were using expect.js, not chai. Ignore most of my comments. :smile:

tests/frontend/helper.js
rhansen: OK, I figured it out: `helper.waitFor()` will throw an uncatchable exception unless you register a callback via the `.fail()` method. Please see the new commits I pushed.

tests/frontend/specs/chat.js
```diff
@@ -2,134 +2,86 @@ describe("Chat messages and UI", function(){
...
-    var $settingsButton = chrome$(".buttonicon-settings");
-    $settingsButton.click();
+    await helper.showChat();
 
-    //get the chat selector
-    var $stickychatCheckbox = chrome$("#options-stickychat");
+    let failed = false;
+    // simulate a keypress of enter (to send an empty message)
+    await helper.sendChatMessage("{enter}")
+      .catch(function(){failed = true});
```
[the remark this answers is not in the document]
rhansen: Nevermind—we're not using Chai.

tests/frontend/specs/timeslider_labels.js
rhansen: Nevermind—I thought we were using Chai, not expect.js.

tests/frontend/specs/timeslider_labels.js
rhansen: Nevermind—I thought we were using Chai, not expect.js.

tests/frontend/specs/timeslider_labels.js
rhansen: Nevermind—I thought we were using Chai, not expect.js.

tests/frontend/specs/timeslider_labels.js
rhansen: Nevermind—I thought we were using Chai, not expect.js.

tests/frontend/specs/timeslider_follow.js
webzwo0i: This introduces a commit of "clear the pad". Probably the same can be done with `sendkeys('{selectall}{del}')`.
Will look into it.

tests/frontend/specs/helper.js
```diff
@@ -339,4 +400,66 @@ describe("the test helper", function(){
...
+      })
+    })
+
+    it(".edit() defaults to send an edit to the first line", async function(){
+      let firstLine = helper.textLines()[0];
+      await helper.edit("line")
+      expect(helper.textLines()[0]).to.be(`line${firstLine}`);
+    })
+
+    it(".edit() to the line specified with parameter lineNo", async function(){
```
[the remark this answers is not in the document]
webzwo0i: Throws. The edit method is brittle right now, because it adds content where the caret is, without taking the caret position into account via parameter, but for the current tests it's probably enough.

tests/frontend/specs/helper.js
```diff
@@ -339,4 +400,66 @@ describe("the test helper", function(){
...
+    before(function(cb){
+      helper.newPad(function(){
+        cb();
+      })
+    })
+
+    it(".textLines() returns the text of the pad as strings", async function(){
+      let lines = helper.textLines();
+      let defaultText = helper.defaultText();
+      expect(lines).to.be.an('object');
```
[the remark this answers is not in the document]
webzwo0i: will still be an object even with `.get()`

tests/frontend/specs/helper.js
```diff
@@ -339,4 +400,66 @@ describe("the test helper", function(){
...
+      // @todo
+      // final "\n" is added automatically, but my understanding is this should happen
+      // only when the default text does not end with "\n" already
+      expect(lines.join("\n")+"\n").to.equal(defaultText);
+    })
+
+    it(".linesDiv() returns the text of the pad as div elements", async function(){
+      let lines = helper.linesDiv();
+      let defaultText = helper.defaultText();
+      expect(lines).to.be.an('object');
```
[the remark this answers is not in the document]
webzwo0i: will still be an object even with .get()

tests/frontend/helper/methods.js
```diff
@@ -0,0 +1,222 @@
...
+ * The default pad text transmitted via `clientVars`
+ *
+ * @returns {string}
+ */
+helper.defaultText = function(){
+  return helper.padChrome$.window.clientVars.collab_client_vars.initialAttributedText.text;
+}
+
+/**
+ * Sends a chat `message` via `sendKeys`
```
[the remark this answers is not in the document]
webzwo0i: no

tests/frontend/specs/timeslider_numeric_padID.js
webzwo0i: will do

tests/frontend/specs/timeslider_numeric_padID.js
webzwo0i: will do

tests/frontend/specs/timeslider_labels.js
webzwo0i: will do

tests/frontend/specs/timeslider_labels.js
webzwo0i: afaik I tested it and it turned out to be not so safe as it should be (maybe the calculations are wrong and it actually clicks outside?) will test again

tests/frontend/specs/timeslider_labels.js
webzwo0i: will do

tests/frontend/specs/timeslider_follow.js
webzwo0i: `$('body').empty()`

webzwo0i:
Thanks @rhansen !!! 
I will go through the rest of your review soon.

@JohnMcLear 
I need to do some last tests here, so don't merge it yet.

JohnMcLear:
Noted.

tests/frontend/specs/helper.js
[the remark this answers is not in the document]
rhansen: True, but `to.be.an('array')` ensures that it's not just any kind of object—it must be an Array. That would catch bugs like forgetting to call `.get()` on the jQuery object.

tests/frontend/specs/timeslider_follow.js
rhansen: `sendkeys('{selectall}{del}')` would make it easier to understand the test, so that would be my preference if it works.

JohnMcLear:
@webzwo0i after resolve/rebase is this ready to go?

rhansen:
I just rebased the branch, though AFAIK @webzwo0i still wants to make some changes.

JohnMcLear:
@webzwo0i what's the status buddy?

tests/frontend/specs/timeslider_follow.js
webzwo0i: I couldn't get this to work with sendkeys and opted for `helper.selectLines()` to select everything and clear it with an delete keystroke

tests/frontend/specs/timeslider_follow.js
webzwo0i: OK. Yes this is not clear. It will put the string into the pad, depending on the caret position and depending on the `lineNum` parameter (defaults to the first line).
It is enough for the tests here but I need to investigate how the other tests use sendkeys to find out what should be supported when "editing" a pad.

tests/frontend/specs/timeslider_follow.js
webzwo0i: The test is intended to catch a special bug:
at rev 0 delete all content
When you are in the timeslider at rev 1 (empty pad with a single new line) going to rev 0, the changeset will be something like: insert a lot of characters, including X new lines (depending on how many new lines are in the default text).

The current behaviour was to get no. of lines from the changeset and theffollow logic was: at rev 1, go to line no. X and apply the changeset to get rev 0. This obviously failed because there was no "line X" at rev 1 and the lines of revision 0 were never displayed. That's why I check the line contents.

But you are right, it's in the timeslider_follow suite and should include a test for following the edits. I extended the test case to cover that

tests/frontend/specs/timeslider_follow.js
webzwo0i: See above

tests/frontend/specs/chat.js
```diff
@@ -2,134 +2,86 @@ describe("Chat messages and UI", function(){
...
-    //click on the settings button to make settings visible
-    var $settingsButton = chrome$(".buttonicon-settings");
-    $settingsButton.click();
+    await helper.showChat();
 
-    //get the chat selector
-    var $stickychatCheckbox = chrome$("#options-stickychat");
+    let failed = false;
+    // simulate a keypress of enter (to send an empty message)
+    await helper.sendChatMessage("{enter}")
```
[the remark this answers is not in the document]
webzwo0i: Nice catch! Done.

tests/frontend/specs/helper.js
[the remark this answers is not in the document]
webzwo0i: checking with Array.isArray now

tests/frontend/specs/timeslider_follow.js
```diff
@@ -1,55 +1,118 @@
...
 
+/**
+ * @param {number} lineNum 
+ * @returns {boolean} scrolled to the lineOffset?
+ */
+function hasFollowedToLine(lineNum) {
+  let scrollPosition = helper.contentWindow().$("#editorcontainerbox")[0].scrollTop;
+  let lineOffset = helper.contentWindow().$('#innerdocbody').find(`div:nth-child(${lineNum})`)[0].offsetTop;
+
+  return Math.abs(scrollPosition - lineOffset) < 1;
```
rhansen, line 116: I think this will only be true if the number of lines of content in the pad is at least `lineNum` more than can be displayed in the visible area at once. The test is only using 3 lines of content, which is probably too few to make a scrollbar appear let alone be able to scroll down to the top of the 3rd line.

webzwo0i: Oh. I always have the scrollbar displayed when in timeslider. Can you verify? (tested on android/chromium/firefox)

webzwo0i: I don't understand CSS (or whatever makes this possible) enough to be able to find the reason why, but the scrollbar always has some "space"/"room" downwards. so for me it can be scrolled even on pads with no/few lines.

Is there something special to my setup?!

rhansen: It looks like the extra blank room at the bottom is equal to the height of the timeslider toolbar. I also don't know enough about CSS to know why that is. As long as the offset to the top of the third line is less than the height of the timeslider toolbar then I think it'll work. However, the test might be fragile and hard to fix if it depends on the height of the toolbar. I'm not sure how to make it better.

JohnMcLear:
@webzwo0i any chance we can wrap this up today?  I'm really keen to get tests wrapped up so I can more easily merge the work from other contributors.

webzwo0i:
I finished it up, run travis multiple times and flipped the expects to see if the code runs :-) There is still some flakiness regarding pad reload/timeouts with different tests (that are not touched here) for no apparent reason (I hope to fix that with https://github.com/ether/etherpad-lite/issues/4344).

One last thing is the responsivness test which is not working. I am with you, that disabling is bad practice (at least skipping it, so we see that it's not running), but we have an issue open https://github.com/ether/etherpad-lite/issues/4274

I figured out what this test originally did: the scheduling of the event handler for keyup was changed and this test basically checks if there is a delay between two consecutive events that is caused by the editor incorporating the changes. This probably can be done only with firing two keyup events without keydown/keypress instead of hundreds. Also this test probably never actually inserted any strings (there is no indication that the line no. count has any impact on performance) or ensure the editor has finished incorporating the big amount of data+author coloring before triggering the events. Imo the events are largely no-ops.

JohnMcLear:
> . Also this test probably never actually inserted any strings (there is no indication that the line no. count has any impact on performance)

I'm pretty sure it did.  That was the point of the test no? :dango:

webzwo0i:
https://github.com/ether/etherpad-lite/pull/1833/files#diff-0787f7a473ec90ce36e88cfbc59daef5e4c5d848e036b491ebe4dba9b8b135d3R66 was the original test and it immediately returned without checking if new content has been added. (it was only fixed recently with 53b80d628029991879b82dd87ade0b89360f1553 but that check also does not work, because when the condition in https://github.com/ether/etherpad-lite/blob/develop/tests/frontend/specs/responsiveness.js#L57 is true, than the later check if the length is greater than amount is also immediately true).

So if you okay with that I would delay re-enabling responsivness test until I have time to work on https://github.com/ether/etherpad-lite/issues/4274

I need to get the import/export/contentcollector stuff ready before I can focus on that, but I certainly will!

JohnMcLear:
Ack, all good :)  Tnx!

webzwo0i:
seeing an error with timeslider_follow in https://travis-ci.org/github/ether/etherpad-lite/jobs/737990904#L2279
investigating now

PR #4410, opened 2020-10-11
CSP: Move `index.html` inline code to separate `.js` file

rhansen:
Addresses #4401.

### DIFF
--- a/src/static/js/index.js
+++ b/src/static/js/index.js
@@ -1,0 +1,61 @@
+// @license magnet:?xt=urn:btih:8e4f440f4c65981c5bf93c76d35135ba5064d8b7&dn=apache-2.0.txt Apache-2.0
+/**
+ * Copyright 2011 Peter Martischka, Primary Technology.
+ * Copyright 2020 Richard Hansen
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


rhansen:
@brunob Before you merge https://framagit.org/infini/ep_infini/-/merge_requests/1 would you test it against this PR? I think it'll still work but I'm not 100% sure because both attempt to set the `onsubit`/`onclick` handlers. We might have to change https://framagit.org/infini/ep_infini/-/merge_requests/1 to use `eejsBlock_indexWrapper` instead.

JohnMcLear:
https://github.com/ether/etherpad-lite/pull/4410/files#diff-b7ed03066e63b832507c09098fed57d4L186 I think is relatively important or was like 3 years ago.  Did this never get wide adoption?   https://www.gnu.org/software/librejs/free-your-javascript.html for details.

If it isn't worth including I don't mind it being dropped.

rhansen:
> https://github.com/ether/etherpad-lite/pull/4410/files#diff-b7ed03066e63b832507c09098fed57d4L186 I think is relatively important or was like 3 years ago.

I added back the annotation.

> Did this never get wide adoption? https://www.gnu.org/software/librejs/free-your-javascript.html for details.

This is the first time I've heard about it.

I just tried that plugin. It is quite terrible:

* It blocks unknown scripts by default, so sites that haven't yet added the annotations (which is most sites) will break.
* There is no way to put it in a report-only mode. If you whitelist a site (so that it still functions) you can't tell what scripts would have been blocked had the site not been whitelisted.
* There's no way to tell it to whitelist everything.
* Annotating scripts is not very dev-friendly. (The poor documentation is only partly to blame.)

I applaud the goal, but it'll never achieve wide adoption unless they change the default behavior to not block.

brunob:
@rhansen just tested it and the overriding of the handlers doesn't work anymore.

rhansen:
Bummer. Thanks for testing @brunob. I'll revise the GitLab merge request and get back to you.

PR #4421, opened 2020-10-19
better follow contents

webzwo0i:
There is a small edge case regarding the follow pad contents functionality:
Currently the changeset that is applied to get the next revision is parsed and the lines property of the first opcode is used to find the line where the change happens.

Given:
pad content rev 0: [default Pad text]
pad content rev 1: [''] (delete everything)
pad content rev 2: ['Test line','',''] (insert "Test line\n\n"
pad content rev 3: ['Test line','','Another test line'] (insert "Another test line" at line 3)

When you go from revision 2 to 1 (changeset: `Z:c<b|2-b$`) the lines property is 2.
When you go from revision 0 to 1 (changeset: `Z:bk<bj|7-bj$`) the lines property is 7.

I think it would be better if we'd scroll to line 1 in both cases. Probably it's only relevant for the "delete everything" edge case, but I need to know to which line we should scroll for https://github.com/ether/etherpad-lite/pull/4408

This PR uses alines to get the first line that has changed length/attributes and will scroll to that line. Only when all alines before/after are the same, the opcode's lines property is used (someone replaces a character that she edited before, so alines before/after are the same)

It's okay if this PR is not merged, because it's probably an edge case. Just let me know so I can adjust the values in https://github.com/ether/etherpad-lite/pull/4408

### DIFF
--- a/src/static/js/broadcast.js
+++ b/src/static/js/broadcast.js
@@ -245,6 +245,7 @@ function loadBroadcastJS(socket, sendSocketMsg, fireWhenAllScriptsAreLoaded, Bro
       BroadcastSlider.setSliderPosition(revision);
     }
 
+    let oldAlines = padContents.alines.slice();
     try
     {
       // must mutate attribution lines before text lines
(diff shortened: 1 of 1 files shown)


webzwo0i:
Ha, after sending the PR I realize: Comparing two arrays is slower than parsing the changeset... So if you agree, that it can only happen when clearing the pad we should probably prefer the current code. :-D

JohnMcLear:
So should we use this PR or??  :D  Sorry but I don't understand the context of "current" code, do you mean current in develop or current in this PR? :dango:

webzwo0i:
:) Current in develop. But let's leave this open until evening (I want to focus on contentcollector during the day) and I will do a final test if it has got performance implications. (Test probably should be: create a pad with thousands of lines and have a hundred edits at the bottom of it, hit the play button in the timeslider right before those hundred edits happen and stop the time. If it's one or two seconds slower compared to current develop it's okay, but if it's 10 seconds it's probably not)

webzwo0i:
I created a pad with ~8.5k empty lines and ~640 revisions at the bottom and hit the play button. With this PR, for every revision, it should compare the strings in every single of those 8.5k lines. I ran the test multiple times and the overall number of revisions that have been processed are included below:

Chromium 83.0.4103.116-1~deb10u3 (CPU usage nearly 100%):
develop: 578022 ms / 3772 revisions = 153 ms per revision
this PR: 521020 ms / 3134 revisions = 166 ms per revision

Firefox 81.0-2 (CPU usage ~40%):
develop: 231744 ms / 1914 revisions = 121 ms per revision
this PR: 232672 ms / 1914 revisions = 121 ms per revision

Conclusion: Chromium needs way more CPU power while being significant slower and this PR has a negative impact of about ~10%. In Firefox there is no performance degradation.

I think this PR is okay.

PR #4423, opened 2020-10-21
WIP: Cookies: Cookie namespace - Note that this is potentially a breaking change

JohnMcLear:
Just prefixes ``session_sid`` with ``ep_`` to stop cookie namespace collisions.

Note that the scope of this change is limited to only one cookie, not all cookies.  All cookies may be preferable for some use cases.

This could break some instance where..

1. Reverse proxy expects ``express_sid`` value
1. Site already uses ``express_sid`` and continuation of session after update is expected to continue (user will be required to auth again).  This might be a non-issue but I haven't tested.


rhansen (approved):
I rebased onto latest `develop`. LGTM.

JohnMcLear:
Closing as stale.

PR #4427, opened 2020-10-22
tests: relaxed constraint on the number of interval triggers

webzwo0i:
it errored in https://travis-ci.org/github/ether/ep_table_of_contents/jobs/738010629#L2313

Not optimal but CI is probably not super accurate when running in a shared environment


webzwo0i:
it's failing on a frontend test that normally works... Should be safe to merge.

Issue #4417, opened 2020-10-16
Cannot find commit when git has detached HEAD


andrewda:
**Describe the bug**
When git has a detached HEAD (e.g. `git checkout tags/1.8.6`), Etherpad logs the following warning:

```
[2020-10-16 17:40:07.271] [WARN] console - Can't get git version for server header
ENOENT: no such file or directory, open '/opt/etherpad-lite/.git/962b615cf971234ca86edf8b887ddac132d'
```

**To reproduce**
1. Checkout a tag: `git checkout tags/1.8.6`
2. Run Etherpad: `./bin/run.sh`

**Expected behavior**
The latest git commit should be found, even when there's a detached HEAD. Currently, Etherpad is checking `.git/HEAD` for a ref, and using that ref to get the commit:

https://github.com/ether/etherpad-lite/blob/1dba121aff4526f4d7392ee18014195ad9daeebc/src/node/utils/Settings.js#L425-L428

However, when there's a detached HEAD, `.git/HEAD` contains the commit we're looking for.

PR #4431, opened 2020-10-22
utils: Fix commit checking on detached HEAD

andrewda:
Closes https://github.com/ether/etherpad-lite/issues/4417

This makes it possible for Etherpad to find the latest commit when the HEAD is detached. Not sure what the right way to go about testing this is, or if tests are needed.

### DIFF
--- a/src/node/utils/Settings.js
+++ b/src/node/utils/Settings.js
@@ -423,8 +423,12 @@ exports.getGitCommit = function() {
       rootPath += '/.git';
     }
     var ref = fs.readFileSync(rootPath + "/HEAD", "utf-8");
-    var refPath = rootPath + "/" + ref.substring(5, ref.indexOf("\n"));
-    version = fs.readFileSync(refPath, "utf-8");
+    if (ref.startsWith("ref: ")) {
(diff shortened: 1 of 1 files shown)


rhansen:
Thank you for the fix!

One way to improve it further would be to run `git rev-parse --short HEAD` and use its output (see [`child_process.execFile`](https://nodejs.org/api/child_process.html#child_process_child_process_execfile_file_args_options_callback)). Do you have the time to make that change? If not we can merge this as-is.

(Don't worry about tests.)

andrewda:
Sure, I can do that! I think that would also require adding `git` as a dependency -- currently the Docker container doesn't have it installed, and I'm not sure if that would break any backwards-compatibility.

rhansen:
> Sure, I can do that! I think that would also require adding `git` as a dependency -- currently the Docker container doesn't have it installed

Good point. Much of the `.git` directory is in `.dockerignore` so running `git rev-parse` inside the container might not work anyway. I think what you've got now is probably the best approach.

An alternative is to set up a clean/smudge filter that writes the Git revision to a file. That would require users to do some one-time setup after cloning the repo (e.g., run a script) which is annoying, though it would make it possible to exclude all of `.git` from the Docker image.

PR #4434, opened 2020-10-25
pad: Don't paste when middle-clicking on a link

rhansen:


JohnMcLear:
Nice!  Didn't know about auxclick

Issue #4422, opened 2020-10-20
Alt F9 is broken


JohnMcLear:
As per accessibility docs alt f9 should focus on toolbar and left/right keys should nav through.

CSS is missing for the focus and left/right doesn't work because the focus is on a button in an a href in a li..  Afaik the focus would need to be on the list item to be able to left/right toggle through??

Bumped @seballot because he's a ninja at these types of things :1st_place_medal:

rhansen:
I did some testing and it looks like it is actually working (Firefox 81):
* Select some text
* Hit Alt-F9 (there is no visual indicator that anything happened, other than the cursor might disappear)
* Hit Tab to move to the italic button (no visual indicator)
* Hit space to activate the italic button. (With the colibris skin there's no visual indicator other than the text becoming italic. With the no-skin skin the italic button becomes darker.)

So I think the problems are:
* No visual indicator showing that a button has focus.
* No visual indicator in colibris showing that a button is in the toggled state (bold on, italic on, etc.).

JohnMcLear:
Left and right arrow keys should nav through without tab key no?

rhansen:
Left and right don't work. Only tab.

JohnMcLear:
Wiki docs say Left/right arrow keys should work, is that only if a specific browser setting is enabled?

rhansen:
¯\_(ツ)_/¯

seballot:
Hi ! I made the css fix for bouton to be outline when focused

> No visual indicator in colibris showing that a button is in the toggled state (bold on, italic on, etc.).

There is a visual indicator, but it's light :)

![image](https://user-images.githubusercontent.com/17404254/97104076-31797d00-16b1-11eb-827b-2a367c9a45d9.png)

JohnMcLear:
I'm closing this with the hope that accessibility is okay w/ tab

PR #4436, opened 2020-10-25
Display outline when toolbar button focused refs #4422

seballot:
Refs #4422


JohnMcLear:
Awesome work thanks!!

Issue #4378, opened 2020-09-30
API: listAll... failed


Joggel72:
**Describe the bug**
I've tried to use the listAll... methods.
Bull methods received the following response:
_{"code":2,"message":"createHTTPError.notFound is not a constructor","data":null}_

There is no entry in the log for these requests.
Other requests are working.

**To Reproduce**
Using curl:
_curl "http://localhost:9001/api/1/listAllGroups?apikey=477d..." _

**Expected behavior**
code: 0 and list of elements

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Win10, Debian
- Version 1.8.4 (6a0f73d)
 - API-Version 1.2.14

JohnMcLear:
Does it work on develop?

Joggel72:
In the log I can see this:
 Etherpad is running in Development mode.

JohnMcLear:
Please try develop branch

Joggel72:
I've tried to use it, but start failed.

I've get the zip from https://github.com/ether/etherpad-lite/archive/develop.zip unzipped the actuell version in a new directory. Then called in directory src 
_npm install_
called _start.bat_
and got
C:\tmpWork\etherpad-lite-develop>node node_modules\ep_etherpad-lite\node\server.js
internal/modules/cjs/loader.js:969
  throw err;
  ^

Error: Cannot find module 'C:\tmpWork\etherpad-lite-develop\node_modules\ep_etherpad-lite\node\server.js'
←[90m    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:966:15)←[39m
←[90m    at Function.Module._load (internal/modules/cjs/loader.js:842:27)←[39m
←[90m    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)←[39m
←[90m    at internal/main/run_main_module.js:17:47←[39m {
  code: ←[32m'MODULE_NOT_FOUND'←[39m,
  requireStack: []
}

Where are the configuration files for this version?

webzwo0i:
bin/installOnWindows.bat before start.bat

YSelfTool:
@Joggel72 listAllGroups is not available in API version 1. You need to pass version >= 1.1, just pass 1.2.14 (the current one):
`curl …/api/1.2.14/listAllGroups?…`, then it will work.
@JohnMcLear Something is wrong with the import of http-errors, leading to this "notFound is not a constructor" error when calling "new createHTTPError.notFound".

[...]

PR #4437, opened 2020-10-25
Use correct constructor for 404,501 error handlers

anttiviljami:
Fixes error message mentioned in #4378, upgrades to `http-errors@1.8.0`, utilise new isHttpError utility introduced by the new version of the package.


JohnMcLear:
Awesome work thanks!!

PR #4440, opened 2020-10-26
WIP: ace.js: fix the logic to inline CSS

webzwo0i:
ace.js contains logic to inline CSS. When ace.js is delivered, Minify.js parses ace.js and searches for `$$INCLUDE_CSS` strings and appends the content of those files to the end of ace.js. When the client loads & executes ace.js to build the iframes, it uses those CSS to create style/link tags.

That logic broke when custom skins where introduced some time ago. Also the advent of the workaround using version strings for cache busting prevents it from working. This PR fixes the code to parse the `$$INCLUDE_CSS` lines again, so that some HTTP requests are saved. atm only the `iframe_editor.css` file gets inlined, because it does not contain a version string. (Actually it's included twice, because there is no code to remove multiple inlines of the same file - that's also fixed in this PR.)

It's WIP because there are two issues:
1. `src/static/css/pad.css` and `src/static/skins/skinName/pad.css` are currently included in the outer and inner iframes + in pad.html for the parent document of those two iframes. Is this necessary? I ask because even after this PR there are two HTTP requests to those files from coming from pad.html. My solution would be to add code to ace.js that also adds CSS to the parent document.
2. For some reason I don't know yet, the paths for fonts are wrong now. (They should be something like `../../../static/font/...` but the requests go to `/font/...`. Would you agree, when I embed the fonts directly into CSS using Base64?

It probably only affects performance on first load,

### DIFF
--- a/src/node/utils/Minify.js
+++ b/src/node/utils/Minify.js
@@ -1,6 +1,7 @@
 /**
- * This Module manages all /minified/* requests. It controls the
- * minification && compression of Javascript and CSS.
+ * @file
+ * This Module manages all /static/* requests. It controls the
+ * minification and compression of Javascript and CSS.
  */
 
 /*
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Cc @seballot

JohnMcLear:
Looking good @webzwo0i , keep it up :+1:

PR #4453, opened 2020-10-30
Speed up and fix travis

webzwo0i:
- set default exit code of the frontend tests (remote_runner.js) to 1. In case saucelabs runner aren't started (limit of parallel worker) the callback of the worker that is pushed to `async#queue` won't be called. This results in `queue#drain` to never be called and the script just silently exits with a default exit code (cf https://travis-ci.com/github/ether/ep_comments/jobs/413945558)

- Increase the "no messages received" timeout of travis from 10 to 15 minutes. (frontend tests would start to exceed the limit very soon) - I did not see builds that hang for more than ten minutes recently, but if this would happen again and 15 minutes delay for a hanging test would be too much, we certainly can lower it again.

- libreoffice is only used during backend tests at the moment. instead of installing it on every job just do this for the 2 backend jobs. also cleanup travis hooks that are overwritten by job specific ones 

- fix flaky helper test: unfortunately setInterval seems flaky on saucelabs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ jobs:
         - "bin/installDeps.sh"
         - "export GIT_HASH=$(git rev-parse --verify --short HEAD)"
       script:
-        - "travis_wait 20 tests/frontend/travis/runner.sh"
+        - "travis_wait 15 tests/frontend/travis/runner.sh"
     - name: "Run the Backend tests without Plugins"
       install:
         - "sudo add-apt-repository -y ppa:libreoffice/ppa"
@@ -77,7 +77,7 @@ jobs:
         - *install_plugins
         - "export GIT_HASH=$(git rev-parse --verify --short HEAD)"
       script:
-        - "travis_wait 20 tests/frontend/travis/runner.sh"
+        - "travis_wait 15 tests/frontend/travis/runner.sh"
     - name: "Lint test package-lock.json"
       install:
         - "npm install lockfile-lint"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -53,9 +39,13 @@ jobs:
...
         - "export GIT_HASH=$(git rev-parse --verify --short HEAD)"
       script:
-        - "tests/frontend/travis/runner.sh"
+        - "travis_wait 15 tests/frontend/travis/runner.sh"
     - name: "Run the Backend tests without Plugins"
       install:
+        - "sudo add-apt-repository -y ppa:libreoffice/ppa"
+        - "sudo apt-get update"
+        - "sudo apt-get -y install libreoffice"
+        - "sudo apt-get -y install libreoffice-pdfimport"
```
rhansen, line 48: Maybe factor this out to avoid duplication?  Something like this:

```yaml
_install_libreoffice: &install_libreoffice >-
  sudo add-apt-repository -y ppa:libreoffice/ppa &&
  sudo apt-get update &&
  sudo apt-get -y install libreoffice libreoffice-pdfimport

jobs:
  include:
    - name: "Run the Backend tests without Plugins"
      install:
        - *install_libreoffice
        - "bin/installDeps.sh"
        # ...
```

rhansen (approved):
LGTM.

.travis.yml
JohnMcLear: I decided to merge prior to making this change, at least we can get some runs in today and then make this change when things are stable

PR #4455, opened 2020-10-31
WIP: fix truncated travis log

webzwo0i:
Travis log is truncated on nearly every frontend job. I observed it for a long time on about 1/10 of the builds, but now it seems to happen on nearly every build that runs the frontend tests.

The problem is probably related to the way travis parses the logs to remove secure environment vars - but I am not 100% sure about this. (The assumption is based on https://github.com/travis-ci/travis-ci/issues/6018#issuecomment-370023238 )
I asked on https://travis-ci.community/t/build-log-truncated/10395 and maybe someone from Travis will show us a way, but until then I think I found a workaround - the remote_runner's log from the frontend tests is now redirected to a file and in an after_script hook I output that file at a rate of 100 lines/sec.

The exit code of the jobs is not changed by adding an after_script hook so it should be save. It worked two times now, while before it was truncated 10 times in a row.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
         - "export GIT_HASH=$(git rev-parse --verify --short HEAD)"
       script:
         - "travis_wait 15 tests/frontend/travis/runner.sh"
-      #output 300 lines at once to prevent travis from truncating the log
+      #output 100 lines at once to prevent travis from truncating the log
       after_script:
         - "cat /tmp/remote_runner.log | (lines=0; while read line; do echo -e \"${line}\"; if [ $((++lines % 100)) -eq 0 ]; then sleep 1; fi; done)"
     - name: "Run the Backend tests without Plugins"
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ jobs:
         - "travis_wait 15 tests/frontend/travis/runner.sh"
       #output 100 lines at once to prevent travis from truncating the log
       after_script:
-        - "cat /tmp/remote_runner.log | (lines=0; while read line; do echo -e \"${line}\"; if [ $((++lines % 100)) -eq 0 ]; then sleep 1; fi; done)"
+        - "travis_fold end after_script && cat /tmp/remote_runner.log | (lines=0; while read line; do echo -e \"${line}\"; if [ $((++lines % 100)) -eq 0 ]; then sleep 1; fi; done)"
     - name: "Run the Backend tests without Plugins"
       install:
         - "sudo add-apt-repository -y ppa:libreoffice/ppa"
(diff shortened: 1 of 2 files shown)


webzwo0i:
The reason it's inside after_script was, that when I included it directly into the remote_runner script it did not work.
Unfortunately user experience is worse now, because you cannot ctrl+f "failed" to see the failing tests, but need to unfold the after_script hook's messages at the bottom.

Including it into the script commands would change the exit code...

rhansen:
Doesn't this make it impossible to follow progress as it's running?

> (The assumption is based on [travis-ci/travis-ci#6018 (comment)](https://github.com/travis-ci/travis-ci/issues/6018#issuecomment-370023238) )

That was quite a while ago; has the rollout completed?

If the problem is process termination before the output has a chance to be filtered and written then adding a sleep at the end should probably fix it. If the problem is writing too fast then we should be able to slow down the output. Can you try those first?

webzwo0i:
> Doesn't this make it impossible to follow progress as it's running?

When running the tests in a local environment, runner.sh/remote_runner.js is not touched and the progress should be always available. The current behaviour when running in CI was that after a worker (ie. all the tests for a specific browser) finishes, the logs from that worker are sent back to Travis at once. So there is always a delay before you can see the progress on Travis. However, that delay is now as long as the time it takes to run the slowest worker.

Probably it's useful to watch the progress. The side effect would be that the console outputs are outputted slow enough for Travis to not truncate the log. But the remote_runner.js has to be modified to constantly log (at the moment, it waits until finished) and to not log the same output multiple times.

So don't merge yet, I will try to go that road and it should make this PR unnecessary.

> > (The assumption is based on [travis-ci/travis-ci#6018 (comment)](https://github.com/travis-ci/travis-ci/issues/6018#issuecomment-370023238) )
> 
> That was quite a while ago; has the rollout completed?

Probably not. At the bottom of the issue is another comment from July. However, I am not sure why slowing down the logging actually helps :-)

> If the problem is process termination before the output has a chance to be filtered and written then adding a sleep at the end should probably fix it. If the problem is writing too fast then we should be able to slow down the output. Can you try those first?

I tried a lot of different approaches (not touching the exit code, while adding sleeps at different places), the one in this PR is the first that seems to consistently work. However, I only tried sleeping in the shell scripts or in .travis.yml, but I did not try to delay the callback in remote_runner.js with `setTimeout(function(){callback()}, 5000)`. If the bug is only triggered, when the script that *produces* the output exits too fast, then it would explain, why adding sleep outside of that script does not have any effect.

I like the approach to log the test results as they happen, so I will try that

webzwo0i:
I resolved the folding issue and it now automatically unfolds the output so it is easy to ctrl+f "failed" to get failing tests.
Downside still is that a log is only seen after *all* browsers finish.

As an alternative I made https://github.com/ether/etherpad-lite/compare/fix-truncation-in-travis
This basically prints the log of all browsers every five seconds. Advantage is you can watch the tests as they are run without going to saucelabs' website. Another advantage is that you have some kind of correlation between the tests and etherpad's log. (You can more or less correlate the test that caused an error/warning in etherpad's log - it's not perfect though. Making it perfect, e.g. correlate every test to specific output in etherpad's log, would be to hard for now.)

The downside is that the logs from different browsers are mixed, which make them a little bit harder to read.

https://travis-ci.com/github/ether/etherpad-lite/jobs/428017252 is an example of the log output of the second option. (There is an issue where empty lines are printed sometimes, ignore that for now. If you choose that option I'm gonna fix it up)

So please choose what you like most. I tend to prefer the second option

JohnMcLear:
I don't mind waiting as long as it's accurate and truthful.  Faster is better and having them out of order is no biggy as long as you can tell which browser it is somehow...

webzwo0i:
okay then I close this. PR with logging every 5 seconds is at https://github.com/ether/etherpad-lite/pull/4479

PR #4458, opened 2020-11-01
remove unnecessary conditional statement

gabrielmcoelho:

remove unnecessary conditional statement

if animationState evaluates to -1 or 0, it would end up in a conditional statement that assign animationState's value to itself. Since this is redundant, it is better to remove this conditional, to avoid an extra check

> Screenshots/GIFs


rhansen:
Thank you!

PR #4462, opened 2020-11-02
Remove `src/static/js/jquery.js` from `.gitignore`

### LINKED ISSUE #4460: Error: The module at "ep_etherpad-lite/static/js/jquery" does not exist.
**Describe the bug**
After signing in and creating a pad, I see the following error when trying to load the newly-created pad:
Screen Shot 2020-11-01 at 8 05 22 AM

**To Reproduce**
I've done a fresh build of Etherpad latest 1.8.6

 - OS: Mac 
 - Browser Chrome
 - Version 78

**Additional context**
Steps to build were to create a docker-compose file and a run script, set the source files as the context, make a custom settings.json, and run `docker-compose build`. That all seems to work, but when loading a pad the above error occurs.

rhansen:
That file is checked in so there's no reason to ignore it.

Fixes #4460.

cc @mcblum

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@ var/dirty.db
 bin/convertSettings.json
 *~
 *.patch
-src/static/js/jquery.js
 npm-debug.log
 *.DS_Store
 .ep_initialized


JohnMcLear:
I was thinking if there is some proper method to get jQuery instead of writing file to SRC.  Something like npm install of it...

mcblum:
Hey thanks! Appreciate your looking at this so quickly.

PR #4464, opened 2020-11-02
Use `contentEditable` for all browsers

### LINKED ISSUE #4456: Support Ctrl (or Shift) + click on link to open it in a background tab (or a new window)
* * *

name: Feature request
title: Support Ctrl + click on link to open it in a background tab
labels: enhancement
assignees: 

* * *

I use pads for working collaboratively on lists of links. I tend to open several links at once in background tabs, then go through them. But opening links from Etherpad in a background tab only works via right click – if the context menu isn't disabled (it is e.g. on board.net).

I'd like links to behave like standard links. In particular, Ctrl + click on a link should open it in a new background tab, and Shift + click should open it in a new window (by default, in most browsers).

Sidenote: Etherpad should not try to emulate Ctrl/Shift + click but delegate the handling to the browser. There are browser settings and extensions that modify link interaction, and those should still work as expected. Effectively, links in pads should be as close to plain, non-scripted `<a target="_blank" href="...">` tags as possible.

**Plugin?**
A previous issue on various other link-related things but also including this (#3533) was closed with a mention of plugins, so I guess it might be possible. However, I believe core Etherpad should mirror the standard behaviour of links throughout the web and only *non-standard* behaviour would be plugin-worthy.

rhansen:
This makes it possible to disable `contentEditable` for certain elements in some circumstances (e.g., on links so that users can click on them normally).

Addresses (but does not fix) #4456.

Also some cleanups (**multiple commits**):
* Delete commented-out code
* Replace `setClassPresence(x, ...)` with `x.classList.toggle(...)`

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -1103,10 +1103,6 @@ function Ace2Inner(){
 
   var idleWorkTimer = makeIdleAction(function()
   {
-
-    //if (! top.BEFORE) top.BEFORE = [];
-    //top.BEFORE.push(magicdom.root.dom.innerHTML);
-    //if (! isEditable) return; // and don't reschedule
     if (inInternationalComposition)
     {
       // don't do idle input incorporation during international input composition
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Nice to remove a lot of legacy cruft!

PR #4465, opened 2020-11-03
editor: throw error and dont hang browser if selection is no longer present

JohnMcLear:
Details at: https://github.com/ether/ep_comments/issues/133

Currently it semi-gracefully handles things, I think core should be more responsible tho and throw a gritter msg if a plugin tries to modify a non existing selection...

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -2245,25 +2245,14 @@ function Ace2Inner(){
 
     // Failing to add attribute to non present selection, see
     // https://github.com/ether/ep_comments/issues/133 for details.
-    var notPresent = false;
     try{
       var line = rep.lines.atIndex(rep.selEnd[0]);
     }catch(e){
-      notPresent = true;
+      throw new Error("Failed to add attribute to selection that does not exist");
     }
-    if(notPresent){
-      parent.parent.$.gritter.add({
-        // (string | mandatory) the heading of the notification
-        title: 'Selection no longer present',
-        // (string | mandatory) the text inside the notification
-        text: 'Unable to modify this content because the selection is no longer present in the pad',
-        // (bool | optional) if you want it to fade out on its own or just $
-        sticky: false,
-        // (int | optional) the time you want it to be alive for before fad$
-        time: '4000'
-      });
-      throw new Error("Failed to add attribute to selection");
-      return;
+    // we know the line exists so let's see if the character exists
+    if(rep.selEnd[1] > rep.lines.atIndex(rep.selEnd[0]).width){
+      throw new Error("Failed to add attribute to selection that does not exist");
     }
 
     documentAttributeManager.setAttributesOnRange(rep.selStart, rep.selEnd, [
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -2240,7 +2240,31 @@ function Ace2Inner(){
 
   function setAttributeOnSelection(attributeName, attributeValue)
   {
+    // if there isn't a selection
     if (!(rep.selStart && rep.selEnd)) return;
+
+    // Failing to add attribute to non present selection, see
+    // https://github.com/ether/ep_comments/issues/133 for details.
+    var notPresent = false;
+    try{
+      var line = rep.lines.atIndex(rep.selEnd[0]);
+    }catch(e){
+      notPresent = true;
+    }
+    if(notPresent){
+      parent.parent.$.gritter.add({
+        // (string | mandatory) the heading of the notification
+        title: 'Selection no longer present',
+        // (string | mandatory) the text inside the notification
+        text: 'Unable to modify this content because the selection is no longer present in the pad',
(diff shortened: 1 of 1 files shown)


src/static/js/ace2_inner.js
```diff
@@ -2240,8 +2240,21 @@ function Ace2Inner(){
...
+    // if there isn't a selection
     if (!(rep.selStart && rep.selEnd)) return;
 
+    // Failing to add attribute to non present selection, see
+    // https://github.com/ether/ep_comments/issues/133 for details.
+    try{
+      rep.lines.atIndex(rep.selEnd[0]);
+    }catch(e){
+      throw new Error("Failed to add attribute to selection that does not exist");
+    }
```
rhansen, line 2252: The skip list code is not well written so I don't 100% understand what's going on, but wouldn't it be equivalent to do this instead:

```suggestion
    if (rep.selEnd[0] >= rep.lines.length()) throw new Error("Selection end is out of bounds");
```

rhansen:
Two questions:

* How does this solve the bug? This change adds exceptions, but weren't errors already being thrown?
* Wouldn't it be better to update the selection when the pad text is updated?

JohnMcLear:
Before browser would hang completely.  Now we reject the change and throw an error.

JohnMcLear:
Bump @rhansen

rhansen:
> Bump @rhansen

Apologies, slipped through.

> You can't update rep if no content exists.

I'm not sure what you mean. Suppose the state looks like this:

>> the [quick brown] fox jumped over the lazy dog

where "[" and "]" indicate the start and end of the selection. Then another user deletes "quick brown". What happens? Does `rep.selEnd` get updated so that it looks like this?

>> the [] fox jumped over the lazy dog

Or is `rep.selEnd` left alone so that the state looks like this?

>> the [ fox jumped] over the lazy dog

If it's the former, then I'm not sure why anything needs to be updated in Etherpad core. If it's the latter, then shouldn't we fix it to be the former?

JohnMcLear:
Imagine you have 

``hello world`` and do rep.selStart [0,0], [0,10] that would be ``hello worl``

that would be fine.

but if the text is modified under you to ``hello`` and you do rep.selStart [0,0], [0,10] it would fail as 0,10 would no longer exist...

Only plugins can make this error, the editor never can (magically).

This PR stops plugins from hanging the browser but instead fires an error.  It's because of ep_comments having an additional dialogue that may try end content that no longer exists..

rhansen:
If the text is modified to `hello`, shouldn't Etherpad (core) automatically update `rep.selEnd` to [0, 5]?

JohnMcLear:
It should, but the time rep is captured is when the user clicks on the editbar button afaik, lemme check that tho.

JohnMcLear:
```

// Create form to add comment
EpComments.prototype.createNewCommentFormIfDontExist = function (rep) {
  const data = this.getCommentData();

  // If a new comment box doesn't already exist, create one
  data.changeFrom = parseMultiline(this.getSelectedText(rep));
  newComment.insertNewCommentPopupIfDontExist(data, (comment, index) => {
    if (comment.changeTo) {
      data.comment.changeFrom = comment.changeFrom;
      data.comment.changeTo = comment.changeTo;
    }
    data.comment.text = comment.text;

    this.saveComment(data, rep);
  });
};
```

```
// Save comment
EpComments.prototype.saveComment = function (data, rep) {
  this.socket.emit('addComment', data, (commentId, comment) => {
    comment.commentId = commentId;

    this.ace.callWithAce((ace) => {
      ace.ace_performSelectionChange(rep.selStart, rep.selEnd, true);
      ace.ace_setAttributeOnSelection('comment', commentId);
    }, 'insertComment', true);

    this.setComment(commentId, comment);
    this.collectComments();
  });
};
```

It looks like rep is passed from when the comment popup form is created..  So yeah, I guess we could make it get rep at time of setting attribute, is that available in context at that point?  I wonder if we can just do ``ace_getRep()`` or something..

rhansen:
>  So yeah, I guess we could make it get rep at time of setting attribute, is that available in context at that point? I wonder if we can just do `ace_getRep()` or something..

I think that's worth trying.

Adding sanity checks to Etherpad core (what this PR does) is also worth doing. Did you see my suggested change?

src/static/js/ace2_inner.js
JohnMcLear: It's been a while but afaik this was my first approach and it didn't work..  Perhaps it is because it makes the assumption that en entire line has been removed where if you had..
```
a
bc
```

and only c is removed it would not work?  I'd need to check.  Will put some time into this tomorrow.

rhansen: `rep.selEnd[0]` is the line number and `rep.selEnd[1]` is the column number, so your example of deleting `c` shouldn't affect the `rep.selEnd[0]` check.

JohnMcLear:
Fixed in https://github.com/ether/ep_comments/commit/86e89a3b44676bb7b6b0d2e80abc28a2c9674c69

PR #4470, opened 2020-11-04
Fix/empty lines

joassouza:
This PR fixes the bug of losing spaces when importing HTML. 
This is a particular problem when the line exported has line attributes.

### DIFF
--- a/src/node/utils/ImportHtml.js
+++ b/src/node/utils/ImportHtml.js
@@ -14,24 +14,18 @@
  * limitations under the License.
  */
 
-const log4js            = require('log4js');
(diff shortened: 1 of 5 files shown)


webzwo0i:
Thanks!
Can you add an example/test for the case that there are line attributes?
Maybe this should be in contentcollector, so that copy&paste is also handled?

I noticed two things, are they intentional? When importing html, I get a lot of &nbsp; at the beginning, probably because of tags without text at the beginning of the html file. Maybe better start at the body tag?

And the second is -  this PR fixes one additional edge case: the space in `<em>good </em>idea` is no longer swallowed, which is super, but it inserts an &nbsp; there, when I would expect a normal space?

PR #4471, opened 2020-11-05
bugfix/translations: Fix missing arg handling in html10n.js

ilmartyrk:
When translation expects arguments to be replaced, but the arguments are missing big red error message is displayed. I believe that that kind of translation errors should be only visible in the console just as errors when a value is missing for the current translation key in the users current selected language. Also added style update to gritter.css as long error messages where overflowing the error popup and hard to read and close by the user:
overflowin_error

### DIFF
--- a/src/static/css/pad/gritter.css
+++ b/src/static/css/pad/gritter.css
@@ -31,6 +31,8 @@
 .gritter-item .gritter-content {
   flex: 1 auto;
   text-align: center;
+  width: 95%;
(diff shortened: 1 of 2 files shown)


JohnMcLear:
This looks mergable, jus twaiting on test results.

Also please see my other comments on other issues/prs I have for you <3

ilmartyrk:
@JohnMcLear thanks, sorry for long delays for my answers, will now check

PR #4491, opened 2020-11-16
Fix gzip not triggered for packages

webzwo0i:
A mime type mismatch, too.

The responseCache caches all headers, including `Content-Encoding`. In case a client does not set `Accept-Encoding: gzip` while another client requested the resource before and had this header set, the resource would be delivered uncompressed, but the `Content-Encoding: gzip` header would be present.

I think this is wrong, so now the `Content-Encoding` header gets deleted in that case.

### DIFF
--- a/src/node/utils/caching_middleware.js
+++ b/src/node/utils/caching_middleware.js
@@ -196,7 +196,7 @@ CachingMiddleware.prototype = new function () {
         var statusCode = responseCache[cacheKey].statusCode;
 
         var pathStr = CACHE_DIR + 'minified_' + cacheKey;
-        if (supportsGzip && (headers['content-type'] || '').match(/^text\//)) {
+        if (supportsGzip && (headers['content-type'] || '').match(/^application\/javascript/)) {
           pathStr = pathStr + '.gz';
           headers['content-encoding'] = 'gzip';
         }
--- a/src/node/utils/caching_middleware.js
+++ b/src/node/utils/caching_middleware.js
@@ -199,6 +199,9 @@ CachingMiddleware.prototype = new function () {
         if (supportsGzip && (headers['content-type'] || '').match(/^application\/javascript/)) {
           pathStr = pathStr + '.gz';
           headers['content-encoding'] = 'gzip';
+        } else {
+          // ensure responseCache is updated
+          delete headers['content-encoding'];
         }
 
         var lastModified = (headers['last-modified']
--- a/src/node/utils/caching_middleware.js
+++ b/src/node/utils/caching_middleware.js
@@ -20,7 +20,6 @@ var fs = require('fs');
 var path = require('path');
 var zlib = require('zlib');
 var settings = require('./Settings');
-var semver = require('semver');
 var existsSync = require('./path_exists');
 
 /*
@@ -192,16 +191,14 @@ CachingMiddleware.prototype = new function () {
         res.write = old_res.write || res.write;
         res.end = old_res.end || res.end;
 
-        var headers = responseCache[cacheKey].headers;
+        let headers = {};
+        Object.assign(headers, (responseCache[cacheKey].headers || {}));
         var statusCode = responseCache[cacheKey].statusCode;
 
         var pathStr = CACHE_DIR + 'minified_' + cacheKey;
         if (supportsGzip && /^application\/javascript/.test(headers['content-type'])) {
           pathStr = pathStr + '.gz';
           headers['content-encoding'] = 'gzip';
-        } else {
-          // ensure responseCache is updated
-          delete headers['content-encoding'];
         }
 
         var lastModified = (headers['last-modified']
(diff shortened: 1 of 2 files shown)


src/node/utils/caching_middleware.js
```diff
@@ -196,9 +196,12 @@ CachingMiddleware.prototype = new function () {
         var statusCode = responseCache[cacheKey].statusCode;
 
         var pathStr = CACHE_DIR + 'minified_' + cacheKey;
-        if (supportsGzip && (headers['content-type'] || '').match(/^text\//)) {
+        if (supportsGzip && (headers['content-type'] || '').match(/^application\/javascript/)) {
           pathStr = pathStr + '.gz';
           headers['content-encoding'] = 'gzip';
+        } else {
+          // ensure responseCache is updated
+          delete headers['content-encoding'];
```
rhansen, line 204: Rather than clear it here, wouldn't it be better to not store the header in the cache to begin with? I think that would make the code easier to understand.

src/node/utils/caching_middleware.js
```diff
@@ -196,7 +196,7 @@ CachingMiddleware.prototype = new function () {
         var statusCode = responseCache[cacheKey].statusCode;
 
         var pathStr = CACHE_DIR + 'minified_' + cacheKey;
-        if (supportsGzip && (headers['content-type'] || '').match(/^text\//)) {
+        if (supportsGzip && (headers['content-type'] || '').match(/^application\/javascript/)) {
```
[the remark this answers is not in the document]
webzwo0i: thanks! I changed it to a normal string comparison, because content-type allows only a single media type, so a regex shouldn't be necessary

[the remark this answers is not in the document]
webzwo0i: Oh, lol. Of course it's something like `application/javascript; charset=utf-8`...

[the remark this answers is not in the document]
webzwo0i: That shouldn't be necessary, because the code path is only taken when requesting packages or javascript files from plugins (paths that start with `/javascripts/`). Am I missing something?

src/node/utils/caching_middleware.js
webzwo0i: I tried it in https://github.com/ether/etherpad-lite/pull/4491/commits/142cf0b6301e7bbfd275b4fb00f9d351afe02350

src/node/utils/caching_middleware.js
```diff
@@ -192,11 +191,12 @@ CachingMiddleware.prototype = new function () {
         res.write = old_res.write || res.write;
         res.end = old_res.end || res.end;
 
-        var headers = responseCache[cacheKey].headers;
+        let headers = {};
+        Object.assign(headers, (responseCache[cacheKey].headers || {}));
```
rhansen, line 195: Questions/comments:
* Is `responseCache[cacheKey]` guaranteed to be non-null here?
* I think the object spread operator is easier to read vs. `Object.assign()`.
* It would be good to add a comment explaining why a shallow copy is made.

Suggested change that addresses the above:
```suggestion
        const {[cacheKey]: {headers: cachedHeaders = {}} = {}} = responseCache;
        // Make a shallow copy of the cached headers so that we can add/change headers without
        // affecting the cache.
        const headers = {...cachedHeaders};
```

src/node/utils/caching_middleware.js
```diff
@@ -192,11 +191,12 @@ CachingMiddleware.prototype = new function () {
...
         var statusCode = responseCache[cacheKey].statusCode;
 
         var pathStr = CACHE_DIR + 'minified_' + cacheKey;
-        if (supportsGzip && (headers['content-type'] || '').match(/^text\//)) {
+        if (supportsGzip && /^application\/javascript/.test(headers['content-type'])) {
```
rhansen, line 199: Should there be an end-of-string anchor?

```suggestion
        if (supportsGzip && /^application\/javascript$/.test(headers['content-type'])) {
```

rhansen (approved):

src/node/utils/caching_middleware.js
```diff
@@ -192,11 +191,12 @@ CachingMiddleware.prototype = new function () {
...
         res.end = old_res.end || res.end;
 
-        var headers = responseCache[cacheKey].headers;
+        let headers = {};
+        Object.assign(headers, (responseCache[cacheKey].headers || {}));
         var statusCode = responseCache[cacheKey].statusCode;
 
         var pathStr = CACHE_DIR + 'minified_' + cacheKey;
-        if (supportsGzip && (headers['content-type'] || '').match(/^text\//)) {
+        if (supportsGzip && /application\/javascript/.test(headers['content-type'])) {
```
rhansen, line 199: Maybe this should just be:

```suggestion
        if (supportsGzip && headers['content-type'] === 'application/javascript') {
```

webzwo0i: That was my second thought, too, but content-type can/should contain `charset`, too.

rhansen: Oh right, good point.

PR #4493, opened 2020-11-17
safari 14

JohnMcLear:
Replace Safari 12 with Safari 13..??


rhansen:
LGTM. What about macOS 11.0 with Safari 14.0?

JohnMcLear:
> 
> 
> LGTM. What about macOS 11.0 with Safari 14.0?

Afaik no Sauce Labs support yet for this config.

PR #4499, opened 2020-11-20
import/export: Kill soffice spawned process after 120 seconds

JohnMcLear:


rhansen:
What bug is this fixing?

JohnMcLear:
I emailed you a private file which can hang/crash libreoffice.  Because it makes libreoffice hang you have to kill it.  This kills the bad process.

rhansen:
Ah, gotcha. Given that it is unusual for OpenOffice to hang, perhaps increase the timeout? Maybe 120 seconds? We don't want the timeout to fire just because it's a large document.

JohnMcLear:
> 
> 
> Ah, gotcha. Given that it is unusual for OpenOffice to hang, perhaps increase the timeout? Maybe 120 seconds? We don't want the timeout to fire just because it's a large document.

changed to 120 👍

PR #4506, opened 2020-11-22
GitHub workflows

webzwo0i:
To bridge the period until Travis is working again.
This also adds a small improvement to remote_runner.js: better error message in case not all saucelabs runner have been started.

One thing that confuses me is https://github.com/ether/etherpad-lite/blob/develop/.travis.yml#L84
I kept this in the workflow for now. While running the plugin tests is good to check compatiblity with latest changes in etherpad, it could also be useful to run etherpad's tests when the plugins are installed.

I noticed that it runs the workflows twice after opening the PR, one time for the branch push and one time for the PR. I tried to change that behaviour: now they run on push to any branch and on PRs, when they come from an external repo branch - with the exception of the frontend tests: they only run on push due to saucelabs secrets


rhansen:
`tests/ratelimit/testlimits.sh` is an empty file. Is that intentional?

I wonder if it would be better to combine everything into a single `.yml` file:
  * There's some duplication between the files, so combining them would allow us to remove the duplication.
  * GitHub doesn't support cross-file dependencies (e.g., only run test foo if test bar from another file passed), so putting everything in the same file would allow us skip later tests if a test fails. I don't know if we would ever use dependencies, but it might be good to have the option.

JohnMcLear:
Wow this is great work!  Just need it to work 👍

webzwo0i:
@rhansen 
It's only changing file mode for `testlimits.sh`.

RE single file: Not sure what's best here. I thought it's beneficial to see the workflows directly in the "checks" view. When they are all combined together in one workflow, I need to click on "Details". I can live with both solutions.

RE executing based on another workflow: I found https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_run

PR #4510, opened 2020-11-23
checkPlugins: Revise `npmpublish.yml`

rhansen:
* Add commentary explaining why things are done the way they are.
  * Delete steps that were added for debugging.
  * Pass `--no-save` when installing `ep_etherpad-lite`.


rhansen:
@JohnMcLear Does `ETHERPAD_NPM_V` need to be bumped to 2?

JohnMcLear:
It was no version before, now it's v1, if you make a change it will need bumping up a version.

PR #4515, opened 2020-11-25
frontend tests: fix trailing comma in runner.js

webzwo0i:
another small typo that leads to hanging pads on saucelabs for some browsers

### DIFF
--- a/tests/frontend/runner.js
+++ b/tests/frontend/runner.js
@@ -143,7 +143,7 @@ $(() => {
   // http://stackoverflow.com/questions/1403888/get-url-parameter-with-jquery
   const getURLParameter = function (name) {
     return decodeURI(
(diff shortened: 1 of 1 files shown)


rhansen:
Awesome find @webzwo0i! How did you figure this out? Were you able to access the Sauce Labs console logs somehow?

There are other instances of trailing commas. I fixed the ESLint config to prohibit such commas: https://github.com/ether/eslint-config-etherpad/commit/673ab07acfdf9df5eb6e6580703e7715c7cb1354. I also re-ran `eslint --fix` with the rule change to fix all of the instances; see PR #4516.

Anyway, I think a better way to fix this particular instance would be this:

```javascript
  const getURLParameter = (name) => (new URLSearchParams(location.search)).get(name);
```

JohnMcLear:
My dislike for trailing commas is justified

PR #4519, opened 2020-11-25
Translate select

ilmartyrk:
Switching languages didn't translate dropdowns. Didn't add test in this pull as couldn't find a dropdown without plugins. But will add a test to `ep_font_color `and `ep_font_size plugins`.


JohnMcLear:
More awesome work thanks!!

PR #4526, opened 2020-11-28
pad: New eejsBlock hook for the permissionDenied div

rhansen:
Multiple commits:
* css: Delete absolute `line-height`
* css: Apply font formatting to the text, not the high-level div
* pad: Add a hook for the permission denied block


JohnMcLear:
Cc @seballot

PR #4529, opened 2020-11-29
Fix whitespace

JohnMcLear:
Screenshot_20201129-024327


JohnMcLear:
I'm merging as this is progress but I'm mindful it might not be the ultimate fix.

Issue #4536, opened 2020-12-02
Plugins can't be updated to version 0.(x+1) via /admin when they where installed at version 0.x


webzwo0i:
One could argue this is a bug, but on the other hand it's also kind of a feature or even intended behavior. Re-label it if needed.

npm's default value that we don't change for the [save-prefix](https://docs.npmjs.com/cli/v6/using-npm/config#save-prefix) is
```
#npm config get save-prefix
^
```

When we add a plugin at version 0.1.2 it gets added to package.json with a semver of `^0.1.2`, however, versions that start with `0` [will only get patch-level updates](https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004) when configured with a caret.

That's why a plugin cannot be updated from 0.1.2 to 0.2.0 via /admin. We may need a way to issue another semver for installed plugins or to use other npm logic when updating plugins, so the latest plugin can be installed.

I did not test that, but probably it's also affecting updates from 1.0 to 2.0 and it's a bigger change in the UI (maybe have the ability to select the version that should be installed)

JohnMcLear:
I consider this a bug not a feature.

Admins need to be able to do updates without restriction with the exception of  peerDepencies.

Is there any way we can fix this within Etherpad or do we need to tell plugin Devs never to use minor/major?

webzwo0i:
I'm pretty sure we can fix it in etherpad. afaik the caret is added due to a npm default config. Maybe we can set this config option only for etherpad.

I think it's also possible to post-process package.json and remove the caret. But the most straight-forward solution, at least I can imagine this will work, include the plugin version that we want to install in https://github.com/ether/etherpad-lite/blob/c673a597f82903f34dc6e42b8f6ff95f003ec886/src/static/js/pluginfw/installer.js#L50

So this bug here can probably be fixed, but I should open up another issue "improving plugin installation via /admin" because we need the ability to specifiy a version somehow, although I am not sure what a good UI would look like:
Suppose you are on etherpad 1.8.6 and use a plugin ep_plugin@0.1.20. The latest 0.1.x release is 0.1.29. There is also a new version 0.4, but you want to stick to 0.1.x. Currently you can only install the latest version of a plugin.

JohnMcLear:
My general thought here is you will use cli of you really want to manage your plugin versions.  Imho admin is for quick updates/wins and cli is if you want to get specific...

[...]

PR #4543, opened 2020-12-05
WIP:explicitly install the latest version of a plugin, fix for #4536

webzwo0i:


webzwo0i:
need to test further

JohnMcLear:
fwiw i was able to install latest 0.1.7 to 0.2.1 w/ out this patch somehow ;/

webzwo0i:
are you running npm 7.x? maybe they changed something. For me it will install 0.1.20, when I have 0.1.7 installed

JohnMcLear:
node version is 15.1.0

webzwo0i:
weird. node v15.3.0 and npm 6.14.8, when I manually install version 0.1.7 and update via /admin I always get 0.1.20.

JohnMcLear:
Merging anyway because it shouldn't do any harm

PR #4545, opened 2020-12-05
regenerate package-lock.json

webzwo0i:
the package-lock.json from the other PR somehow contained rehype-format although it was removed from package.json :-)


webzwo0i:
weirdness! the package-lock.json did not contain rehype-format, but node still complained, probably because my node_modules directory was not deleted

Issue #4546, opened 2020-12-05
Attempting to restart server using /admin/settings hangs instance


JohnMcLear:
Caused by https://github.com/ether/etherpad-lite/commit/a000a93dc644aa32f5ad9a0385837981f57edc44#diff-e0209b1fab75aa27962acd6cac97eeaa8fd2f41a4e61d4cd0b913d1206b1f7d0R61

Replicate by click Restart in /admin/settings

Server will restart after all tabs/windows are closed...

Noticed bug due to not being able to update plugins using admin interface.

webzwo0i:
Currently we only call cancel, so no new connections are allowed, but the remaining connections stay alive.
The fix is imo something like (not sure if something special needs to be done for socket.io's websocket):
- store sockets that connect ('connection' event)
- on restart/ctrl+c in the console, call `cancel` to stop new connections
- call the `end` method on all sockets, to send a TCP FIN
- wait for some time (e.g. some seconds, maybe less on ctrl+c) to give the clients the opportunity to gracefully terminate the sessions, then call `destroy` on every remaining socket

PR #4547, opened 2020-12-05
"/admin" frontend test coverage

JohnMcLear:
Does front end testing of admin pages if a ``disablePasswordRequirementForAdminUI`` boolean  is set to ``true`` in ``settings.json``

Coverage for https://github.com/ether/etherpad-lite/issues/4546

# Docs to add once merged.

Also any plugins that call a test ``adminWHATEVER.js`` will be enabled with this setting.  This means if developers want to add test coverage for their plugins admin UI they can :)  That will need documenting once this is merged.  This also means that if someone calls their plugin file name adminMyawesomePadTest.js it will be ignored/disabled.

# To test 
1. ``"disablePasswordRequirementForAdminUI":true,`` in your settings.json
1. Visit /tests/frontend in your browser

OR

1. run the CI tests which automatically enable the ``disablePasswordRequirementForAdminUI`` boolean

# Note that...

``disablePasswordRequirementForAdminUI`` is always disabled by default.


webzwo0i:
Please see my commit here: https://github.com/ether/etherpad-lite/commit/4905ecdd745c62f2643ca556ff6af38b8bf11b70
If you like it, cherry-pick into this branch. 

It adds a small test if saving&reloading settings does actually work and does real basic auth, so skipping the auth checks is not necessary anymore.

I have one problem though, that jquery somehow can't get the items sometimes:
https://github.com/ether/etherpad-lite/commit/4905ecdd745c62f2643ca556ff6af38b8bf11b70#diff-20bd0ec7961a57df2cefa5d362008165301225a71bdc7c132eb763dbfcd9442eR28

I don't know why :-( It works in other situations, like getting `helper.admin$('.settings').val()` !?

Another problem: Because the tests now actively change settings, multiple runner interfere with each other, so tests sometimes fail again :-)

JohnMcLear:
It makes sense to favor new-admin-test-coverage over this.

PR #4554, opened 2020-12-08
Feature: Scroll to Line number based on Hash IE http://foo.com/p/bar#L10 will scroll to line 10.

JohnMcLear:
# How to use (tm)
Clicking on side div line number to set line number hash and then on visiting scrolling back into view.

- [x] Agree on UX.
- [x] Agree this should be in core.
- [x] Tests
- [x] Docs?
- [x] Correct Y position when padding-top is not visible.
- [x] Some linting
- [x] Someone please check my test spec because I think I messed something up

Future TODOs:
 - [ ] Anchor Support IE ``#foo`` will visit ``<h1>foo</h1>`` ?


webzwo0i:
agreed this is useful to be in core and works for me.

any reason you are not using code from ace2_inner.js or scroll.js? Something like https://github.com/ether/etherpad-lite/blob/develop/src/static/js/ace2_inner.js#L4076, although this means you need to set rep? (Like setting the caret to the beginning of the line - maybe the editor will automatically scroll there?)

JohnMcLear:
I can use ``setrep``. Will try that.  Good suggestion.

Afaik I don't use it because sometimes using ace inner code can get a little spaghetti where just interacting with outer is clean.  Will see.

JohnMcLear:
@webzwo0i give that a pull and test.  Now it focused caret too..  But note that I'm still using new logic to handle focus because trying to use ``ace.ace_scrollSelectionIntoView();`` after exposing it did not work.  I also prefer this UX 👍

madix93:
Hello,

We've been using some pads for a very long time and line numbers can change over time. So being able to set an anchor would be useful too.

Or, even better, being able to set an an anchor in /having anchor on the table of contents entries. So that, https://foo.com/p/bar#EntryOneOfToc will scroll directly to EntryOneOfToc. 

For example, there is a table of contents on this pad <https://pad.april.org/p/radio_cause_commune>. It would be useful to be able to have an URL going directly to an entry of the table of contents. And, we could share this url by email for example.

All the best,

webzwo0i:
Try with a setTimeout, (compare with https://github.com/ether/etherpad-lite/commit/b83589fc73a20826f618eb62476ece159f2941d8 )

The default behavior with `ace.ace_scrollSelectionIntoView();` is probably that the line will be the last line in the view (not the first one).

In the commit above the line that has the caret is not the first visible line in the editor, but when we now start typing or scroll to the right, no additional scrolling happens. Code that incorporates the offset of the first line and the offsetTop of the iframe is commented in the code. Can't we use the offset of the first line and the offset of the line we want to go to, to determine how much we need to scroll? IE if the offset of the first line is 30 (while scrollY is 0) and we want to go to the offset 1000, we need to scroll 1000 - 30.

There is also `Scroll.prototype._getPaddingTopAddedWhenPageViewIsEnable` that returns the top padding of `ace_outer`.

However, there is still one big problem: When we have a lot of lines and want to go to a line in the middle of the pad, it won't work unless we increase the timeout (firefox needs a larger timeout than chromium). I can't put much time into better understanding the ACE code and don't understand what exactly has not finished loading.

Maybe the method needs to be triggered in another handler.

One other thing is, do we need to take into account lines with line attributes? Because the caret (first char in a line) is at `[y, 1]` on a line with line attributes and at `[y, 0]` on a line without attributes. Might be irrelevant when using `setSelection`.

I'm okay with the way you solved it, so don't mind my code, unless you have a good idea if there is another place to call the method from

JohnMcLear:
So ..

1. test with long pads.
2. test with long lines with varying line heights etc.

Ack, will do.

JohnMcLear:
Tested pad with 5K lines and it works fine.

Tested content with varying line height and it works fine.

Tested in Chrome and FF.

JohnMcLear:
@webzwo0i can you check my test coverage please?

tests/frontend/specs/scrollTo.js
```diff
@@ -0,0 +1,45 @@
...
+
+  it('Scrolls down to Line 4', function (done) {
+    this.timeout(10000);
+    const chrome$ = helper.padChrome$;
+    helper.waitFor(() => {
+      const topOffset = parseInt(chrome$('iframe').first('iframe')
+          .contents().find('#outerdocbody').css('top'));
+      return (topOffset >= 100);
+    });
+    done();
```
webzwo0i, line 21: I think this should be
```
helper.waitFor....
}).done(done)
```
or
```
async function () {
...
await helper.waitForPromise...
```

tests/frontend/specs/scrollTo.js
```diff
@@ -0,0 +1,45 @@
...
+    this.timeout(10000);
+  });
+
+  it('Does NOT change scroll', function (done) {
+    this.timeout(10000);
+    const chrome$ = helper.padChrome$;
+    helper.waitFor(() => {
+      const topOffset = parseInt(chrome$('iframe').first('iframe')
+          .contents().find('#outerdocbody').css('top'));
+      return (!topOffset); // no css top should be set.
```
webzwo0i, line 41: see above, also applies to here

webzwo0i:
RE test: I added a comment there. Also, if you happen to know how to get the visible lines from the viewport, then that's probably more accurate?

JohnMcLear:
It's something I'm gonna have to bake in.

JohnMcLear:
Okay this is good for final review and merge.

PR #4560, opened 2020-12-10
pad: Fix wrong variable name in global exception handler

rhansen:
This fixes a bug introduced in commit c845d985e0d6014f2c2474564c59f33cdc552de6.


JohnMcLear:
oof.

PR #4561, opened 2020-12-11
missing await in createClearStartAtext

automatist:
Missing await in call to this._pad.getInternalRevisionAText(rev). Function returns a promise (see call in line 81 of same file). This bug breaks the createDiffHTML API call (how I discovered it).

> Screenshots/GIFs


JohnMcLear:
Thanks! 👍

PR #4563, opened 2020-12-14
URL regex improvements

rhansen:
Multiple commits:
* lint: Fix some straightforward ESLint errors
* tests: New `helper.clearPad` method
* tests: Expand the tests in `urls_become_clickable.js`
* Factor out common URL regular expression code
* Exclude `?`, `!`, and `)` from last character of URL
* Add `'` and `*` to acceptable URL characters


JohnMcLear:
Much cleaner :)  Thanks!!

PR #4570, opened 2020-12-15
lint: AttributeManager

JohnMcLear:


rhansen:
[Line 361](https://github.com/ether/etherpad-lite/blob/219c0c3186abee243ec8d1a8e891d07c48d56f8f/src/static/js/AttributeManager.js#L361) is highly suspect for a few reasons:
1. The callback passed to `map()` ([line 355](https://github.com/ether/etherpad-lite/blob/219c0c3186abee243ec8d1a8e891d07c48d56f8f/src/static/js/AttributeManager.js#L355)) is a normal function, not an arrow function. Thus, the value of `this` on line 361 depends on how the callback is called. In this case, the callback is called  without an explicit context object (it is not called as a method nor is it called via `apply()`, `bind()`, or `call()`). This smells foul—why is `this` used when there is no explicit context object? I'm guessing it's a bug; that `this` was intended to refer to the AttributeManager object.
2. Let's assume that the use of `this` was intentional even though there is no explicit context. Because the callback is called without any explicit context, the value of `this` depends on whether the function is in strict mode or not. Before this commit, it was in sloppy mode so `this` referred to the "global this" object (a.k.a. `window`). After this commit, the function is in strict mode so `this` is undefined and `this.author` will throw an exception. That's probably not a desired behavior change. Why didn't the tests see this? Is the test coverage insufficient?
3. The return value on that line is `[attributeName, this.author]`. I don't fully understand the code, but I think `attributeName` should be `attrib[0]`. Doesn't `attributeName` cause attribute corruption? It will cause `attributeName` to appear multiple times in the `map()` output, possibly with different values, and it silently discards the `author` attribute.

rhansen:
> After this commit, the function is in strict mode so `this` is undefined and `this.author` will throw an exception. That's probably not a desired behavior change. Why didn't the tests see this? Is the test coverage insufficient?

Actually it looks like that change is causing tests to fail, so that's good at least.

rhansen:
I force-pushed some fixes to the branch. Please review my commits, and if they look good to you do a rebase and merge (**not** a squash and merge).

PR #4571, opened 2020-12-15
lint: broadcast_revisions

JohnMcLear:
Need help with a global 👍

### DIFF
--- a/src/static/js/broadcast.js
+++ b/src/static/js/broadcast.js
@@ -199,6 +199,7 @@ function loadBroadcastJS(socket, sendSocketMsg, fireWhenAllScriptsAreLoaded, Bro
   };
 
   const loadedNewChangeset = (changesetForward, changesetBackward, revision, timeDelta) => {
+    const revisionInfo = window.revisionInfo;
     const broadcasting = (BroadcastSlider.getSliderPosition() === revisionInfo.latest);
     revisionInfo.addChangeset(
         revision, revision + 1, changesetForward, changesetBackward, timeDelta);
@@ -264,7 +265,7 @@ function loadBroadcastJS(socket, sendSocketMsg, fireWhenAllScriptsAreLoaded, Bro
 
   const goToRevision = (newRevision) => {
     padContents.targetRevision = newRevision;
-    const path = revisionInfo.getPath(padContents.currentRevision, newRevision);
+    const path = window.revisionInfo.getPath(padContents.currentRevision, newRevision);
 
     hooks.aCallAll('goToRevisionEvent', {
       rev: newRevision,
@@ -405,7 +406,7 @@ function loadBroadcastJS(socket, sendSocketMsg, fireWhenAllScriptsAreLoaded, Bro
             Changeset.moveOpsToNewPool(data.forwardsChangesets[i], pool, padContents.apool);
         const backwardcs =
             Changeset.moveOpsToNewPool(data.backwardsChangesets[i], pool, padContents.apool);
-        revisionInfo.addChangeset(astart, aend, forwardcs, backwardcs, data.timeDeltas[i]);
+        window.revisionInfo.addChangeset(astart, aend, forwardcs, backwardcs, data.timeDeltas[i]);
       }
       if (callback) callback(start - 1, start + data.forwardsChangesets.length * granularity - 1);
     },
--- a/src/static/js/broadcast.js
+++ b/src/static/js/broadcast.js
@@ -34,6 +34,7 @@ const hooks = require('./pluginfw/hooks');
 // These parameters were global, now they are injected. A reference to the
 // Timeslider controller would probably be more appropriate.
 function loadBroadcastJS(socket, sendSocketMsg, fireWhenAllScriptsAreLoaded, BroadcastSlider) {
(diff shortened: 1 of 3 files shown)


rhansen:
> Need help with a global +1

Fixed. Please take a look.

JohnMcLear:
Right I learned...  

https://github.com/ether/etherpad-lite/pull/4571/commits/c5c60175d1a57294b46ba225c89bf7051bae2ed8 -- We still want to use ``arguments`` instead of passing ``...args``?

``splice.apply`` is fine?  I thought that was erroring for me..

https://github.com/ether/etherpad-lite/pull/4571/commits/c5c60175d1a57294b46ba225c89bf7051bae2ed8#diff-7db26e6410bd870477ccfc66fd084e0ce0d134b9554267ca8c6aa301fb64aebcR486 I don't see how this doesn't error but okay...

But wait then we turn back to using ``...args`` on https://github.com/ether/etherpad-lite/pull/4571/commits/b85bfc7dc2b7833093590d9670e4de4e47083fd4 so which do we prefer?  

https://github.com/ether/etherpad-lite/pull/4571/commits/783469aea87254e7d854d76f503d4d690223ab15 is much cleaner than my approach, +1 that, I think I was just a bit lost in the mix and should have spent more time on this specific function.

https://github.com/ether/etherpad-lite/pull/4571/commits/ce3ab66035a7edf4a5b21d07b59109943a42bf5e#diff-7db26e6410bd870477ccfc66fd084e0ce0d134b9554267ca8c6aa301fb64aebcR474 I prefer your approach here but I can't promise I will remember to use it and I think it's beyond the scope of lint fixing for this round?  Can we agree that these are bonuses if I remember but not put the effort in this round?

It's bad to use ``window.whatever`` but it's sometimes a necessary evil until we can refactor these out?

Gun to replace ``self.`` with ``this.`` if possible.  Do you want this done on this round of linting?  Should we add warnings for self if that's the case?

Thanks for the help and lessons :)

rhansen:
Some background behind why I made that "Revert some incorrect lint fixes" commit: If a change introduces a bug, good version control practice is to revert the change then re-do it properly. Benefits:

  * It's easier to review a revert followed by a re-do than it is to review a fix of a buggy change. (It's hard to see the net change when looking at a fix of a buggy change). This is particularly true if there have been some overlapping changes between the original buggy change and the revert of the buggy change.
  * The re-do can be cherry-picked by itself.
  * `git blame` provides more useful results.
  * If the re-do is also buggy you just revert and try again—you don't end up with a difficult to understand fix for a fix for a fix.

In this case I didn't revert the whole branch, just the parts that were buggy.

I expect all of these commits to be squashed into one commit before this PR is merged, so the end result will be `...args` and `for (const [k, v] of Object.entries(obj))` if you decide to keep all of my changes.

> 
> [ce3ab66#diff-7db26e6410bd870477ccfc66fd084e0ce0d134b9554267ca8c6aa301fb64aebcR474](https://github.com/ether/etherpad-lite/commit/ce3ab66035a7edf4a5b21d07b59109943a42bf5e#diff-7db26e6410bd870477ccfc66fd084e0ce0d134b9554267ca8c6aa301fb64aebcR474) I prefer your approach here but I can't promise I will remember to use it and I think it's beyond the scope of lint fixing for this round? Can we agree that these are bonuses if I remember but not put the effort in this round?

I think it's a good idea to leave some lint errors unresolved for now if their resolution is complicated or risky. The most important thing is to not incorrectly resolve lint errors.

> It's bad to use `window.whatever` but it's sometimes a necessary evil until we can refactor these out?

`window.foo` is no worse than top-level `var foo`. From JavaScript's perspective they are 99% equivalent (that last 1% is unimportant for our purposes). From a readability perspective `window.foo` is arguably better because the reader immediately knows that the `window` object is being polluted. Getting rid of all of the `window.foo` (and `var foo`) globals is not trivial, so it's definitely OK to leave those alone until the code can be refactored.

> Gun to replace `self.` with `this.` if possible. Do you want this done on this round of linting? Should we add warnings for self if that's the case?

I'm fine with deferring changes from `self` to `this` until a later time. I made those changes last so that they could be easily discarded or moved to another PR. The reason I included those changes at all is because the squashed commit would fix the prefer-arrow-functions lint error by changing `self` to `this`, not by changing functions to arrow functions.

JohnMcLear:
Yea obviously happy to take your changes as you know better than me and they are way more readable than the original.

So one thing to consider is that I commit a lot, like every few minutes, because my life is absolute chaos, I jump between devices because I'm often doing lots of other things while trying to hack on Etherpad.  Because of this I will commit a breaking change then push it so I don't lose it when going onto another device.  This means that I'm used to squashing (because it sanitizes my chaos).  Hope you can appreciate/understand that and apologies for the noise! :)  Also worth noting that it's kinda embarrassing from my perspective that I'm not writing "working" lint fixes, so having a commit that's "undone broken code" is not ideal, but I'm not precious and totally fine with being the _noob_ in the room as long as I'm learning on the way ;)

Appreciate the feedback, will take it on-board and begin really focusing on improving the Etherpad git history.

JohnMcLear:
If you are happy with this, merge away  :)  LGTM! :)

PR #4573, opened 2020-12-15
lint: chat

JohnMcLear:
If username isn't set I noticed it doesn't pass "" value as chat name but some weird value..  I think this is because underscore is now being required..


rhansen:
> If username isn't set I noticed it doesn't pass "" value as chat name but some weird value.. I think this is because underscore is now being required..

Yes, that is the cause. The `require('./underscore')` line shouldn't be there; the `_` variable is actually an alias of `html10n.get`, not underscore.js.

I pushed some fixes; please take a look.

JohnMcLear:
Great that's super useful to learn thanks.

PR #4576, opened 2020-12-16
Lint: pluginfw tsort.js

HMarzban:
Lint `/src/static/js/pluginfw/tsort.js`

### DIFF
--- a/src/static/js/pluginfw/tsort.js
+++ b/src/static/js/pluginfw/tsort.js
@@ -70,7 +70,7 @@ const tsortTest = () => {
   ];
 
   let sorted = tsort(edges);
-  // console.log(sorted);
+  console.log(sorted);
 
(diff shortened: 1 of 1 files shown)


src/static/js/pluginfw/tsort.js
```diff
@@ -6,6 +6,7 @@
  *
  * @returns Array : topological sorted list of IDs
  **/
+'use strict';
```
JohnMcLear, line 9: minor thing but I have been putting this as line 1.  I'm not sure if it's something we lint for tho as long as it's there and before any logic :)

JohnMcLear (approved):

src/static/js/pluginfw/tsort.js
HMarzban: Not necessary, but I pulled it up.

PR #4577, opened 2020-12-16
lint: pad prefix files

JohnMcLear:
Lint of the rest of the "pad" files.

### DIFF
--- a/src/static/js/pad.js
+++ b/src/static/js/pad.js
@@ -569,9 +569,9 @@ const pad = {
       if (!pad.padOptions.view) {
         pad.padOptions.view = {};
       }
-      for (const k in opts.view) {
-        pad.padOptions.view[k] = opts.view[k];
-        padcookie.setPref(k, opts.view[k]);
+      for (const [k, v] of Object.entries(opts.view)) {
+        pad.padOptions.view[k] = v;
+        padcookie.setPref(k, v);
       }
       padeditor.setViewOptions(pad.padOptions.view);
     }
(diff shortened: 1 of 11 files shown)


JohnMcLear:
Need help/suggestions on how to best handle ``clientVars``

https://github.com/ether/etherpad-lite/pull/4577/commits/c4ab92112b3d6d659a01f278e80fb31e9e1339b4#diff-e13d29dae930cc2fb95633bbba052d9d7e990f33feb2a6404c181d27d0f8d14cR292

rhansen:
> Need help/suggestions on how to best handle `clientVars`

For assignments, doing `window.clientVars = foo;` is probably the best option for now (until the code can be refactored to something cleaner).

I rebased onto latest `develop` and pushed some fixes and refinemens; please take a look.

A common mistake I see is in the fixes to [guard-for-in](https://eslint.org/docs/rules/guard-for-in) errors. Code that uses [`for...in`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) is almost always wrong; the code should probably use [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) with [`Object.entries()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries), [`Object.keys()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys), or [`Object.values()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values) instead. For example, code that looks like this:

```javascript
for (const k in obj) {
  const v = obj[k];
  // do something with k and v...
}
```

should usually be fixed by doing this:

```javascript
for (const [k, v] of Object.entries(obj)) {
  // do something with k and v...
}
```

Fixes like the above technically change the behavior, but that's the point: The behavior of `for..in` is almost never desired.

JohnMcLear:
Tnx for fixes.

PR #4580, opened 2020-12-17
Lint: timeslider

HMarzban:
Lint `/src/static/js/timeslider.js`

### DIFF
--- a/src/static/js/timeslider.js
+++ b/src/static/js/timeslider.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /**
  * This code is mostly from the old Etherpad. Please help us to comment this code.
  * This helps other people to understand this code better and helps them to improve it.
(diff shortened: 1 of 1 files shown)


src/static/js/timeslider.js
```diff
@@ -28,11 +30,81 @@ const Cookies = require('./pad_utils').Cookies;
 const randomString = require('./pad_utils').randomString;
 const hooks = require('./pluginfw/hooks');
 
-let token, padId, export_links;
+let token, padId, exportLinks, socket, changesetLoader, BroadcastSlider;
```
rhansen, line 33: If I recall correctly, the `socket` and `BroadcastSlider` vars need to be on the `globalThis` object (`window`) because other `.js` files (and maybe code in the HTML?) reference them. Rather than scope them to this file, use `window.socket` and `window.BroadcastSlider` to ensure that these lint fixes don't change behavior.

I realize that it is gross to do `window.foo`, but cleaning that up is out of scope for a lint fix PR.

src/static/js/timeslider.js
```diff
@@ -44,14 +116,14 @@ function init() {
 
     // ensure we have a token
     token = Cookies.get('token');
-    if (token == null) {
+    if (!token) {
```
rhansen, line 119: Undo this change; it's not a lint fix.

src/static/js/timeslider.js
```diff
@@ -28,11 +30,81 @@ const Cookies = require('./pad_utils').Cookies;
 const randomString = require('./pad_utils').randomString;
 const hooks = require('./pluginfw/hooks');
 
-let token, padId, export_links;
+let token, padId, exportLinks, socket, changesetLoader, BroadcastSlider;
+const fireWhenAllScriptsAreLoaded = [];
```
rhansen, line 34: Moving the code around makes it difficult to review changes. Would you move these back to the order they were originally in? (That might not be possible if the original order throws `ReferenceError`.) ESLint will complain about `no-use-before-define`, but those are warnings not errors.

src/static/js/timeslider.js
```diff
@@ -71,90 +143,35 @@ function init() {
...
+    exportLinks = $('#export > .exportlink');
 
     $('button#forcereconnect').click(() => {
       window.location.reload();
     });
 
-    exports.socket = socket; // make the socket available
-    exports.BroadcastSlider = BroadcastSlider; // Make the slider available
+    // exports.socket = socket;
+    // exports.BroadcastSlider = BroadcastSlider;
```
rhansen, line 163: Please don't introduce commented-out code.

rhansen (commented):
Please rebase onto `develop`.

src/static/js/timeslider.js
JohnMcLear: Weird that I did...  I wonder why..  Thanks for spotting ...  Will investigate why/how this happened because I know not to.

src/static/js/timeslider.js
JohnMcLear: afaik this was to fix an error (probably not ``initialized``), I didn't address lint warnings..

src/static/js/timeslider.js
JohnMcLear: Oh...  This wasn't my PR :)  I'm an idjit 📦

PR #4583, opened 2020-12-18
Ignore head tag on import

webzwo0i:
import should only use body. removes a workaround that results in empty pads if an empty title element is present.

The PR is based on another PR that enables the tests for importexport, so it needs rebasing in case the other gets merged in first. alternatively merge this without squashing, then the other PR can be closed.

It might be useful in the future to parse style tags inside <head> on import though

### DIFF
--- a/tests/backend/specs/api/importexport.js
+++ b/tests/backend/specs/api/importexport.js
@@ -1,9 +1,12 @@
+'use strict';
 /*
  * ACHTUNG: there is a copied & modified version of this file in
  * <basedir>/tests/container/spacs/api/pad.js
  *
  * TODO: unify those two files, and merge in a single one.
  */
+
+/* eslint-disable max-len */
 
 const common = require('../../common');
 const supertest = require(`${__dirname}/../../../../src/node_modules/supertest`);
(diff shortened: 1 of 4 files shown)


tests/backend/specs/api/importexport.js
```diff
@@ -1,18 +1,20 @@
+'use strict';
 /*
  * ACHTUNG: there is a copied & modified version of this file in
  * <basedir>/tests/container/spacs/api/pad.js
  *
  * TODO: unify those two files, and merge in a single one.
  */
 
+/* eslint-disable max-len */
```
JohnMcLear, line 9: I'm trying really hard not to use disables btw :)

tests/backend/specs/api/importexport.js
```diff
@@ -54,13 +56,24 @@ const testImports = {
...
+  },
 };
 
 describe(__filename, function () {
   Object.keys(testImports).forEach((testName) => {
     const testPadId = makeid();
-    test = testImports[testName];
-    describe('createPad', function () {
+    const test = testImports[testName];
+    if (test.disabled) {
```
JohnMcLear, line 71: What's going on here?   where are the docs / settings / info for this value?

PR #4584, opened 2020-12-18
Delete dead code

rhansen:
Multiple commits:
* Delete dead SERVER_MESSAGE and guest handling code. None of this code seems to be reachable. Hopefully no plugins expect it to exist.
* pad import/export: Delete dead functions


JohnMcLear:
Note that I already had done this as part of pad-star-dot-star branch so it will conflict pretty heavily.

PR #4599, opened 2020-12-21
change scroll-behavior to auto

webzwo0i:
@JohnMcLear can you test if you can still replicate the failing timeslider_follow test?
It changes the scroll-behavior from smooth to auto. After that the test is never failing for me. I've seen that the smooth scrolling sometimes stops before reaching the offset. With auto I can't reproduce this anymore.

I also increased the number of lines that are in the pad view below the changed line. consider a pad with 40 lines and a change in the last line. To meet the test assumption, the last line should be the first visible line in the pad. That's not the case, when the display size is large enough. With a lot of lines after the changed line, so that the whole view port consists of lines, it will successfully scroll the changed line into the position of the "first visible line"


JohnMcLear:
Tested, confirmed working!  Please rebase to fix broadcast.js conflict 👍

JohnMcLear:
Let's go! :)

PR #4606, opened 2020-12-22
stale code: removed excanvas which was Internet Explorer support for a lack of canvas being supported

JohnMcLear:
excanvas should have been removed as part of removing IE cruft.  This PR removes it 👍


JohnMcLear:
Bump @rhansen :)

PR #4615, opened 2020-12-24
enable our errorhandling for all routes

webzwo0i:
This makes errorhandling the last route that's registered in the `expressCreateServer` hook. Because the hooks are called in reverse order this means it must move to the upper most position.

With this in place, errorhandling is called on every `next(new Error('somethings forbidden').status = 403)` etc. Before it was only called when the route, in which the next handler is called with an error happens to have been registered before errorhandler (e.g. this was not the case for any static routes, caching_middleware etc).

Now it forwards all errors to express' internal error handler, while up to now it would convert any error that reaches the errorhandler to an 500. The error logging has been removed, as this is already the default with the internal error handler.

New is, that now any error is counted in the stats metrics.

### DIFF
--- a/src/node/hooks/express/errorhandling.js
+++ b/src/node/hooks/express/errorhandling.js
@@ -2,17 +2,22 @@
 
 const stats = require('../../stats');
 
-exports.expressCreateServer = (hook_name, args, cb) => {
-  exports.app = args.app;
-
-  // Handle errors
+/**
+ * Express already comes with an error handler that is attached
+ * as the last middleware. Within all routes it's possible to call
+ * `next(err)` where `err` is an Error object. You can specify
+ * `statusCode` and `statusMessage`.
+ * For more details see "The default error handler" section on
+ * https://expressjs.com/en/guide/error-handling.html
+ *
+ * This method is only used for metrics
+ *
+ */
+exports.expressCreateServer = (hookName, args, cb) => {
   args.app.use((err, req, res, next) => {
-    // if an error occurs Connect will pass it down
-    // through these "error-handling" middleware
-    // allowing you to respond however you like
-    res.status(500).send({error: 'Sorry, something bad happened!'});
-    console.error(err.stack ? err.stack : err.toString());
-    stats.meter('http500').mark();
+    const status = err.statusCode || err.status;
+    stats.meter(`http${status}`).mark();
+    next(err);
   });
 
   return cb();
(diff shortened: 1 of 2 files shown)


src/node/hooks/express/errorhandling.js
```diff
@@ -2,17 +2,22 @@
...
   args.app.use((err, req, res, next) => {
-    // if an error occurs Connect will pass it down
-    // through these "error-handling" middleware
-    // allowing you to respond however you like
-    res.status(500).send({error: 'Sorry, something bad happened!'});
-    console.error(err.stack ? err.stack : err.toString());
-    stats.meter('http500').mark();
+    const status = err.statusCode || err.status;
+    stats.meter(`http${status}`).mark();
+    next(err);
```
rhansen, line 20: I'm a bit concerned about the `status` variable being unset/wrong. How about this instead:

```suggestion
    next(err);
    // Get the status code after next() has had a chance to update it.
    stats.meter(`http${res.statusCode || 500}`).mark();
```

rhansen (commented):
Apologies for totally missing this PR until now.

src/node/hooks/express/errorhandling.js
webzwo0i: In my test (I put `throw new Error('bad bad');` right after `args.app.get('/', (req, res) => {` in specialpages.js) I get a response status code of 200 - which feels strange. (The res object is probably prepared in advance, but I thought it was enough to only throw the error without explicitly setting the error code of the response.)

I'm pretty sure I tested it in December but I probably always set a status myself, so I did not notice that it also can be undefined. (like: `err=new Error('bad bad');err.status=400;return next(err)`)

Right now we don't have any further error handler, so `next(err)` triggers express' default error handler (which prints something on console.error)

I think it's best if we had some test coverage here, but we would probably need to implement some kind of throwing route handler. What do you think if instead of using `res.statusCode` we use `err.statusCode || err.status || 500`? I'm not sure where the 200 I see is coming from (my wget gets a 500 though), but I think we should expect the err object to have the proper status code

rhansen: > In my test ... I get a response status code of 200 - which feels strange.

That's unfortunate. :slightly_frowning_face:

rhansen: I found a reliable way to get the HTTP status code and replaced your commit with the new approach. Please take a look.

PR #4619, opened 2020-12-26
change_user_name test: refactor

webzwo0i:
This test sporadically failed in CI
Don't merge yet. locall it seemes to work in firefox/chrome, but not in CI. Need to debug tomorrow


webzwo0i:
OK this is ready now. I also added a commit that fixes the content-type of the spec files, so better don't squash those two (unrelated) commits together.

JohnMcLear:
Nice and clean, thanks! 👍

Issue #4621, opened 2020-12-27
Etherpad page up/page down has a substandard user experience..


JohnMcLear:
# Test research findings.

1. Page down from line 0
	notepad
		Goes to one line above expected line
	winword
		Incorrect, caret in wrong place
	online word
		same as winword
	google docs
		Incorrect
	etherpad
		Incorrect, Goes to one line above expected line
	libreoffice
		Really weird, caret and focus in wrong place
	textarea

2. page down then up from caret at [0, whatever]
	notepad
		Correct
	winword
		Correct
	online word
		Correct
	google docs
		Correct
	etherpad
		Wrong, loses a line
	libreoffice
		Correct	

3.  page up / down over content that is a huge length line on 0,0
	notepad
		Correct
	winword
		Focus is correct but caret is wrong
	online word
		Same as Winword
 	google docs
		Incorrect (2 lines too far down
	etherpad
		Completely broken
	libreoffice
		Really weird, caret and focus in wrong place		

4. page down / up where notion of page break exists
	notepad
		Notion doesn't exist
	winword
		Really bizarre user experience and loses historical references
	online word
		Nearly correct (not that winword is not correct!) .. On various page downs/page up if caret lands on blank line the historical reference position is lost.
	google docs
		Loses historical references on page up after just two normal page downs with no specific line breaks etc being hit.
	etherpad
		Notion doesn't exist
	libreoffice
		Correct

5. caret location on X axis is maintained on page down/up
	notepad
		Broken, goes to character 0 on blank line
	winword
		Correct!  Very impressive, keeps indentation even after hitting blank line and visiting last character in document.
	online word
		Broken, goes to character 0 on blank line
	google docs
		Broken, goes to character 0 on blank line
	etherpad
		Broken, goes to character 0 on blank line
	libreoffice
		Broken, goes to character 0 on blank line

6. Transition / lag
	notepad
		Fast
	winword
		Smooth, Fast.
	online word
		Jerky.
	google docs
		Smooth-ish, Fast-ish.  Page identifier on scroll bar begins to land behind page location on page up/down event.
	etherpad
		Slow, Jerky.
	libreoffice
		Fast.
	
7. Page down on long text codent
	etherpad
		Broken.
	all others
		Fine.
		
# Some rules of page up / down
- [ ] Caret should always go back to position on page up after page down event (consider if this should be position after others editing or new position take into account edits [much harder to implement].
[...]

PR #4622, opened 2020-12-27
Page up / page down behavior rewrite

JohnMcLear:
My WIP fixes for https://github.com/ether/etherpad-lite/issues/4621

- [x] Page down works properly
  - [x] Test coverage
- [x] Single paragraph page up puts char at [0,0]
  - [x] Test coverage
- [x] Single paragraph page down puts char at ``[1, line.length]``;
  - [x] Test coverage
- [x] Page up returns to previous line (last location assuming no other edits performed around area or arrow keys etc pressed)  I think we have a bug here around getVisibleLines logic.
  - [x] Test coverage
- [ ] Support for long lines (character position wrapping)
  - [ ] Test coverage (partially done)
- [x] Maintain X caret offset on page down/up events
  - [x] Test coverage
- [x] Page down works from view instead of rep (if rep is not present in viewport?) 
  - [x] Test coverage (mostly done, might wanna add more coverage).
- [x] Support for Shift page up/down events.
  - [x] Test coverage
- [x] Press and hold page up / page down


webzwo0i:
I did some tests yesterday before your pushes. Maybe it's already outdated, but here are my observations 

- I increased the zoom level quite a bit and noticed the caret is sometimes placed off by one outside the viewport (bottom and top) (scrollNodeVerticallyIntoView leaves it up to the browser to scroll the line into view, not sure if this is related)
- when I zoomed, the line height is always ~24 pixels. It confused me, but maybe is correct? Using ep_font_size the line height changes.
- we should include coverage in connection with ep_font_size and wrapped lines
- keeping page down/up pressed -> browser is hanging, so there should be a fast path

And the most important: I'm not sure if the viewport coordinates top/bottom are accurate. the bottom was x and the top coordinate of a line was above x, but I saw the background color of that line. Is this intended/some CSS internals re background I don't understand?

JohnMcLear:
Yeah they will be outdated mostly as I did a complete rewrite.

Good point re press and hold.  I totally missed that...

JohnMcLear:
@webzwo0i can we get this finished up and merged please? :)

JohnMcLear:
Bump @webzwo0i

PR #4624, opened 2020-12-28
page up/down: add test coverage

JohnMcLear:


webzwo0i:
A yeah this is good to merge. It's also showing that the page down logic is slower than page up (>1sec compared to <100ms)

webzwo0i:
it only checks if the line changed in the right direction, but it does not check the x coordinate inside that line. maybe it should? (ie pressing page up in the middle of the first line should jump to the first character in that line, pressing page down in the last line should go to the last character of the document while in all other cases it should stay at the same x offset?)

PR #4639, opened 2021-01-08
editor: fix enter key keep line in view

### LINKED ISSUE #4636: Scroll viewport when pressing enter on the last line of viewport
If the caret is on the last line that's visible in the viewport, pressing enter won't move the viewport one line down. If you press enter two times, then the viewport moves one line.

JohnMcLear:
Fixes #4636

I was able to replicate the bug one time, then I applied this patch and it's gone but now I can't replicate it ?! 🔢


webzwo0i:
maybe a caching issue, ie press ctrl+shift+r?

JohnMcLear:
Browser always runs with cache disabled, I must be doing something weird..  Can you grab a quick video of how to replicate?  I think it's prolly caret location IE if line is empty / end of empty line / beginning etc.

PR #4647, opened 2021-01-14
lint: contentcollector and domline

JohnMcLear:
Linting of both files.

Also includes a fix for ``lineAttributes.img`` which erroneously were able to be carried onto the next line...  

**We need a proper fix for this.**  

The bug can be replicated by using API to try and upload an image when image support is enabled via a plugin.  Also can be see on the image support branch.

### DIFF
--- a/src/static/js/contentcollector.js
+++ b/src/static/js/contentcollector.js
@@ -267,9 +267,8 @@ const makeContentCollector = function (
     const attributes = [
       ['lmkr', '1'],
       ['insertorder', 'first'],
-    ].concat(
-        _.map(state.lineAttributes, (value, key) => [key, value])
-    );
+      ...Object.entries(state.lineAttributes)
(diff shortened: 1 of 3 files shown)


JohnMcLear:
Ready for review 👍

rhansen:
I'm currently reviewing this. During the review I noticed a bunch of pre-existing bugs (not caused by this PR) that slowed me down, but I should be finished soon.

JohnMcLear:
Great thanks :)  The low hanging lint fixes will arrive tonight hopefully.

rhansen:
I pushed a bunch of commits for you to review before you squash them. I reverted the non-lint changes so that they can be put in their own pull request.

JohnMcLear:
Backend tests are broken for import.  Will investigate why and bugfix.

PR #4671, opened 2021-01-22
security: bumping socketio version due to vulnerability

JohnMcLear:


webzwo0i:
@rhansen Do you have any input on lockfileVersion in package-lock.json? If I understand your commit (see below) correctly, then you decided to keep lockfileVersion 1, because LTS npm versions would regenerate the package-lock.json on `npm i`. We are using `npm ci` in ./bin/installDeps.sh, so that won't overwrite it (but installing a plugin will overwrite the package-lock.json in the main directory if that would be lockfileVersion 2 and we use version 1).

todo before merging:
- [ ] find out if hanging frontend tests are caused by this
- [ ] find out if we should use lockfileVersion 1 or 2. 2 is a backwards compatible version, see https://github.com/npm/cli/blob/653769de359b8d24f0d17b8e7e426708f49cadb8/docs/content/configuring-npm/package-lock-json.md#lockfileversion but @rhansen did revert to version 1 in the past, see https://github.com/ether/etherpad-lite/commit/9a15af78b6bcc7b96bc0263d67a6a43ea3b96d96
- [ ] find out if we are affected by https://github.com/socketio/engine.io/commit/001ca62cc4a8f511f3b2fbd9e4493ad274a6a0e5 I don't think so, though
- [ ] don't know why, but I still see `Access-Control-Allow-Origin: '*'` in the responses of the /socket.io/ path https://github.com/socketio/socket.io/commit/f78a575f66ab693c3ea96ea88429ddb1a44c86c7

rhansen:
> * find out if we should use lockfileVersion 1 or 2.

It doesn't particularly matter which version we choose, but we should be consistent. Ideally we would have a GitHub workflow that fails if rebuilding the lock files results in a diff from what is checked in.

The main reason I prefer v1 is because many users will be using the LTS version of npm which prints out a warning when it sees a v2 lock file. The warning is benign, but most users won't know that.

JohnMcLear:
I changed to lockfile v1

JohnMcLear:
building a v1 lockfile gives lint errors regarding http instead of https, I think now is a good time to migrate to v2 and we swallow the warnings.

rhansen:
According to `src/package-lock.json` we've been using socket.io 2.4.1 since commit ee158b0fe5dfdc8201057b5edb40847727958b03.

rhansen:
> * don't know why, but I still see `Access-Control-Allow-Origin: '*'` in the responses of the /socket.io/ path [socketio/socket.io@f78a575](https://github.com/socketio/socket.io/commit/f78a575f66ab693c3ea96ea88429ddb1a44c86c7)

https://github.com/socketio/socket.io/commit/f78a575f66ab693c3ea96ea88429ddb1a44c86c7 was reverted in 2.4.1 (see https://github.com/socketio/socket.io/commit/a1690509470e9dd5559cec4e60908ca6c23e9ba0).

rhansen:
> * find out if we are affected by [socketio/engine.io@001ca62](https://github.com/socketio/engine.io/commit/001ca62cc4a8f511f3b2fbd9e4493ad274a6a0e5) I don't think so, though

We're not affected because we [disable the socket.io cookie](https://github.com/ether/etherpad-lite/blob/81b860bc352e926b0b098ec8c3b9246a21ef2f99/src/node/hooks/express/socketio.js#L39-L58).

rhansen:
> * find out if hanging frontend tests are caused by this

All tests are passing so I don't think so...

webzwo0i:
thanks for your awesome research on this!

PR #4672, opened 2021-01-22
contentcollector fixes

rhansen:
Multiple commits:
* lint: Fix some straightforward ESLint errors
* tests: Re-enable import/export test that is now working
* tests: Delete invalid contentcollector test
* tests: Fix invalid HTML in contentcollector tests
* tests: Delete erroneous `describe()` calls
* tests: Use `assert.deepEqual()` to simplify equality checks
* tests: Replace "expected" with "want", "received" with "got"
* tests: Fix missing call to `done` callback
* openapi: Fix error logging
* contentcollector: Delete unused `domInterface` parameter
* contentcollector: Document the `dom` object
* contentcollector: Factor out call to `.toLowerCase()`
* contentcollector: Fix Element attribute accesses
* contentcollector: Rename `dom` functions for consistency with DOM spec
* contentcollector: Avoid `for..in` iteration of object properties
* contentcollector: Fix iteration over child Nodes
* contentcollector: Delete unnecessary truthiness check
* contentcollector: Invert logic to improve readability
* contentcollector: Use destructuring to improve readability
* contentcollector: Fix collectContentLineText hook
* contentcollector: Delete unnecessary parentheses
* contentcollector: Fix parent node access
* contentcollector: Fix Element tag name fetch
* contentcollector: Skip over non-Text, non-Element Nodes
* contentcollector: Simplify child node access
* contentcollector: Delete unnecessary `dom` functions

cc @webzwo0i

### DIFF
--- a/doc/api/hooks_client-side.md
+++ b/doc/api/hooks_client-side.md
@@ -421,7 +421,20 @@ Things in context:
 4. text - the text for that line
 
 This hook allows you to validate/manipulate the text before it's sent to the
-server side. The return value should be the validated/manipulated text.
+server side. To change the text, either:
+
+* Set the `text` context property to the desired value and return `undefined`.
+* (Deprecated) Return a string. If a hook function changes the `text` context
+  property, the return value is ignored. If no hook function changes `text` but
+  multiple hook functions return a string, the first one wins.
+
+Example:
+
+```
+exports.collectContentLineText = (hookName, context) => {
+  context.text = tweakText(context.text);
+};
+```
 
 ## collectContentLineBreak
 
--- a/src/node/hooks/express/openapi.js
+++ b/src/node/hooks/express/openapi.js
@@ -391,9 +391,9 @@ const defaultResponseRefs = {
 
 // convert to a dictionary of operation objects
 const operations = {};
-for (const resource in resources) {
-  for (const action in resources[resource]) {
-    const {operationId, responseSchema, ...operation} = resources[resource][action];
+for (const [resource, actions] of Object.entries(resources)) {
+  for (const [action, spec] of Object.entries(actions)) {
+    const {operationId, responseSchema, ...operation} = spec;
 
     // add response objects
     const responses = {...defaultResponseRefs};
--- a/src/node/hooks/express/openapi.js
+++ b/src/node/hooks/express/openapi.js
@@ -623,7 +623,7 @@ exports.expressCreateServer = (hookName, args, cb) => {
             } else {
               // an unknown error happened
               // log it and throw internal error
-              apiLogger.error(err);
+              apiLogger.error(err.stack || err.toString());
               throw new createHTTPError.InternalError('internal error');
             }
           });
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -1139,7 +1139,7 @@ function Ace2Inner() {
 
       lastDirtyNode = (lastDirtyNode && isNodeDirty(lastDirtyNode) && lastDirtyNode);
       if (firstDirtyNode && lastDirtyNode) {
-        const cc = makeContentCollector(isStyled, browser, rep.apool, null, className2Author);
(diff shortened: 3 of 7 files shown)


JohnMcLear:
``npx mocha ../tests/backend/specs/contentcollector.js`` failing

rhansen:
> `npx mocha ../tests/backend/specs/contentcollector.js` failing

Fixed.

JohnMcLear:
The big change here I'm just double checking about is the removal of ``cheerio``, will check which plugins this might break if any before merging 👍🏽

JohnMcLear:
```
./ep_announce/node_modules/eslint/package.json:84:    "cheerio": "^0.22.0",
./ep_webrtc/node_modules/eslint/package.json:84:    "cheerio": "^0.22.0",
./ep_limit_maximum_number_of_users_on_pad/node_modules/eslint/package.json:84:    "cheerio": "^0.22.0",
./ep_google_analytics/node_modules/eslint/package.json:84:    "cheerio": "^0.22.0",
./ep_copypad/node_modules/eslint/package.json:84:    "cheerio": "^0.22.0",
./ep_sticky_attributes/node_modules/eslint/package.json:84:    "cheerio": "^0.22.0",
./ep_defaultPadText/node_modules/eslint/package.json:84:    "cheerio": "^0.22.0",
./ep_offline_edit/node_modules/eslint/package.json:84:    "cheerio": "^0.22.0",
./ep_comments_page/node_modules/eslint/package.json:84:    "cheerio": "^0.22.0",
./ep_font_family/node_modules/eslint/package.json:84:    "cheerio": "^0.22.0",
```
The most concerning ones are (ep_comments are ep_readonly_guests)
```
./ep_comments_page/exportHTML.js:3:const $ = require('ep_etherpad-lite/node_modules/cheerio');
./ep_comments_page/exportHTML.js:23:  // https://github.com/cheeriojs/cheerio/issues/1031
./ep_readonly_guest/index.js:3:const $ = require('cheerio');

```

So

1) Why does ``eslint`` require ``cheerio``?
2) How will we handle the transition?  Surely we should keep cheerio in for a few versions until plugins have had time to migrate to JSDom?
3) We should provide either PRs to projects that need to migrate AND/OR docs (quick guides for what to change from>to).
4) Why have our tests not picked up that these projects will fail?

rhansen:
> * Why does `eslint` require `cheerio`?

Looks like eslint has a way to generate a report in HTML. But whether eslint uses cheerio or not doesn't affect us, especially since it's a dev dependency

> * How will we handle the transition?  Surely we should keep cheerio in for a few versions until plugins have had time to migrate to JSDom?

The only plugins that have to change are those that call `collectContent()` (are there any?). Otherwise, they can continue to use cheerio without any issues, so long as cheerio is in their dependencies and they're not piggybacking off of Etherpad's cheerio dependency. (In other words, `require('cheerio')` is fine, `require('ep_etherpad-lite/node_modules/cheerio')` is not.)

> * Why have our tests not picked up that these projects will fail?

As far as I know, the only plugin this affects is ep_comments_page which will be fixed by https://github.com/ether/ep_comments_page/pull/186.

I'm going to move the cheerio to jsdom commit to a separate PR so that it doesn't hold up merging the other commits in this PR.

rhansen:
> I'm going to move the cheerio to jsdom commit to a separate PR so that it doesn't hold up merging the other commits in this PR.

I removed that commit from this PR. Once this PR is merged I'll open a new PR for switching from cheerio to jsdom.

JohnMcLear:
re ran core dumped test

rhansen:
> re ran core dumped test

I've been seeing that a lot lately. There must be a bug in node.js that we're somehow triggering. I don't think it can be a bug in our code because it shouldn't be possible for javascript code to cause node.js to abort.

rhansen:
Re-ran the tests yet again and this time they passed. ¯\\\_(ツ)\_/¯

JohnMcLear:
Merging with the knowledge another PR is coming for the cheerio > jsdom

PR #4679, opened 2021-01-23
Fix dependency installation in GitHub workflows

rhansen:
Multiple commits:
* Update top-level `package-lock.json`
* GitHub workflows: Break list of plugins across multiple lines
* GitHub workflows: Synchronize plugin lists
* GitHub workflows: Install Etherpad deps after installing plugins


webzwo0i:
LGTM
the failing backend test for exportHTML is not caused by this PR afaik

PR #4680, opened 2021-01-26
Fixed possible runtime error if no buttons are present - especially read only mode

bartekwitczak:
Simple bugfix if no buttons are present in toolbar. 

Details: Runtime error occurs with message `Cannot set property 'grouping' of undefined`


JohnMcLear:
Awesome stuff thanks :)  We need test coverage in for this too, how were you able to replicate the problem?  If you can provide STR we can include some testing :)  Thanks again!

PR #4681, opened 2021-01-26
tests: Load times / Performance stats & testing

JohnMcLear:
# What problem does this PR solve?

Answer the questions of:

1. Did a commit/pr/merge slow a part of Etherpad down?
1. Did our work on minification/caching improve page load times.
1. Is maxAge et al being set properly on files.

# How do we solve this problem?

1. Hard coded timeouts for most back & front end tests.
1. Additional CI test coverage for ``minify`` ``false`` & ``true``
1. Low Hanging Linting of the front & back end plugins
1. Exposes to ``stats``:
   - timing of execution (IE time between ``start > minification > accepting http requests``)stats at ``/stats``
   - browser performance IE request speeds (upon installation ``ep_performance_test_hooks``)
   - transferred to the browser file sizes (upon installation ``ep_performance_test_hooks``)

# Tasks:

- [x] Research pre-existing solutions
- [x] Get mocha timing data from CI
- [x] Decide on data endpoints for what really matters
- [x] Create MVP using statically set timeouts (90% done, need to do perf specific frontend tests as per below)
   - [x] Minify set to true performance specific tests / stats into ep_performance_test_hooks.
- [ ] If maxAge is set to 0 and minify is false, then ensure modifications to files are available to pads.

# Tests

- [x] minify true, maxAge 0 etc, first browser visit and then refresh page and see what performance is then.
- [x] Initial page load time[time to get response to first http request] (backend) -- ``firstHttpResponseDuration``
- [x] Time to load all files (frontend) document.ready() -- ``filesLoadedDuration`` from ``performance`` object
- [x] Initial time to edit (when a user can start editing) (frontend) postAceInit? -- ``postAceInitDuration``
- [x] Overall Time/Size of delivered files

# Frontend > Backend architecture for graphing?

We pass frontend test runner data(https://github.com/ether/ep_performance_test_hooks) to the backend.   The backend stat data is available in ``/stats`` and can be consumed from there.  ep_prometheus consumes the stats and pushes them onto ``/metrics`` which can be consumed in Grafana by Prometheus 👍🏽 - I will be setting this up tomorrow on https://video.etherpad.com

# Questions 
## Should we redo the tests with maxAge 0 too?  I think not, I think we will notice if caching is being applied when it shouldn't as it will break our dev cycle.

# Weirdness still to resolve.
- [ ] Tests in iframe appear to have 7* slower response times over those running in the browser - https://github.com/ether/ep_performance_test_hooks/blob/master/static/tests/frontend/specs/tests.js#L837 
- [ ] I have to use ``window.top`` to get the stats between hooks.  See the test specs. :(

# AOB and notes
``window.performance``
image

https://github.com/adamgruber/mochawesome - solution to get JSON of test runner - useful for getting server side test ``duration``.

## How to get browser spceific data
 - https://developer.mozilla.org/en-US/docs/Web/API/Resource_Timing_API/Using_the_Resource_Timing_API

## V2 How to gather benchmark baselines

Could we get a "baseline" and feed that into the this.timeout?

How to test against previous benchmark figures? this.timeout? The idea is that we use % based changes IE alert if more than a 20% increase in time, this and a baseline should do it!  This might have to be a v2 thing tho...

> CI can be ~30% slower, we should get a baseline for this and feed this into test durations.

> One commit could have a huge performance improvement but if we are using averages (say of 100 previous commits) it would never really be consumed. Perhaps 3-5 commits is better? The best thing IMHO is to just get 1 commit history for comparison but ensure that commit has the "baseline" introduced by CI being slower/faster.


bin/getTestRunnerTimings.sh
```diff
@@ -0,0 +1,2 @@
+# Requires gron
+for file in src/mochawesome-report/mochawesome*.json; do gron $file; done | grep -F 'results[' | grep -F 'tests[' | grep -E '(fullTitle|duration)' | sed -r 's/.*= //' | sed -r 's/[;"]//g' | paste -d '$' - - | awk -F'$' '{if ($1 > durations[$2]) durations[$2] = $1} END{for (i in durations) print durations[i], i}' | sed 's/^[^0-9].*/0 \0/'
```
rhansen, line 2:
```suggestion
#!/bin/sh
# Requires gron
for file in src/mochawesome-report/mochawesome*.json; do
  gron $file
done \
  | grep -F 'results[' \
  | grep -F 'tests[' \
  | grep -E '(fullTitle|duration)' \
  | sed -r 's/.*= //' \
  | sed -r 's/[;"]//g' \
  | paste -d '$' - - \
  | awk -F'$' '{if ($1 > durations[$2]) durations[$2] = $1} END{for (i in durations) print durations[i], i}' \
  | sed 's/^[^0-9].*/0 \0/'
```

tests/backend/specs/api/api.js
```diff
@@ -11,7 +12,8 @@ const supertest = require(`${__dirname}/../../../../src/node_modules/supertest`)
 const settings = require(`${__dirname}/../../../../src/node/utils/Settings`);
 const api = supertest(`http://${settings.ip}:${settings.port}`);
 
-const validateOpenAPI = require(`${__dirname}/../../../../src/node_modules/openapi-schema-validation`).validate;
+const validateOpenAPI =
+    require(`${__dirname}/../../../../src/node_modules/openapi-schema-validation`).validate;
```
rhansen, line 16:
```suggestion
const validateOpenAPI = require('ep_etherpad-lite/node_modules/openapi-schema-validation').validate;
```

tests/backend/specs/api/api.js
```diff
@@ -34,13 +37,15 @@ describe(__filename, function () {
...
+      this.timeout(15000);
       api
           .get('/api/openapi.json')
           .expect((res) => {
             const {valid, errors} = validateOpenAPI(res.body, 3);
             if (!valid) {
               const prettyErrors = JSON.stringify(errors, null, 2);
-              throw new Error(`Document is not valid OpenAPI. ${errors.length} validation errors:\n${prettyErrors}`);
+              throw new Error(`Document is not valid OpenAPI.
+                  ${errors.length} validation errors:\n${prettyErrors}`);
```
rhansen, line 48:
```suggestion
              throw new Error('Document is not valid OpenAPI. ' +
                              `${errors.length} validation errors:\n${prettyErrors}`);
```

tests/backend/specs/api/api.js
```diff
@@ -21,6 +23,7 @@ const testPadId = makeid();
 describe(__filename, function () {
   describe('API Versioning', function () {
     it('errors if can not connect', function (done) {
+      this.timeout(150);
```
rhansen, line 26: 150ms is a really short timeout. Intentional? Ditto for the timeouts added below.

rhansen (commented):
Please move the lint fixes to a separate PR.

bin/getTestRunnerTimings.sh
JohnMcLear: I decided against including the gron requirement/command in this PR :)

tests/backend/specs/api/api.js
JohnMcLear: Yea, it's intentional, these timeouts are due to catch changes that make processes/execution slower, they are as close to execution time(on a slow device) as possible.

JohnMcLear:
> 
> 
> Please move the lint fixes to a separate PR.

Got any pro tips for a quick way to do this w/ out going through each file again?

tests/frontend/runner.js
```diff
@@ -69,7 +69,7 @@ $(() => {
       }, 60000 * 3);
 
       const medium = test.slow() / 2;
-      test.speed = test.duration > test.slow()
+      test.speed = test.duratoin > test.slow()
```
JohnMcLear, line 72: wtf? 📄

JohnMcLear:
I ended up sending separate PRs for the majority of this.  The only bit I didn't include was CI because we didn't have a method to run enough test runners AND to get the data and push that data to anything useful.  Our best option might be just monitoring a box such as video.etherpad.com and see how various commits impact on load times?  Further conversation needed but significant progress made.

PR #4683, opened 2021-01-27
HTML import: Replace cheerio with jsdom to simplify contentcollector

rhansen:
Cheerio provides jQuery-like objects but they wrap DOM Node-like objects that are not 100% API compatible with the DOM spec. Because of this, contentcollector, which is used in browsers and in Node.js during HTML import, has until now needed to support two different APIs. This commit modifies HTML import to use jsdom instead of cheerio and simplifies contentcollector.

cc @webzwo0i


JohnMcLear:
Can we please provide an example for plugins to refer to when they want to replace cheerio w/ JSDom too?

For example: https://github.com/ether/ep_comments_page/blob/master/index.js - the replacement here should be quite straight forward and easy to replicate across other plugins.

webzwo0i:
Makes sense to me! we used jsdom-noncontextifiy in the past but switched to cheerio. Based on the old [PR](https://github.com/ether/etherpad-lite/pull/2341) this was due to performance on import and robustness with broken HTML. The latter is a good candidate for another issue - we already install tidy so we can use it on import, too or we don't do anything and expect clean HTML. Import performance should probably be solved when we use another worker process.

I have some more contentcollector tests that I would like to run against this branch, but I don't expect anything to break.

rhansen:
> Can we please provide an example for plugins to refer to when they want to replace cheerio w/ JSDom too?

I don't think it's a good idea to replace cheerio with jsdom elsewhere. The two packages have overlapping functionality but different goals. For contentcollector, jsdom is more appropriate because `collectContent()` is processing individual DOM Elements from a full HTML document. For plugins, cheerio is usually more appropriate because it's better for tweaking HTML snippets from an eejs block.

rhansen:
> I have some more contentcollector tests that I would like to run against this branch, but I don't expect anything to break.

Have you had a chance to run your tests @webzwo0i?

JohnMcLear:
Bump RE conflicts btw :)

JohnMcLear:
Looks like we're experiencing some test fails on this PR btw

JohnMcLear:
I think this one needs rebasing and tests re-running :)

rhansen:
There's a bug that I haven't yet figured out, so it'll be a bit before this is ready.

rhansen:
> There's a bug that I haven't yet figured out, so it'll be a bit before this is ready.

jsdom doesn't support [HTMLImageElement.currentSrc](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/currentSrc) which breaks ep_image_upload. PR citizenos/ep_image_upload#49 will fix ep_image_upload to work with jsdom. Other plugins may have similar problems.

Issue #4684, opened 2021-01-28
Node.js SIGABRT (abort) due to EBADF from epoll_ctl() call in libuv uv__io_poll()


rhansen:
Lately I've noticed frequent node (v14.15.4) crashes when running backend tests. The crashes are caused by a [call to abort() in libuv](https://github.com/nodejs/node/blob/04509d6847a2d164b223a77aed0c41fe9ca60e57/deps/uv/src/unix/linux-core.c#L254) due to `errno` of `EBADF` ("bad file descriptor") set by a call to `epoll_ctl()` in `uv__io_poll()`. The crashes always seem to happen after the mocha tests finish running (I think) but before nyc prints out coverage stats. It doesn't always happen, so it's probably a race condition somewhere. I poked at the core file a bit but didn't see a smoking gun. Here's the stack trace:

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff783c859 in __GI_abort () at abort.c:79
#2  0x0000000000962331 in uv__io_poll (loop=loop@entry=0x7fffdeffcad8, timeout=<optimized out>) at ../deps/uv/src/unix/linux-core.c:254
#3  0x000000000137c418 in uv_run (loop=0x7fffdeffcad8, mode=UV_RUN_ONCE) at ../deps/uv/src/unix/core.c:385
#4  0x0000000000abce1d in node::worker::Worker::Run() ()
#5  0x0000000000abcf28 in node::worker::Worker::StartThread(v8::FunctionCallbackInfo<v8::Value> const&)::{lambda(void*)#1}::_FUN(void*) ()
#6  0x00007ffff7a14609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#7  0x00007ffff7939293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

If anyone out there is familiar with libuv and can help us debug, that would be greatly appreciated.

webzwo0i:
afaik there are only two worker threads involved atm, MinifyWorker and maybe handling soffice.

Looking at some of the failing backend runs I get the impression that it's not nyc but mocha that's having the problems, because the coredump msg is printed before mocha's test summary.

In mocha's changelog: `#4465: Worker processes guaranteed (as opposed to "very likely") to exit before Mocha does; fixes a problem when using nyc with Mocha in parallel mode (@boneskull)`
[...]

PR #4688, opened 2021-01-29
tests: Stop using nyc

rhansen:
There are some problems with nyc:
* The coverage numbers aren't useful in our case because most of the code is executed outside the test process (the test code is mostly API client logic).
* nyc messes with line numbers, which makes it much harder to debug problems.
* We're seeing frequent SIGABRT crashes while nyc is printing the results table. I'm not sure if nyc is the cause of the crashes, or if it's making a race condition worse, or if the crashes have nothing to do with nyc, but we don't lose much by removing it so we might as well see if the crash frequency improves.

Addresses #4684.

### DIFF
--- a/src/package-lock.json
+++ b/src/package-lock.json
@@ -101,181 +101,11 @@
         "@babel/highlight": "^7.10.4"
       }
     },
-    "@babel/core": {
-      "version": "7.12.10",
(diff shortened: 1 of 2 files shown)


JohnMcLear:
https://github.com/ether/etherpad-lite/runs/1790450309?check_suite_focus=true ;(
   No dice in solving #4684

PR #4667, opened 2021-01-21
Low Hanging Lint

JohnMcLear:
A huge, very squashable PR that only does easy/low hanging lint issues.  Removing a lot of headache and overhead so further PRs can be more focused.

Backend tests will fail due to a bug present in develop.

The only thing I'm not sure on is the TidyHTML tests coverage which had to be commented out as the tests seemed relatively broken.  I'm going to look at that but this can be merged anyway so it doesn't start getting huge conflicts.

PR #4695, opened 2021-02-01
Low hanging lint frontend tests

JohnMcLear:


rhansen (approved):
LGTM. I pushed a commit for squashing for you to review. All of these commits can be squashed together into a single commit. (The reason I wanted PR #4667 split into separate commits is to make it easier to find a bug with `git bisect`, but we'll never need to `git bisect` these changes.)

PR #4698, opened 2021-02-01
lint: json.js

JohnMcLear:
- [x] Line 419 left to do :)

Please check I got Line 419 right :)

### DIFF
--- a/bin/doc/json.js
+++ b/bin/doc/json.js
@@ -248,11 +248,11 @@ const processList = (section) => {
   switch (section.type) {
     case 'ctor':
     case 'classMethod':
-    case 'method':
+    case 'method': {
       // each item is an argument, unless the name is 'return',
       // in which case it's the return value.
       section.signatures = section.signatures || [];
-      var sig = {};
+      const sig = {};
       section.signatures.push(sig);
       sig.params = values.filter((v) => {
         if (v.name === 'return') {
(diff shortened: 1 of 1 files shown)


bin/doc/json.js
```diff
@@ -246,13 +246,11 @@ const processList = (section) => {
   // depending on the section type, the list could be different things.
 
   switch (section.type) {
-    case 'ctor':
-    case 'classMethod':
```
rhansen, line 250: Why were these two lines removed? Their removal changes the behavior.

bin/doc/json.js
```diff
@@ -486,7 +485,7 @@ const finishSection = (section, parent) => {
 // Not a general purpose deep copy.
 // But sufficient for these basic things.
 const deepCopy = (src, dest) => {
-  Object.keys(src).filter((k) => !dest.hasOwnProperty(k)).forEach((k) => {
+  Object.keys(src).filter((k) => !Object.prototype.hasOwnProperty.call(k)).forEach((k) => {
```
rhansen, line 488:
```suggestion
  Object.keys(src).filter((k) => !Object.prototype.hasOwnProperty.call(dest, k)).forEach((k) => {
```

bin/doc/json.js
JohnMcLear: My mistake, thanks for spotting!

bin/doc/json.js
JohnMcLear: Wow that was a big one to miss..  Can't beat 11pm sloppy work 🖌️

PR #4700, opened 2021-02-02
lint: frontend specs higher hanging

JohnMcLear:
For squash.

### DIFF
--- a/tests/frontend/specs/select_formatting_buttons.js
+++ b/tests/frontend/specs/select_formatting_buttons.js
@@ -39,7 +39,7 @@ describe('select formatting buttons when selection has style applied', function 
     const originalHTML = helper.padInner$('body').html();
     const $undoButton = helper.padChrome$('.buttonicon-undo');
     $undoButton.click();
-    await helper.waitForPromise(() => originalHTML !== helper.padInner$('body').html());
+    await helper.waitForPromise(() => helper.padInner$('body').html() !== originalHTML);
   };
 
   const testIfFormattingButtonIsDeselected = function (style) {
@@ -97,7 +97,7 @@ describe('select formatting buttons when selection has style applied', function 
     e.which = key.charCodeAt(0); // I, U, B, 5
     inner$('#innerdocbody').trigger(e);
     await helper.waitForPromise(
-      () => originalHTML !== helper.padInner$('body').html());
+        () => helper.padInner$('body').html() !== originalHTML);
   };
 
   STYLES.forEach((style) => {
--- a/tests/frontend/specs/select_formatting_buttons.js
+++ b/tests/frontend/specs/select_formatting_buttons.js
@@ -39,7 +39,7 @@ describe('select formatting buttons when selection has style applied', function 
     const originalHTML = helper.padInner$('body').html();
     const $undoButton = helper.padChrome$('.buttonicon-undo');
     $undoButton.click();
-    await helper.waitFor(() => originalHTML !== helper.padInner$('body').html());
+    await helper.waitForPromise(() => originalHTML !== helper.padInner$('body').html());
   };
 
   const testIfFormattingButtonIsDeselected = function (style) {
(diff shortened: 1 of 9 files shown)


tests/frontend/specs/select_formatting_buttons.js
```diff
@@ -102,8 +107,8 @@ describe('select formatting buttons when selection has style applied', function
         applyStyleOnLineAndSelectIt(FIRST_LINE, style, done);
       });
 
-      after(function () {
-        undo();
+      after(async function () {
+        await undo();
```
webzwo0i, line 111: because the undo function is not async, does this work?

JohnMcLear: Didn't I make undo async?

tests/frontend/specs/select_formatting_buttons.js
```diff
@@ -125,33 +130,28 @@ describe('select formatting buttons when selection has style applied', function
...
-    });
-
-    // clean the style applied
-    after(function () {
-      applyStyleOnLine(style, FIRST_LINE);
-    });
 
     it('selects the style button', function (done) {
+      applyStyleOnLine(style, FIRST_LINE);
+      helper.waitForPromise(() => isButtonSelected(style) === true);
```
webzwo0i, line 136: needs an await, or it will run the code below helper.waitForPromise even if the Promise is rejected (afaik this would result in a green test, with an printed error in case the Promise is rejected)

tests/frontend/specs/xxauto_reconnect.js
```diff
@@ -34,9 +34,11 @@ describe('Automatic pad reload on Force Reconnect message', function () {
   });
 
   context('and user clicks on Cancel', function () {
-    beforeEach(function () {
+    beforeEach(async function () {
       const $errorMessageModal = helper.padChrome$('#connectivity .userdup');
       $errorMessageModal.find('#cancelreconnect').click();
+      await helper.waitFor(
+          () => helper.padChrome$('#connectivity .userdup').is(':visible') === true);
```
webzwo0i, line 41: this needs await helper.waitForPromise

rhansen:
I pushed a bunch of fixes and refinements to be squashed.

rhansen:
Can you move non-lint changes (timeout changes, asyncification, addition of waits for HTML changes) to a separate PR? Or at least put them in separate commits.

JohnMcLear:
Doh, I thought you wanted a commit per file when linting.

The async et al was lint fixes, or at least it was things that I had to change to please eslint so I consider those lint changes, is that not something we're considering is the case?  My bad if so.  It makes sense if asyncify should be a separate commit tho as it's changing the behavior of the test, not just the readability. 

I think with the effort required to split this into separate commits vs maintainability I think we're good to squash merge this entire branch as is with the knowledge that if we identify a test as failing we have a single commit we can bisect back to and go from there.  There are very few behavior changes to each file so identifying the test should be elementary (it will be the one failing) and then looking at the history should show just a few lines modified of that file...

In the future I will endeavor to commit behavior changes in a separate commit to readability stuff.  Either way this is good progress so thanks :)

JohnMcLear:
Closed in favor of https://github.com/ether/etherpad-lite/pull/4718/files - handling the conflicts was tooooo much ;\

rhansen:
> Doh, I thought you wanted a commit per file when linting.

Only for non-test code to make it easier to bisect a problem. There's no need to bisect changes to tests because changes to tests can never break Etherpad itself.

> The async et al was lint fixes, or at least it was things that I had to change to please eslint so I consider those lint changes

Are you referring to [mocha/no-synchronous-tests](https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-synchronous-tests.md)? That rule is satisfied if the function takes a done callback, so there's no need to convert such functions to async in order to appease that rule.

> Closed in favor of https://github.com/ether/etherpad-lite/pull/4718/files - handling the conflicts was tooooo much ;\

Please try to get familiar with rebase—#4718 is missing all of the fixes I pushed to this PR's branch.

JohnMcLear:
Ack, noted and will do, rebase is on my todo for this coming week.

PR #4707, opened 2021-02-03
Restructure of folders

JohnMcLear:
# Todo:
- [x] Test require paths
- [x] Update bin scripts to use correct require paths

# As per this conversation

> I think the first step is to move bin/ and tests/ to src/ because we can add symlinks in the old locations to mostly avoid negative fallout
installing plugins to plugins/ is more disruptive
but we can support both node_modules/ and plugins/node_modules/ for a while until everyone has updated scripts and whatnot to use the new location
actually I wonder how much plugins/ would help. If we want plugins to keep their peer dependency on ep_etherpad-lite, npm will still want to recursively ensure that all of ep_etherpad-lite's dependencies are installed which means it might still be slow.
But maybe not as slow—currently npm is hoisting dependencies from src/node_modules to node_modules and it wouldn't do that if plugins were in plugins/node_modules
I'm not too worried about a PR for moving bin/ and tests/ conflicting with other PRs because Git is pretty good at detecting moves. We would need to rebase PRs, but we should be doing that regularly anyway
JohnMcLear01/24/2021
I tried a rebase this week and it/I failed, I still need to get skilled up, it's 100% a user error.
how would symlinks work on windows based deployments btw?
rhansen01/24/2021
I found rebase to be difficult to grok until I (1) understood Git's object model, and (2) understood how the commands manipulated that model
good question about symlinks on windows. it's been >15y since i've used windows so I don't know how symlinks might work
we currently have a symlink checked into the repo: node_modules/ep_etherpad-src is a checked-in symlink to ../src
i haven't heard any windows users complaining about that
so maybe it works just fine


rhansen:
I pushed a couple of commits. Please take a look.

src/bin/plugins/checkPlugin.js
```diff
@@ -429,7 +429,7 @@ fs.readdir(pluginPath, (err, rootFiles) => {
   // The ep_etherpad-lite peer dep must be installed last otherwise `npm install` will nuke it. An
   // absolute path to etherpad-lite/src is used here so that pluginPath can be a symlink.
   execSync(
-      `${npmInstall} --no-save ep_etherpad-lite@file:${__dirname}/../../src`, {stdio: 'inherit'});
+      `${npmInstall} --no-save ep_etherpad-lite@file:${__dirname}/../../`, {stdio: 'inherit'});
```
rhansen, line 432: Shouldn't this still be `../../src`?

JohnMcLear: ``src`` broke when I ran it, changed it to the ``../../`` and it worked...  Didn't go further into than that!

rhansen: Oops, I misread the code. `__dirname` is the directory containing the `checkPlugin.js` script, not the plugin's directory. Your change is correct.

PR #4712, opened 2021-02-05
restructure: Prefix `bin/` and `tests/` with `src/`

rhansen:
This is a follow-up to commit 2ea8ea12754c92ea06f22428f1e86341dcf00d99.


JohnMcLear:
The files that modify ``settings.json`` IE ``runnreBackend.sh`` / ``runnerFrontend.sh`` et al are pointing to the wrong location.

```
try cd "${MY_DIR}/../../../"
```
should be
```
try cd "${MY_DIR}/../../../../"
```
?

JohnMcLear:
n1 thanks :)

PR #4723, opened 2021-02-07
ace2_inner: Delete special arrow key handling

rhansen:
I can't see any reason this would be necessary, and it appears to not behave as intended (`scroll.scrollWhenPressArrowKeys()` is not invoked after a continuously held arrow key is finally let up).


JohnMcLear:
Will look into this, must be a reason this logic existed, prolly some weird IE6 thing but will check.

rhansen:
Friendly ping.

JohnMcLear:
Will look at this today :)

JohnMcLear:
Thanks, apologies for delay in merge :)

PR #4729, opened 2021-02-07
tests: disabled checks shouldnt change behavior

JohnMcLear:


rhansen:
This fix isn't quite right because you're testing for lack of change and `waitForPromise()` will immediately succeed even if the HTML later changes. An approach I've seen elsewhere: Invert the check (so continue to use `!==` ) but fail the test if the `waitForPromise()` check succeeds and pass the test if the `waitForPromise()` check fails. Use a short timeout so that it doesn't slow down the tests too much, but long enough to notice any changes.

src/tests/frontend/specs/ordered_list.js
```diff
@@ -32,13 +32,12 @@ describe('ordered_list.js', function () {
...
         });
       });
-
       context('and pad shortcut is disabled', function () {
         beforeEach(async function () {
           const originalHTML = helper.padInner$('body').html();
           makeSureShortcutIsDisabled('cmdShiftN');
           triggerCtrlShiftShortcut('N');
-          await helper.waitForPromise(() => helper.padInner$('body').html() !== originalHTML);
+          await helper.waitForPromise(() => helper.padInner$('body').html() === originalHTML);
```
rhansen, line 40: This is what I mean:

```suggestion
          try {
            // The HTML should not change. Briefly wait for it to change and fail if it does change.
            await helper.waitForPromise(() => helper.padInner$('body').html() !== originalHTML, 500);
          } catch (err) {
            // We want the test to pass if the above wait timed out. (If it timed out that
            // means the HTML never changed, which is a good thing.)
            // TODO: Re-throw non-"condition never became true" errors to avoid false positives.
          }
          // This will fail if the above `waitForPromise()` succeeds.
          assert.equal(helper.padInner$('body').html(), originalHTML);
```

JohnMcLear:
updated w/ proposed fix.

PR #4731, opened 2021-02-08
wip - checkPlugins: remove travis.yml and add github badges if they aren't present

JohnMcLear:
This is the first step to just using Github badges instead of Travis.  We will have another step that will remove Travis at a later date.

- [x] only run on autofix


rhansen:
I don't see any change in this PR that would add GitHub actions if not present. Isn't that already the case on `develop`?

JohnMcLear:
My bad, badges not actions, and don't merge yet please :)

rhansen:
* Delete `src/bin/plugins/lib/travis.yml`.
* The PR title says "remove travis" but it looks like it's just not adding travis anymore (if it's currently there it doesn't touch it).

JohnMcLear:
done

rhansen (approved):

Issue #4748, opened 2021-02-11
OL numbers are broken and so are tests


JohnMcLear:
- [x] Bisect to find offending - f0a77cb98cd57a99934a2132a204bcf90d396812
  - [x] fix logic.
- [x] Fix test

PR #4749, opened 2021-02-11
Fix OL list items not increasing

JohnMcLear:
#4748 is a bug report where ordered list items were not properly formatted due to a lazy check.  The lazy check was ``x == 1`` and when we linted we changed it to ``x === 1``, the ``1`` was coming from a string check so it was typeOf string which meant when we did a more strict check it never returned true causing line number snot to incremented.

This PR.

* Fixes bug by parsing the string to an integer
* Fixes test by adding a new test which covers this edge case, although I'm not sure why it wasn't failing before on other tests..   we can investigate why once this is merged and we have order restored so can release.

### DIFF
--- a/src/static/js/domline.js
+++ b/src/static/js/domline.js
@@ -101,7 +101,7 @@ domline.createDomLine = (nonEmpty, doesWrap, optBrowser, optDocument) => {
             postHtml = `</li></ul>${postHtml}`;
           } else {
             if (start) { // is it a start of a list with more than one item in?
-              if (parseInt(start[1]) === 1) { // if its the first one at this level?
+              if (Number.parseInt(start[1]) === 1) { // if its the first one at this level?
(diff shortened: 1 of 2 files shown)


src/static/js/domline.js
```diff
@@ -101,7 +101,7 @@ domline.createDomLine = (nonEmpty, doesWrap, optBrowser, optDocument) => {
             postHtml = `</li></ul>${postHtml}`;
           } else {
             if (start) { // is it a start of a list with more than one item in?
-              if (start[1] === 1) { // if its the first one at this level?
+              if (parseInt(start[1]) === 1) { // if its the first one at this level?
```
rhansen, line 104: Nit:

```suggestion
              if (Number.parseInt(start[1]) === 1) { // if its the first one at this level?
```

rhansen (approved):

PR #4750, opened 2021-02-11
security: limit socketio to 1M chars

JohnMcLear:
Temporary band aid for SocketIO security issue.

### DIFF
--- a/src/node/hooks/express/socketio.js
+++ b/src/node/hooks/express/socketio.js
@@ -56,6 +56,7 @@ exports.expressCreateServer = (hookName, args, cb) => {
      *   https://github.com/socketio/socket.io/issues/2276#issuecomment-147184662 (not totally true, actually, see above)
      */
     cookie: false,
+    maxHttpBufferSize: 10E3,
   });
 
   io.use((socket, next) => {
--- a/src/tests/frontend/specs/chat_load_messages.js
+++ b/src/tests/frontend/specs/chat_load_messages.js
@@ -8,7 +8,7 @@ describe('chat-load-messages', function () {
     this.timeout(60000);
   });
 
-  it('adds a lot of messages', function (done) {
+  it('adds a lot of messages', async function () {
     const chrome$ = helper.padChrome$;
     const chatButton = chrome$('#chaticon');
     chatButton.click();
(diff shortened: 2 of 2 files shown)


webzwo0i:
LGTM. Not sure why tests are failing though

rhansen (approved):

Exploder98:
@rhansen @JohnMcLear Hi! It seems that this change has limited the websocket message size to 10 KB, instead of 1M (which is what the title and commit messages say). This pretty much broke `ep_image_upload` because most images are larger than 10 KB. In addition, trying to upload an image that is larger than that causes a reconnect loop, because the server closes the socket and the client tries to repeatedly upload the added image.

JohnMcLear:
Whoa that's a really good spot!  Thanks!  Weird that I didn't spot that in testing or that ep_image_upload didn't fire an error off..

1. Going to increase to 1Mb.
2. Going to include ep_image_upload in CI.

webzwo0i:
yeah awesome! Even more wtf-why-didn't-I-saw-this-coming: If pasting too many characters it's in an endless loop, too. It's probably the same problem as with ep_image_upload, the client assembles the changeset, puts it into it's "to be committed" queue and there you go.

We need a way to communicate the limit to the clients and the client to split it's changeset. For image_upload we need a workaround like indicate "partial image" not to be rendered.
I know, it's hacky, but increasing the limit will defeat it's purpose. I'm not sure if we can do anything else, e.g. limit the number of packets that are allowed per message - that's probably something to discuss with engine.io people?

I tested on a fairly fast machine (i7-7700) and 1M full of empty messages costs ~15 seconds full CPU.

JohnMcLear:
Oh boy.

So here is my suggestion..

ep_image_upload could modify a settings exposed message limit value, meaning we can keep the initial baseline value low(say 10e3) and if the instance wants to support images then it takes on additional threat/risk threshold and we document that in the README.  This solves 95% of the problem, as the majority of the instances don't care for images and 10k chars is probably fine?

2) Then go ahead and work on splitting messages.  But do this work once we have left SocketIO behind?  I just don't want to put any more effort into Socket IO because it feels like we take 1 step forward and 10 steps back always...

JohnMcLear:
Also see https://github.com/ether/etherpad-lite/pull/4760

rhansen:
> ep_image_upload could modify a settings exposed message limit value

I think it would be better for ep_image_upload to expose a new POST endpoint and use XMLHttpRequest to upload the image.

JohnMcLear:
Image upload does do that if that setting is enabled but it also allows for images to be inserted as base64 into the pad.

PR #4751, opened 2021-02-11
WIP: eejs update to support latest ejs

JohnMcLear:
This patch will mean that plugins source wont render :(

### DIFF
--- a/src/node/eejs/index.js
+++ b/src/node/eejs/index.js
@@ -50,7 +50,11 @@ exports._exit = (b, recursive) => {
 exports.begin_block = (name) => {
   exports.info.block_stack.push(name);
   exports.info.__output_stack.push(exports.info.__output.concat());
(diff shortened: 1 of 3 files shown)


webzwo0i:
because of https://github.com/mde/ejs/pull/470
`__output` is no array anymore. however, I tried to convert the code to work with a string (slice instead of splice, no join, += instead of push etc) but it does not work. weird

JohnMcLear:
yea I tried the same ;(

webzwo0i:
Giving it another try now. The problem is we need access to that (internal) variable, probably for our custom begin/end blocks. If it's not working, we should better understand the security issue (I don't completely understand this atm): I don't think we're using user-generated render/filenames or if we do, we can sanitize it ourselves.

Because you asked regarding replacing ejs: I do not have this on my agenda for new packaging, as I hope there will be a working loader.

PR #4755, opened 2021-02-11
stats: gather startup stats

JohnMcLear:
This is a move from the previous stats/performance PR.

The idea is we expose stats for startup events so we can review how our efforts are actually impacting instances.

### DIFF
--- a/src/node/hooks/express/static.js
+++ b/src/node/hooks/express/static.js
@@ -5,8 +5,12 @@ const plugins = require('../../../static/js/pluginfw/plugin_defs');
 const CachingMiddleware = require('../../utils/caching_middleware');
 const Yajsml = require('etherpad-yajsml');
 const _ = require('underscore');
+const stats = require('../../stats');
 
 exports.expressCreateServer = (hookName, args, cb) => {
+  const expressDurations = {};
+  stats.gauge('expressDurations', () => expressDurations);
+  const preMinification = Date.now();
   // Cache both minified and static.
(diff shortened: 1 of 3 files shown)


src/node/hooks/express/static.js
```diff
@@ -15,6 +19,9 @@ exports.expressCreateServer = (hookName, args, cb) => {
   // file-specific hacks for ace/require-kernel/etc.
   args.app.all('/static/:filename(*)', minify.minify);
 
+  expressDurations.minification = Date.now() - preMinification;
```
rhansen, line 22: This is going to be super tiny duration because no processing is done between `preMinification` and this line—the above is only registering callbacks to be called later.

rhansen (approved):
Overall looks good, but I wonder about the scope of these metrics. I think it might be more effective to start with a single overall startup time metric, and maybe add a few specific duration metrics for things we know contribute heavily to startup time such as plugin loading. If we think overall time is too high then we can do profiling to see what the heavy contributors are. If it's OK, we can just keep an eye on it as we make changes to make sure it doesn't get out of control.

JohnMcLear:
Ugh, merge conflicts due to the PromiseAll from the minification merge :\

rhansen:
> Ugh, merge conflicts due to the PromiseAll from the minification merge :\

The `expressConfigure` and `expressCreateServer` hooks are run in parallel now so this is a genuine semantic conflict. Rather than two separate metrics for the two hooks you'll probably want a single metric that covers both.

PR #4757, opened 2021-02-12
tests: Try all supported node versions during backend tests

webzwo0i:
We should run our tests with all node versions that we currently support. This PR uses v10/12/14/15 for our backend tests. Maybe adding different versions to the admin tests would be clever too, as they use npm

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -135,7 +135,7 @@ jobs:
 
     - name: Install Etherpad plugins
       run: >
-        npm install
+        npm install --no-save
         ep_align
         ep_author_hover
         ep_cursortrace
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -13,13 +13,18 @@ jobs:
     name: Linux without plugins
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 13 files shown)


rhansen:
We have a bunch of tests that use `assert.match()` or `assert.doesNotMatch()`, which were added in Node.js 12. Those will have to be modified first.

We may want to instead just drop support for Node.js.

webzwo0i:
I added `assert.match` and `assert.doesNotMatch` to a assert-helper file that we can require instead of builtin assert.
Also running different node versions now for the admin frontend tests to check if npm works in all cases.

RE dropping support for nodejs v10: I don't have an overview around different linux distributions, but Buster (debian's current stable) only has v10. EOL of v10 is 2021-04-30. After that dat4 I'm +1 for dropping support. However, if we can afford it (ie we depend it for tests etc) I would try to keep backwards compatibility as long as possible.

rhansen:
I rebased onto latest `develop` and made some fixes. I think this is very close to done; I think #4790 might have to be merged first to fix Node.js 15 runs.

JohnMcLear:
Please rebase

JohnMcLear:
Sorry, please rebase again, should be good to go after :)

PR #4758, opened 2021-02-12
WIP: remove pad_utils.js from pad.html and include escapeHTML inline

JohnMcLear:
I'm not sure this is actually working..  

I still see pad_utils.js 307 but now to pad.js?


webzwo0i:
A test (ie. putting `throw new Error('broken')` into a file that is loaded before `pad.js:init` (as this calls `padutils.setupGlobalExceptionHandler()`) to cause an error during load) shows, that `padutils` is not defined.
Also the method is called `escapeHtml`, not `escapeHTML`. So this should never have worked. It was introduced in https://github.com/ether/etherpad-lite/pull/3647 and fixed in https://github.com/ether/etherpad-lite/commit/695c2d2e84b3f0046b8aaa715db21867465466db Notice the commit message: imo the right thing would be to configure require earlier, but the commit message says this does not work.

Also I'm not sure if in this context (JS context that sets HTML context) we really need to replace characters or if encodeURI is enough or if we even could match the source/URL argument like `[a-zA-Z0-9\.\?=]`. The code is used to display line and source of an error during load, so we can do some kind of whitelist.

Let me push another PR that loads require earlier

webzwo0i:
Just an update:
The pad_utils removal is finished in https://github.com/ether/etherpad-lite/compare/use-less-requests?expand=1#diff-a8c600f1edd6ff1da416450e43cc70a13fef8ec23455da111d2c8e131315aabd 
I'm doing more package juggling/require reordering on that branch. The goal is to have 1 pad.js and 1 timeslider.js and 1 other package that contains all the stuff that's shared between pad and timeslider. This way we can save some requests and bandwidth without changing internals of yajsml or require-kernel

JohnMcLear:
Closed this as @webzwo0i is landing these PRs

PR #4760, opened 2021-02-12
socketio: include image upload frontend test coverage and increase socketio limit to something workable

JohnMcLear:
The initial value was 10k, @Exploder98 rightly pointed out this value is far too low.

@webzwo0i I hope this is still inline with your testing on this issue?


rhansen:
The ep_upload_image tests are failing with this PR but I merged it anyway to get the 1MiB fix.

Issue #4443, opened 2020-10-27
Possible page load improvements


webzwo0i:
I just checked the network tab and the following requests are not optimal, while https://github.com/ether/etherpad-lite/pull/4442 is the most important one:

- socket.io has no max-age header. instead it uses if-none-match header and gets a 304.
- locales.json has no max-age header. instead it uses if-none-match header but still gets 200. (why?)
- plugin-definitions.json has no max-age header. 
- favicon.ico has no max-age header. response uses etag, but the request does not use if-none-match.

imo the two more interesting:
- pad_utils.js is included in pad.html to use `escapeHTML` (so no code duplication happens?). It results in a 307 redirect to the timeslider.js package (!). It's also packaged in pad.js, however timeslider.js is defined later in tar.json so the latter is picked. This means, whenever a normal pad is visited the timeslider package (100kb+) is downloaded after following the redirect. The timeslider package however is cached, so after a first load it's served from browser cache. However, the 307 redirect is still present. It's easy to put pad_utils.js in an own package or to put pad.js after timeslider.js in tar.json - but I think frontend tests are necessary first. (pad_utils.js is used when showing a meaningful error message) Alternatively we could inline all the code for escapeHTML and drop pad_utils from pad.html (https://github.com/ether/etherpad-lite/blob/develop/src/templates/pad.html#L446)

- client_plugins.js also results in a 307, redirecting to ace2_common.js I am not sure from where it's included, may be https://github.com/ether/etherpad-lite/blob/develop/src/static/js/ace.js#L266

JohnMcLear:
I'm going to make something that gets timing performance (startup, page load, availability to edit etc) and push this as a test.

JohnMcLear:
https://github.com/ether/etherpad-lite/pull/4681 now has test coverage in for performance, obviously values in the plugin will need modifying to catch the faults you have spotted.  I'm gonna consider my input on this closed off as you have the tools now for your minification/caching work :)

JohnMcLear:
@webzwo0i did you have a PR you wanted to land for this?

JohnMcLear:
Adding my findings.

 > socket.io has no max-age header. instead it uses if-none-match header and gets a 304. **<- confirmed.**
 > locales.json has no max-age header. instead it uses if-none-match header but still gets 200. (why?) <- **<- confirmed.**
 > plugin-definitions.json has no max-age header.  <- **<- confirmed.**
[...]

PR #4761, opened 2021-02-12
performance: maxAge for favicon and plugin definitions

JohnMcLear:
Resolve favicon/ plugin-definition issue at https://github.com/ether/etherpad-lite/issues/4443

### DIFF
--- a/src/node/hooks/express/static.js
+++ b/src/node/hooks/express/static.js
@@ -65,7 +65,7 @@ exports.expressCreateServer = async (hookName, args) => {
       clientPlugins[name] = {...plugins.plugins[name]};
       delete clientPlugins[name].package;
     }
-    res.header('Content-Type', 'application/json; charset=utf-8');
+    res.setHeader('Content-Type', 'application/json; charset=utf-8');
     res.setHeader('Cache-Control', `public, max-age=${settings.maxAge}`);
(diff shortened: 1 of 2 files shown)


src/node/hooks/express/static.js
```diff
@@ -66,6 +66,7 @@ exports.expressCreateServer = async (hookName, args) => {
       delete clientPlugins[name].package;
     }
     res.header('Content-Type', 'application/json; charset=utf-8');
```
rhansen, line 68: Looks like `.header()` is an old Express 3.x method. Might as well update to the new hotness while you're here:

```suggestion
    res.setHeader('Content-Type', 'application/json; charset=utf-8');
```

rhansen (approved):

webzwo0i:
FYI favicon cache does not work 100% under chrome, but it seems it's not our fault: https://bugs.chromium.org/p/chromium/issues/detail?id=969033
When reloading a page it requests the favicon, but when opening a new tab after the cache is filled, it does not. Firefox behaves correctly after this patch.

The solution for plugin-definition.json is incomplete, as we instruct jquery to not cache AJAX requests. It does so by appending the current time as query (?_=...). After adding randomVersionString instead I noticed, that the version string is not regenerated after a plugin is installed via /admin.

PR #4765, opened 2021-02-12
editor: update lines with zero height on sidediv

### LINKED ISSUE #4764: Consider hidden lines during sidediv update
**Describe the bug**

The `updateLineNumbers` function is called to calculate the size of each line after editing/removing a text and updates its respective line-height on the `#sidediv` container. The problem happens when we have a scenario where it is possible to have lines with zero height. An example would be a plugin that allows toggling the visualization of lines created by a specific author, hiding all other lines from other authors. In this scenario, the line height within `#sidediv` is not updated properly. See the code below:

https://github.com/ether/etherpad-lite/blob/73d31b12a868fbc4c478c4a2d3ed7230d178df55/src/static/js/ace2_inner.js#L3860-L3866

The `if(lineHeights[currentLine])` condition is only true for height ​​greater than zero. However, hidden lines would have 0 height, which would not satisfy the `if` condition and therefore not update its height in `#sidediv`.

To help the understanding, the image below shows the line number on the left side and the `#sidediv` HTML code on the right side.

image

ingoncalves:
Our proposal is to simply modify the `if` condition, allowing us to update lines with zero height.
This change does not have any side effects for the current cases of updating the `#sidediv`.

Closes #4764

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -3858,7 +3858,7 @@ function Ace2Inner() {
     // Apply height to existing sidediv lines
     currentLine = 0;
     while (sidebarLine && currentLine <= lineNumbersShown) {
-      if (lineHeights[currentLine]) {
+      if (lineHeights[currentLine] !== undefined) {
         sidebarLine.style.height = `${lineHeights[currentLine]}px`;
       }
       sidebarLine = sidebarLine.nextSibling;


JohnMcLear:
If you visit /tests/frontend do all the other tests pass with this change made?

src/static/js/ace2_inner.js
```diff
@@ -3858,7 +3858,7 @@ function Ace2Inner() {
     // Apply height to existing sidediv lines
     currentLine = 0;
     while (sidebarLine && currentLine <= lineNumbersShown) {
-      if (lineHeights[currentLine]) {
+      if (lineHeights[currentLine] !== undefined) {
```
rhansen, line 3861: This is more idiomatic:

```suggestion
      if (lineHeights[currentLine] != null) {
```

rhansen (approved):

rhansen:
Merged as commit 3ff244f7a046af0916d6cdc4186a4fc4602741ad.

src/static/js/ace2_inner.js
ingoncalves: By definition, an `Array` returns `undefined` if the given index can not be found. So I think `undefined`  is the more suitable option.

rhansen: Yes it returns `undefined`, but loose comparison with `null` is more idiomatic (with good reasons) and matches the style of the rest of Etherpad (most of it anyway).

PR #4740, opened 2021-02-09
[Snyk] Security upgrade socket.io from 2.4.1 to 3.0.0

JohnMcLear:
<h3>Snyk has created this PR to fix one or more vulnerable packages in the `npm` dependencies of this project.</h3>

merge advice

#### Changes included in this PR

- Changes to the following files to upgrade the vulnerable dependencies to a fixed version:
    - src/package.json
    - src/package-lock.json

##### With an upgrade:
Severity                   | Priority Score (*)                   | Issue                   | Breaking Change                   | Exploit Maturity
:-------------------------:|-------------------------|:-------------------------|:-------------------------|:-------------------------
high severity  |  **696/1000**  <br/> **Why?** Proof of Concept exploit, Has a fix available, CVSS 7.5  | Denial of Service (DoS) <br/>[SNYK-JS-ENGINEIO-1056749](https://snyk.io/vuln/SNYK-JS-ENGINEIO-1056749) |  Yes  | Proof of Concept 

(*) Note that the real score may have changed since the PR was raised.

<details>
  <summary><b>Commit messages</b></summary>
  </br>
  <details>
    <summary>Package name: <b>socket.io</b></summary>
    The new version differs by 57 commits.</br>
    <ul>
      <li><a href="https://snyk.io/redirect/github/socketio/socket.io/commit/1af3267e3f5f7884214cf2ca4d5282d620092fb0">1af3267</a> chore(release): 3.0.0</li>
      <li><a href="https://snyk.io/redirect/github/socketio/socket.io/commit/02951c4391aabf5cdf746633b909e2b673324dc7">02951c4</a> chore(release): 3.0.0-rc4</li>
      <li><a href="https://snyk.io/redirect/github/socketio/socket.io/commit/54bf4a44e9e896dfb64764ee7bd4e8823eb7dc7b">54bf4a4</a> feat: emit an Error object upon middleware error</li>
      <li><a href="https://snyk.io/redirect/github/socketio/socket.io/commit/aa7574f88471aa30ae472a5cddf1000a8baa70fd">aa7574f</a> feat: serve msgpack bundle</li>
      <li><a href="https://snyk.io/redirect/github/socketio/socket.io/commit/64056d6616ced0e4655d513b603d3566b38ec8fd">64056d6</a> docs(examples): update TypeScript example</li>
[...]

PR #4767, opened 2021-02-13
socket.io: Manually track client connections/disconnections

rhansen:
This change is required for socket.io 3.x because in 3.x `io.sockets.clients()` no longer returns all client Socket objects.

Necessary (but probably not sufficient) before #4740 can be merged.

### DIFF
--- a/src/node/hooks/express/socketio.js
+++ b/src/node/hooks/express/socketio.js
@@ -1,31 +1,49 @@
 'use strict';
 
+const events = require('events');
 const express = require('../express');
+const log4js = require('log4js');
 const proxyaddr = require('proxy-addr');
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Frontend test failing were image upload test errors, I'm not sure why it's erroring, I expected https://github.com/ether/etherpad-lite/pull/4760/files would fix the test but clearly not.  Will review.

rhansen:
> Frontend test failing were image upload test errors, I'm not sure why it's erroring

The ep_upload_image test image is only ~300KiB so it should work. ¯\\\_(ツ)\_/¯

JohnMcLear:
After rebasing/merging would you prefer I click delete or do you want me to leave that to you?

PR #4771, opened 2021-02-13
deps: use ci --no-optional flags, this might break some things

JohnMcLear:
might break some things as it requires npm 6.31.4


rhansen:
This could break users. It might be better to just change our CI tests to use `--no-optional`.

JohnMcLear:
I had a think and I'm not sure how it can break users, unless they use optional sqlite?

rhansen:
> unless they use optional sqlite?

Exactly. I'm one of those users that uses sqlite. :slightly_smiling_face: 

Though I always explicitly do `npm i sqlite3` so this wouldn't affect me personally.

JohnMcLear:
Let's try with it in for a few days :P  See if someone complains

PR #4778, opened 2021-02-14
maxHttpBufferSize as a setting

JohnMcLear:
By default we limit insertion size to ~10k.

ep_image_upload modifies this value in https://github.com/citizenos/ep_image_upload/commit/aac99c998780f1fd45554e3956e8213dcfee8f94 to a much larger number.

This is a stop gap solution until we chop socket messages up instead of allowing large ones to be sent.


rhansen (approved):
I pushed a commit containing changes for squash. If the changes look good to you, feel free to squash merge.

PR #4779, opened 2021-02-14
Bugfix for latest version of EJS

redhog:
This PR works around the breakage that happens if we upgrade to the latest version of EJS, which changes the type of __output from array to string.

### DIFF
--- a/src/node/eejs/index.js
+++ b/src/node/eejs/index.js
@@ -49,19 +49,18 @@ exports._exit = (b, recursive) => {
 
 exports.begin_block = (name) => {
   exports.info.block_stack.push(name);
-  exports.info.__output_stack.push(exports.info.__output.concat());
-  exports.info.__output.splice(0, exports.info.__output.length);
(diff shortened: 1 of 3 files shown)


JohnMcLear:
Some EEJS notes we could add to the index.js or a README.md in the folder

```
/*
 * EEJS is Similar to Django template engine where you can have multiple templates
 * and define blocks within those that can have content modified.
  Example

    test.html
      A
      <% e.begin_block("modals"); %>
        123
      <% e.end_block(); %>
    B

    test2.html
       <%  require("test.html"); %>
      <% e.begin_block("modals"); %>
        456
      <% e.end_block(); %>

    output of test2.html
    A
    456
    B
    
    
    After a block is rendered to text (content):
        
        const args = {content, renderContext};
         hooks.callAll(`eejsBlock_${name}`, args);
         
         is called and can modify content before it's used
    
    // contents of test with the ..... replaced with whatever is in test2.html

```

rhansen (approved):

PR #4781, opened 2021-02-15
fix for caching plugin-definitions

webzwo0i:
see https://github.com/ether/etherpad-lite/pull/4761#issuecomment-779016536

Sets the randomVersionString earlier when delivering pad.html/timeslider.html and prevents jquery from adding a query parameter to circumvent the cache.

### DIFF
--- a/src/node/utils/Settings.js
+++ b/src/node/utils/Settings.js
@@ -42,20 +42,6 @@ const _ = require('underscore');
 exports.root = absolutePaths.findEtherpadRoot();
 console.log(`All relative paths will be interpreted relative to the identified Etherpad base dir: ${exports.root}`);
 
-/*
- * At each start, Etherpad generates a random string and appends it as query
- * parameter to the URLs of the static assets, in order to force their reload.
- * Subsequent requests will be cached, as long as the server is not reloaded.
- *
- * For the rationale behind this choice, see
- * https://github.com/ether/etherpad-lite/pull/3958
- *
- * ACHTUNG: this may prevent caching HTTP proxies to work
- * TODO: remove the "?v=randomstring" parameter, and replace with hashed filenames instead
- */
-exports.randomVersionString = randomString(4);
-console.log(`Random string used for versioning assets: ${exports.randomVersionString}`);
-
 /**
  * The app title, visible e.g. in the browser window
  */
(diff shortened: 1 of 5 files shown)


webzwo0i:
Not 100% sure if the behavior is correct, that randomVersionString is not reset after installing plugins.

src/templates/pad.html
```diff
@@ -476,6 +476,7 @@ <h1>Skin Builder</h1>
   <script type="text/javascript">
       // @license magnet:?xt=urn:btih:8e4f440f4c65981c5bf93c76d35135ba5064d8b7&dn=apache-2.0.txt
       var clientVars = {};
+      clientVars.randomVersionString = '<%=settings.randomVersionString%>';
```
rhansen, line 479: Won't this be overwritten when the `CLIENT_VARS` message is sent to the client? See: https://github.com/ether/etherpad-lite/blob/f0a6236a0db683fc4ed5eed29e04e7ee11897e1f/src/static/js/pad.js#L297

It would probably be better to set the value in the `CLIENT_VARS` message here: https://github.com/ether/etherpad-lite/blob/f0a6236a0db683fc4ed5eed29e04e7ee11897e1f/src/node/handler/PadMessageHandler.js#L1038

src/node/utils/Settings.js
```diff
@@ -784,6 +784,11 @@ exports.reloadSettings = () => {
     // using Unix socket for connectivity
     console.warn('The settings file contains an empty string ("") for the "ip" parameter. The "port" parameter will be interpreted as the path to a Unix socket to bind at.');
   }
+
+  // regenerate randomVersionString so clients get a fresh plugin-definition.json after plugin
+  // install/uninstall. This also means that most of the resources are rerequested on reload and
+  // plugin installation.
+  exports.randomVersionString = randomString(4);
```
rhansen, line 791: Move the comment, assignment, and log message from [lines 45-57](https://github.com/ether/etherpad-lite/blob/f0a6236a0db683fc4ed5eed29e04e7ee11897e1f/src/node/utils/Settings.js#L45-L57) to here:

```suggestion
  /*
   * At each start, Etherpad generates a random string and appends it as query
   * parameter to the URLs of the static assets, in order to force their reload.
   * Subsequent requests will be cached, as long as the server is not reloaded.
   *
   * For the rationale behind this choice, see
   * https://github.com/ether/etherpad-lite/pull/3958
   *
   * ACHTUNG: this may prevent caching HTTP proxies to work
   * TODO: remove the "?v=randomstring" parameter, and replace with hashed filenames instead
   */
  exports.randomVersionString = randomString(4);
  console.log(`Random string used for versioning assets: ${exports.randomVersionString}`);
```

src/templates/pad.html
webzwo0i: Will change to assigning in one line.
It's intended that it will be replaced later. The reason is that `clientVars` is pushed after receiving `CLIENT_READY` from a client, but clients need to request `plugin-definition.json` before they send their `CLIENT_READY` message.

rhansen:
@webzwo0i I pushed some commits to be squashed; please take a look.

JohnMcLear:
Please rebase :)

rhansen:
> Please rebase :)

Done. I also squashed while I was at it.

Issue #4787, opened 2021-02-15
NPM issue with a fresh Windows Prebuilt Package?


alexeisenhart:
**Describe the bug**

I first followed the "Manually install on Windows" directions to set up Etherpad on Windows for development then I ran into the issue that I'll describe below. I assumed that I set something up wrong, so I trashed that and started from scratch with the "Prebuilt Windows package" steps and the error happened again.

Below you'll find the command line output from running start.bat as an administrator, using an installation that was freshly extracted from etherpad-lite-win.zip and unmodified. Is there any other information that I can supply that would be helpful to debug this issue?

```
C:\Windows\system32>cd c:\dev\W\VisualStudio.com\Etherpad

c:\dev\W\VisualStudio.com\Etherpad>start.bat

c:\dev\W\VisualStudio.com\Etherpad>node node_modules\ep_etherpad-lite\node\server.js
[2021-02-15 10:18:12.137] [DEBUG] console - Running on Node v12.20.2 (minimum required Node version: 10.17.0)
[2021-02-15 10:18:12.149] [DEBUG] AbsolutePaths - c:\dev\W\VisualStudio.com\Etherpad\node_modules\ep_etherpad-lite does not end with "src"
[2021-02-15 10:18:12.150] [INFO] console - All relative paths will be interpreted relative to the identified Etherpad base dir: c:\dev\W\VisualStudio.com\Etherpad
[2021-02-15 10:18:12.150] [INFO] console - Random string used for versioning assets: 3d78ab94
[2021-02-15 10:18:12.151] [DEBUG] AbsolutePaths - Relative path "settings.json" can be rewritten to "c:\dev\W\VisualStudio.com\Etherpad\settings.json"
[2021-02-15 10:18:12.151] [DEBUG] AbsolutePaths - Relative path "credentials.json" can be rewritten to "c:\dev\W\VisualStudio.com\Etherpad\credentials.json"
[2021-02-15 10:18:12.157] [INFO] console - settings loaded from: c:\dev\W\VisualStudio.com\Etherpad\settings.json
[2021-02-15 10:18:12.158] [INFO] console - No credentials file found in c:\dev\W\VisualStudio.com\Etherpad\credentials.json. Ignoring.
[2021-02-15 10:18:12.159] [INFO] console - Using skin "colibris" in dir: c:\dev\W\VisualStudio.com\Etherpad\src\static\skins\colibris
[...]

PR #4791, opened 2021-02-16
WIP: Windows Server CI

JohnMcLear:
Due to https://github.com/ether/etherpad-lite/issues/4787 it's clear we need some CI for running Etherpad Server on Windows server and run some backend tests(no need for frontend).

Todo: 
 - [x] Fix 4787
 - [x] Test if mocha timeouts are long enough
 - [x] Test runner
 - [x] Fix tests that fail in Windows ``importExportGetPost.js``  <-- they don't fail on my local windows machine! 🍡 
 - [x] Revert 306390c509d8b89693ccdf929c99f7d280f70afe
 - [x] Fiesta


JohnMcLear:
Note that before this is merged the removal of tests should be taken back out, they are removed to make testing the PR easier.

``mocha --timeout 120000 --recursive tests/backend/specs ../node_modules/ep_*/static/tests/backend/specs`` from runBackendTests.sh or whatever might not be a long timeout because the test runner starts at the same time as the node process (intentional).

Once the windows build is fixed we can finish this PR off.

PR #4793, opened 2021-02-16
Socket query string

JohnMcLear:
Currently we can't scale Etherpad due to a lack of ability to shard based on padIds.  By introducing padId into the connecting query we can do routing.

Still to do:
- [x] timeslider
- [x] client libraries
- [x] plugins that use socket direct (ep_comments)
- [x] ratelimit test

To discuss:
admin, plugins, settings

Note: This may or may not be the way we solve the sharding problem, I expect it will be but this is experimental and I wont be offended if this code never lands in develop :)  Other approaches are not using socketio, using socketio rooms, using namespace etc.


src/static/js/pad.js
```diff
@@ -218,7 +218,14 @@ const sendClientReady = (isReconnect, messageType) => {
...
 
 const handshake = () => {
+  let padId = document.location.pathname.substring(document.location.pathname.lastIndexOf('/') + 1);
+  // unescape neccesary due to Safari and Opera interpretation of spaces
+  padId = decodeURIComponent(padId);
+
+  // padId is used here for sharding / scaling.  We prefix the padId with padId: so it's clear
+  // to the proxy/gateway/whatever that this is a pad connection and should be treated as such
   socket = pad.socket = socketio.connect(exports.baseURL, '/', {
+    query: `padId=${padId}`,
```
rhansen, line 228: The value needs to be URI encoded. The easiest way to do that is to let socket.io encode it for you:

```suggestion
    query: {padId},
```

src/static/js/timeslider.js
```diff
@@ -52,7 +52,9 @@ const init = () => {
       Cookies.set('token', token, {expires: 60});
     }
 
-    socket = socketio.connect(exports.baseURL);
+    socket = socketio.connect(exports.baseURL, '/', {
+      query: `padId=${padId}`,
+    });
```
rhansen, line 57:
```suggestion
    socket = socketio.connect(exports.baseURL, '/', {query: {padId}});
```

rhansen (commented):
Also need to do the following:
* Update `src/tests/backend/specs/socketio.js`
* Remove any mention of pad ID from message payloads. Alternatively, check that the pad IDs match.
* Perform auth checks against the query parameter.

JohnMcLear:
RE

> Remove any mention of pad ID from message payloads. Alternatively, check that the pad IDs match.
> Perform auth checks against the query parameter.

I'd rather not change this logic for now, but certainly once this is stable and tested we can add validation then...  This query value is for sharding and wasn't designed to replace any Etherpad internals.

JohnMcLear:
One issue I found with this approach is that it is that static assets will come from any backend, which is probably fine but worth noting :)

JohnMcLear:
Ready for a second review, this may never be implemented in any production environment but having it in develop will be useful.

PR #4803, opened 2021-02-17
tests: Restore `runnerBackend.sh`

rhansen:
`runnerBackend.sh` was deleted in commit 7dae5e3db8b68da09a064f21aa837ccdcdb94b08 but plugins still need it until their GitHub workflow definitions have been updated.


jkh925 (commented):
Hi

Issue #4527, opened 2020-11-28
Line numbers should line up with baseline


rhansen:
Currently line numbers are positioned near the top of the line, so if you use `ep_font_size` to make a tall line then the line number is rendered too far up:

![screenshot](https://user-images.githubusercontent.com/1612291/100494104-4a3fdd00-310c-11eb-8d1f-488a0d0f7e19.png)

JohnMcLear:
Cc @seballot

seballot:
Sorry, no time to contribute to etherpad right now...

JohnMcLear:
No rush.  I have another good challenge for you too with another plugin but it can wait some weeks.  Thanks man!!

JohnMcLear:
Ready for review: https://github.com/ether/etherpad-lite/pull/4825 cc @seballot please let me know if I did it wrong :)

JohnMcLear:
![image](https://user-images.githubusercontent.com/220864/108592495-07398280-7366-11eb-99c0-32269aeabf58.png)

Happy with this?  The code isn't as clean as the image and I'm worried about additional browser processing cost..  But it was never gonna come for free...

rhansen:
Looks awesome, thanks!

seballot:
I made some tests re https://github.com/citizenos/ep_image_upload/issues/35 and yes it's not related to the image_upload plugin

https://video.etherpad.com/p/test2

![image](https://user-images.githubusercontent.com/17404254/109616278-440b3380-7b35-11eb-9e83-0e6f95ab7ebb.png)

JohnMcLear:
Yea it's defo core and 99% related to our patch :/

seballot:
No it's not :) seems to come from a commit after our patch (if I checkout on our patch it works fine, but if I check out develop it's buggy)

I guess it's coming from your new loading image !! (just kidding)

Any idea?

JohnMcLear:
Will look

JohnMcLear:
cc @rhansen 
```
a17f9bf3cfc745a44d0e57b77912e346ffd3ce1c is the first bad commit
commit a17f9bf3cfc745a44d0e57b77912e346ffd3ce1c
Author: Richard Hansen <rhansen@rhansen.org>
Date:   Thu Feb 25 21:11:30 2021 -0500

    ace: Build the outer and inner iframes programmatically

    This makes the code easier to read and it silences Chrome's
    `document.write()` warning:
[...]

PR #4825, opened 2021-02-19
include lineHeight property in sidebar elements

JohnMcLear:
This fixes the issue as described in #4527

This has the added advantage of making the full line number element clickable(before it was just 24px of the element) to ensure a positive UX for the ``?lineNumber`` URL endpoint.  

It also makes it more obvious that a click action can happen based on the hover.

In the image below line 1 side div element is being hovered

image

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -72,8 +72,8 @@ function Ace2Inner() {
   let sideDivInner;
 
   const initLineNumbers = () => {
-    const htmlOpen = '<div id="sidedivinner" class="sidedivinner"><div><span class="line-number">1';
-    const htmlClose = '</span></div></div>';
+    const htmlOpen = '<div id="sidedivinner" class="sidedivinner"><div><div class="line-number">1';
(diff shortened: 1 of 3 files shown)


rhansen:
Ideally the baseline of the line number would line up with the baseline of the paragraph's top line, like this:

```
1   Lorem ipsum dolor sit amet, consectetur
    adipiscing elit, sed do eiusmod tempor
    incididunt ut labore et
2   dolore magna aliqua.
```

Not vertically centered like this:

```
    Lorem ipsum dolor sit amet, consectetur
1   adipiscing elit, sed do eiusmod tempor
    incididunt ut labore et
2   dolore magna aliqua.
```

JohnMcLear:
I don't think that's doable but will take a look.

JohnMcLear:
I spotted a weird bug w/ this I want to deal with.

image

JohnMcLear:
Ready for review

image

seballot:
Hi ! please see #4831

JohnMcLear:
Closing in favor of the @seballot

Issue #4818, opened 2021-02-18
CI doesn't test full import process


JohnMcLear:
It should

PR #4827, opened 2021-02-19
tests: fix importexport frontend tests

JohnMcLear:
Fixes https://github.com/ether/etherpad-lite/issues/4818 as much as possible without OS level control.

### DIFF
--- a/.github/workflows/frontend-admin-tests.yml
+++ b/.github/workflows/frontend-admin-tests.yml
@@ -50,8 +50,8 @@ jobs:
       id: environment
       run: echo "::set-output name=sha_short::$(git rev-parse --short ${{ github.sha }})"
 
-    - name: Write custom settings.json with loglevel WARN
-      run: "sed 's/\"loglevel\": \"INFO\",/\"loglevel\": \"WARN\",/' < settings.json.template > settings.json"
+    - name: Create settings.json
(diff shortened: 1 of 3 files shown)


rhansen:
There are a number of issues with these tests:
  * They aren't any better than the backend tests because they don't actually test the form event handlers. (None of these tests would have caught the recently fixed import bug.) Proper tests would emulate selecting a file and clicking the submit button.
  * There are issues with the `helper.waitFor()` calls: They should be `await helper.waitForPromise()`, and the body should return a boolean not throw.
  * The `$.ajax()` `async: false` option should be replaced with `await`.
  * Not all of the expected strings look right.
  * The string comparisons don't allow for different attribute order (e.g., `<ol start="1" class="list-number1">` vs. `<ol class="list-number1" start="1">`).

I will push a commit that redoes the tests.

rhansen:
I pushed a commit that redoes the testing approach. I converted the first two test cases; the remainder still need to be converted.

rhansen:
Woohoo, I think this is finally ready. I'm pretty sure the frontend test failure is a flaky test.

JohnMcLear:
yea we have the issue where random tests are failing again, it was stable for a while but it's been introduced again.  If you watch the sauce labs runner you will see a browser (usually FF 52) randomly just stop loading pads.

PR #4830, opened 2021-02-20
Move vendors 

JohnMcLear:
LGTM should not consider vendor code quality as ours.  This PR asks LGTM to ignore the vendors folder and moves vendor files to vendors..

This may break plugins!


webzwo0i:
also significantly improves readability of our code structure! best would be to migrate vendors to npm packages - but I think not all of them ship as npm packages right now :-(
Not related to this issue - I noticed we don't use much jquery in core, except gritter/farbtastic/nice-select. Though most of the plugins use jquery, some of them maybe only because we make it available in core. jquery is a big number for bandwidth consumption (for first time access after a restart/reload)

JohnMcLear:
jQuery is used extensively in plugins, but these can be modernized over the years.

Agreed RE modules; I looked and we have some modifications of most libraries, gritter, nice-select etc. ;\

PR #4831, opened 2021-02-20
editor: line numbers alignment

seballot:
Hi !

I made small change to your PR @JohnMcLear :

- small UI improvement
- make it compatible with author_neat (keeping span and using inline-block property)
- squashing the three commits into one

Nice work btw ! it's not perfect with ep_headings because there is also some margin involved, but still it's a good improvement !


JohnMcLear:
This PR is going to lose my commit history but I don't mind :)  It's just worth noting.

PR #4838, opened 2021-02-21
Lgtm bugfixes

JohnMcLear:
Various low hanging issues LGTM raised.  Mostly pointless declarations and tests of always true/false values.


rhansen:
I pushed some new commits for squash. Please take a look, and if they look OK to you then feel free to squash and merge.

PR #4841, opened 2021-02-21
Instead of remove JSONP API this might serve as temp patch

JohnMcLear:
Fix the LGTM security issue:

> This part of the regular expression may cause exponential backtracking on strings starting with '$["' and containing many repetitions of 'a"]["'.
> This part of the regular expression may cause exponential backtracking on strings starting with '$['' and containing many repetitions of 'a'][''.

https://lgtm.com/projects/g/ether/etherpad-lite/snapshot/a6141937b719967b0fc6c16925cd91e6d39002f2/files/src/node/hooks/express/isValidJSONPName.js?sort=name&dir=ASC&mode=heatmap

A further consideration is to drop JSONP entirely.

### DIFF
--- a/src/node/hooks/express/isValidJSONPName.js
+++ b/src/node/hooks/express/isValidJSONPName.js
@@ -67,7 +67,7 @@ const RESERVED_WORDS = [
   'yield',
 ];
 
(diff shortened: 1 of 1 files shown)


rhansen:
I'm not very familiar with this code, but this PR makes me very nervous. I'm worried that the relaxed regex will introduce an XSS vulnerability.

PR #4845, opened 2021-02-22
caretPosition: Simplify `getPosition()`

rhansen:
Multiple commits:
* caretPosition: Invert condition in `getPosition()` for readability
* caretPosition: Delete unused var in `getPosition()`
* caretPosition: Delete no-op `Range.detach()` call
* caretPosition: Delete pointless logic in `getPosition()`
* caretPosition: Clarify comment in `getPosition()`


JohnMcLear:
Much cleaner, tnx :)  Will get this tested and merged

PR #4846, opened 2021-02-22
plugins: Better fix for LGTM security warning

rhansen:


JohnMcLear:
I think LGTM might still panic about this but let's see :)

JohnMcLear:
Totally weird, I swear I took this approach and still got errors but LGTM is a little weird and probably needs some configuration / tweaking..

PR #4847, opened 2021-02-22
CI: Delete old Sauce Connect logic

rhansen:
@JohnMcLear Do any plugins use `sauce_tunnel.sh` for their CI?


JohnMcLear:
> 
> 
> @JohnMcLear Do any plugins use `sauce_tunnel.sh` for their CI?

Nope :)  All plugins that have frontend test coverage are only tested on PR on ``ether/`` afaik.

PR #4856, opened 2021-02-24
deps: Update ueberdb2 to work around dirty DB bug

rhansen:
Addresses but is not a complete fix for #4684.


JohnMcLear:
We will probably need to modify this as I merge in more of the ueberdb fixes, ensuring we get the minor version change for redis.

PR #4864, opened 2021-02-24
Windows CI Installer

JohnMcLear:
This PR introduces CI builds of a windows installer(using NSIS) .

It builds an executable that installs Etherpad and runs it.

There are obvious steps to make once this has been merged. But I'd suggest on each release we include both the .zip and the .exe and allow users to have a portable zip or an installed executable.

https://github.com/ether/etherpad_nsis

This was a relatively rushed project (4 hours) and I didn't want to spend any more time on it so it will need a foster parent to maintain it :)

props to @joncloud for https://github.com/joncloud/makensis-action-test and the nsis team that while have a horrible UX make relatively easy to use and rapid tools.

Note for review: I'm using linux to build the windows executable, this may need to be reviewed and we might want to switch to Windows if we can confirm building on linux causes a problem.

> Screenshots/GIFs

### DIFF
--- a/.github/workflows/windows-installer.yml
+++ b/.github/workflows/windows-installer.yml
@@ -10,12 +10,16 @@ jobs:
     if: |
          (github.event_name != 'pull_request')
          || (github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id)
-    name: Building Windows Zip and Executable Installer on Linux without plugins
-    runs-on: ubuntu-latest
+
+    name: Build Zip & Exe
+    runs-on: windows-latest
 
     steps:
-    - name: 'Install makensis (apt)'
-      run: sudo apt update && sudo apt install -y nsis nsis-pluginapi
+    - uses: msys2/setup-msys2@v2
+      with:
+        path-type: inherit
+        install: >-
+          zip
 
     - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-installer.yml
```diff
@@ -0,0 +1,54 @@
...
+on: [push, pull_request]
+
+jobs:
+  build:
+    # run on pushes to any branch
+    # run on PRs from external forks
+    if: |
+         (github.event_name != 'pull_request')
+         || (github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id)
+    name: Building Windows Zip and Executable Installer on Linux without plugins
```
rhansen, line 13: Please keep this brief:
```suggestion
    name: Build
```

.github/workflows/windows-installer.yml
```diff
@@ -0,0 +1,54 @@
...
+    runs-on: ubuntu-latest
```
rhansen, line 14: With `runs-on: ubuntu-latest`, npm will create `src/node_modules/.bin/*` according to Linux's needs, not Windows's needs. Use `windows-latest` instead with msys2 shell. See `windows-zip.yml`.

PR #4870, opened 2021-02-25
avoid manually including require-kernel in ace.js

webzwo0i:


webzwo0i:
It results in another request that should be served from cache (as the same file, require-kernel.js, is already downloaded from pad.html before pad.js and thus ace.js is loaded). It will save some bits on the wire due to not inlining the code twice and also reduces some complexity.
require-kernel is needed to resolve files in pad.html and also in the inner iframes. However, at the moment we cannot simply reuse the require-kernel from the parent window that is defined during loading pad.html. If we do so, we get different values for `window` as soon as the ACE is initialized in the iframe. We can probably workaround this be either changing the way we access `window` in ace2_inner.js (ie when using require-kernel from the parent window, the window property in ace2_inner.js points to the parent, when it should point to ace_inner iframe) or by avoiding including ace2_inner.js to early, however the last option will result in additional HTTP requests.

Because we want to not put too much time into resolving problems that are caused by require-kernel I think we should probably move on and don't try to reuse require from the parent window. However if someone has a good idea or wants to reproduce my findings, I will definitely help. It's possible that I tried to solve more than one problem at a time (ie I was working on reducing the number of HTTP requests) and might included ace2_inner.js too early myself during the endeavour

webzwo0i:
if anyone stumbles upon `Ace2Editor.EMBEDED`: this is still needed in case minify is true for inlining CSS (https://github.com/ether/etherpad-lite/pull/4440). However, that's also some logic we can remove as soon as we switch bundler

JohnMcLear:
Hell of a brilliant piece of work mate.  Will do a full review today.

webzwo0i:
:-D I'm a little bit concerned about "pretty pad names", ie serving `/padname` instead of `/p/padname` due to the link being made to `../static/`. Maybe we should include test coverage for this?

JohnMcLear:
Okay let's make an issue to cover for pretty pad names

src/static/js/ace.js
```diff
@@ -196,15 +194,10 @@ const Ace2Editor = function () {
           `../static/skins/${clientVars.skinName}/pad.css?v=${clientVars.randomVersionString}`);
 
       pushStyleTagsFor(iframeHTML, includedCSS);
-
-      if (!Ace2Editor.EMBEDED || !Ace2Editor.EMBEDED[KERNEL_SOURCE]) {
-        // Remotely src'd script tag will not work in IE; it must be embedded, so
-        // throw an error if it is not.
-        throw new Error('Require kernel could not be found.');
-      }
+      iframeHTML.push(`<script type="text/javascript" src="../static/js/require-kernel.js?v=${clientVars.randomVersionString}"></script>`);
```
rhansen, line 197: I'm surprised this works. I would have expected something like the following:

```javascript
const script = document.createElement("script");
script.type = 'text/javascript';
script.src = `../static/js/require-kernel.js?v=${clientVars.randomVersionString}`
document.head.appendChild(script);
```

PR #4873, opened 2021-02-26
reversecompat: underscore

JohnMcLear:
Backward compatibility for plugins that were written when underscore lived at src/static/js/underscore.js.

### DIFF
--- a/src/node/utils/tar.json
+++ b/src/node/utils/tar.json
@@ -83,4 +83,5 @@
   , "pluginfw/hooks.js"
   ]
 , "jquery.js": ["jquery.js"]
+, "underscore.js": ["underscore.js"]
 }
--- a/src/node/utils/Minify.js
+++ b/src/node/utils/Minify.js
@@ -146,6 +146,14 @@ const minify = async (req, res) => {
     filename = 'js/vendors/jquery.js';
   }
 
+  // Backward compatibility for plugins that were written when underscore lived at
+  // src/static/js/underscore.js.
+  if (['js/underscore.js', 'plugins/ep_etherpad-lite/static/js/underscore.js']
+      .indexOf(filename) !== -1) {
+    logger.warn(`request for deprecated underscore path: ${filename}`);
+    filename = 'js/vendors/underscore.js';
+  }
+
   /* Handle static files for plugins/libraries:
      paths like "plugins/ep_myplugin/static/js/test.js"
      are rewritten into ROOT_PATH_OF_MYPLUGIN/static/js/test.js,


src/node/utils/tar.json
```diff
@@ -83,4 +83,5 @@
   , "pluginfw/hooks.js"
   ]
 , "jquery.js": ["jquery.js"]
+, "underscore.js": ["underscore.js"]
```
webzwo0i, line 86: I think this should instead include:
`
  , "vendors/underscore.js"
  , "$underscore/underscore.js"
` (or `$underscore` I can't remember) so that underscore itself (and not only our require-wrapper at vendors/underscore.js) is included.

Also I'm not sure if this (and the jquery.js thing above) is really needed. Should this help plugins when they require jquery.js or underscore.js directly from core, while they don't use anything else (like ace2_common that already bundles those dependencies)?

The internals are something like this:
It does not matter if we require a file via script tag (like `/javascripts/lib/ep_etherpad-lite/static/js/THE_FILE?callback=require.define` - notice the important callback) or via calling require like `require('THE_FILE')` or probably better `require('ep_etherpad-lite/static/js/THE_FILE')`. The former has the advantage of being faster. There is no client-side code run to find out that we need to download the file. Also the browser directly handles fetching the resource (under the "no more than 2 parallel requests" paradigm).

In that scenario we need `THE_FILE` to be packaged like you did above or have `THE_FILE` being included in a package (e.g. ace2_common) in which case yajsml takes care of serving the appropriate package (side note: we can always directly fetch the file, if we don't include `javascripts/lib`). This is suboptimal in cases where we are only interested in underscore, but not in the rest of the package, as we would get a lot of extra code we did not want.

In normal pads we have two instances of require-kernel that don't share their caches. The require-kernel in the main window contains nearly all of the packages and the require-kernel of the inner iframe contains at least ace2_common and ace2_inner. And this is why I think the above additional packages are not necessary: all plugins I can think of either run there code in the inner iframe or the main window and thus they already have access to require-kernel's internal cache (which already includes underscore and jquery).

webzwo0i: Hmm but if you tested it and it did not work without those, than ignore my comment. It's possible I did not consider something, e.g. that we would need to serve underscore two times (old and new path of our wrapper) in a package, so that it's found if a plugin requests it under the old path.

Yeah, probably this is what the patch is about, so it's needed so we don't need to include the files more than once :-D

JohnMcLear: Happy to see a better patch if you can craft one, mine was just quickly tested because I had a schedule for the day already -_-

PR #4874, opened 2021-02-26
allow etherpadHooks from ep_performance_hooks plugin to be global

webzwo0i:
makes mocha happy


rhansen:
I don't see `etherpadHooks` anywhere. What's the context?

webzwo0i:
from ep_performance_hooks

rhansen:
If `etherpadHooks` is a plugin thing, why does this need to be fixed in core? What goes wrong without this change? Please at least add a comment.

Issue #4875, opened 2021-02-26
Clarification on backwards compatibility for jquery/underscore


webzwo0i:
If I understand the changes correctly then:
- if plugins use underscore in their backend code, they need to update from `require('ep_etherpad-lite/static/js/underscore')` to `require('ep_etherpad-lite/static/js/vendors/underscore')`
- if plugins do `require('underscore')` in their backend code, it's best if they'd define a dependency on underscore in their package.json. We could however add our src directory to `module.path` during startup, so that even for plugins that live under `./node_modules` a dependency that lives in `./src/node_modules` can be found without using `ep_etherpad-lite` prefix
- if plugins use jquery in their backend or frontend code, they should `require('ep_etherpad-lite/static/js/rjquery')` and don't need to change anything.
- if plugins use `require('jquery')` in their backend code, they should define a dependency on jquery in their package.json
- if plugins link to files in their frontend code with `<script src="/static/plugins/ep_plugin/static/js/underscore"` they need to update the path

So to summarize: whenever plugins use underscore in their backend code they need to update their code anyway (cp. failing tests in https://github.com/ether/etherpad-lite/runs/1988391793?check_suite_focus=true#step:13:35)

In https://github.com/ether/etherpad-lite/blob/develop/src/node/utils/Minify.js I would rather only check for "/static/js/jquery" and deliver "/static/js/vendors/jquery" - that should catch all different cases for linking in src-attributes.
If required for backend code usage or via `require()` in the frontend, the "rjquery" file should always be used and no update is needed.

Also I would move underscore back to the old position. It's really just a wrapper not the vendors file. And I would remove the changes to tar.json as jquery/underscore does not need to be a package on it's own.

Did I miss something?

rhansen:
(Replying mostly for the benefit of users and plugin authors who might come across this bug.)

> If I understand the changes correctly then:
> 
> * if plugins use underscore in their backend code, they need to update from `require('ep_etherpad-lite/static/js/underscore')` to `require('ep_etherpad-lite/static/js/vendors/underscore')`

Not anymore thanks to your PR #4876.

> 
> * if plugins do `require('underscore')` in their backend code, it's best if they'd define a dependency on underscore in their package.json.

They would *have* to define a dependency in their `package.json`.

> We could however add our src directory to `module.path` during startup, so that even for plugins that live under `./node_modules` a dependency that lives in `./src/node_modules` can be found without using `ep_etherpad-lite` prefix

[...]

PR #4876, opened 2021-02-26
move underscore to its old place and remove unnecessary packages

webzwo0i:
for https://github.com/ether/etherpad-lite/issues/4875

I think this is better. The underscore wrapper again lives at ./src/static/js and not in vendors directory, as it's really just a wrapper. If we don't do this, all plugins that use underscore from core would need to be updated anyway. So if we really want to move that file we should probably include a warning in the plugin devs helper script for a while?

It also removes the imo unnecessary packages in tar.json. Checking the filename in Minify.js only helps plugins when they link to `/static/js/jquery.js` in their script tags. If they use jquery with `require` they should always use rjquery - which is still at its old place.

@rhansen what do you think? Did I miss something fundamental?


src/node/utils/Minify.js
```diff
@@ -141,19 +141,11 @@ const minify = async (req, res) => {
 
   // Backward compatibility for plugins that were written when jQuery lived at
   // src/static/js/jquery.js.
-  if (['js/jquery.js', 'plugins/ep_etherpad-lite/static/js/jquery.js'].indexOf(filename) !== -1) {
+  if (filename.endsWith('js/jquery.js')) {
```
rhansen, line 144: This matches `foojs/jquery.js`, which I don't think is a good idea. I think it's better to leave this line as-is.

src/node/utils/tar.json
```diff
@@ -82,6 +82,4 @@
   , "pluginfw/shared.js"
   , "pluginfw/hooks.js"
   ]
-, "jquery.js": ["jquery.js"]
```
rhansen, line 85: I coulda swore this line was necessary to support plugins that have frontend code that does `require('jquery')`, but maybe not.

rhansen (approved):

src/node/utils/Minify.js
webzwo0i: OK good point

src/node/utils/tar.json
webzwo0i: I would expect `require('ep_etherpad-lite/static/js/jquery')` in this case, which will work and `require('jquery')` to resolve to a local dependency of the plugin.

PR #4881, opened 2021-02-27
performance: Page load improvements by way of removal of unnecessary 307 redirections

webzwo0i:
removes 4 307s redirections.

It also ensures that there is no request without a `versionString` made from the inner iframe, because it explicitly downloads the needed files via script tag now. (Before the change, the files were loaded on demand via require-kernel which did not append the `versionString` which could have resulted in problems during upgrades.)

### DIFF
--- a/src/node/utils/tar.json
+++ b/src/node/utils/tar.json
@@ -2,6 +2,9 @@
   "pad.js": [
     "pad.js"
   , "pad_utils.js"
+  , "$js-cookie/src/js.cookie.js"
+  , "security.js"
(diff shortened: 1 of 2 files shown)


webzwo0i:
One thing that's still problematic: We have some code duplication in our packages. I'm trying some reunion out now, but those will get into another PR if they work.

Nevertheless, contrary to what one would expect, the overall code size is reduced in this PR a little bit, too, because the 307s resulted in (unnecessary) package downloads.

PR #4888, opened 2021-02-28
tests: run a long test

JohnMcLear:
Designed to catch issues that only appear X duration.

We might want to introduce some "chaos" into the load test tool or do we think this will be sufficient?

### DIFF
--- a/.github/workflows/load-test.yml
+++ b/.github/workflows/load-test.yml
@@ -28,7 +28,7 @@ jobs:
       run: sudo npm install -g etherpad-load-test
 
     - name: Run load test
-      run: src/tests/frontend/travis/runnerLoadTest.sh
+      run: src/tests/frontend/travis/runnerLoadTest.sh 25
 
   withplugins:
     # run on pushes to any branch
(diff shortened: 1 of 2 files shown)


.github/workflows/load-test.yml
```diff
@@ -78,6 +78,34 @@ jobs:
     - name: Install all dependencies and symlink for ep_etherpad-lite
       run: src/bin/installDeps.sh
 
-    # configures some settings and runs npm run test
+    # configures some settings and runs npm run test for 25 seconds with 50 authors
```
rhansen, line 81: This comment is likely to drift out of sync with the code. I'd say leave out the "for 25 seconds with 50 authors". If you want it to be more readable then use more verbose flags like `--duration=25 --users=50`.

rhansen (approved):
Leave out chaos for now. I agree that it would be good to introduce chaos in the future, but that's really hard to do properly: A general rule about testing is it should be easy to reproduce a failure. That's difficult when there's randomness or extreme complexity.

.github/workflows/load-test.yml
JohnMcLear: why would it drift out of sync?

rhansen: It's common for people to bump a number and forget to update the comment to match. Then that leaves people scratching their heads.

.github/workflows/load-test.yml
```diff
@@ -80,4 +80,32 @@ jobs:
...
+        node-version: 12
+
+    - name: Install all dependencies and symlink for ep_etherpad-lite
+      run: src/bin/installDeps.sh
+
+    - name: Install etherpad-load-test
+      run: sudo npm install -g etherpad-load-test
+
+
+    # configures some settings and runs npm run test for 5000 seconds(50 mins) with 5 authors
```
JohnMcLear, line 109: What about this one @rhansen  ?

rhansen: Delete it.

PR #4894, opened 2021-02-28
tests:  Frontend test Windows ZIP

JohnMcLear:
This PR introduces Frontend testing within Github actions!

We're depending a lot on saucelabs recently and that's fine but sometimes we just want to quickly do a frontend simple test on a weird environment (IE windows build) so this PR solves that problem.

Things to note.

1. It still builds the windows .zip if the cypress tests fail.
1. It does not add any heavy deps to Etherpad as cypress must be installed in CI.
1. Cypress is responsible for running the Etherpad instance.

It's up to us how much we use this or not, I know it introduces a bunch of technical debt but I tried to keep that a minimum by compartmentalizing things and documenting where required.


.github/workflows/windows-zip.yml
```diff
@@ -65,11 +65,12 @@ jobs:
     - name: Extract Etherpad
       run: 7z x etherpad-lite-win.zip -oetherpad
 
-    - name: list
-      run: dir etherpad
+    - name: Install Cypress
+      run: npm install cypress -g
```
rhansen, line 69: Why install it globally? Isn't superuser access required for that?

.github/workflows/windows-zip.yml
```diff
@@ -65,11 +65,12 @@ jobs:
...
 
     - name: Run Etherpad
       run: |
             cd etherpad
             node node_modules\ep_etherpad-lite\node\server.js &
             curl --connect-timeout 10 --max-time 20 --retry 5 --retry-delay 10 --retry-max-time 60 --retry-connrefused http://127.0.0.1:9001/p/test
+            cd src\tests\frontend\cypress && cypress run --spec cypress\integration\test.js
```
rhansen, line 76: Either wrap the two commands in the PowerShell equivalent of a subshell, or put the two commands on separate lines to make it easier to understand that one will have to add `cd ..\..\..\..` after Cypress exits if more commands will be added.

```suggestion
            cd src\tests\frontend\cypress
            cypress run --spec cypress\integration\test.js
```

src/tests/frontend/cypress/cypress/integration/test.js
```diff
@@ -0,0 +1,27 @@
...
+    ($iframe) => new Cypress.Promise((resolve) => {
+      $iframe.ready(() => {
+        resolve($iframe.contents().find('body'));
+      });
+    }));
+
+describe(__filename, function () {
+  it('Pad content exists', async function () {
+    cy.visit('http://127.0.0.1:9001/p/test', {timeout: 60000});
+    cy.wait(2000);
```
rhansen, line 16: I don't think this is necessary—the docs say that Cypress automatically retries until timeout or the condition is satisfied.

src/tests/frontend/cypress/cypress/integration/test.js
```diff
@@ -0,0 +1,27 @@
...
+    cy.get('iframe[name="ace_outer"]', {timeout: 10000}).iframe()
+        .find('.line-number:first', {timeout: 10000})
+        .should('have.text', '1');
+    cy.wait(500);
```
rhansen, line 20: ditto

src/tests/frontend/cypress/cypress/README.md
```diff
@@ -0,0 +1,10 @@
+# Cypress Etherpad guide
```
rhansen, line 1: Why the `cypress/cypress` directory structure? Why not move everything in `src/tests/frontend/cypress/cypress` to `src/tests/frontend/cypress`?

.github/workflows/windows-zip.yml
JohnMcLear: so I can access the binary easily.  And we have enough privilege in CI

.github/workflows/windows-zip.yml
JohnMcLear: I seem to remember that's how I did it at first but something failed, will commit and see if it was related or not.

src/tests/frontend/cypress/cypress/integration/test.js
JohnMcLear: The issue is that on initial page load Etherpad is still building cache so we want to wait for it to be fully built...  

HTTP connectivity !== Pad Availability.  I check for Pad Availability.

JohnMcLear: I can drop them and see, but I am pretty sure the wait is required.  I assume you are referring to the timeouts not being necessary?

rhansen: I'm referring to the `cy.wait()` calls. See:
https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting

src/tests/frontend/cypress/cypress/README.md
JohnMcLear:
```
C:\etherpad-lite\src\tests\frontend\cypress>cypress run ..\integration\test.js --config-file cypress.json

Can't run because no spec files were found.

We searched for any files inside of this folder:

C:\etherpad-lite\src\tests\frontend\cypress\cypress\integration
```

I will run the binary from a level up but it's messy ;\

JohnMcLear:
Ready for final review.

PR #4900, opened 2021-03-01
Import plugins less warny

JohnMcLear:
Currently Etherpad warns about too many elements that are actually supported (spans, strong etc).

This PR puts the arrays separate from blockElems (I could use a ... to create but it's nice to have an array that can be copy/pasted between the two files without concern).

The outcome of this PR is that admins will have less warnings in their console and the only warnings they should see will be the ones for the plugins they actually need.

If anything todays bug report (of a lot of noise) is a good indicator that this functionality is working as expected

### DIFF
--- a/src/static/js/contentcollector.js
+++ b/src/static/js/contentcollector.js
@@ -64,8 +64,25 @@ const makeContentCollector = (collectStyles, abrowser, apool, className2Author) 
     li: 1,
   };
 
+  // supportedElems are Supported natively within Etherpad and don't require a plugin
+  const supportedElems = ['div',
+    'br',
+    'p',
+    'pre',
+    'li',
+    'author',
+    'lmkr',
+    'insertorder',
+    'strong',
+    'ul',
+    'ol',
+    'span',
+    'font',
+    'i'];
+
   hooks.callAll('ccRegisterBlockElements').forEach((element) => {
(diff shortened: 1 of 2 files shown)


src/static/js/contentcollector.js
```diff
@@ -64,8 +64,25 @@ const makeContentCollector = (collectStyles, abrowser, apool, className2Author)
     li: 1,
   };
 
+  // supportedElems are Supported natively within Etherpad and don't require a plugin
+  const supportedElems = ['div',
```
rhansen, line 68: Can you export this and import it from `ImportEtherpad.js` (or vice-versa)?

rhansen, line 68:
```suggestion
  const supportedElems = [
    'div',
```

src/static/js/contentcollector.js
```diff
@@ -64,8 +64,25 @@ const makeContentCollector = (collectStyles, abrowser, apool, className2Author)
...
+    'li',
+    'author',
+    'lmkr',
+    'insertorder',
+    'strong',
+    'ul',
+    'ol',
+    'span',
+    'font',
+    'i'];
```
rhansen, line 81:
```suggestion
    'i',
  ];
```

JohnMcLear:
Ready for final review

Issue #4898, opened 2021-03-01
Hitting up all endpoints creates a session


JohnMcLear:
Even hitting up ``/stats`` creates database noise by the way of a session.

Where do we actually need sessions to be created other than for ```/p/...``` ?  Do we need sessions for static files for example?

rhansen:
It's needed for `/admin` and `/socket.io` too. The `ep_readonly_guest` and `ep_openid_connect` plugins add their own endpoints that need it; I suspect other plugins do too. It'd probably be better to maintain a list of paths that don't need it rather than a list of paths that do. See https://github.com/ether/etherpad-lite/blob/a79f9efdb405d3b518974e63ab9ca9eb153597c9/src/node/hooks/express/webaccess.js#L12-L19

webzwo0i:
I'm +1 for a whitelist, that don't need a session (and/or should not set a cookie).
The actual pad data is send together with CLIENT_VARS with socket.io, so `/p/*` probably also doesn't need it.

Not sure how to test this, though. We could parse dirty.db after frontend tests have been run, but that would mean some effort (ie we could store some info in the tests for those that delete a cookie -> create a new session and then we can probably derive the number of expected sessions from the number of frontend tests).

JohnMcLear:
It's easy to test..

CI bring up instance
curl x number of requests at y endpoint, expect z lines in dirty.db w/ session | ``test cat /var/dirty.db | grep session | wc -l -lt 50 -gt 10`` etc. check, change endpoint, check again.

I will make some tests for this, should be easy because I'm in CI land every day atm

webzwo0i:
My concern was more related to new endpoints being added and not capturing those in the tests. But this is probably a more general problem re "how to write tests that still work in the future".

I extracted all the routes with
```
$ cat tests/backend/specs/express-routes.js 
'use strict';

const common = require('../common');
const errorhandling = require('../../../node/hooks/express/errorhandling');

describe(__filename, function () {
  this.timeout(30000);
  let endpoints;

  before(async function () {
    await common.init();
    endpoints = errorhandling.app._router.stack.map(layer => layer.regexp);
    console.log(endpoints);
  });
  it('when requesting endpoints no session should be created', async function () {
  });
  it('when requesting endpoints no cookies should be sent', async function () {
  });
});
```
[...]

PR #4911, opened 2021-03-02
tests: rather course (but it works) test coverage for session creation.

JohnMcLear:
*will fail intentionally*

In response to https://github.com/ether/etherpad-lite/issues/4898

### DIFF
--- a/src/tests/backend/specs/api/httpRouteSessionCreation.js
+++ b/src/tests/backend/specs/api/httpRouteSessionCreation.js
@@ -3,8 +3,11 @@
 const assert = require('assert').strict;
 const common = require('../../common');
 const settings = require('../../../../node/utils/Settings');
-const staticPathsRE = require('../../../../node/hooks/express/webaccess').staticPathsRE;
+const shouldNotCreateExpressSession =
+    require('../../../../node/hooks/express/webaccess').staticPaths;
 const fs = require('fs');
+const SessionStore = require('../../../../node/db/SessionStore');
+const store = new SessionStore;
 let agent;
 
 const shouldCreateExpressSession = [
@@ -14,22 +17,6 @@ const shouldCreateExpressSession = [
   '/ep_example',
   '/admin',
 ];
-
-const shouldNotCreateExpressSession = [
-  '/',
-  '/api/',
-  '/favicon.ico',
-  '/locales.json',
-  '/pluginfw/plugin-definitions.json',
-  '/static/js/pad.js',
-  '/stats/',
-];
-
-const getDatabaseSize = () => {
-  const dbFile = settings.dbSettings.filename;
-  const database = fs.readFileSync(settings.dbSettings.filename, 'utf8');
-  return database.split('\n').length;
-}
 
 describe(__filename, function () {
   before(async function () { agent = await common.init(); });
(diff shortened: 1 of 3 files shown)


src/tests/backend/specs/api/httpRouteSessionCreation.js
```diff
@@ -0,0 +1,62 @@
...
+  '/',
+  '/api/',
+  '/favicon.ico',
+  '/locales.json',
+  '/pluginfw/plugin-definitions.json',
+  '/static/js/pad.js',
+  '/stats/',
+];
+
+const getDatabaseSize = () => {
```
rhansen, line 23: This is fragile. Instead, implement the [`Store.length() method`](https://www.npmjs.com/package/express-session#storeallcallback) in the SessionStore class in `SessionStore.js`.

src/tests/backend/specs/api/httpRouteSessionCreation.js
```diff
@@ -0,0 +1,62 @@
...
+    if (settings.dbType !== 'dirty') this.skip;
+
+    this.timeout(100);
+
+    for (const endpoint of shouldNotCreateSession) {
+      it(endpoint, async function () {
+        const previousCount = getDatabaseSize();
+        await agent.get(endpoint)
+        .expect(200)
+        .expect(() => {
```
rhansen, line 42: Also test that the server does not reply with a `express_sid` cookie.

src/tests/backend/specs/api/httpRouteSessionCreation.js
```diff
@@ -0,0 +1,62 @@
...
+        })
+      });
+    }
+
+    for (const endpoint of shouldCreateSession) {
+      const previousCount = getDatabaseSize();
+      it(endpoint, async function () {
+        await agent.get(endpoint)
+        .expect(200)
+        .expect(() => {
```
rhansen, line 54: Also test that the server replies with an `express_sid` cookie.

webzwo0i:
just so we don't forget them: atm also any 404s will cause a session creation/cookie sent

src/tests/backend/specs/api/httpRouteSessionCreation.js
JohnMcLear: I made an attempt at this but it didn't work.  I'm unsure how to access ``Store`` as it's the parent of ``SessionStore`` (which is extending ``Store``)

src/tests/backend/specs/api/httpRouteSessionCreation.js
```diff
@@ -0,0 +1,62 @@
...
+const common = require('../../common');
+const settings = require('../../../../node/utils/Settings');
+const fs = require('fs');
+let agent;
+
+const shouldCreateSession = [
+  '/p/foo',
+  '/p/foo/export/html',
+];
+const shouldNotCreateSession = [
```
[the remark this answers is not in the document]
JohnMcLear: Done :)

src/tests/backend/specs/api/httpRouteSessionCreation.js
JohnMcLear: Done

src/tests/backend/specs/api/httpRouteSessionCreation.js
webzwo0i: If I read it correctly, we would need to add our own `length` method and increase/decrease it on set/destroy

JohnMcLear:
Oh but the link shows a method already exists for length...??

webzwo0i:
Yep, but we don't implement `Store.all` which is needed to get all the active sessions (that's needed by `length`). My statement was not quite correct: As per @rhansen link we would need a method `all` that returns all active sessions as an array. Thus we would not need to implement our own increase/decrease.

JohnMcLear:
Ack, no problems, will try sort that before I do proxy stuff 👍🏽

JohnMcLear:
Wont ``sessions.length`` only include most recent and not persist?  Else we would need to get all at startup...  It's fine if it is since startup but it's gonna be confusing if someone else uses it....

src/tests/backend/specs/api/httpRouteSessionCreation.js
JohnMcLear: I think this is done, but might need polish.

JohnMcLear:
I had a quick stab at this but I feel like I'm on the wrong track so will wait for input from @rhansen before spending any more time :)

PR #4920, opened 2021-03-04
Various CSS fixes

rhansen:
Multiple commits:
* CSS: Fix class name for outer iframe `<html>` tag
* CSS: Use `auto` for iframe body height
* CSS: Delete bogus `<link>` tag
* CSS: Add comment to `no-skin/pad.css` to silence warning

### DIFF
--- a/src/static/css/iframe_editor.css
+++ b/src/static/css/iframe_editor.css
@@ -4,7 +4,7 @@
 
 @import url('./lists_and_indents.css');
 
-html.inner-editor {
+html.outer-editor, html.inner-editor {
   height: auto !important;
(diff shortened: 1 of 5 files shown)


JohnMcLear:
Cc @seballot

JohnMcLear:
Do these fixes work with the other skin?  Worth asking.

rhansen:
Yes, I tested both skins.

webzwo0i:
Me too. I would have never come up with this solution. Great! Chrome did not work before when setting `height:auto` to outerdocbody, but now it does. However, my CSS skills are not good enough to assess this :)

PR #4921, opened 2021-03-04
add more endpoints that do not need a session

webzwo0i:
cc @rhansen 

I automatically extracted the routes and I think those should not create a session.

Left out are `/admin.*`, `/ro/.*`, `/p/.*` and `/` for the moment. I think `/` should also not create a session though.

### DIFF
--- a/src/node/hooks/express/webaccess.js
+++ b/src/node/hooks/express/webaccess.js
@@ -10,8 +10,7 @@ const readOnlyManager = require('../../db/ReadOnlyManager');
 hooks.deprecationNotices.authFailure = 'use the authnFailure and authzFailure hooks instead';
 
 const staticPathsRE = new RegExp(`^/(?:${[
-  'api',
-  'api/.*',
+  'api(?:/.*)?',
   'favicon\\.ico',
   'ep/pad/connection-diagnostic-info',
   'javascript',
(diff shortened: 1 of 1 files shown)


src/node/hooks/express/webaccess.js
```diff
@@ -10,12 +10,21 @@ const readOnlyManager = require('../../db/ReadOnlyManager');
 hooks.deprecationNotices.authFailure = 'use the authnFailure and authzFailure hooks instead';
 
 const staticPathsRE = new RegExp(`^/(?:${[
+  'api',
   'api/.*',
```
rhansen, line 14:
```suggestion
  'api(?:/.*)?',
```

src/node/hooks/express/webaccess.js
```diff
@@ -10,12 +10,21 @@ const readOnlyManager = require('../../db/ReadOnlyManager');
...
   'javascripts/.*',
+  'jserror/?',
   'locales\\.json',
+  'locales/.*',
+  'rest/.*',
   'pluginfw/.*',
+  'robots.txt',
   'static/.*',
+  'stats/?',
+  'tests/frontend.*'
```
rhansen, line 27:
```suggestion
  'tests/frontend(?:/.*)?'
```

rhansen (approved):

PR #4944, opened 2021-03-13
WIP: Drop original responsiveness test and replace with collaborative clients.

JohnMcLear:
To replace https://github.com/ether/etherpad-lite/pull/1764

Using Cypress to actually have collaborative clients for the first time ever.  Getting this functionality available will be a huge milestone in our testing if we can make it work.

We still ❤️ Sauce Labs but Cypress might be a better fit here.


src/tests/frontend/cypress/integration/collaborate.js
```diff
@@ -0,0 +1,29 @@
...
+
+Cypress.Commands.add('iframe', {prevSubject: 'element'},
+    ($iframe) => new Cypress.Promise((resolve) => {
+      $iframe.ready(() => {
+        resolve($iframe.contents().find('body'));
+      });
+    }));
+
+describe(__filename, () => {
+  it('Pad content exists', () => {
```
rhansen, line 11: All 4 browsers are all doing the same thing, yes? I think it would be better if this collaboration test tested collaboration:

1. browser 1 types "hello world 1" while the other three wait for that to appear
2. browser 2 types "hello world 2" while the other three wait for that to appear
3. etc.

.github/workflows/collaboration.yml
```diff
@@ -0,0 +1,50 @@
...
+    name: Test
+    runs-on: windows-latest
+
+    steps:
+    - uses: msys2/setup-msys2@v2
+      with:
+        path-type: inherit
+        install: >-
+          zip
+
```
rhansen, line 22: Unused:

```suggestion
```

.github/workflows/collaboration.yml
```diff
@@ -0,0 +1,50 @@
...
+      run: npm install cypress -g
+
+    - name: Run Etherpad & Tests
+      run: |
+            node src\node\server.js &
+            cd src\tests\frontend
+            cypress run --spec cypress\integration\collaborate.js --config-file cypress\cypress.json &
+            cypress run --spec cypress\integration\collaborate.js --config-file cypress\cypress.json &
+            cypress run --spec cypress\integration\collaborate.js --config-file cypress\cypress.json &
+            cypress run --spec cypress\integration\collaborate.js --config-file cypress\cypress.json
```
rhansen, line 44: Do these instances share the same cookie state? If so, Etherpad will kick out 3 of them.

JohnMcLear:
Cypress couldn't keep up here w/ our needs so had to drop attempting to use it for this job ;(  I'm sure it could do what we want in the future but not having a video feed etc. is too painful for debug.

PR #4954, opened 2021-03-17
Fix editor height issue

webzwo0i:
This fixes the problem of the clientHeight in ace_inner iframe being as large as the whole document instead of only the visible viewport. Please don't squash as it also fixes some other things.

@rhansen 
I deleted the poll condition as it was the source of loading problems in ~20% with chromium. The debug output was:
```
["Ace2Editor.init()"]
["Ace2Editor.init() waiting for outer frame"]
(2) ["Ace2Editor.init() load event on", Window]
["Ace2Editor.init() outer frame ready"]
["Ace2Editor.init() waiting for inner frame"]
["Ace2Editor.init() polling"]
["Ace2Editor.init() poll condition became true"]
["Ace2Editor.init() inner frame ready"]
["Ace2Editor.init() waiting for require kernel load"]
```

In the other 80% "poll condition became true" never happened. Do you know which browsers needed that, as our frontend tests are passing?

### DIFF
--- a/src/static/js/ace.js
+++ b/src/static/js/ace.js
@@ -263,14 +263,6 @@ const Ace2Editor = function () {
     requireKernel.src =
         absUrl(`../static/js/require-kernel.js?v=${clientVars.randomVersionString}`);
     innerDocument.head.appendChild(requireKernel);
-    // Pre-fetch modules to improve load performance.
-    for (const module of ['ace2_inner', 'ace2_common']) {
-      const script = innerDocument.createElement('script');
-      script.type = 'text/javascript';
-      script.src = absUrl(`../javascripts/lib/ep_etherpad-lite/static/js/${module}.js` +
-                          `?callback=require.define&v=${clientVars.randomVersionString}`);
-      innerDocument.head.appendChild(script);
-    }
     const innerStyle = innerDocument.createElement('style');
     innerStyle.type = 'text/css';
     innerStyle.title = 'dynamicsyntax';
--- a/src/static/js/ace.js
+++ b/src/static/js/ace.js
@@ -176,7 +176,7 @@ const Ace2Editor = function () {
     this.importText(initialCode);
 
     const includedCSS = [
-      '../static/css/iframe_editor.css',
+      `../static/css/iframe_editor.css?v=${clientVars.randomVersionString}`,
       `../static/css/pad.css?v=${clientVars.randomVersionString}`,
       ...hooks.callAll('aceEditorCSS').map(
           // Allow urls to external CSS - http(s):// and //some/path.css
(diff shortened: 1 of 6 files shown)


webzwo0i:
I had no luck with test coverage. All I can do is check the clientHeight, but I'm unable to find a test for "Is all the pad content visible". (After reverting the auto height commit clientHeight is correct, but it won't grow the editor, so it would be good to have a regression test for this, too. I seem to be unable to mouse click anywhere inside the iframe)

JohnMcLear:
Looks great work man

src/static/js/ace.js
```diff
@@ -281,14 +263,6 @@ const Ace2Editor = function () {
...
         absUrl(`../static/js/require-kernel.js?v=${clientVars.randomVersionString}`);
     innerDocument.head.appendChild(requireKernel);
-    // Pre-fetch modules to improve load performance.
-    for (const module of ['ace2_inner', 'ace2_common']) {
-      const script = innerDocument.createElement('script');
-      script.type = 'text/javascript';
-      script.src = absUrl(`../javascripts/lib/ep_etherpad-lite/static/js/${module}.js` +
-                          `?callback=require.define&v=${clientVars.randomVersionString}`);
-      innerDocument.head.appendChild(script);
-    }
```
webzwo0i, line 291: Removing it here is not good, as we now fetch those resources automatically (without version string). So we need to find a proper fix for the "require not found" caused by those lines.

webzwo0i:
Making it draft, because deleting ace2_inner/ace2_common results in those packages get downloaded automatically via require-kernel and thus they don't have a version string anymore, which is bad for caching. So we need another fix, maybe loading them after require has been defined.

rhansen:
> In the other 80% "poll condition became true" never happened. Do you know which browsers needed that, as our frontend tests are passing?

At least Safari, but I think Chrome too. However, with your `iframe.srcdoc = '<!DOCTYPE html>'';` change it looks like the load event now fires on those browsers so polling is unnecessary.

webzwo0i:
this depends on https://github.com/ether/etherpad-require-kernel/pull/3 now. I need to revert the scrdoc change, as this is already in develop.

JohnMcLear:
published require kernel update

rhansen:
In the revert commit, would you update the commit message to explain why it needs to be reverted?

webzwo0i:
I wanted to add the message now, but then I realized the srcdoc patch was enough for scrolling and line number links to work. It works with and without https://github.com/ether/etherpad-lite/commit/470f40d7db01b68856edf47ac7a5befd50b50340, so I gonna keep it as I can't see any disadvantage.

I'm closing this and do a PR with the other commits in a separate branch.

PR #4957, opened 2021-03-17
import/export: Improve error handling, promisify

rhansen:
Multiple commits:
* tests: Increase import/export test timeouts
* tests: Promisify import/export tests
* tests: Use `assert` to simplify import/export tests
* import/export: Delete unnecessary comments
* Abiword: Reset stdout buffer when starting abiword
* ExportHandler: Replace unnecessary exception with `return`
* import/export: Use Error objects for errors, not strings
* ExportHandler: Pass the error unmodified
* import/export: Spelling fix: "convertor" -> "converter"
* import/export: On export error return 500 instead of crashing
* Abiword: Don't call the callback if null
* Abiword: Reduce log spam
* Abiword: Fix logging of conversion failure
* Abiword: Use the async-provided callback to signal errors
* LibreOffice: Add missing `fileExtension` property on intermediate step
* LibreOffice: Use consistent intermediate filename
* LibreOffice: Use `async.series` to properly handle conversion errors
* LibreOffice: Use the async-provided callback to signal errors
* import/export: Promisify Abiword and LibreOffice conversion
* LibreOffice: Log conversion errors

### DIFF
--- a/src/node/handler/ExportHandler.js
+++ b/src/node/handler/ExportHandler.js
@@ -98,8 +98,7 @@ exports.doExport = async (req, res, padId, readOnlyId, type) => {
           settings.soffice != null ? require('../utils/LibreOffice')
           : settings.abiword != null ? require('../utils/Abiword')
           : null;
-      // @TODO no Promise interface for converters (yet)
-      await util.promisify(converter.convertFile)(srcFile, destFile, type);
+      await converter.convertFile(srcFile, destFile, type);
     }
 
     // send the file
--- a/src/node/handler/ExportHandler.js
+++ b/src/node/handler/ExportHandler.js
@@ -33,17 +33,10 @@ const util = require('util');
 const fsp_writeFile = util.promisify(fs.writeFile);
 const fsp_unlink = util.promisify(fs.unlink);
 
-let convertor = null;
-
-// load abiword only if it is enabled
-if (settings.abiword != null) {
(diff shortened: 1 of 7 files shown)


JohnMcLear:
can we please modify the timeouts, not entirely remove them?  I don't mind if they are longer but they are present for a purpose :)

rhansen:
Mocha has a default timeout so it won't hang or anything. I prefer to leave out timeouts unless there's a reason to change the default because they're hard to tune and make tests flaky.

JohnMcLear:
The test timeouts are for detecting if a change in an implementation slows down the functionality.

rhansen:
It's better to have purpose-built benchmarks for testing slowdows. Using timeouts for this purpose isn't particularly helpful because the tests will fail on slow machines even if the change did not slow things down, and the tests will pass on fast machines even if the change did slow things down.

JohnMcLear:
> It's better to have purpose-built benchmarks for testing slowdows. Using timeouts for this purpose isn't particularly helpful because the tests will fail on slow machines even if the change did not slow things down, and the tests will pass on fast machines even if the change did slow things down.

While that may be the case can we please not make it part of the conversation of this PR to keep things simple and inline with the rest of the testing specs?

rhansen:
Bumped the timeout to 1s.

rhansen:
This is ready for review @JohnMcLear.

PR #4960, opened 2021-03-19
ace: Use iframe `srcdoc` property to refine frame load logic

rhansen:
This seems to fix "null is not an object (evaluating 'browserSheet.insertRule')" errors on Safari.

cc @webzwo0i


webzwo0i:
Great. I think we also need to revert the height patch, but I will do this in my other open PR.

PR #4963, opened 2021-03-20
cssmanager: Refactor CSS manager creation to avoid race condition

### LINKED ISSUE #4961: TypeError: null is not an object (evaluating 'browserSheet.insertRule')
Client-side exception at [src/static/js/cssmanager.js#L64](https://github.com/ether/etherpad-lite/blob/cb9f6d67768da5c87651c03bcc24e019574121aa/src/static/js/cssmanager.js#L64) on latest develop (cb9f6d67768da5c87651c03bcc24e019574121aa). Seems to only happen on Safari. It doesn't always happen, so it appears to be a race condition. I had hoped that PR #4960 would fix it, but it did not.

rhansen:
Multiple commits:
* cssmanager: Simplify iteration over style sheets
* cssmanager: Throw if no matching style sheet is found
* Ace2Inner: Promisify `init()`
* pad_editor: Promisify `init()`
* cssmanager: Refactor CSS manager creation to avoid race condition

Fixes #4961 

cc @webzwo0i


JohnMcLear:
Looks good to me!  Thanks!

PR #4966, opened 2021-03-21
bump require-kernel dependency

webzwo0i:


webzwo0i:
need to check if admin tests are failing or if they were blocked by too many saucelabs requests

PR #4968, opened 2021-03-21
add info for plugin devs regarding version string

webzwo0i:
not 100% if it makes sense to add this now, because it will probably change when switching bundler.

Now nearly all of our (core) resources have the version string attached, so on update there shouldn't be problems with stale entries anymore. However, some plugins link to static resources directly and this could result in stale entries after plugin updates.


webzwo0i:
I better put this into the wiki

PR #4970, opened 2021-03-22
Changeset: Return a new op object by default when iterating

rhansen:
Reusing the same op object for each iteration can result in very weird behaviors because previously yielded op objects will get a surprise mutation.

It is unclear why the code was written to reuse the same object. There was no comment, nor is there a commit message providing rationale (it has behaved this way since the very first commit). Perhaps the objects were reused to improve performance (fewer object allocations that need to be garbage collected). I do expect this change to reduce performance somewhat, but not enough to warrant reverting this commit.

### DIFF
--- a/src/static/js/Changeset.js
+++ b/src/static/js/Changeset.js
@@ -126,10 +126,9 @@ exports.opIterator = (opsStr, optStartIndex) => {
     return result;
   };
   let regexResult = nextRegexMatch();
-  const obj = exports.newOp();
 
-  const next = (optObj) => {
-    const op = (optObj || obj);
+  const next = (optOp) => {
+    const op = optOp || exports.newOp();
     if (regexResult[0]) {
       op.attribs = regexResult[1];
       op.lines = exports.parseNum(regexResult[2] || 0);


JohnMcLear:
🐙

webzwo0i:
You're supposed to `cloneOp` or `copyOp` if you want to reuse the op in case you call `next` without argument. Maybe a comment on `opIterator` would be enough?

I did a small test with etherpad-load-test and I cannot see any difference in performance. Maybe we should check what code uses next without arguments, as this is a very hot path (I expect tens of thousands executions per second on moderately used instances, but most of them could be not relevant due to calling the function with an op). A timeslider test would also be interesting, as I expect this is where clients do most of opIterations.

I also think this code was due to performance reasons and I don't think your patch has problematic consequences, so LGTM.

rhansen:
A comment is not enough, as the [principle of least surprise](https://en.m.wikipedia.org/wiki/Principle_of_least_astonishment) clearly applies here. If a caller wants to avoid a memory allocation by reusing an op object, the caller should provide the op object to reuse.

Issue #4978, opened 2021-03-26
Text sync problems with multiple users editing the same pad


ingoncalves:
We recently noticed a problem with text synchronization between users. Apparently randomly, **a user stops synchronizing the editions in the text, and after reloading the page, all the text is lost forever.**

By stressing the problem, we were able to establish one of the possible scenarios to reproduce the problem in a very consistent way.

**To Reproduce**
1. Two or more users on the same pad
2. User 1 has a problematic connection (slow or with packet loss)
3.  User 1 types composite characters (eg: `' + e = é`)
4. User 2 types any characters simultaneously with User 1
5. User 1 stops receiving changes from User 2 and submitting changes to the server
7. User 2 stops receiving changes from User 1
8. User 1 refreshes the page and loses his editions

**Expected behavior**
- All users have the same text

**Video**
This video shows the bug in action, were I follow the same steps described above. The left user is the **User 1** (running in a VM) and the right one is the **User 2** (running in the host machine). I use a simple robot to type characters in the pad for both users. I also use a network conditioner to simulate bad conditions. I left the Chrome Dev Tools opened in the Network tab, were we can the message flow through the socket, filtering messages with `changset` info.

We can notice that at certain moment, User 1 stops sending his changesets (green messages). After refreshing the page, he loses all his editions. The User 2 is not affected and keeps his editions after reloading the page.

https://user-images.githubusercontent.com/5985312/112691689-6fdfb780-8e5c-11eb-9134-80bd3262274a.mp4

**Server (please complete the following information):**
 - Etherpad version: 1.8.13

**Desktop (please complete the following information):**
 - OS: macOS Big Sur 11.2.3, macOS Catalina 10.15.7 and Ubuntu 14.04
 - Browser: Chrome 89.0.4389.90; Safari Version 13.1.3 and 14.0.3, Firefox 72.0.2 and 87.0

PR #4979, opened 2021-03-26
fix: Dequeue messages when committing

ingoncalves:
**This PR proposes a fix to the issue https://github.com/ether/etherpad-lite/issues/4978**

Let's try to explain this issue.

After a thorough analysis of the code, we found that the problem occurs in the message management performed by [collab_client.js](https://github.com/ether/etherpad-lite/blob/1.8.6/src/static/js/collab_client.js). The messages handler uses a `msgQueue` queue to store all messages coming from the server. There is also a state control that admits the values ​​`IDLE` and` COMMITING`. **The synchronization problem occurs when there is a combination of the state equal to `COMMITING` with the non-empty queue.**

Every time a user makes a change to the text, a new changeset is generated, but the changes are not yet applied to the base text, being suspended until confirmation from the server. So, if the state is `IDLE` (initial value), the user sends a `USER_CHANGES` message to the server and changes its state to `COMITTING`. 
https://github.com/ether/etherpad-lite/blob/5db0c8d1cfc3f354a5a8badce6d5c3890d428e98/src/static/js/collab_client.js#L151-L158
The server then receives and processes the message, and finally responds to the author user with a message `ACCEPT_COMMIT` and to the other users with the message` NEW_CHANGES`. When the user receives an `ACCEPT_COMMIT` message, he commits the changeset to the base text and sets its status to` IDLE` again.

It is worth mentioning that when a new message from the server is received, the user checks the message queue. If the queue is not empty, the user queues the message and postpones its processing. The special case is when the user receives a message `NEW_CHANGES`, where in addition to checking the status of the queue, he also checks if he is during a character composition. If so, the message is also queued.
https://github.com/ether/etherpad-lite/blob/5db0c8d1cfc3f354a5a8badce6d5c3890d428e98/src/static/js/collab_client.js#L222

This mechanism works very quickly and the queue is almost never larger than 1. However, with a slow or problematic connection, the user may be in the `COMMITING` state waiting for the` ACCEPT_COMMIT` from the server and, concurrently, he receives messages ` NEW_CHANGES` during a song. In this case, the message will be queued and the status will continue as `COMMITING`. When receiving `ACCEPT_COMMIT`, the queue will not be empty and this message will also be queued, without any change of state. In this special condition, `ACCEPT_COMMIT` will no longer be processed and status will not be changed. Consequently, new changes will not be sent to the server, since the sending only occurs when the file is `IDLE`.
This loop occurs in the line `102`
https://github.com/ether/etherpad-lite/blob/5db0c8d1cfc3f354a5a8badce6d5c3890d428e98/src/static/js/collab_client.js#L94-L105

**The solution we propose is to process all messages in the queue whenever the function `handleUserChanges` is called with `state == COMMITING`.** We add one more condition to this `if/else` statement, after checking network issues.
```js
//...
} else if (msgQueue.length > 0) {
  dequeueMessages();
} else {
//...
```

### DIFF
--- a/src/static/js/collab_client.js
+++ b/src/static/js/collab_client.js
@@ -97,6 +97,10 @@ const getCollabClient = (ace2editor, serverVars, initialUserInfo, options, _pad)
         setChannelState('DISCONNECTED', 'slowcommit');
       } else if (state === 'COMMITTING' && msgQueue.length === 0 && (t - lastCommitTime) > 5000) {
         callbacks.onConnectionTrouble('SLOW');
+      } else if (msgQueue.length > 0) {
+        // in slow or bad network conditions, there might be enqueued messages
+        // while the state is still COMMITTING
+        dequeueMessages();
       } else {
         // run again in a few seconds, to detect a disconnect
         setTimeout(wrapRecordingErrors('setTimeout(handleUserChanges)', handleUserChanges), 3000);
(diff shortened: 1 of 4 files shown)


rhansen:
Thanks for this! I need to do some studying to understand this change. In the meantime would you split the commit into two:
1. A commit that factors out the duplicate `ACCEPT_COMMIT` handling logic into a separate `acceptCommit()` function.
2. The rest of your changes.

rhansen:
Superseded by #4983.

PR #4992, opened 2021-04-06
escape userId before setting it as HTML attribute

webzwo0i:


rhansen:
It would be better to safely build the DOM nodes:

```javascript
        const chatMsg = $('<p>')
            .attr('data-authorId', msg.userId)
            .addClass(authorClass)
            .append($('<b>').text(authorName))
            .append(document.createTextNode(' '))
            .append($('<span>')
                .addClass('time')
                .addClass(authorClass)
                .text(ctx.timeStr))
            .append(document.createTextNode(` ${ctx.text}`));
        if (isHistoryAdd) chatMsg.insertAfter('#chatloadmessagesbutton');
        else $('#chattext').append(chatMsg);
```

rhansen (approved):

pcworld:
Hi, thanks for taking care of this!

I have a bit of feedback regarding the handling of security fixes:
The gap between publicly committing this and making a new release (or security announcement) was quite long.
I would suggest that in the future, you either make a new release within a short time frame, potentially only containing the security fix, or you wait until before a release is imminent before publicly pushing security fixes.
That way, malicious users do not get to see the security bug in advance and use that time to build and deploy an exploit.

On another note, I have looked quite a bit at this, but haven't found a way to exploit it. I could not find where the `userId` can be controlled by users of the platform, as far as I could backtrack the code it is generated as a random string and stored in a server-side session storage.
What am I missing?
(The only candidate place I have seen is the [HTTP API's createSession endpoint](https://etherpad.org/doc/v1.8.14/#index_createsessiongroupid-authorid-validuntil), but I haven't looked further into it whether this only accepts existing or arbitrary authorIDs.)

rhansen:
> I have looked quite a bit at this, but haven't found a way to exploit it.

I don't see a way to exploit it either. @webzwo0i Was this change motivated by a report from an analysis tool?

PR #5002, opened 2021-04-12
Fix read only pad access with authentication

pcworld:
Before this PR, webaccess.checkAccess saved the authorization in
user.padAuthorizations[padId] with padId being the read-only pad ID,
however later stages, e.g. in PadMessageHandler, use the real pad ID for
access checks. This led to authorization being denied.

This commit fixes it by only storing and comparing the real pad IDs and
not read-only pad IDs.

This fixes test case "authn user readonly pad -> 200, ok" (which is also introduced in this PR) in
src/tests/backend/specs/socketio.js.

---

Note that I haven't really understood the design principles behind the authorization architecture. In particular, the `padAuthorizations` being part of the user session seems really weird to me and is the root cause behind the bug being fixed by this PR.

### DIFF
--- a/src/node/db/SecurityManager.js
+++ b/src/node/db/SecurityManager.js
@@ -22,6 +22,7 @@
 const authorManager = require('./AuthorManager');
 const hooks = require('../../static/js/pluginfw/hooks.js');
 const padManager = require('./PadManager');
+const readOnlyManager = require('./ReadOnlyManager');
 const sessionManager = require('./SessionManager');
 const settings = require('../utils/Settings');
 const webaccess = require('../hooks/express/webaccess');
@@ -55,6 +56,15 @@ exports.checkAccess = async (padID, sessionCookie, token, userSettings) => {
   }
 
   let canCreate = !settings.editOnly;
+
+  if (readOnlyManager.isReadOnlyId(padID)) {
+    canCreate = false;
+    padID = await readOnlyManager.getPadId(padID);
+    if (padID == null) {
+      authLogger.debug('access denied: read-only pad ID for a pad that does not exist');
+      return DENY;
+    }
+  }
 
   // Authentication and authorization checks.
   if (settings.loadTest) {
--- a/src/node/hooks/express/webaccess.js
+++ b/src/node/hooks/express/webaccess.js
@@ -70,14 +70,19 @@ const checkAccess = async (req, res, next) => {
   // This helper is used in steps 2 and 4 below, so it may be called twice per access: once before
   // authentication is checked and once after (if settings.requireAuthorization is true).
   const authorize = async () => {
-    const grant = (level) => {
+    const grant = async (level) => {
       level = exports.normalizeAuthzLevel(level);
       if (!level) return false;
       const user = req.session.user;
       if (user == null) return true; // This will happen if authentication is not required.
       const encodedPadId = (req.path.match(/^\/p\/([^/]*)/) || [])[1];
       if (encodedPadId == null) return true;
-      const padId = decodeURIComponent(encodedPadId);
+      let padId = decodeURIComponent(encodedPadId);
+      if (readOnlyManager.isReadOnlyId(padId)) {
+        // pad is read-only, first get the real pad ID
+        padId = await readOnlyManager.getPadId(padId);
+        if (padId == null) return false;
+      }
       // The user was granted access to a pad. Remember the authorization level in the user's
       // settings so that SecurityManager can approve or deny specific actions.
       if (user.padAuthorizations == null) user.padAuthorizations = {};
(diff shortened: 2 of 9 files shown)


src/node/hooks/express/webaccess.js
```diff
@@ -70,28 +70,34 @@ const checkAccess = async (req, res, next) => {
...
       if (!level) return false;
       const user = req.session.user;
       if (user == null) return true; // This will happen if authentication is not required.
       const encodedPadId = (req.path.match(/^\/p\/([^/]*)/) || [])[1];
       if (encodedPadId == null) return true;
-      const padId = decodeURIComponent(encodedPadId);
+      let padId = decodeURIComponent(encodedPadId);
+      if (readOnlyManager.isReadOnlyId(padId)) {
+        // pad is read-only, first get the real pad ID
+        padId = await readOnlyManager.getPadId(padId);
```
pcworld, line 83: The alternative to performing normalization here would be ensuring that checkAccess() is never called with read-only pad IDs. As far as I can tell, this would currently only affect src/node/padaccess.js (which does not get the real pad ID before querying checkAccess) which is used by pad export.
However, I have decided to perform normalization here in order to prevent similar bugs in the future if developers would forget to "normalize" the pad ID elsewhere.

src/node/db/SecurityManager.js
```diff
@@ -70,6 +71,13 @@ exports.checkAccess = async (padID, sessionCookie, token, userSettings) => {
     // Note: userSettings.padAuthorizations should still be populated even if
     // settings.requireAuthorization is false.
     const padAuthzs = userSettings.padAuthorizations || {};
+    if (readOnlyManager.isReadOnlyId(padID)) {
+      // pad is read-only, first get the real pad ID; some callers of checkAccess() first fetch the real pad ID,
+      // but some do not
+      padID = await readOnlyManager.getPadId(padID);
```
pcworld, line 77: The alternative to performing normalization here would be ensuring that securityManager.checkAccess() is never called with read-only pad IDs. As far as I can tell, this would currently only affect src/node/padaccess.js (which does not get the real pad ID before querying checkAccess) which is used by pad export.
However, I have decided to perform normalization here in order to prevent similar bugs in the future if developers would forget to "normalize" the pad ID elsewhere.

webzwo0i:
@rhansen Any idea why PadMessageHandler checks auth on the r/w padId when actually requesting a readonly pad address? (for handleMessage and SWITCH_TO_PAD) Looks like old code?!

src/node/db/SecurityManager.js
```diff
@@ -70,6 +71,13 @@ exports.checkAccess = async (padID, sessionCookie, token, userSettings) => {
...
+      if (padID === null) return DENY;
```
rhansen, line 78:
```suggestion
      if (padID == null) return DENY;
```

rhansen, line 78:
```suggestion
      if (padID == null) return DENY;
```

src/node/hooks/express/webaccess.js
```diff
@@ -70,28 +70,34 @@ const checkAccess = async (req, res, next) => {
...
+        if (padId === null) return false;
```
rhansen, line 84:
```suggestion
        if (padId == null) return false;
```

rhansen:
> @rhansen Any idea why PadMessageHandler checks auth on the r/w padId when actually requesting a readonly pad address? (for handleMessage and SWITCH_TO_PAD) Looks like old code?!

All auth checks should use the regular (read+write) pad ID. There are two main reasons:
* The consistency should make it easier to understand the code. (The auth code is already super difficult to understand; imagine how much worse it would be if access grants were tracked separately for read-only pad IDs.)
* It should make it harder for auth plugin authors to accidentally introduce an auth bypass bug in their plugin. For example, suppose there is a plugin that limits access to pads whose IDs begin with "private" to a specific group of people. If pad IDs were not normalized and the plugin forgot to convert read-only IDs to read+write IDs then a malicious user could access the pad via the read-only ID (if they could figure out the read-only ID).

@pcworld I have not yet reviewed the changes to the tests; I will do that later today. The non-test changes look good. Thank you!

rhansen:
@pcworld Please take a look at the commits I pushed to your branch. If they look OK to you, I'll squash the fixup commits into your commits.

pcworld:
> @pcworld Please take a look at the commits I pushed to your branch. If they look OK to you, I'll squash the fixup commits into your commits.

Your changes look fine to me. I haven't looked _too_ closely at them though. I suggest that a third person looks over it.

webzwo0i:
LGTM and also thanks for the clarification

webzwo0i (approved):

PR #5006, opened 2021-04-13
fix(perf): Disable wtfnode dump by default

Chocobozzz:
Consumes a lot of CPU (because of https://github.com/myndzi/wtfnode/blob/master/index.js#L123) so it's better to enable it on purpose in production.

### DIFF
--- a/settings.json.docker
+++ b/settings.json.docker
@@ -463,6 +463,11 @@
    */
   "loadTest": "${LOAD_TEST:false}",
 
+  /**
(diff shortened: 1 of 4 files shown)


webzwo0i:
thanks!

webzwo0i:
I did merge it already, but now I think defaulting to enabling wtfnode would be better. Large instances will likely disable it, but most other instances will have it enabled

Issue #5000, opened 2021-04-09
authorID and groupID created by API does not exist, in following request


a-r-m-i-n:
In the past week I've integrated Etherpad Lite to a PHP application. The group pads get created using the API and then I create a session, save the cookie and show the group pad in iframe.

Everything works like charm (when using latest version 1.8.13). I'm using Docker for local development and also for the staging server. On the staging server I had weird API errors.

Basically I'm using those API calls (in this order):
- `createAuthorIfNotExistsFor` -> I get a valid authorID (a.xyz)
- `listSessionsOfAuthor` 

Suddenly I get **error code 4** (which is invalid or missing apikey) with the message "**authorID does not exist**", when calling `listSessionsOfAuthor` route. This only occures for the first time a portal user get in touch with Etherpad API. I was able to reconstruct it with a second user, but then I wasn't able to reproduce it anymore (with other, new users) on Staging environment. Locally I'm entirely unable to reconstruct this issue (with same Docker containers).

Same with:
- `createGroupIfNotExistsFor` -> I get a valid groupID (g.zzz)
- `createGroupPad`

**Error 4: groupID does not exist.**

Just for the records: We are talking about the authorID and groupID, which the API gave me one request earlier. Also, the apikey is given and valid in query params. 

Because this issue does not appear everytime, I guess there is an issue with the database itself (I'm on MySQL)? Concurrent requests, maybe. Do you ever had similar issues? Do I need to add sleep commands between my requests?

For the API requests in PHP I'm using this lib: https://packagist.org/packages/tomnomnom/etherpad-lite-client

Thanks in advance!

a-r-m-i-n:
🎉 #5000

a-r-m-i-n:
Here I got another example:
![image](https://user-images.githubusercontent.com/1275298/114199975-0d56e300-9955-11eb-9652-efc86c3c7dc4.png)

webzwo0i:
Interesting! And congrats for 5000th issue. :-)

I would recommend running etherpad with DEBUG logging enabled (loglevel: DEBUG in settings.json) This should log db queries and tell you if the data is read from cache/dirty cache or database.
The queries are serialized on one connection, so there is no concurrency. Additionally, we use a buffer/cache from which entries are normally served, unless they are served for the first time (then they will hit the database). So in theory, if the problem only happens when you want to get back a value, that you have just written, we probably have a problem with serving dirty values.

a-r-m-i-n:
[...]

PR #5011, opened 2021-04-17
await in AuthorManager.js

webzwo0i:
Should partially fix https://github.com/ether/etherpad-lite/issues/5000

createGroupIfNotExistsFor/createGroup seems to be using the correct code already, so I'm not sure yet what happened

### DIFF
--- a/src/tests/backend/specs/regression-db.js
+++ b/src/tests/backend/specs/regression-db.js
@@ -1,0 +1,31 @@
+'use strict';
+
+const AuthorManager = require('../../../node/db/AuthorManager');
+const assert = require('assert').strict;
+const common = require('../common');
+const db = require('../../../node/db/DB');
+
+describe(__filename, function () {
+  let setBackup;
+
+  before(async function () {
+    await common.init();
+    setBackup = db.set;
+
+    db.set = async (...args) => {
+      // delay db.set
+      await new Promise((resolve) => { setTimeout(() => resolve(), 500); });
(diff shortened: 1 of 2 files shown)


webzwo0i:
making draft, as it breaks other tests. The new test runs, but other tests that need the database fail now. (maybe the delayed set is still run?)

webzwo0i:
I had code in that checked for the size of db.buffer._cache before the test runs. It showed that there was an globalAuthor entry (probably from one of the last tests?). After the test finishes, I check db.buffer._cache again to see if there are now two values inside (1 from before and 1 from running my test). Unexpectedly, there is only one!

src/tests/backend/specs/regression-db.js
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  after(async function () {
+    // ensure db is closed after all writes have been written
+    await waitForDB();
+    db.set = tmp;
+    if (standalone) {
+      await db.shutdown();
+    }
+  });
```
rhansen, line 55: None of this is necessary if you do `await common.init();`.

webzwo0i: I do use `common.init()` now and removed everything except `db.set = tmp` to undo the delayed set. Thanks for pointing that out!

rhansen:
I pushed a couple of fixup commits; please take a look. If they look OK to you I will squash them into their respective commits and merge.

webzwo0i:
looking great!

PR #5024, opened 2021-05-05
Revert "PadMessageHandler: Use a `Map` for `sessioninfos`"

rhansen:
Switching to a Map broke ep_webrtc and maybe other plugins.

This reverts commit eeead4643792d2e7a10aef935112bab8e620375c.

### DIFF
--- a/src/node/db/API.js
+++ b/src/node/db/API.js
@@ -810,12 +810,15 @@ exports.createDiffHTML = async (padID, startRev, endRev) => {
 
 exports.getStats = async () => {
   const sessionInfos = padMessageHandler.sessioninfos;
-  const map = function* (it, fn) { for (const i of it) yield fn(i); };
-  const activePads = new Set(map(sessionInfos.values(), ({padId}) => padId));
+
+  const sessionKeys = Object.keys(sessionInfos);
+  const activePads = new Set(Object.entries(sessionInfos).map((k) => k[1].padId));
+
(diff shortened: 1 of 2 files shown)


webzwo0i:
From the list of plugins that are publicly available via static.etherpad.org on 22th April the following seem to use sessioninfos:
ep_auth_author
ep_auth_citizenos
ep_delete_after_delay
ep_profile_modal
ep_rtc
ep_webhooks
ep_webrtc
ep_wrtc_heading

In the past people complained whenever something that was part of `exports` of a file changed in a backwards incompatible way without incrementing the major version. Maybe this change is something for a 2.x branch? We don't have code yet to install/find the last compatible version of a plugin in case we want to support two versions of a plugin (one that works with 1.x and one for 2.x) and do this in a npm-compatible way. I expect this will make it hard in the future to do similar breaking changes...

Regarding this PR which is very useful imo, can we get around it with ES proxys, internally using a Map?
Something like:
```
  class sessioninfos {
    constructor() {   
      this.map = new Map();
      return new Proxy(this, {
        get: (obj, key) => {
          return this.map.get(key);
        },
        set: (obj, key, val) => {
          return this.map.set(key,val);
        }
      })
    }
  }
  exports.sessioninfos = new sessioninfos();
```

In case this is not possible I can issue PRs to the plugins that use sessioninfos so they use either object or map depending on the version in package.json

PR #5051, opened 2021-06-06
Settings: Support `null` and `undefined` environment var substitutions

### LINKED ISSUE #5048: Connecting to Redis without authentication appears impossible
**Describe the bug**
It appears to be impossible to connect to a Redis database without authentication. Attempting to do so produces the following error in node-redis:

```
Error: node_redis: The AUTH command contains a invalid argument type.
Only strings, dates and buffers are accepted. Please update your code to use valid argument types.
    at RedisClient.internal_send_command (/opt/etherpad-lite/src/node_modules/redis/index.js:835:39)
    at RedisClient.auth (/opt/etherpad-lite/src/node_modules/redis/lib/individualCommands.js:216:20)
    at RedisClient.create_stream (/opt/etherpad-lite/src/node_modules/redis/index.js:244:14)
    at new RedisClient (/opt/etherpad-lite/src/node_modules/redis/index.js:121:10)
    at Object.exports.createClient (/opt/etherpad-lite/src/node_modules/redis/index.js:1023:12)
    at exports.Database.init (/opt/etherpad-lite/src/node_modules/ueberdb2/databases/redis_db.js:55:23)
    at internal/util.js:308:30
    at new Promise (<anonymous>)
    at Object.bound  [as init] (internal/util.js:307:12)
    at exports.Database.init (/opt/etherpad-lite/src/node_modules/ueberdb2/lib/CacheAndBufferLayer.js:267:24)
```

While I haven't debugged it properly, the background seems to be that node-redis [checks for `auth_pass !== undefined`](https://github.com/NodeRedis/node-redis/blob/4f85030e42da2eed6a178e54994330af5062761e/index.js#L241) to decide whether to send the auth command, but Etherpad passes (via ueberDB) `null`, at least when configured via environment variables (as evidenced by the console warning: `Environment variable "DB_PASS" does not contain any value for configuration key "password", and no default was given. Returning null.`). That then leads to the crash above on the type checking in node-redis's `internal_send_command` function.

**To Reproduce**
1. Build Docker container: `docker build --tag foo/etherpad .`
2. Run Docker with settings to connect to a Redis server that does not require authentication, i.e. with `DB_PASS` unspecified: `docker run -e 'DB_TYPE=redis' -e 'DB_HOST=example.org' foo/etherpad`

In my specific case, I have Redis in another container and exposed via a Docker network, but I don't think that is relevant because the crash happens before it actually attempts to connect to the server anyway.

**Expected behavior**
A connection is established without attempting authentication.

 - Etherpad version: 3bca8528
 - OS: Debian buster
 - Docker: 18.09.1 (from Debian repos)
 - Node.js version (`node --version`): 14.17.0
 - npm version (`npm --version`): 6.14.13

rhansen:
Multiple commits:
* Settings: Support null and undefined env var substitutions
* Docker: Explicitly default env var substitutions to null
* Settings: Deprecate null as the default default value
* Docker: If `DB_*` env var is unset, remove the corresponding setting
* tests: Add tests for `settings.json` parsing

Fixes #5048

@JustAnotherArchivist: Would you test this PR to see if it fixes the bug you reported?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,19 @@
   connections to the MySQL/MariaDB server (by default) instead of 1. This might
   cause Etherpad to crash with a "ER_CON_COUNT_ERROR: Too many connections"
   error if your server is configured with a low connection limit.
+* Changes to environment variable substitution in `settings.json` (see the
+  documentation comments in `settings.json.template` for details):
+  * An environment variable set to the string "null" now becomes `null` instead
+    of the string "null". Similarly, if the environment variable is unset and
+    the default value is "null" (e.g., `"${UNSET_VAR:null}"`), the value now
+    becomes `null` instead of the string "null". It is no longer possible to
+    produce the string "null" via environment variable substitution.
+  * An environment variable set to the string "undefined" now causes the setting
+    to be removed instead of set to the string "undefined". Similarly, if the
+    environment variable is unset and the default value is "undefined" (e.g.,
+    `"${UNSET_VAR:undefined}"`), the setting is now removed instead of set to
+    the string "undefined". It is no longer possible to produce the string
+    "undefined" via environment variable substitution.
 
 ### Notable enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@
     `"${UNSET_VAR:undefined}"`), the setting is now removed instead of set to
     the string "undefined". It is no longer possible to produce the string
     "undefined" via environment variable substitution.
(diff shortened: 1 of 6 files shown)


JustAnotherArchivist:
Yes, I can confirm that 6c2f31a5 fixes #5048. I did not test any other setups though, e.g. authenticated Redis or other databases. Especially the latter should really be tested in case one of the other database libraries expects `null` rather than `undefined`.

rhansen:
Thanks for testing! I'm not too concerned about thoroughly testing other databases. If there's a problem we can always revert and retry.

PR #5064, opened 2021-06-13
update package-lock

webzwo0i:


rhansen:
How was this generated? I don't get a diff when I run:

```shell
rm src/package-lock.json && (cd src && npm i)
```

webzwo0i:
Actually in the way you wrote. I think I used npm 6.4.12

PR #5068, opened 2021-06-14
deps: Drop support for old Node.js versions

rhansen:
Multiple commits:
* deps: Drop support for Node.js < 12.13.0
* tests: Replace Node.js v15 with v16
* lint: Update eslint-config-etherpad and friends


webzwo0i:
thanks, merging. just for the record: the failing 'upgrade from latest release' workflow is already fixed in a001a13411e377773b1e17ebda04585cb4b05070

PR #5075, opened 2021-06-16
CSS: Fix button icon centering

rhansen:
cc @packardone


webzwo0i:
Just stumbled upon this: I initially thought it's a bug, but I assume it was the original intention to have it centered because of `text-align: center;`

PR #5087, opened 2021-06-17
bin/importSqlFile.js: Read the file line at a time

### LINKED ISSUE #5085: importSqlFile.js does not work for very large SQL dumps
**Describe the bug**

After converting a 2,7 GiB SQLite database to a 1,7 GiB SQL dump, it is not possible to ingest the file with `importSqlFile.js`.

**To Reproduce**

1. Create a large anough SQL dump from an Etherpad instance
2. Run `node importSqlFile.js etherpad.sql`
3. See error

```
[2021-06-17 22:57:04.504] [INFO] console - 0.098        initializing db
[2021-06-17 22:57:04.510] [INFO] console - 0.104        done
[2021-06-17 22:57:04.510] [INFO] console - 0.104        open output file...
/opt/etherpad-lite/src/bin/importSqlFile.js:5
process.on('unhandledRejection', (err) => { throw err; });
                                            ^

Error: Cannot create a string longer than 0x1fffffe8 characters
    at Object.slice (buffer.js:605:37)
    at Buffer.toString (buffer.js:804:14)
    at Object.readFileSync (fs.js:438:41)    at /opt/etherpad-lite/src/bin/importSqlFile.js:75:20    at processTicksAndRejections (internal/process/task_queues.js:95:5) {  code: 'ERR_STRING_TOO_LONG'
}
```

**Expected behavior**

The database dump is imported asynchronously, in so Node can parse the file as a stream. As an explanation I like what was said in https://github.com/jprichardson/node-fs-extra/issues/653#issuecomment-449763179:

> You're dealing with a file that is far to large to load into memory at once. We cannot buffer the file for you, etc. because the size of the file is larger than Node.js itself is capable of creating a string for; actually, larger than Node.js is able to store in memory at once as well. You're going to have to use streams to work with such a big file.

Unfortunately I am not knowledgeable enough with JavaScript to turn the code in

https://github.com/ether/etherpad-lite/blob/485538bd79620f254e1088d84abd8658435fe563/src/bin/importSqlFile.js#L75-L95

into an async version. Is it possible to refactor this code in such a way, that the very large SQL dump can be ingested and migrated into the configured ueberDB store?

rhansen:
This avoids running out of memory if the file is large.

@almereyda Would you test this to see if it fixes your issue?

Fixes #5085

### DIFF
--- a/src/bin/importSqlFile.js
+++ b/src/bin/importSqlFile.js
@@ -48,6 +48,7 @@ const unescape = (val) => {
 (async () => {
   const fs = require('fs');
   const log4js = require('log4js');
+  const readline = require('readline');
   const settings = require('../node/utils/Settings');
   const ueberDB = require('ueberdb2');
 
@@ -71,14 +72,12 @@ const unescape = (val) => {
   await util.promisify(db.init.bind(db))();
   log('done');
 
-  log('open output file...');
-  const lines = fs.readFileSync(sqlFile, 'utf8').split('\n');
+  log(`Opening ${sqlFile}...`);
+  const stream = fs.createReadStream(sqlFile, {encoding: 'utf8'});
 
-  const count = lines.length;
+  log(`Reading ${sqlFile}...`);
   let keyNo = 0;
-
-  process.stdout.write(`Start importing ${count} keys...\n`);
-  lines.forEach((l) => {
+  for await (const l of readline.createInterface({input: stream, crlfDelay: Infinity})) {
     if (l.substr(0, 27) === 'REPLACE INTO store VALUES (') {
       const pos = l.indexOf("', '");
       const key = l.substr(28, pos - 28);
@@ -88,11 +87,9 @@ const unescape = (val) => {
       console.log(`unval: ${unescape(value)}`);
       db.set(key, unescape(value), null);
       keyNo++;
-      if (keyNo % 1000 === 0) {
-        process.stdout.write(` ${keyNo}/${count}\n`);
-      }
+      if (keyNo % 1000 === 0) log(` ${keyNo}`);
     }
-  });
+  }
   process.stdout.write('\n');
   process.stdout.write('done. waiting for db to finish transaction. ' +
                        'depended on dbms this may take some time..\n');


almereyda:
Thank you for the quick reaction! Yes, this allowed to read the file:

```
[2021-06-17 23:53:38.926] [INFO] console - 0.089        initializing db
[2021-06-17 23:53:38.931] [INFO] console - 0.094        done
[2021-06-17 23:53:38.931] [INFO] console - 0.094        Opening /var/etherpad-lite/etherpad.sql...
[2021-06-17 23:53:38.932] [INFO] console - 0.095        Reading /var/etherpad-lite/etherpad.sql...

done. waiting for db to finish transaction. depended on dbms this may take some time..
[2021-06-17 23:53:55.033] [INFO] console - 16.196       finished, imported 0 keys.
```

Unfortunately it appears this did not load any data.

rhansen:
> Unfortunately it appears this did not load any data, due to the formatting of my dump.

You may have better luck feeding the SQL directly to your database using your database's own tools.

This script does not look terribly useful. I wonder where the input file is expected to come from, as the script expects a very specific format. I'm going to merge this PR anyway; it's better than nothing. If you have suggestions for improving this script please feel free to open a new bug or send a PR.

src/bin/importSqlFile.js
```diff
@@ -71,14 +72,12 @@ const unescape = (val) => {
...
+  log(`Opening ${sqlFile}...`);
+  const stream = fs.createReadStream(sqlFile, {encoding: 'utf8'});
 
-  const count = lines.length;
+  log(`Reading ${sqlFile}...`);
   let keyNo = 0;
-
-  process.stdout.write(`Start importing ${count} keys...\n`);
-  lines.forEach((l) => {
+  for await (const l of readline.createInterface({input: stream, crlfDelay: Infinity})) {
```
almereyda, line 80: It appears this loop is never entered.
Placing a

    console.log('enter')

or

    log('enter')

in the line below is never evaluated, in this case shown on the screen.

src/bin/importSqlFile.js
```diff
@@ -71,14 +72,12 @@ const unescape = (val) => {
...
     if (l.substr(0, 27) === 'REPLACE INTO store VALUES (') {
```
almereyda, line 81: In my case this would be

```js
if (l.substr(0, 26) === 'INSERT INTO store VALUES (') {
```

src/bin/importSqlFile.js
```diff
@@ -71,14 +72,12 @@ const unescape = (val) => {
...
       const pos = l.indexOf("', '");
```
almereyda, line 82: Here I find the following pattern to be present as a deliminator in the SQLite dump:

```js
const pos = l.indexOf("','");
```

src/bin/importSqlFile.js
```diff
@@ -71,14 +72,12 @@ const unescape = (val) => {
...
       const key = l.substr(28, pos - 28);
```
almereyda, line 83: With the whole string moving one character to the left, due to `INSERT` having one letter less than `REPLACE`, here I use:

```
const key = l.substr(27, pos - 27);
```

src/bin/importSqlFile.js
rhansen: If I make the following change I see each line in the log, so it does appear to be reading the file correctly:
```diff
diff --git a/src/bin/importSqlFile.js b/src/bin/importSqlFile.js
index 148503e8..4146463d 100644
--- a/src/bin/importSqlFile.js
+++ b/src/bin/importSqlFile.js
@@ -78,6 +78,7 @@ const unescape = (val) => {
   log(`Reading ${sqlFile}...`);
   let keyNo = 0;
   for await (const l of readline.createInterface({input: stream, crlfDelay: Infinity})) {
+    console.log(l);
     if (l.substr(0, 27) === 'REPLACE INTO store VALUES (') {
       const pos = l.indexOf("', '");
       const key = l.substr(28, pos - 28);
```

almereyda:
The syntactic modifications for valid parsing of the SQL text produced by an SQLite dump have ultimately been documented in https://github.com/ether/etherpad-lite/issues/5085#issuecomment-863665749

PR #5090, opened 2021-06-18
Don't create author on empty token

webzwo0i:
If you have objections regarding dropping that one Promise.all call I can revert that commit.

The important patch in this PR is that it won't generate/update timestamps of authors when there is no token


rhansen:
> If you have objections regarding dropping that one Promise.all call I can revert that commit.

I'm not sure whether the change is correct, but either way the code is harder to read and understand. I would rather err on the side of readability and instead find an alternative approach to reduce latency (e.g., move the DB commits to a background buffer flush).

src/node/db/AuthorManager.js
```diff
@@ -181,41 +185,42 @@ exports.createAuthor = async (name) => {
 
 /**
  * Returns the Author Obj of the author
- * @param {String} author The id of the author
+ * @param {String} authorID The id of the author
```
rhansen, line 188: Nit: Use `authorId` instead of `authorID`. Rationale: Style guides used by many companies require lowercase `d` because it facilitates conversion to snake case by tools that bridge languages (every uppercase letter is converted to an underscore followed by that letter in lowercase). We don't use such tools (yet), but consistency with those style guides has value.

src/node/db/AuthorManager.js
```diff
@@ -107,6 +107,10 @@ exports.doesAuthorExists = exports.doesAuthorExist;
  * @param {String} token The token
  */
 exports.getAuthor4Token = async (token) => {
+  // e.g. in case of an export of a public pad, a token is not needed
+  if (token == null) {
+    return;
+  }
```
rhansen, line 113: I don't think the comment is necessary, and I would explicitly return `null`:

```suggestion
  if (token == null) return null;
```

src/node/db/AuthorManager.js
```diff
@@ -107,6 +107,10 @@ exports.doesAuthorExists = exports.doesAuthorExist;
  * @param {String} token The token
  */
 exports.getAuthor4Token = async (token) => {
+  // e.g. in case of an export of a public pad, a token is not needed
+  if (token == null) {
```
rhansen, line 111: Wouldn't it be better to change the caller to not call this function if `token` is `null`?

src/node/db/AuthorManager.js
webzwo0i: you mean only in the documentation? won't that cause jsdoc etc to complain that it does not know the signature?
but anyway, can do.

src/node/db/AuthorManager.js
webzwo0i: I thought about this but choosed to do it here, because there are many callers of that function in plugins. but now, that I write this line, I think I need to check the (plugin) callers anyway, because until now `getAuthor4Token` never returned null/undefined, so plugins could break when they don't expect null/undefined

src/node/db/AuthorManager.js
rhansen: Both documentation and code.

src/node/db/AuthorManager.js
webzwo0i: I checked all plugins that are available via static.etherpad.org:
The following seem to be save (explicitly checking for token before calling `getAuthor4Token`):
ep_auth_author
ep_codepad
ep_headerauth
ep_httpauth_author
ep_idea_review
ep_jwt
ep_ldapauth
ep_pomerium
ep_termin_app_integration

The following are save because they explicitly test the message.type to be `CLIENT_READY` (and PadMessageHandler checks for token in the `CLIENT_READY` case):
ep_openid_connect
ep_readonly_guest

The following could be vulnerable, need to check again tomorrow:
ep_cloudron and ep_hash_auth. They seem to register the `handleMessage` hook, while not checking if they parse an `CLIENT_READY` message. Other messages like `USER_CHANGES` don't have a token in `context.message`. I wonder if this is problematic ie it looks to me as if they get the same "undefined" author (which is a real authorId) for any requests...

rhansen: Thinking about it more, I think the best approach would be to throw if given a falsy value (after updating plugins as necessary):
```javascript
  if (!token) throw new Error('missing token');
```

Issue #5100, opened 2021-07-03
Frontend tests seem to be cached


webzwo0i:
Tested with ep_font_size, when I change the contents of frontend test files, I need to restart Etherpad to get a fresh version

rhansen:
I'm able to reproduce if I use the default `minify` and `maxAge` settings (`true` and `21600`, respectively). If I disable minify then it works as expected. Perhaps this is a general (not test-specific) bug in `caching_middleware.js`?

webzwo0i:
OK, I think I'm able to find it then. Will bump if not

webzwo0i:
Bisect identified https://github.com/ether/etherpad-lite/pull/5050/commits/e4f011df760b2eb6cd1142a14a4dbd17c4eed2b6 as the first commit with this behavior.

Before this commit an ETag header was sent and no Cache-Control header. The ETag changed when the test file changed, so the browser receives the most recent version.
Now two things happen:
1. No ETag header is sent anymore, but a Cache-Control header with a max-age based on settings.json's maxAge setting.
2. The file is cached in ./var by caching_middleware.js and files are never updated. Because there is no version string attached in the request, the file is even not updated on restart, unless browser cache is disabled.

webzwo0i:
See https://github.com/ether/etherpad-lite/pull/5102 and https://github.com/ether/etherpad-require-kernel/pull/6
The first is not really helpful, but for clarity (and to save some bytes in browser cache when excessively testing). If you think it's superfluous please reject. :-)
The second is a workaround - it just appends some bytes (based on the date of the request) in case test specs are requested.

rhansen:
What would it take to fix this properly? Fixing this properly would also benefit non-test files too, right?

PR #5102, opened 2021-07-04
set max-age to 0 when serving test specs

webzwo0i:


rhansen:
I reopened this thinking it would be worth finishing up, but thinking about it more I would prefer a general fix to the underlying problem reported in #5100.

Issue #5005, opened 2021-04-13
Etherpad crashs when querying stats API with timeout


0x46616c6b:
**Describe the bug**

We wrote a [Prometheus Exporter for Etherpad](https://github.com/systemli/prometheus-etherpad-exporter) which aggregates the metrics from `/stats` and if an api key is configured also the [`getStats`](https://etherpad.org/doc/v1.8.4/#index_getstats) data from the API. The exporter runs independently and greps every 15 seconds the data from Etherpad. After Etherpad restarted the API call takes more time and run into a timeout which kills Etherpad entirely. So you end up into a endless crashloop until the exporter is stopped or no timeout is configured.

**To Reproduce**

1. Start Etherpad
2. Trigger the API Endpoint `getStats` with a "short" timeout

**Expected behavior**

Etherpad should not crash if the HTTP Client closes the connection to the API Endpoint.

**Server (please complete the following information):**
 - Etherpad version: 1.8.13
 - OS: Debian Buster
 - Node.js version (`node --version`): v14.16.1
 - npm version (`npm --version`): 6.14.12

**Additional context**

```
Apr 12 23:22:22 etherpad1 systemd[1]: Started Etherpad-lite, the collaborative editor..
Apr 12 23:22:22 etherpad1 nodejs[24651]: #033[36m[2021-04-12 23:22:22.373] [DEBUG] console - #033[39mRunning on Node v14.16.1 (minimum required Node version: 10.17.0)
Apr 12 23:22:22 etherpad1 nodejs[24651]: #033[32m[2021-04-12 23:22:22.491] [INFO] console - #033[39mAll relative paths will be interpreted relative to the identified Etherpad base dir: /home/etherpad/etherpad-lite
Apr 12 23:22:22 etherpad1 nodejs[24651]: #033[36m[2021-04-12 23:22:22.492] [DEBUG] AbsolutePaths - #033[39mRelative path "settings.json" can be rewritten to "/home/etherpad/etherpad-lite/settings.json"
Apr 12 23:22:22 etherpad1 nodejs[24651]: #033[36m[2021-04-12 23:22:22.493] [DEBUG] AbsolutePaths - #033[39mRelative path "credentials.json" can be rewritten to "/home/etherpad/etherpad-lite/credentials.json"
[...]

PR #5106, opened 2021-07-04
await more queries / handle Inactivity timeouts

webzwo0i:
The relevant issue is https://github.com/ether/etherpad-lite/issues/5005

I still see crashes when configuring a READ LOCK. At one point we may need to shutdown Etherpad when there are lots of failing queries, however I think we need to be more robust.

The scenarios are probably one of either:
- read queries for large data or that scan the whole table can timeout
- all pool connection are used with long queries, when new queries arrived, so the latter timeout at one point

When loading pads when a READ LOCK is active, I see crashes due to:
- globalAuthor can't be set in `createAuthor`
- pad state can't be read (don't know which caller)

Theoretically any query can randomly fail (when running with READ or WRITE LOCK). In practice I assume that when all connections are used and long queries are running, at one point all new queries start to timeout and writes might be retried and at some point they start working again. Or they continue failing. The problem is, any of the database queries can start failing so we need to handle that at a lot of places.

### DIFF
--- a/src/node/db/SessionStore.js
+++ b/src/node/db/SessionStore.js
@@ -17,18 +17,12 @@ const logger = log4js.getLogger('SessionStore');
 module.exports = class SessionStore extends Store {
   get(sid, fn) {
     logger.debug(`GET ${sid}`);
-    DB.db.get(`sessionstorage:${sid}`, (err, sess) => {
-      if (sess) {
-        sess.cookie.expires = ('string' === typeof sess.cookie.expires
-          ? new Date(sess.cookie.expires) : sess.cookie.expires);
-        if (!sess.cookie.expires || new Date() < sess.cookie.expires) {
-          fn(null, sess);
-        } else {
-          this.destroy(sid, fn);
-        }
-      } else {
-        fn();
-      }
+    DB.db.get(`sessionstorage:${sid}`, (err, s) => {
+      if (err != null) return fn(err);
+      if (!s) return fn(null);
+      if (typeof s.cookie.expires === 'string') s.cookie.expires = new Date(s.cookie.expires);
+      if (s.cookie.expires && new Date() >= s.cookie.expires) return this.destroy(sid, fn);
+      fn(null, s);
     });
   }
 
--- a/src/node/db/SessionStore.js
+++ b/src/node/db/SessionStore.js
@@ -1,38 +1,40 @@
 'use strict';
-/*
- * Stores session data in the database
- * Source; https://github.com/edy-b/SciFlowWriter/blob/develop/available_plugins/ep_sciflowwriter/db/DirtyStore.js
- * This is not used for authors that are created via the API at current
- *
- * RPB: this module was not migrated to Promises, because it is only used via
- *      express-session, which can't actually use promises anyway.
- */
 
 const DB = require('./DB');
 const Store = require('express-session').Store;
 const log4js = require('log4js');
+const util = require('util');
 
 const logger = log4js.getLogger('SessionStore');
 
-module.exports = class SessionStore extends Store {
-  get(sid, fn) {
+class SessionStore extends Store {
+  async _get(sid) {
     logger.debug(`GET ${sid}`);
-    DB.db.get(`sessionstorage:${sid}`, (err, s) => {
-      if (err != null) return fn(err);
-      if (!s) return fn(null);
-      if (typeof s.cookie.expires === 'string') s.cookie.expires = new Date(s.cookie.expires);
-      if (s.cookie.expires && new Date() >= s.cookie.expires) return this.destroy(sid, fn);
-      fn(null, s);
-    });
+    const s = await DB.get(`sessionstorage:${sid}`);
+    if (!s) return;
+    if (typeof s.cookie.expires === 'string') s.cookie.expires = new Date(s.cookie.expires);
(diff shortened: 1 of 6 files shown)


webzwo0i:
Please dont squash.

What I did was essentially set the `queryTimeout` in ueberdb to a very low value (1-20ms), start Etherpad and then set a WRITE LOCK to the store table. Before this PR, Etherpad crashes with "Query inactivity timeout" when trying to load a pad. As said above, any query (read and write) can suddenly fail and we should handle that, however this will be a huge task. The PR is a start, as it covers some central functions.

src/node/db/SessionStore.js
```diff
@@ -27,7 +27,7 @@ module.exports = class SessionStore extends Store {
           this.destroy(sid, fn);
         }
       } else {
-        fn();
+        fn(err);
```
rhansen, line 30: Check for error early in the function:

```javascript
  get(sid, fn) {
    logger.debug(`GET ${sid}`);
    DB.db.get(`sessionstorage:${sid}`, (err, sess) => {
      if (err != null) return fn(err);
      if (!sess) return fn();
      sess.cookie.expires = typeof sess.cookie.expires === 'string'
        ? new Date(sess.cookie.expires) : sess.cookie.expires;
      if (sess.cookie.expires && new Date() >= sess.cookie.expires) return this.destroy(sid, fn);
      fn(null, sess);
    });
  }
```

src/node/hooks/express/padurlsanitize.js
```diff
@@ -11,7 +11,13 @@ exports.expressCreateServer = (hookName, args, cb) => {
...
     }
 
-    const sanitizedPadId = await padManager.sanitizePadId(padId);
+    let sanitizedPadId;
+    try {
+      sanitizedPadId = await padManager.sanitizePadId(padId);
+    } catch {
+      res.status(500).send('Query inactivity timeout');
+      return;
+    }
```
rhansen, line 20: Let Express handle the error by passing it to `next`:

```javascript
  args.app.param('pad', (req, res, next, padId) => {
    (async () => {
      // ensure the padname is valid and the url doesn't end with a /
      if (!padManager.isValidPadId(padId) || /\/$/.test(req.url)) {
        res.status(404).send('Such a padname is forbidden');
        return;
      }

      const sanitizedPadId = await padManager.sanitizePadId(padId);

      if (sanitizedPadId === padId) {
        // the pad id was fine, so just render it
        next();
      } else {
        // the pad id was sanitized, so we redirect to the sanitized version
        const realURL =
            encodeURIComponent(sanitizedPadId) + new URL(req.url, 'http://invalid.invalid').search;
        res.header('Location', realURL);
        res.status(302).send(`You should be redirected to <a href="${realURL}">${realURL}</a>`);
      }
    })().catch((err) => next(err || new Error(err)));
  });
```

src/node/db/SecurityManager.js
```diff
@@ -57,86 +57,96 @@ exports.checkAccess = async (padID, sessionCookie, token, userSettings) => {
...
-    authLogger.debug('access denied: must have an HTTP API session to access private group pads');
+    return grant;
+  } catch (err) {
+    if (err.message === 'Query inactivity timeout') {
+      authLogger.debug('access denied: a necessary database query has timed out.');
+    } else {
+      authLogger.debug('access denied: an error occurred', err.message);
+    }
     return DENY;
   }
```
rhansen, line 151: I don't think this is a good idea. If something throws, it would be better to let the error propagate, ultimately resulting in a 500 error to the user. Treating internal errors as unauthorized requests could have unintended side effects, such as triggering automatic firewall rules.

If you want to keep the debug logging I suggest doing this:

```suggestion
  } catch (err) {
    authLogger.debug(err.stack || err.toString());
    throw err;
  }
```

src/node/db/SessionStore.js
webzwo0i: will do

src/node/hooks/express/padurlsanitize.js
webzwo0i: Actually I tried this first and it didn't work, but yes your proposed change is definitely the idiomatic one. Will do.

src/node/db/SecurityManager.js
webzwo0i: Good point re semantic difference between 403-like responses and 500. I'll fix the callers then.

webzwo0i:
Thank you! Whenever I press "Make PR" I'm pretty sure that it's readable and idiomatic, but you always identify a lot of problematic areas.

src/node/db/SecurityManager.js
rhansen: PR #5171 should fix the biggest caller.

rhansen: And now the `/ro/:id` handler is gone: 58bd96ce8f1e054e10e769506660ccb103c22a44

rhansen:
I rebased onto latest `develop` and made some changes to address the feedback I provided (plus other stuff):
  * dropped the last commit
  * squashed the first two commits together and made the following changes to the resulting commit:
    * moved a Promise into the `Promise.all()`
    * added more missing `await`s
  * new commits:
    * simplify `Pad.copy()`
    * Promisify `SessionStore`
  * Errors are passed to Express rather than handled directly

PR #5107, opened 2021-07-05
error log uncaught exception

webzwo0i:
In situations with multiple error stack traces (like https://github.com/ether/etherpad-lite/issues/5005), when some of them are benign, it's hard to identify the bad stack trace unless using DEBUG loglevel.

With this PR we can clearly see all uncaught exceptions.

### DIFF
--- a/src/node/server.js
+++ b/src/node/server.js
@@ -113,7 +113,7 @@ exports.start = async () => {
     stats.gauge('memoryUsageHeap', () => process.memoryUsage().heapUsed);
 
     process.on('uncaughtException', (err) => {
-      logger.debug(`uncaught exception: ${err.stack || err}`);
+      logger.error(`uncaught exception: ${err.stack || err}`);
       exports.exit(err);
     });
     // As of v14, Node.js does not exit when there is an unhandled Promise rejection. Convert an


src/node/server.js
```diff
@@ -113,7 +113,7 @@ exports.start = async () => {
     stats.gauge('memoryUsageHeap', () => process.memoryUsage().heapUsed);
 
     process.on('uncaughtException', (err) => {
-      logger.debug(`uncaught exception: ${err.stack || err}`);
+      logger.error(`uncaught exception: ${err.stack || err}`);
```
rhansen, line 116: The full stack is already logged with `logger.error` at line 222.

webzwo0i: The purpose of this PR is to identify the stack trace that's causing the crash. Instances normally don't run on DEBUG logging and in case an error occurs that is not handled and that crashes the instance, we can only hope that no other error was thrown. If there are multiple errors, we don't know which one of the exceptions was fatal.

webzwo0i: Oh I see what you mean now. Makes sense. I'll close this, maybe I'll add a small note to line 222 "This error was causing the crash"

PR #5112, opened 2021-07-09
Bump jquery to 3.6.0

### LINKED ISSUE #4196: Handle deprecation warnings after jquery update
I prepared a [gist with integrated versions of jquery-migrate](https://gist.github.com/webzwo0i/f93fbb8f06a85686abcdeff08298d01d) that replace src/static/js/jquery.js

That's a jquery-plugin that should help people with transitioning to new jquery versions. There are two versions of it in the gist atm: one for the old jquery version before @JohnMcLear's update last month and one for the version after update.

The process would be to go back to the old version of last month, replace it with the version in the gist, let frontend tests run and look through all the deprecation warnings (there aren't many with the 1.12.4-version). Best would also be to temporarily activate some of the disabled frontend tests. Even though they might be wrong, they probably trigger additional code paths. It would be also nice to have the process include two browsers, e.g. join the same pad with firefox and chromium and write something because this also triggers other code paths compared to only running the frontend tests.

After warnings are silenced the whole process needs to be repeated with the updated 3.0.0 version.

I am not very skilled in clientside webdev so for me it probably would take a little bit too long, because I would need to learn a lot about browser internals etc to understand and verify the nature of some of the warnings, but I think it would be helpful to actually fix them, especially because 3.0.0 is already outdated and when we update to a recent jquery version, stuff could break.

The reason I stumbled across this is that they made document-ready handler async in 3.0 and I suspect this to have an effect on our frontend tests and page setup in general. Unfortunately though, jquery-migrate does not log warnings in the areas I suspected. (RE frontend tests I found two things that made them unreliable and one that prevented travis to get the logs and I am now testing with jquery 3.1.1 (cf https://blog.jquery.com/2016/07/07/jquery-3-1-0-released-no-more-silent-errors/) because with 3.0 errors from ready-handler are swallowed. Will finish on the weekend)

webzwo0i:
Draft, as I need to clean up the commits a little bit.

A lot of the stuff that was replaced (e.g. event handlers to use the more generic `.trigger` or `.on`/`.off` syntax) would still work with 3.6.0, but is deprecated since 3.0 so I thought it's a good idea to change it.

The most important change is regarding onclick handler and there are more places in Etherpad that need to be checked (they probably have no test coverage yet): The relevant jquery commit our code starts failing with is https://github.com/jquery/jquery/commit/669f720edc4f557dfef986db747c09ebfaa16ef5
After trying to debug this for some time I decided to follow the [advice not to use inline event handlers](https://github.com/jquery/jquery/wiki/Won%27t-Fix#inline-event-handlers).

I'll probably delete the farbtastic change - we should refactor this anyway and use an alternative

The changes to language.js and `helper.edit` are due to observations that these tests seem to be flaky. I'll run some more tests and might revert those commits, too.

Fixes #4196

### DIFF
--- a/src/static/js/admin/plugins.js
+++ b/src/static/js/admin/plugins.js
@@ -164,7 +164,7 @@ $(document).ready(() => {
     if (data.query.offset === 0) search.results = [];
     search.messages.hide('nothing-found');
     search.messages.hide('fetching');
-    $('#search-query').removeAttr('disabled');
+    $('#search-query').prop('disabled', false);
 
     console.log('got search results', data);
 
--- a/src/static/js/admin/plugins.js
+++ b/src/static/js/admin/plugins.js
@@ -112,15 +112,15 @@ $(document).ready(() => {
 
   const updateHandlers = () => {
     // Search
-    $('#search-query').unbind('keyup').keyup(() => {
+    $('#search-query').off('keyup').on('keyup', () => {
       search($('#search-query').val());
     });
 
     // Prevent form submit
-    $('#search-query').parent().bind('submit', () => false);
+    $('#search-query').parent().on('submit', () => false);
 
     // update & install
-    $('.do-install, .do-update').unbind('click').click(function (e) {
+    $('.do-install, .do-update').off('click').on('click', function (e) {
       const $row = $(e.target).closest('tr');
       const plugin = $row.data('plugin');
       if ($(this).hasClass('do-install')) {
@@ -134,7 +134,7 @@ $(document).ready(() => {
     });
 
     // uninstall
-    $('.do-uninstall').unbind('click').click((e) => {
+    $('.do-uninstall').off('click').on('click', (e) => {
       const $row = $(e.target).closest('tr');
       const pluginName = $row.data('plugin');
       socket.emit('uninstall', pluginName);
@@ -143,14 +143,14 @@ $(document).ready(() => {
     });
 
     // Sort
-    $('.sort.up').unbind('click').click(function () {
+    $('.sort.up').off('click').on('click', function () {
       search.sortBy = $(this).attr('data-label').toLowerCase();
       search.sortDir = false;
       search.offset = 0;
       search(search.searchTerm, search.results.length);
       search.results = [];
     });
-    $('.sort.down, .sort.none').unbind('click').click(function () {
+    $('.sort.down, .sort.none').off('click').on('click', function () {
       search.sortBy = $(this).attr('data-label').toLowerCase();
       search.sortDir = true;
       search.offset = 0;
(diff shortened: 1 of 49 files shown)


src/static/js/vendors/farbtastic.js
```diff
@@ -33,7 +33,7 @@ $._farbtastic = function (container, options) {
   fb.linkTo = function (callback) {
     // Unbind previous nodes
     if (typeof fb.callback == 'object') {
-      $(fb.callback).unbind('keyup', fb.updateValue);
+      $(fb.callback).off('keyup', fb.updateValue);
```
rhansen, line 36: I prefer to minimize diff to the upstream version so that newer versions can be merged in with fewer conflicts.

src/static/js/chat.js
```diff
@@ -38,15 +38,18 @@ exports.chat = (() => {
...
-    stickToScreen(fromInitialCall) {
+    stickToScreen: (fromInitialCall) => {
+      if ($('#options-stickychat').prop('checked')) {
+        $('#options-stickychat').prop('checked', false);
+      }
       if (pad.settings.hideChat) {
         return;
       }
-      this.show();
+      exports.chat.show();
```
rhansen, line 52: Why make these changes?

rhansen (commented):
Thank you for this!

The changes to `jquery.js` in the "jQuery: refactor mouseup event shorthand" commit should be reverted.

src/static/js/admin/plugins.js
```diff
@@ -164,7 +164,7 @@ $(document).ready(() => {
     if (data.query.offset === 0) search.results = [];
     search.messages.hide('nothing-found');
     search.messages.hide('fetching');
-    $('#search-query').removeAttr('disabled');
+    $('#search-query').prop('disabled', false);
```
rhansen, line 167: Properties and attributes are subtly different. Why avoid `.removeAttr()`? It doesn't appear to be deprecated.

src/static/js/chat.js
```diff
@@ -38,11 +38,14 @@ exports.chat = (() => {
...
       setTimeout(() => {
-        $('#chatinput').focus();
+        $('#chatinput').trigger('focus');
       }, 100);
     },
     // Make chat stick to right hand side of screen
     stickToScreen(fromInitialCall) {
+      if ($('#options-stickychat').prop('checked')) {
+        $('#options-stickychat').prop('checked', false);
+      }
```
rhansen, line 48: Why is this here?

rhansen (commented):
I rebased and cleaned up the commits a bit. Please take a look.

JohnMcLear:
Bump @webzwo0i -- any update on this?

webzwo0i:
thanks for the reminder, will look into this again

src/static/js/admin/plugins.js
SamTV12345: According to the docs: 

> It is almost always a mistake to use .removeAttr( "checked" ) on a DOM element. The only time it might be useful is if the DOM is later going to be serialized back to an HTML string. In all other cases, .prop( "checked", false ) should be used instead.

PR #5113, opened 2021-07-10
tests: fix unaccessible sendkeys

webzwo0i:
My idea in https://github.com/ether/etherpad-lite/pull/5103 didn't work, but I have some new insights: I can reproduce this reliably in Chromium when I disable cache and use a remote server, so there is some latency between requests.

sendkeys depends on jQuery to be loaded first. When cache is enabled, jQuery is downloaded via https://github.com/ether/etherpad-lite/blob/develop/src/tests/frontend/index.html#L18 and all later requests are cached, so when issuing the get requests with Promise.all, jQuery will immediately load and sendkeys will work properly.

However, when in a non-cached, higher latency scenario, jQuery will almost always be received after sendkeys was received, because the size of the jQuery file is larger (uncompressed 257kb vs 19kb), so sendkeys gets executed before jQuery was loaded and in essence, sendkeys is not available. Also other tests that require sendkeys will fail. It works again as soon as cache is enabled.

I verified the load sequence with some console.log messages that were placed at the top of the IIFE in sendkeys and jquery respectivily. You can see the load sequence of a failing test below. In cases when the test was successful, jquery preceeds sendkeys at the "<----helper.init" mark.
```
Navigated to http://localhost:8000/tests/frontend/?grep=large%20font
jquery.js:14 evaluating jquery, window is  <----src/tests/frontend/index.html
require-kernel.js:10 [Deprecation] Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/.
VM6899:380 evaluating sendkeys, window is <----helper.init
VM6900:14 evaluating jquery, window is <----helper.init
ace2_common.js?callb…efine&v=d6dcf1fc:27 evaluating jquery, window is  <----src/templates/pad.html
VM6915:380 evaluating sendkeys, window is <----src/templates/pad.html
VM6930:29 evaluating jquery, window is ace_inner <----getFrameJQuery
VM6932:15 evaluating jquery, window is ace_outer <----getFrameJQuery
VM6933:380 evaluating sendkeys, window is ace_outer <----getFrameJQuery
VM6935:380 evaluating sendkeys, window is ace_inner <----getFrameJQuery
```

The reason why it's only failing in Safari when running on SL is still unclear to me (probably Safari has another cache strategy).

### DIFF
--- a/src/tests/frontend/helper.js
+++ b/src/tests/frontend/helper.js
@@ -4,19 +4,6 @@ const helper = {};
 
 (() => {
   let $iframe;
-  const jsLibraries = {};
-
-  helper.init = async () => {
-    [
-      jsLibraries.jquery,
-      jsLibraries.sendkeys,
-    ] = await Promise.all([
-      $.get('../../static/js/vendors/jquery.js'),
-      $.get('lib/sendkeys.js'),
-    ]);
-    // make sure we don't override existing jquery
-    jsLibraries.jquery = `if (typeof $ === 'undefined') {\n${jsLibraries.jquery}\n}`;
-  };
 
   helper.randomString = (len) => {
     const chars = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
@@ -28,20 +15,29 @@ const helper = {};
     return randomstring;
   };
 
-  const getFrameJQuery = ($iframe) => {
-    /*
-      I tried over 9001 ways to inject javascript into iframes.
-      This is the only way I found that worked in IE 7+8+9, FF and Chrome
-    */
+  const getFrameJQuery = async ($iframe, includeJquery = false, includeSendkeys = false) => {
     const win = $iframe[0].contentWindow;
     const doc = win.document;
 
-    // IE 8+9 Hack to make eval appear
-    // https://stackoverflow.com/q/2720444
-    win.execScript && win.execScript('null');
+    const load = async (url) => {
+      const elem = doc.createElement('script');
+      elem.setAttribute('src', url);
+      const p = new Promise((resolve, reject) => {
+        const handler = (evt) => {
+          elem.removeEventListener('load', handler);
+          elem.removeEventListener('error', handler);
+          if (evt.type === 'error') return reject(new Error(`failed to load ${url}`));
+          resolve();
+        };
+        elem.addEventListener('load', handler);
+        elem.addEventListener('error', handler);
+      });
+      doc.head.appendChild(elem);
+      await p;
+    };
 
-    win.eval(jsLibraries.jquery);
-    win.eval(jsLibraries.sendkeys);
+    if (!win.$ && includeJquery) await load('../../static/js/vendors/jquery.js');
+    if (!win.bililiteRange && includeSendkeys) await load('../tests/frontend/lib/sendkeys.js');
 
     win.$.window = win;
     win.$.document = doc;
@@ -125,7 +121,7 @@ const helper = {};
     // set new iframe
     $('#iframe-container').append($iframe);
     await new Promise((resolve) => $iframe.one('load', resolve));
-    helper.padChrome$ = getFrameJQuery($('#iframe-container iframe'));
+    helper.padChrome$ = await getFrameJQuery($('#iframe-container iframe'), true, true);
     helper.padChrome$.padeditor =
         helper.padChrome$.window.require('ep_etherpad-lite/static/js/pad_editor').padeditor;
     if (opts.clearCookies) {
@@ -141,8 +137,10 @@ const helper = {};
       if (opts._retry++ >= 4) throw new Error('Pad never loaded');
       return await helper.aNewPad(opts);
     }
-    helper.padOuter$ = getFrameJQuery(helper.padChrome$('iframe[name="ace_outer"]'));
-    helper.padInner$ = getFrameJQuery(helper.padOuter$('iframe[name="ace_inner"]'));
+    helper.padOuter$ = await getFrameJQuery(
+        helper.padChrome$('iframe[name="ace_outer"]'), true, false);
+    helper.padInner$ = await getFrameJQuery(
+        helper.padOuter$('iframe[name="ace_inner"]'), true, true);
 
     // disable all animations, this makes tests faster and easier
     helper.padChrome$.fx.off = true;
@@ -184,8 +182,8 @@ const helper = {};
     $('#iframe-container iframe').remove();
     // set new iframe
     $('#iframe-container').append($iframe);
-    $iframe.one('load', () => {
-      helper.admin$ = getFrameJQuery($('#iframe-container iframe'));
+    $iframe.one('load', async () => {
+      helper.admin$ = await getFrameJQuery($('#iframe-container iframe'), true, false);
     });
   };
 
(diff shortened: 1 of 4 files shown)


rhansen:
Is `$.get()` evaluating the script it is fetching? It should only be fetching the raw bytes so that they can be evaluated later.

webzwo0i:
Yes, this is the problem. I can't bisect but I made a test script to check this and it appears jquery until 1.3 does not execute it, while 1.4+ does. The docs say:
https://api.jquery.com/jquery.ajax/
`
The type of data that you're expecting back from the server. If none is specified, jQuery will try to infer it based on the MIME type of the response (an XML MIME type will yield XML, in 1.4 JSON will yield a JavaScript object, in 1.4 script will execute the script, and anything else will be returned as a string). The available types (and the result passed as the first argument to your success callback) are:
`
and the changelog for 1.4 https://web.archive.org/web/20150119205819/http://jquery14.com/day-01/jquery-14
`
If the response to an Ajax request is returned with a JSON mime type (application/json), the dataType defaults to “json” (if no dataType is specified). Additionally, if the response to an Ajax request is returned with a JavaScript mime type (text/javascript or application/x-javascript) , the dataType defaults to “script” (if no dataType is specified), causing the script to automatically execute.
`

In my test scenario making the call `$.get(url, null, null, 'text')` stops the file from being automatically executed, but I need to test if this is also the case in Etherpad. Which solution would you prefer if this turns out to work? From what I see in helper.js we eval the file later in the ace_inner/ace_outer context, so I'd vote for the `$.get(url, null, null, 'text')` approach.

rhansen:
I think I would prefer a third approach: inject `<script>` elements into the document. That would preserve line numbers in stack traces which is nice (though it's unlikely we'll ever need to dig into their sources).

webzwo0i:
I tried this just now (adding script tags with src attribute in `getFrameJQuery`) but it seems to take some time until those scripts are evaluated and waiting with helper.waitForPromise in this method does not seem to work either. Will try tomorrow if it helps to inline the code directly into script tags without using `src` attributes.

In the meantime, this PR now includes working code. The frontend admin tests are failing but I've seen the error for some time now, flaky test imo (maybe to short timeout)

webzwo0i:
I literally spend hours, because I thought it would be worthwhile to completely drop jquery inclusion in outer/inner and container. First it seemed to work, but I got problems with events (for reasons totally unknown to me). It worked with https://github.com/jquery/jquery-simulate and I could fix some failed tests. But it turned out that 1. jquery-simulate doesn't seem to be actively maintained and 2. there are problems i.e. focus/blur and sendkeys' special commands (like `{enter}` etc) didn't work anymore. I will give this another try once jquery is bumped to 3.6 just to be sure that it wasn't a jquery bug, but in the mean time imo this PR can be merged.

One remaining problem is the sometimes failing admin frontend test. Still figuring out what's broken :(

src/tests/frontend/helper.js
```diff
@@ -185,7 +180,7 @@ const helper = {};
     // set new iframe
     $('#iframe-container').append($iframe);
     $iframe.one('load', () => {
-      helper.admin$ = getFrameJQuery($('#iframe-container iframe'));
+      helper.admin$ = getFrameJQuery($('#iframe-container iframe'), false, false);
```
rhansen, line 183: Missing `await`:

```suggestion
      helper.admin$ = await getFrameJQuery($('#iframe-container iframe'), false, false);
```

rhansen, line 183: Why shouldn't it load jQuery here?

src/tests/frontend/index.html
```diff
@@ -16,6 +16,7 @@
 
     <script src="../../static/js/require-kernel.js"></script>
     <script src="../../static/js/vendors/jquery.js"></script>
+    <script src="lib/sendkeys.js"></script>
```
rhansen, line 19: Isn't this handled by the `includeSendkeys` parameter?

src/tests/frontend/helper.js
```diff
@@ -28,20 +15,26 @@ const helper = {};
...
+      scriptElem.setAttribute('src', '../../static/js/vendors/jquery.js');
+      doc.head.appendChild(scriptElem);
+      await helper.waitForPromise(() => win.$, 5000);
+    }
+    if (!win.bililiteRange && includeSendkeys) {
+      const scriptElem = doc.createElement('script');
+      scriptElem.setAttribute('src', '../tests/frontend/lib/sendkeys.js');
+      doc.head.appendChild(scriptElem);
+      await helper.waitForPromise(() => win.$ && win.bililiteRange, 5000);
+    }
```
rhansen, line 37: My preference would be to create a Promise that resolves when the `load` event fires on the script element or rejects when the `error` event fires, though that is more complicated.

For example (untested):
```suggestion
    const load = async (url) => {
      const elt = doc.createElement('script');
      elt.setAttribute('src', url);
      const p = new Promise((resolve, reject) => {
        const handler = (ev) => {
          elt.removeEventListener('load', handler);
          elt.removeEventListener('error', handler);
          if (ev.type !== 'load') return reject(new Error(`failed to load ${url}`));
          resolve();
        };
        elt.addEventListener('load', handler);
        elt.addEventListener('error', handler);
      });
      doc.head.appendChild(elt);
      await p;
    };
    if (!win.$ && includeJquery) await load('../../static/js/vendors/jquery.js');
    // sendkeys.js depends on jQuery, so it cannot be loaded until jQuery has finished loading.
    if (!win.bililiteRange && includeSendkeys) await load('../tests/frontend/lib/sendkeys.js');
```

src/tests/frontend/index.html
webzwo0i: That's why I wanted to get rid of the 3 additional instances of jquery. We have separate jq instances on the test main site (index.html that starts mocha), in pad.html, ace2inner and enforced by (obsolete helper.init or) the logic in this PR that ensures it's also available as a global in ace_outer iframe.

It's kind of confusing which instance is used. When running the admin tests we don't need sendkeys because all text fields are manipulated with `val()` and jquery is already included via script elements (however, testing if this is the case wouldn't hurt now that getFrameJquery also checks if it's already defined).

I can remove this, as it won't make a difference (sendkeys isn't used in the context of the "main site" and prefetching the file won't have any performance impact). I think getting rid of three additional jquery instantions every time helper.newPad is called would make some small performance improvement, but it's something for another PR.

Btw: In the long term I'd like to get rid of jquery in core entirely, because it's very large compared to our code. But as this means getting rid of farbtastic and other plugins (most of them seem to be very outdated) and some plugins require jquery to be in core, this is some kind of bigger task.

src/tests/frontend/helper.js
webzwo0i: Will do this. Not sure if there is some delay when loading one of the two files, so that there is any chance `win.$` or `win.bililiteRange` is defined, but not fully loaded yet, but your suggestion will prevent this from happening, so it's the better solution.

src/tests/frontend/helper.js
webzwo0i: not needed, as it's already src'ed in a script tag. However, to be sure I changed it to "true, false" so that it will check if win.$ is present and download jquery if not.

src/tests/frontend/helper.js
webzwo0i: done

src/tests/frontend/helper.js
webzwo0i: thanks

src/tests/frontend/index.html
webzwo0i: btw this was needed. `includeSendkeys` is not considered in the index.html context (where helper.js is included)

webzwo0i:
all tests green - for the moment :) - so merging

PR #5115, opened 2021-07-13
Fix settings.useMonospaceFontGlobal

Lastpixl:
When settings.useMonospaceFontGlobal is set to `true`, it sets the default
font to 'monospace'. This font seems to have been removed in
a5164dad430d99f1fc0a7fd045c37ce56e70efa6.

This commit sets the default font to "RobotoMono" which is a valid
option.

Tested in a Docker environment, setting `PAD_OPTIONS_USE_MONOSPACE_FONT`
to `true`

> Screenshots/GIFs

### DIFF
--- a/src/static/js/pad.js
+++ b/src/static/js/pad.js
@@ -338,7 +338,7 @@ const handshake = () => {
 
       // If the Monospacefont value is set to true then change it to monospace.
       if (settings.useMonospaceFontGlobal === true) {
-        pad.changeViewOption('padFontFamily', 'monospace');
+        pad.changeViewOption('padFontFamily', 'RobotoMono');
       }
       // if the globalUserName value is set we need to tell the server and
       // the client about the new authorname


rhansen:
`monospace` is a [standard generic font](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#generic-name) that is supported in all browsers, so it should still work. If it's not working, then something unexpected is happening and I'd prefer to figure out the root cause.

`monospace` was removed from the [list of font options](https://github.com/ether/etherpad-lite/blob/336d48add73a7d22191d810cc68855ecd72c11a5/src/templates/pad.html#L153) in commit 7aa7e02e273c66936516f1ba87f435e4fe206c75. If you add it back does it work as expected?

rhansen:
Oh, I see what's happening. It's not setting the font directly, but changing the font setting. So your change is the proper fix. (We may want to restore generic fonts like `serif` and `monospace`, but that should be done in a separate PR.)

Issue #5037, opened 2021-05-19
Bold text loses its format when pasting


maa-sw:
**Describe the bug**
Whenever I paste a bolded text, the bold disappears.  

**To Reproduce**
1. Go to any live instance from the list here (https://github.com/ether/etherpad-lite/wiki/Sites-that-run-Etherpad-Lite)
2. Write any text and have it bolded
3. Copy and paste it, the pasted text will be bolded for a mere second and then the bold disappear. 

**Expected behavior**
I expect the pasted text to be bolded.

**Screenshots**
![Peek 2021-05-19 12-18](https://user-images.githubusercontent.com/69582629/118789390-66dbf700-b89d-11eb-905c-f52b04bfa85a.gif)

**Server (please complete the following information):**
 - Etherpad version: latest image from docker (etherpad/etherpad:latest)

**Desktop (please complete the following information):**
 - OS: [Ubuntu 20.04]
 - Browser [Chrome]
 - Version [90.0.4430.85 (Official Build) (64-bit)]

webzwo0i:
Can you test if applying https://github.com/ether/etherpad-lite/issues/4229#issuecomment-681032006 helps?

maa-sw:
> Can you test if applying [#4229 (comment)](https://github.com/ether/etherpad-lite/issues/4229#issuecomment-681032006) helps?

Sorry, I'm honestly not sure what I should be applying here.

jinbullsushil:
Hi, 
I am facing the same issue. 
Did you solved the "bold text loses its format when pasting"?

webzwo0i:
Afaik the patch never got applied yet, mainly because I didn't understand why it works when removing the CSS for `b` tags.

joassouza:
Just adding...The same bug is going to happen with any selection attribute.
For example, using the plugin `ep_font_color`. Note that after the pasting, part of the text loses the red color

https://user-images.githubusercontent.com/4945372/125976964-223bfc6f-1993-4e95-a68a-efe23eb2217f.mov

webzwo0i:
@joassouza Cool, thx for the info! Was it tested with Chrome/Chromium or Firefox? If Firefox works, I assume this is the bug that's described in https://github.com/ether/etherpad-lite/issues/4668

joassouza:
Exactly @webzwo0i! On Firefox works flawlessly!

Issue #4668, opened 2021-01-22
Clipboard API for copy&paste


webzwo0i:
I noticed browser do things differently when copy&pasting, which results in small unexpected behaviour e.g. in ep_font_size. I'm opening this on core because we can do something about it here. Test coverage would be a requirement probably.

When using ctrl+c on selected text, browsers put different markup into the clipboard. e.g. if you have 4 letters of a 60px font and copy 2 from the middle of it
Firefox: `<meta http-equiv="content-type" content="text/html; charset=utf-8"><span class="author-a-z78z0d5qjz81zz81zz77zqz70zvz86zz122zer font-size:60">FO</span>`

Chrome: `<meta http-equiv="content-type" content="text/html; charset=utf-8"><span style="color: rgb(72, 83, 101); font-family: Quicksand, Cantarell, &quot;Open Sans&quot;, &quot;Helvetica Neue&quot;, sans-serif; font-size: 60px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(243, 226, 222); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">FO</span>`

As you can see, Firefox copies the correct span attributes, so on paste everything is perfect while Chrome tries to derive the relevant styling, while ignoring span attributes. This results in loss of authorship info and styling. Chrome copies span attributes as soon as the whole span is selected (or maybe as soon as it's crossing span boundaries), ie when you have a whole line with font size 60 and double click the line to select every text on it, Chrome does the right thing.

Authorship might be a bad example, as we probably should attribute the whole text to the author that pasted it, instead of keeping the original authorship. Font-size might serve as a better example: We could iterate over pasted styles to find out which are relevant to us or we could [fill the clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) with the correct data. Imo the latter option is better.

I'm not sure if this is a Chrome-only issue or if there are more serious bugs than the edge case above.

UchihaSasuke2131:
![Screenshot_2](https://user-images.githubusercontent.com/84627856/119754539-59c89480-bea9-11eb-91b1-dcd836fb5c4e.png)

joassouza:
[...]

PR #5117, opened 2021-07-16
This fixes an issue with lost bold formatting when copying bold text.

webzwo0i:
Fixes https://github.com/ether/etherpad-lite/issues/5037

I'm not sure this is the right way, maybe someone else knows it. I observed the following behavior: When I make a line bold and put it into clipboard, apply the patch, restart/reload, and paste it, the text still looses formatting. So it could be that something in our copy&paste logic is wrong (maybe the serialization that's done in contentcollector or https://github.com/ether/etherpad-lite/issues/4668)


webzwo0i:
Closing this as more research showed that the CSS is probably not the main cause

PR #5149, opened 2021-08-11
Accessibility fix for JAWS screen readers

vwbij:
ace.js: removed the role 'application' from innerDocument.body. JAWS do not read any text from the edit lines if this role is set.

domline.createDomLine: to give JAWS the ability to read the lines correctly, it is required to set the attribute 'aria-live' to 'assertive'.

### DIFF
--- a/src/static/js/domline.js
+++ b/src/static/js/domline.js
@@ -61,6 +61,8 @@ domline.createDomLine = (nonEmpty, doesWrap, optBrowser, optDocument) => {
 
   if (document) {
     result.node = document.createElement('div');
+    // JAWS and NVDA screen reader compatibility. Only needed if in a real browser.
+    result.node.setAttribute('aria-live', 'assertive');
   } else {
     result.node = {
       innerHTML: '',
--- a/src/static/js/ace.js
+++ b/src/static/js/ace.js
@@ -284,7 +284,6 @@ const Ace2Editor = function () {
     // <body> tag
     innerDocument.body.id = 'innerdocbody';
     innerDocument.body.classList.add('innerdocbody');
-    innerDocument.body.setAttribute('role', 'application');
     innerDocument.body.setAttribute('spellcheck', 'false');
     innerDocument.body.appendChild(innerDocument.createTextNode('\u00A0')); // &nbsp;
 


src/static/js/domline.js
```diff
@@ -61,6 +61,9 @@ domline.createDomLine = (nonEmpty, doesWrap, optBrowser, optDocument) => {
 
   if (document) {
     result.node = document.createElement('div');
+    // Set the ARIA role and label for JAWS compatibility. Do it here, because it is only needed and
```
rhansen, line 64: This comment says "Set the ARIA role" but I don't see `role` set anywhere. Is that intentional?

src/static/js/domline.js
```diff
@@ -61,6 +61,9 @@ domline.createDomLine = (nonEmpty, doesWrap, optBrowser, optDocument) => {
...
+    // possible if we have a document and are working in a real browser context.
+    result.node.setAttribute('aria-live', 'assertive');
```
rhansen, line 66: Just to confirm: Is `assertive` the right value here? I'm not an ARIA expert, which is why I'm a bit concerned by [MDN's strong language](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#live_regions):

> `aria-live="assertive"` should only be used for time-sensitive/critical notifications that absolutely require the user's immediate attention.

rhansen (commented):
Thank you!

I pushed some minor fixes to your branch. If they look OK to you I'll squash them into your commit when I merge.

src/static/js/domline.js
vwbij: It is the right value we choose, we tested it and found it as a required value to make JAWS and NVDA work correctly. Before we checked "role=textbox" and "ariaLabel=Line". It was working also, but testing it in the school environment turned out that "assertive" is the best choice.
I am also not a expert in ARIA, it was just checking what makes the screenreader work correctly, and this depends on how the screenreader deals with that values.

src/static/js/domline.js
vwbij: The comment points to the attribute 'aria-live'='assertive', you are right, it has nothing to do with the role. So the comment is not well-formed and may be changed to better understanding.

src/static/js/domline.js
vwbij: The changes you have committed above are OK, thank you! So it is good to let them flow into the merge as you are ready to do it.

vwbij:
Thank you, have a nice evening!

PR #5150, opened 2021-08-13
Work in progress: webpack

### LINKED ISSUE #4820: Dropping require-kernel and yajsml
This is intended as a collection of useful information/discussion re migration to a better bundler like webpack or snowpack.

Old PoC for removing require-kernel: https://github.com/ether/etherpad-lite/pull/3603
Some useful todo that should be covered: https://github.com/ether/etherpad-lite/issues/3505

Maintained plugin for bundling plugins: https://github.com/storytouch/ep_webpack
imo it's best if we'd integrate bundling directly into core and cover plugins. Rebundling should only happen on the first start of core, after updating core or changing files in ./src, or after installing/deinstalling plugins. I'm not sure if we should support `minify:false` any longer. I think proper support for source maps would be enough.

We need to decide if we want a separate build step (it could be run on start/restart from within etherpad) or if we bundle files on demand. It needs to be investigated if the bundler in non-cli mode is as comprehensive as it's cli counterpart.

imo we should also think about moving to ESM. I imagine this to be unrelated to bundlin, ie move from CommonJS to ESM first, then think about bundling (hopefully the bundler comes with a transpiler to optionally support browser that don't support ESM yet - however only IE seems to be affected). The goal should be to enable plugin devs to easily migrate to ESM, maybe an automated script, or ESM wrapper and it would be cool if /admin/plugin would support something like "show/install the most recent compatible version of this plugin". Best would be, if core would support legacy plugins for a longer period. This needs investigation. 
Because we probably need to rethink what we want to export to make this happen, there is a relation to https://github.com/ether/etherpad-lite/issues/4714

Also related:
- removing globals from client-side code
- refactor/ease the page load process (onload/documentReady)
- less inline javascript (#4401)

rhansen:
Still far from complete. Progress:

* [x] bundle pad code
* [x] bundle plugin code
* [x] file name based on content hash (for cache friendliness)
* [x] source map
* [ ] bundle frontend tests
* [ ] watch for changes when not in production mode (to facilitate development)
* [ ] figure out migration plan for plugins that do dynamic `require()` (are there any?)
* [ ] bundle timeslider code
* [ ] bundle inner frame code
* [ ] put vendor code in a separate chunk
* [ ] put shared code (pad, timeslider, inner) in separate chunks
* [ ] split cleanups off into their own commits
* [ ] figure out better way to handle `adoptPluginsFromAncestorsOf()` (can it be removed?)

For future PRs:
* [ ] babel
* [ ] bundle CSS

Fixes #4820

cc @webzwo0i

### DIFF
--- a/src/node/hooks/express/specialpages.js
+++ b/src/node/hooks/express/specialpages.js
@@ -110,6 +110,11 @@ exports.expressCreateServer = async (hookName, args) => {
       path: path.join(settings.root, 'var/js'),
       filename: '[name]-[contenthash].js',
     },
+    resolve: {
+      alias: {
+        'ep_etherpad-lite': path.join(settings.root, 'src'),
+      },
+    },
   });
   const stats = await util.promisify(compiler.run.bind(compiler))();
   console.log(`webpack stats:\n${stats}`);
--- a/src/node/hooks/express/specialpages.js
+++ b/src/node/hooks/express/specialpages.js
@@ -6,9 +6,11 @@ const fs = require('fs');
 const fsp = fs.promises;
 const toolbar = require('../../utils/toolbar');
 const hooks = require('../../../static/js/pluginfw/hooks');
+const plugins = require('../../../static/js/pluginfw/plugin_defs');
 const settings = require('../../utils/Settings');
 const util = require('util');
 const webaccess = require('./webaccess');
+const webpack = require('webpack');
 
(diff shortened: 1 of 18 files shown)


SamTV12345:
Thanks for the initial pitch. This is fixed with esbuild and it's blazingly fast :)

PR #5173, opened 2021-09-14
ux: dont take focus when a pad is embedded in another page

### LINKED ISSUE #4996: Facility to prevent focus theft when embedded (in Element for example)
* * *

name: Prevent focus theft
about: 
title: ''
labels: enhancement
assignees: 

* * *

Please see this issue reported for the Element chat client (for the Matrix chat protocol), which has a facility to embed Etherpad as a widget.

"reports that etherpad widget can steal keyboard focus entirely from MessageComposer #5130"
https://github.com/vector-im/element-web/issues/5130

TL;DR Etherpad takes a significant time to load when entering a room, at which point it steal keyboard input focus (often part-way through typing something), and chaos ensues.

Would be good to have a way to prevent Etherpad from stealing focus.

A change to Element might help, however this is not straightforward according to the dev updates to the issue above.

JohnMcLear:
When a pad is embedded in a page it's improbable you will want to draw focus as such if Etherpad is not the parent then do not pull caret or focus away from the parent page to the editor.

Fixes #4996

### DIFF
--- a/src/static/js/pad.js
+++ b/src/static/js/pad.js
@@ -49,6 +49,7 @@ const socketio = require('./socketio');
 const hooks = require('./pluginfw/hooks');
 
 let receivedClientVars = false;
+let focusOnEditor = true;
 
 // This array represents all GET-parameters which can be used to change a setting.
 //   name:     the parameter-name, eg  `?noColors=true`  =>  `noColors`
@@ -146,13 +147,17 @@ const getParameters = [
       Cookies.set('language', val);
     },
   },
+  {
+    name: 'focusOnEditor',
+    callback: (val) => focusOnEditor = val !== 'false',
+  },
 ];
 
 const getParams = () => {
   // Tries server enforced options first..
   for (const setting of getParameters) {
     const value = clientVars.padOptions[setting.name];
-    if (value.toString() === setting.checkVal) {
+    if (value != null && value.toString() === setting.checkVal) {
       setting.callback(value);
     }
   }
@@ -463,9 +468,7 @@ const pad = {
 
     const postAceInit = () => {
       padeditbar.init();
-      setTimeout(() => {
-        padeditor.ace.focus();
-      }, 0);
+      if (focusOnEditor) setTimeout(() => padeditor.ace.focus(), 0);
       // if we have a cookie for always showing chat then show it
       if (padcookie.getPref('chatAlwaysVisible')) {
         chat.stickToScreen(true); // stick it to the screen
--- a/src/tests/frontend/specs/embed_value.js
+++ b/src/tests/frontend/specs/embed_value.js
@@ -35,6 +35,7 @@ describe('embed links', function () {
       showChat: 'true',
       showLineNumbers: 'true',
       useMonospaceFont: 'false',
+      focusOnEditor: 'false',
     };
 
     // check the url
--- a/src/tests/frontend/specs/focus.js
+++ b/src/tests/frontend/specs/focus.js
@@ -1,0 +1,27 @@
+'use strict';
+
+describe('caret focus', function () {
+  for (const params of [{}, {focusOnEditor: 'true'}, {focusOnEditor: 'false'}]) {
+    const wantFocus = params.focusOnEditor !== 'false';
+
+    describe(`focusOnEditor=${params.focusOnEditor}`, function () {
+      before(async function () {
+        await helper.aNewPad({params});
+      });
+
+      it(`does${wantFocus ? '' : ' not'} focus`, async function () {
+        if (wantFocus) {
+          expect(helper.padChrome$.document.activeElement)
+              .to.be(helper.padChrome$.document.querySelector('iframe[name="ace_outer"]'));
+          expect(helper.padOuter$.document.activeElement)
+              .to.be(helper.padOuter$.document.querySelector('iframe[name="ace_inner"]'));
+          expect(helper.padInner$.document.activeElement)
+              .to.be(helper.padInner$.document.body);
+        } else {
+          expect(helper.padChrome$.document.activeElement)
+              .to.not.be(helper.padChrome$.document.querySelector('iframe[name="ace_outer"]'));
+        }
+      });
+    });
+  }
+});
--- a/doc/api/embed_parameters.md
+++ b/doc/api/embed_parameters.md
@@ -73,3 +73,11 @@ Default: 0
 Focuses pad at specific line number and places caret at beginning of this line
 Special note: Is not a URL parameter but instead of a Hash value
 
+## focusOnEditor
+ * Boolean
+
+Default: true
+
+If true, focus on the editor upon initial page load. Set this to false if the
+pad is embedded in an iframe adjacent to other editable content that should
+retain focus.
--- a/src/static/js/pad_editbar.js
+++ b/src/static/js/pad_editbar.js
@@ -262,7 +262,7 @@ exports.padeditbar = new class {
   }
   setEmbedLinks() {
     const padUrl = window.location.href.split('?')[0];
-    const params = '?showControls=true&showChat=true&showLineNumbers=true&useMonospaceFont=false';
+    const params = '?showControls=true&showChat=true&showLineNumbers=true&useMonospaceFont=false&focusOnEditor=false';
     const props = 'width="100%" height="600" frameborder="0"';
 
     if ($('#readonlyinput').is(':checked')) {


src/static/js/pad.js
```diff
@@ -656,6 +664,7 @@ const pad = {
     pad.determineChatVisibility(isConnected && !isInitialConnect);
     pad.determineChatAndUsersVisibility(isConnected && !isInitialConnect);
     pad.determineAuthorshipColorsVisibility();
+    pad.determineFocus();
```
rhansen, line 667: Seems unnecessary to have a separate function for something so trivial.

src/static/js/pad.js
```diff
@@ -669,6 +678,12 @@ const pad = {
       $('#options-stickychat').prop('checked', false); // set the checkbox for off
     }
   },
+  determineFocus: (asNowConnectedFeedback) => {
```
rhansen, line 681:
```suggestion
  determineFocus: () => {
```

src/static/js/pad.js
```diff
@@ -669,6 +678,12 @@ const pad = {
...
+    const focus = padcookie.getPref('focusOnEditor');
```
rhansen, line 682: I don't think this setting should be persisted in a cookie.

src/static/js/pad.js
```diff
@@ -669,6 +678,12 @@ const pad = {
...
+    if (focus) {
+      padeditor.ace.focus();
```
rhansen, line 684: Is this necessary? `focus()` is already called after pad init.

src/static/js/pad.js
```diff
@@ -146,6 +146,14 @@ const getParameters = [
       Cookies.set('language', val);
     },
   },
+  {
+    name: 'focusOnEditor',
+    checkVal: 'false',
+    callback: (val) => {
+console.log("calling back ", val);
```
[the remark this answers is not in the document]
JohnMcLear: Heh, good spot :P

src/static/js/pad.js
JohnMcLear: is pad init called upon reconnect?

src/static/js/pad.js
rhansen: I mean there's no need to create a new `determineFocus` method:

```suggestion
    if (padcookie.getPref('focusOnEditor')) padeditor.ace.focus();
```

src/static/js/pad.js
rhansen: No, but is there a reason `focus()` needs to be called on reconnect? Seems to me that it would be better to not call it.

src/static/js/pad.js
```diff
@@ -464,7 +471,7 @@ const pad = {
     const postAceInit = () => {
       padeditbar.init();
       setTimeout(() => {
-        padeditor.ace.focus();
+        if (!settings.focusOnEditor) padeditor.ace.focus();
```
rhansen, line 474: Isn't this backwards?

```suggestion
        if (settings.focusOnEditor) padeditor.ace.focus();
```

src/static/js/pad.js
```diff
@@ -146,6 +146,13 @@ const getParameters = [
       Cookies.set('language', val);
     },
   },
+  {
+    name: 'focusOnEditor',
+    checkVal: 'false',
+    callback: (val) => {
+      settings.focusOnEditor = val;
+    },
```
rhansen, line 154: This is broken: it sets `settings.focusOnEditor` to the string `'false'` which is truthy. Here's a fix:

```suggestion
    callback: (val) => settings.focusOnEditor = val !== 'false',
```

src/static/js/pad.js
rhansen, line 474:
```suggestion
        if (settings.focusOnEditor) padeditor.ace.focus();
```

src/static/js/pad.js
JohnMcLear: I'd of thought so but I'm not passionate about it.  My goal of using the additional reconnect was to keep the logic the same as other reconnect logic but I'm easy either way :)

src/static/js/pad.js
JohnMcLear: I agree.

src/node/utils/Settings.js
```diff
@@ -133,6 +133,7 @@ exports.padOptions = {
   noColors: false,
   showControls: true,
   showChat: true,
+  focusOnEditor: true,
```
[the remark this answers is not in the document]
JohnMcLear: I think so yeah, I think the whole settings thing was more for persistence but as we have dropped cookie persistence this can be removed too.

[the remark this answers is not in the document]
rhansen: Pushed a change that removes the setting.

src/static/js/pad.js
rhansen: Pushed a change that removes the cookie.

src/static/js/pad.js
rhansen: I pushed a change so that it doesn't re-steal focus on reconnect.

rhansen:
I pushed a commit to address the remaining comments. Please take a look.

rhansen:
Can you add tests?

JohnMcLear:
Tests in.  I tried various approaches and this seemed the most reliable.

JohnMcLear:
Ugh, it appears Safari doesn't like this approach, that's going to be a world of pain to debug..

src/tests/frontend/specs/focus.js
```diff
@@ -0,0 +1,36 @@
...
+        params: {
+          focusOnEditor: 'true'
+        }
+      });
+    });
+
+    it('does focus when focusOnEditor is set to true', async function () {
+      const activeElement = helper.padChrome$.document.activeElement;
+      console.log(activeElement)
+      await helper.waitForPromise(() => $(activeElement).attr('name') === 'ace_outer');
```
rhansen, line 32: It should already be focused by the time `helper.aNewPad()` resolves, so there's no need for `waitForPromise()` here. Also, shouldn't each frame's `activeElement` be inspected?

```suggestion
      expect(helper.padChrome$.document.activeElement)
          .to.be(helper.padChrome$.document.querySelector('iframe[name="ace_outer"]'));
      expect(helper.padOuter$.document.activeElement)
          .to.be(helper.padOuter$.document.querySelector('iframe[name="ace_inner"]'));
      expect(helper.padInner$.document.activeElement)
          .to.be(helper.padInner$.document.body);
```

src/tests/frontend/specs/embed_value.js
```diff
@@ -35,6 +35,7 @@ describe('embed links', function () {
       showChat: 'true',
       showLineNumbers: 'true',
       useMonospaceFont: 'false',
+      focusOnEditor: 'false'
```
rhansen, line 38:
```suggestion
      focusOnEditor: 'false',
```

src/tests/frontend/specs/focus.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+describe('caret focus', function () {
+  describe('doesnt focus on editor using focusOnEditor URL Param', function () {
+    before(async function () {
+      await helper.aNewPad({
+        params: {
+          focusOnEditor: 'false'
```
rhansen, line 8:
```suggestion
          focusOnEditor: 'false',
```

src/tests/frontend/specs/focus.js
```diff
@@ -0,0 +1,36 @@
...
+        }
```
rhansen, line 9:
```suggestion
        },
```

src/tests/frontend/specs/focus.js
```diff
@@ -0,0 +1,36 @@
...
+      await helper.waitForPromise(() => $(activeElement).hasClass('readwrite'));
+    });
+
+  });
+
+  describe('does focus on editor using focusOnEditor URL Param', function () {
+    before(async function () {
+      await helper.aNewPad({
+        params: {
+          focusOnEditor: 'true'
```
rhansen, line 24:
```suggestion
          focusOnEditor: 'true',
```

src/tests/frontend/specs/focus.js
```diff
@@ -0,0 +1,36 @@
...
+        }
```
rhansen, line 25:
```suggestion
        },
```

src/tests/frontend/specs/focus.js
```diff
@@ -0,0 +1,36 @@
...
+
+describe('caret focus', function () {
+  describe('doesnt focus on editor using focusOnEditor URL Param', function () {
+    before(async function () {
+      await helper.aNewPad({
+        params: {
+          focusOnEditor: 'false'
+        }
+      });
+    });
```
rhansen, line 11: Wait for postAceInit so we know there has been an opportunity to focus:

```suggestion
      const {classList} = helper.padChrome$.document.getElementById('editorcontainer');
      await helper.waitForPromise(() => classList.contains('initialized'));
    });
```

src/tests/frontend/specs/focus.js
```diff
@@ -0,0 +1,36 @@
...
+
+    it('does not focus when focusOnEditor is set to false', async function () {
+      const activeElement = helper.padChrome$.document.activeElement;
+      await helper.waitForPromise(() => $(activeElement).hasClass('readwrite'));
```
rhansen, line 15: Waiting for the `readwrite` class might not be long enough. See my comment for the `before()`.

Also, this doesn't actually test that it does not focus because the `activeElement` variable is set right away but might be updated later. Fix (when combined with my above suggested change):

```suggestion
      expect(helper.padChrome$.document.activeElement)
          .to.not.be(helper.padChrome$.document.querySelector('iframe[name="ace_outer"]'));
```

src/tests/frontend/specs/focus.js
rhansen: #5195 will make this change unnecessary.

rhansen: #5195 is merged.

flotho:
I've tested successfully the focus issue.
Yet I still have this messages in the console :
```
Invalid 'X-Frame-Options' header encountered when loading 'https://XXX.YYY.tld/': 'ALLOW-FROM https://YYY.tld' is not a recognized directive. The header will be ignored
``` 
Yet I didn't set up such an ALLOW FROM directive,  I det up frame ancestors directive.

Do you think it could be related ?

Regards

flotho:
> Nah not related I don't think? Could be X origin if it's trying to query parent frame but idk. Did you have this console msg before??

Not sure, I didn't check.

I'll make additionnal searches later on.

Thanks

flotho:
More info, it works on Chrome but not on FF 88

rhansen:
> More info, it works on Chrome but not on FF 88

I just tested Firefox 88 (on Windows 10) and I couldn't reproduce. Would you please check out this PR, fire up Etherpad, visit http://localhost:9001/tests/frontend/?grep=%3Ccore%3E%20focus%5C.js and report whether the last test passes?

flotho:
Hi,
I think I have a problem with my installation, no tests are provided :
image


image

t3chguy:
It would be great if this could be revived.

Issue #5172, opened 2021-09-14
Text loses style with fn+delete


vitmf:
**Describe the bug**
Pressing `fn+delete` (or `delete`on PCs) to join two lines removes classes of the bottom line text, with the exception of `author-...`

**To Reproduce**
1. Install a plugin which styles text by adding a class to `span`, like `ep_font_color`.
2. Enter some text and style it with the plugin.
3. Go to the end of the line above and press `fn+delete`.
4. Text loses style.

**Expected behavior**
Text should maintain style.

**Screenshots**
![delete](https://user-images.githubusercontent.com/640041/133262082-5e368c8b-fe7f-4330-88b6-2b83f56fd8cb.gif)

**Server (please complete the following information):**
 - Etherpad version: 1.8.14
 - OS: Ubuntu 20.04
 - Node.js version (`node --version`): 16.9.1
 - npm version (`npm --version`): 7.21.1

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Browser: Chrome
 - Version: 92

**Additional context**
Text only loses style because it loses the class. As the styling options in pure etherpad do not work that way, we have to install a plugin to test it.
This problem does not happen in Firefox.

PR #5175, opened 2021-09-15
ace2_inner.js: Do fn+delete customly

vitmf:
'fn+delete' (or 'delete' in PCs) behaviour is inconsistent across browsers. In Chrome,
it might remove the class of the text in the line below when using
fn+delete to join lines. In this PR, we solve this issue by
doing 'fn+delete' customly, as we do 'delete' (or 'backspace' in PCs) customly.
Fixes: #5172.

### DIFF
--- a/src/node/utils/Settings.js
+++ b/src/node/utils/Settings.js
@@ -150,6 +150,7 @@ exports.padShortcutEnabled = {
   altF9: true,
   altC: true,
   delete: true,
+  fnDelete: true,
   cmdShift2: true,
   return: true,
   esc: true,
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -2539,6 +2539,67 @@ function Ace2Inner(editorInfo, cssManagers) {
     }
   };
 
+  const doFnDeleteKey = (optEvt) => {
+    const evt = optEvt || {};
+    if (rep.selStart) {
+      if (isCaret()) {
+        const theLine = caretLine();
+        const theCol = caretColumn();
+        const lineEntry = rep.lines.atIndex(theLine);
+        const lineText = lineEntry.text;
+        const lastColOfLine = lineText.length;
+        const lastLineNum = rep.lines.length() - 1;
+
+        if (theCol === lastColOfLine) {
+          // caret is at the end of line
+          const nextLineEntry = (theLine < lastLineNum && rep.lines.atIndex(theLine + 1));
+          const nextLineMarker = nextLineEntry && nextLineEntry.lineMarker;
+          if (nextLineEntry) {
+            // if it is not the last line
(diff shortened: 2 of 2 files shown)


rhansen:
I think it would be better to remove all special backspace/delete handling code.

@JohnMcLear Do you know why the special backspace handling code exists? At first glance it looks like it could be removed entirely, which would fix this bug (I think) without any downsides.

JohnMcLear:
@rhansen oof I can't remember exactly why, it was probably old IE6 support code or something, I'd say that as long as we have test coverage in (which might be tricky as I bet sendkeys doesn't support this) and tests pass with it removed then we should provide correct UX.

Obvious follow up request:  @vitmf Can you please include some test coverage for this UX?

vitmf:
Is it possible to use a plugin like ep_font_color in the test? I need to set a class in the span surrounding the text.

rhansen:
Can you reproduce the problem by making the text bold or putting it in a bulleted list? Those don't need plugins.

vitmf:
I can't reproduce with these, as they don't set a class in <span>. It seems to only happen in this case.

rhansen:
Interesting. I'd be interested to understand why it only affects `<span>`. Is it a browser thing? Or is there some code somewhere in Etherpad that we can adjust instead?

vitmf:
It happens in Chrome/Chromium/Edge, but not in Firefox. Also it does not happen if there is more than one `<span>` in the same line. Chrome also seems to have problems when dragging elements, but I did not look too closely into it.

rhansen:
This is definitely a browser quirk that we need to work around; see https://jsfiddle.net/fgnubt1a/.

Chrome only seems to mangle the `span` if there is no other visible content before or after the `span`. For example, these all break if the caret is placed at the end of the previous line and the delete key is pressed:

* `<div><span class="color:red">...</span></div>`
* `<div><span class="color:red">...</span><!-- --></div>`
* `<div><span class="color:red>...</span><span class="emptySpan"></span></div>`
* `<div><span class="color:red>...</span><span style="display: none;">x</span></div>`

But the following are OK (except the additional content causes other problems):

* `<div>&zwnj;<span class="color:red">...</span></div>`
* `<div><span class="color:red">...</span>&zwnj;</div>`
* `<div><span class="color:red">...</span><span>&zwnj;</span></div>`

I don't think we would be able to work around the bad behavior by injecting content before or after the `span`. That would likely cause too many other problems.

So I think we're stuck with special backspace and delete handlers. This PR, along with the existing backspace handling code, means we would have a lot of code dedicated to this problem. Also, there's a case that is not yet handled: create a selection that starts and ends on different lines, then replace the selected text by typing or pasting. Can the logic be simplified and made more general so that it can be reused for backspace, delete, and select-and-replace?

PR #5180, opened 2021-09-16
Docker: Expose socketIo.maxHttpBufferSize and dumpOnUncleanExit settings as environment variables

JustAnotherArchivist:
Unlike all other (simple) settings, these two were previously not available as environment variables. This change makes it possible to modify them without having to rebuild the image (cf. 85adaa44).


rhansen:
Thanks!

Issue #5073, opened 2021-06-15
Text export includes attribs


webzwo0i:
See https://github.com/ether/etherpad-lite/issues/5071 for details, citing:
> `getText(padID, [rev])`
> works as expected by taking 'rev' into account.
> But it has a different return schema when 'rev' is specified. It wraps the "text" object in another "text" object, and also adds an "attribs" blob.

introduced by commit e7dc0766fdcc3c8477e97c83393fcc2683fab72e

JohnMcLear:
If we already return ``foo.text.text`` we'd also need to return ``foo.text`` when rev is specified so not ideal to duplicate the data, if we dropped ``foo.text.text``it would break existing installs...  I think we should..
1. Also return ``foo.text`` when rev is specified
2. Docs to say to migrate from ``foo.text.text``(when rev is specified) to ``foo.text``
3. Make a new API version that supports ``foo.text`` for when rev is specified & drop the attribs blob

JohnMcLear:
For posterity..  

No rev
127.0.0.1:9001/api/1/getText?apikey=e0c49272bd0a22fce486dbb7d255b76d5c75ce9b9a59682690521cec8a92c88d&padID=foo is my test URI


```
{"code":0,"message":"ok","data":{"text":"hello world\n\nthis world\n\nis real.\n\n\n"}}
```

With rev

http://127.0.0.1:9001/api/1/getText?apikey=e0c49272bd0a22fce486dbb7d255b76d5c75ce9b9a59682690521cec8a92c88d&padID=foo&rev=2

```
{"code":0,"message":"ok","data":{"text":{"text":"hel\n","attribs":"*0+3|1+1"}}}
```

https://github.com/ether/etherpad-lite/tree/jm-bugfix-api-change-text-export-rev-contents can be used for testing.

JohnMcLear:
Please test with: https://github.com/ether/etherpad-lite/pull/5190

PR #5190, opened 2021-09-24
bugfix: ensure getText only includes text when a revision is specified.

JohnMcLear:
Fix for https://github.com/ether/etherpad-lite/issues/5073

Note that if people have been using ``getText`` with revision set before this would be a breaking change.  I imagine most people can adapt their environments to use this function as per the docs.

I included tests too.

### DIFF
--- a/src/node/db/API.js
+++ b/src/node/db/API.js
@@ -171,8 +171,8 @@ exports.getText = async (padID, rev) => {
     }
 
     // get the text of this revision
-    // getInternalRevisionAText method returns a text.text and attribs so we ensure we only
-    // return the text - details at https://github.com/ether/etherpad-lite/issues/5073
+    // getInternalRevisionAText() returns an atext object but we only want the .text inside it.
+    // Details at https://github.com/ether/etherpad-lite/issues/5073
     const {text} = await pad.getInternalRevisionAText(rev);
     return {text};
   }
--- a/src/node/db/API.js
+++ b/src/node/db/API.js
@@ -173,10 +173,8 @@ exports.getText = async (padID, rev) => {
     // get the text of this revision
     // getInternalRevisionAText method returns a text.text and attribs so we ensure we only
     // return the text - details at https://github.com/ether/etherpad-lite/issues/5073
-    const text = await pad.getInternalRevisionAText(rev);
-    return {
-      text: text.text
-    };
+    const {text} = await pad.getInternalRevisionAText(rev);
+    return {text};
   }
 
   // the client wants the latest text, lets return it to him
--- a/src/node/db/API.js
+++ b/src/node/db/API.js
@@ -171,8 +171,12 @@ exports.getText = async (padID, rev) => {
     }
 
     // get the text of this revision
+    // getInternalRevisionAText method returns a text.text and attribs so we ensure we only
+    // return the text - details at https://github.com/ether/etherpad-lite/issues/5073
     const text = await pad.getInternalRevisionAText(rev);
-    return {text};
+    return {
+      text: text.text
+    };
   }
 
   // the client wants the latest text, lets return it to him
--- a/src/tests/backend/specs/api/pad.js
+++ b/src/tests/backend/specs/api/pad.js
@@ -114,10 +114,13 @@ describe(__filename, function () {
                                 -> getLastEdited(padID) -- Should not be 0
                                 -> appendText(padID, "hello")
                                 -> getText(padID) -- Should be "hello worldhello"
+                                -> getText(padID, rev) -- Should be "hello world
                                  -> setHTML(padID) -- Should fail on invalid HTML
                                   -> setHTML(padID) *3 -- Should fail on invalid HTML
                                    -> getHTML(padID) -- Should return HTML close to posted HTML
                                     -> createPad -- Tries to create pads with bad url characters
+
+
 
   */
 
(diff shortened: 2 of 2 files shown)


src/tests/backend/specs/api/pad.js
```diff
@@ -586,6 +588,31 @@ describe(__filename, function () {
...
 
+  describe('appendText', function () {
+    it('Append text to a pad Id', function (done) {
+      agent.get(`${endPoint('appendText', '1.2.13')}&padID=${testPadId}&text=hello`)
+          .expect((res) => {
+            if (res.body.code !== 0) throw new Error('Pad Append Text failed');
+          })
+          .expect('Content-Type', /json/)
+          .expect(200, done);
+    });
```
rhansen, line 599: Use Promises and `assert`:

```suggestion
    it('Append text to a pad Id', async function () {
      const res = await agent.get(`${endPoint('appendText', '1.2.13')}&padID=${testPadId}&text=hello`)
          .expect(200)
          .expect('Content-Type', /json/);
      assert.equal(res.body.code, 0);
    });
```

src/tests/backend/specs/api/pad.js
```diff
@@ -586,6 +588,31 @@ describe(__filename, function () {
...
+      agent.get(`${endPoint('getText')}&padID=${testPadId}&rev=3`)
+          .expect((res) => {
+            if (res.body.code !== 0) throw new Error('Pad Get Text failed');
+            if (res.body.data.text !== `${text}hello\n`) {
+              throw new Error('getText Revision not returning correct contents');
+            }
+          })
+          .expect('Content-Type', /json/)
+          .expect(200, done);
+    });
```
rhansen, line 614: Use Promises and `assert`:

```suggestion
    it('Gets text on a pad Id at a given revision', async function () {
      const res = await agent.get(`${endPoint('getText')}&padID=${testPadId}&rev=3`)
          .expect(200)
          .expect('Content-Type', /json/);
      assert.equal(res.body.code, 0);
      assert.equal(res.body.data.text, `${text}hello\n`);
    });
```

src/node/db/API.js
```diff
@@ -171,8 +171,12 @@ exports.getText = async (padID, rev) => {
     }
 
     // get the text of this revision
+    // getInternalRevisionAText method returns a text.text and attribs so we ensure we only
+    // return the text - details at https://github.com/ether/etherpad-lite/issues/5073
     const text = await pad.getInternalRevisionAText(rev);
-    return {text};
+    return {
+      text: text.text
+    };
```
rhansen, line 179:
```suggestion
    const {text} = await pad.getInternalRevisionAText(rev);
    return {text};
```

JohnMcLear: Weird, I thought I took that approach initially but it didn't work for some reason, must have been a brain fart.

JohnMcLear: I prefer leaving the tracked comment in, it's something that in the future might stop me from tripping myself up again.

src/tests/backend/specs/api/pad.js
```diff
@@ -586,6 +588,25 @@ describe(__filename, function () {
     });
   });
 
+  describe('appendText', function () {
```
rhansen, line 591: Mocha doesn't like duplicate names:

```suggestion
  describe('appendText again', function () {
```

src/tests/backend/specs/api/pad.js
```diff
@@ -586,6 +588,25 @@ describe(__filename, function () {
...
+    it('Append text to a pad Id', async function () {
```
rhansen, line 592: This is inside a different `description()`, so it might be OK that it shares the same name as another `it()`. But it might be a good idea to change it anyway.

src/tests/backend/specs/api/pad.js
```diff
@@ -586,6 +588,25 @@ describe(__filename, function () {
...
+      const res = await agent.get(`${endPoint('appendText', '1.2.13')}&padID=${testPadId}&text=hello`)
+          .expect(200)
+          .expect('Content-Type', /json/);
+      assert.equal(res.body.code, 0);
+    });
+  });
+
+
+  describe('getText', function () {
```
rhansen, line 601:
```suggestion
  describe('getText of old revision', function () {
```

src/tests/backend/specs/api/pad.js
```diff
@@ -586,6 +588,25 @@ describe(__filename, function () {
...
+    it('Gets text on a pad Id at a given revision', async function () {
+      const res = await agent.get(`${endPoint('getText')}&padID=${testPadId}&rev=3`)
```
rhansen, line 603: Rather than add a new rev 4 and read rev 3, why not just read rev 2?

src/node/db/API.js
```diff
@@ -171,7 +171,9 @@ exports.getText = async (padID, rev) => {
     }
 
     // get the text of this revision
-    const text = await pad.getInternalRevisionAText(rev);
+    // getInternalRevisionAText method returns a text.text and attribs so we ensure we only
+    // return the text - details at https://github.com/ether/etherpad-lite/issues/5073
```
rhansen, line 175:
```suggestion
    // getInternalRevisionAText() returns an atext object but we only want the .text inside it.
    // Details at https://github.com/ether/etherpad-lite/issues/5073
```

src/tests/backend/specs/api/pad.js
JohnMcLear: My logic was that I just like to know that it was this test that made the change.  In heinsight it would have been smarter to make a new pad and just to rev0 and rev1...   So yea, could prolly drop some logic here!

src/tests/backend/specs/api/pad.js
JohnMcLear: Note that if we apply https://github.com/ether/etherpad-lite/pull/5190#discussion_r716116155 then this is mute.

rhansen:
I pushed a commit that simplifies the tests.

I think this is ready to squash and merge, but it's marked as draft. Is there something you wanted to do first?

PR #5194, opened 2021-09-25
CSS: Fix underscore and clear authorship icon alignment

### LINKED ISSUE #5193: Underline and clearAuthorship buttons slightly unaligned
After 251cc7ab324d138b3f05d03ee3ef70707321678e the underline and clearauthorship buttons feel like they are slightly misaligned, ie the lower baseline seems to be "below" the baselines of all other icons. The problem is also on beta.etherpad.com. Not sure what's the cause :-(

cc @rhansen

rhansen:
Now the toolbar looks like this:

image

Fixes #5193

cc @webzwo0i


webzwo0i:
thanks! Looks good now. I'll see if I can fix the underline glyph in the future.

PR #5232, opened 2021-10-16
tests: Run easysync tests with frontend tests

### LINKED ISSUE #4190: Easysync tests aren't run when backend tests are run
Easysync tests.js exists in node folder.  They should be refactored and included in the backend tests.

rhansen:
Multiple commits:
* tests: easysync: Run with the frontend tests
* tests: easysync: Delete commented-out and unused code
* tests: easysync: Fix some ESLint errors
* tests: easysync: Remove unnecessary `Random` class
* tests: easysync: Use mocha `describe()` and `it()`
* tests: easysync: Use `expect.js` for checks
* tests: easysync: Inline some functions that are only used once

The easysync tests could be run with the backend tests, but the tested code is used in browsers so I put the tests with the frontend tests to get browser coverage.

Fixes #4190

cc @webzwo0i

### DIFF
--- a/src/tests/frontend/specs/easysync.js
+++ b/src/tests/frontend/specs/easysync.js
@@ -29,33 +29,21 @@ const AttributePool = require('../../../static/js/AttributePool');
 const randInt = (maxValue) => Math.floor(Math.random() * maxValue);
 
 describe('easysync', function () {
-  const throughIterator = (opsStr) => {
-    const iter = Changeset.opIterator(opsStr);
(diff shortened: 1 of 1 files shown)


webzwo0i:
Looks perfect. I verified coverage results before and after (as backend test though, ie I added expect and run it through nyc) and they are identical when it comes to "what lines are run". They differ when it comes to "how often is this line hit", probably due to randomness. Maybe in the future we should hard code values, testing all possible x instead of `rand(x)`

Issue #440, opened 2012-02-11
cmd-delete should delete entire line, not single word (Mac OS X)


rmondello:
In a standard textarea on Mac OS X, cmd-delete removes all of the text from the insertion point back until the beginning of the line the insertion point is on. In etherpad-lite, a single word is deleted, which is what option-delete does.

In order to match the expected behavior of textareas on Mac OS X, this should be changed.

JohnMcLear:
Mac people are such heavy contributors, I just can't handle the number of pull requests landing to solve this issue despite it only being a one liner...  *grins.

muxator:
@Tristramg, given your recent story with Etherpad, do you happen to have a Mac and want to work on this macOS ergonomy fix?

Otherwise I'll close for inactivity.

JohnMcLear:
I think we can close it :)

PR #5240, opened 2021-10-20
feat(ace2_inner.js): Add keybinding meta-backspace to delete to beginning of line

pjvandehaar:
It would be more natural to add the new code inside the code that handles ctrl-backspace, but then this couldn't delete multiple "tabs" (which are actually 4 spaces).

It would be nice to have this feature on Windows too, but I don't know what keybinding it should have.  Maybe ctrl-shift-backspace?

Fixes https://github.com/ether/etherpad-lite/issues/440

### DIFF
--- a/src/static/js/ace2_inner.js
+++ b/src/static/js/ace2_inner.js
@@ -2482,7 +2482,11 @@ function Ace2Inner(editorInfo, cssManagers) {
         const lineEntry = rep.lines.atIndex(lineNum);
         const lineText = lineEntry.text;
         const lineMarker = lineEntry.lineMarker;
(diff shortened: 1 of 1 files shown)


webzwo0i:
Did you test (manually) that this also works in lists or headings, ie deleting the characters before the caret but keeping the list attribute/heading if you immediately type anything? (I assume it works, because you use linemarker, just want to be sure)

pjvandehaar:
Does this PR need changes, or is it ready to merge?

It works, but I didn't write tests because I don't know how to send command-backspace via `sendkeys()`.  Does it need to also support ctrl-shift-backspace, for Windows?

pjvandehaar:
Thanks for merging!

Issue #5213, opened 2021-10-01
Misuse of "Import" erases all history data


jej:
**Describe the bug**
I see some closed issues about this, but maybe it worths to reopen it.
I suggested to a friend to use etherpad for a collaboration work. After few days and many modifications, she wanted to import an inline image in the text. She clicked on "Import/Export", picked an image file and didn't care about the warning... Of course the pad was scrambled by a binary text, and the image was not inserted inline as expected by the user. The problem is that now the historic does not show the revisions any more. The revisions are browsable with the slider, but current content only is displayed. All previous revisions seems recorded as usual, but not available.
I tested this on https://framapad.org/fr/ . I don't know what version of framapad they use.

**To Reproduce**
1. Write/modify some text, in order to create some revisions.
2. Click on 'Import/Export'.
3. Pickup a binary image file (ex. png/jpg)
4. Import it, acknowledge the warning.
5. The binary data replace the text. Eventually: erase it and add some text. 
6. Look at the history to see if previous revisions are still displayed.

**Expected behavior**
It's not a problem if binary data erase the current version, but at least it should be possible to retrieve the previous content of the pad.
Maybe the GUI should be more careful with import. A mandatory checkbox could make the user aware that **all the content** will be replaced, instead of a simple button.

**Server (please complete the following information):**
 - Etherpad version: ?
 - Here: https://framapad.org/

JohnMcLear:
1.  There is a plugin for forking pads to a new state from a point in time.  ep_fork maybe?  You access the previous revision through the timeslider.  Plugins are visible on the plugin list page.

2. Re importing ux, I would say the expected ux for import in any text editor is to replace the whole document no?

jej:
Thanks for your reply.
1. I didn't know this plugin. I will test for my own use.
2. The problem here is the bug about the timeslider. If the import replace the doc, why the slider allows to browse the revisions but don't show them? I would say that import a doc just create a new revision with the new content. But here the concept of the pad is broken...

JohnMcLear:
1. Gimme 5 mins will find it for you.
2. Import should create a new revision, you are right!  Can you provide exact steps to replicate then here because if you are saying it's not creating a new revision something is very broken!  Please provide steps to replicate on a clean install / no plugins etc.

JohnMcLear:
[...]

PR #5253, opened 2021-10-29
Fix textLinesMutator when dealing with insertions/deletions at the end of pad

### LINKED ISSUE #5214: This .etherpad contents breaks timeslider
```
{"pad:zsx23zLru1V9ZGAG1t7u":{"atext":{"text":".kj lij lkj lkj kl jlkj \n","attribs":"*0+o|1+1"},"pool":{"numToAttrib":{"0":["author","a.DeyjwfIUFrfSDNzv"]},"nextNum":1},"head":13,"chatHead":-1,"publicStatus":false,"savedRevisions":[]},"globalAuthor:a.DeyjwfIUFrfSDNzv":{"colorId":7,"name":null,"timestamp":1633170305516,"padIDs":"zsx23zLru1V9ZGAG1t7u"},"pad:zsx23zLru1V9ZGAG1t7u:revs:0":{"changeset":"Z:1>ay|7+ay$Welcome to Etherpad!\n\nThis pad text is synchronized as you type, so that everyone viewing this page sees the same text. This allows you to collaborate seamlessly on documents!\n\nGet involved with Etherpad at https://etherpad.org\n\nWarning: DirtyDB is used. This is not recommended for production. -- To suppress these warning messages change suppressErrorsInPadText to true in your settings.json\n","meta":{"author":"","timestamp":1633170121014,"pool":{"numToAttrib":{},"attribToNum":{},"nextNum":0},"atext":{"text":"Welcome to Etherpad!\n\nThis pad text is synchronized as you type, so that everyone viewing this page sees the same text. This allows you to collaborate seamlessly on documents!\n\nGet involved with Etherpad at https://etherpad.org\n\nWarning: DirtyDB is used. This is not recommended for production. -- To suppress these warning messages change suppressErrorsInPadText to true in your settings.json\n\n","attribs":"|8+az"}}},"pad:zsx23zLru1V9ZGAG1t7u:revs:1":{"changeset":"Z:az<aw|7-ay*0+2$GH","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170122631}},"pad:zsx23zLru1V9ZGAG1t7u:revs:2":{"changeset":"Z:3>4=2*0+4$EWEW","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170123131}},"pad:zsx23zLru1V9ZGAG1t7u:revs:3":{"changeset":"Z:7>c=6*0+c$PORIOPWIKESR","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170123634}},"pad:zsx23zLru1V9ZGAG1t7u:revs:4":{"changeset":"Z:j>4=i*0|2+2*0+2$\n\nSD","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170124134}},"pad:zsx23zLru1V9ZGAG1t7u:revs:5":{"changeset":"Z:n>a|2=k=2*0|4+a$F\nS\nDF\nSD\n","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170124634}},"pad:zsx23zLru1V9ZGAG1t7u:revs:6":{"changeset":"Z:x>7|6=w*0|3+7$F\n\nSDF\n","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170125134}},"pad:zsx23zLru1V9ZGAG1t7u:revs:7":{"changeset":"Z:14>6|9=13*0|3+6$\nSDF\n\n","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170125634}},"pad:zsx23zLru1V9ZGAG1t7u:revs:8":{"changeset":"Z:1a>2|c=19*0|2+2$\n\n","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170126134}},"pad:zsx23zLru1V9ZGAG1t7u:revs:9":{"changeset":"Z:1c<1b|f-1c|1+1$\n","meta":{"author":"","timestamp":1633170137010}},"pad:zsx23zLru1V9ZGAG1t7u:revs:10":{"changeset":"Z:1>0$","meta":{"author":"","timestamp":1633170137011}},"pad:zsx23zLru1V9ZGAG1t7u:revs:11":{"changeset":"Z:1>1*0+1$.","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170227678}},"pad:zsx23zLru1V9ZGAG1t7u:revs:12":{"changeset":"Z:2>b=1*0+b$kj lij lkj ","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170228177}},"pad:zsx23zLru1V9ZGAG1t7u:revs:13":{"changeset":"Z:d>c=c*0+c$lkj kl jlkj ","meta":{"author":"a.DeyjwfIUFrfSDNzv","timestamp":1633170228677}}}
```

STR is import this .etherpad contents and then click play in timeslider.

Details from https://github.com/ether/etherpad-lite/issues/5213

webzwo0i:
Will clean this up tomorrow (fixing eslint etc). This should fix https://github.com/ether/etherpad-lite/issues/5214 https://github.com/ether/etherpad-lite/issues/5213 and replace https://github.com/ether/etherpad-lite/pull/5215 and https://github.com/ether/etherpad-lite/pull/2911

Also adds some comments and additional coverage.

### DIFF
--- a/src/static/js/Changeset.js
+++ b/src/static/js/Changeset.js
@@ -983,11 +983,17 @@ class TextLinesMutator {
         this._curSplice.push(...newLines);
         this._curLine += newLines.length;
       }
-    } else {
+    } else if (!this.hasMore()) {
       // There are no additional lines. Although the line is put into splice, curLine is not
-      // increased because there may be more chars in the line (newline is not reached).
+      // increased because there may be more chars in the line (newline is not reached). We are
+      // inserting at the end of lines. curCol is 0 as curLine is not in splice.
+      this._curSplice.push(text);
+      this._curCol += text.length;
+    } else {
+      // insert text after curCol
       const sline = this._putCurLineInSplice();
       if (!this._curSplice[sline]) {
+        // TODO should never happen now
         const err = new Error(
             'curSplice[sline] not populated, actual curSplice contents is ' +
             `${JSON.stringify(this._curSplice)}. Possibly related to ` +
--- a/src/static/js/Changeset.js
+++ b/src/static/js/Changeset.js
@@ -908,8 +908,10 @@ class TextLinesMutator {
     let removed = '';
     if (this._isCurLineInSplice()) {
       if (this._curCol === 0) {
+        // First line to be removed is in splice.
         removed = this._curSplice[this._curSplice.length - 1];
         this._curSplice.length--;
+        // Next lines to be removed are not in splice.
         removed += nextKLinesText(L - 1);
         this._curSplice[1] += L - 1;
       } else {
@@ -917,11 +919,13 @@ class TextLinesMutator {
         this._curSplice[1] += L - 1;
         const sline = this._curSplice.length - 1;
         removed = this._curSplice[sline].substring(this._curCol) + removed;
-        this._curSplice[sline] = this._curSplice[sline].substring(0, this._curCol) +
-            this._linesGet(this._curSplice[0] + this._curSplice[1]);
-        this._curSplice[1] += 1;
+        // Is there a line left?
+        const remaining = this._linesGet(this._curSplice[0] + this._curSplice[1]) || '';
+        this._curSplice[sline] = this._curSplice[sline].substring(0, this._curCol) + remaining;
+        this._curSplice[1] += remaining ? 1 : 0;
       }
     } else {
+      // Nothing that is removed is in splice. Implies curCol === 0.
       removed = nextKLinesText(L);
       this._curSplice[1] += L;
     }
--- a/src/static/js/Changeset.js
+++ b/src/static/js/Changeset.js
@@ -977,8 +977,9 @@ class TextLinesMutator {
         this._curLine += newLines.length;
         // insert the remaining chars from the "old" line (e.g. the line we were in
         // when we started to insert new lines)
-        this._curSplice.push(theLine.substring(lineCol));
-        this._curCol = 0; // TODO(doc) why is this not set to the length of last line?
+        const remaining = theLine.substring(lineCol);
+        if (remaining !== '') this._curSplice.push(remaining);
+        this._curCol = 0;
       } else {
         this._curSplice.push(...newLines);
         this._curLine += newLines.length;
--- a/src/tests/frontend/specs/easysync-assembler.js
+++ b/src/tests/frontend/specs/easysync-assembler.js
@@ -92,6 +92,40 @@ describe('easysync-assembler', function () {
     expect(assem.toString()).to.equal('-k');
   });
 
+  it('smartOpAssembler append - op with text', async function () {
+    const assem = Changeset.smartOpAssembler();
+    const pool = poolOrArray([
+      'attr1,1',
+      'attr2,2',
+      'attr3,3',
+      'attr4,4',
+      'attr5,5',
+    ]);
+
+    assem.appendOpWithText('-', 'test', '*3*4*5', pool);
+    assem.appendOpWithText('-', 'test', '*3*4*5', pool);
+    assem.appendOpWithText('-', 'test', '*1*4*5', pool);
+    assem.endDocument();
+    expect(assem.toString()).to.equal('*3*4*5-8*1*4*5-4');
+  });
+
+  it('smartOpAssembler append - op with multiline text', async function () {
+    const assem = Changeset.smartOpAssembler();
+    const pool = poolOrArray([
+      'attr1,1',
+      'attr2,2',
+      'attr3,3',
+      'attr4,4',
+      'attr5,5',
+    ]);
+
+    assem.appendOpWithText('-', 'test\ntest', '*3*4*5', pool);
+    assem.appendOpWithText('-', '\ntest\n', '*3*4*5', pool);
+    assem.appendOpWithText('-', '\ntest', '*1*4*5', pool);
+    assem.endDocument();
+    expect(assem.toString()).to.equal('*3*4*5|3-f*1*4*5|1-1*1*4*5-4');
+  });
+
   it('smartOpAssembler append + op with text', async function () {
     const assem = Changeset.smartOpAssembler();
     const pool = poolOrArray([
--- a/src/tests/frontend/specs/easysync-assembler.js
+++ b/src/tests/frontend/specs/easysync-assembler.js
@@ -102,9 +102,14 @@ describe('easysync-assembler', function () {
       'attr5,5',
     ]);
 
-    assem.appendOpWithText('-', 'test', '*3*4*5', pool);
-    assem.appendOpWithText('-', 'test', '*3*4*5', pool);
-    assem.appendOpWithText('-', 'test', '*1*4*5', pool);
+    padutils.warnDeprecated.disabledForTestingOnly = true;
+    try {
+      assem.appendOpWithText('-', 'test', '*3*4*5', pool);
+      assem.appendOpWithText('-', 'test', '*3*4*5', pool);
+      assem.appendOpWithText('-', 'test', '*1*4*5', pool);
+    } finally {
+      delete padutils.warnDeprecated.disabledForTestingOnly;
+    }
     assem.endDocument();
     expect(assem.toString()).to.equal('*3*4*5-8*1*4*5-4');
   });
@@ -119,9 +124,14 @@ describe('easysync-assembler', function () {
       'attr5,5',
     ]);
 
-    assem.appendOpWithText('-', 'test\ntest', '*3*4*5', pool);
-    assem.appendOpWithText('-', '\ntest\n', '*3*4*5', pool);
-    assem.appendOpWithText('-', '\ntest', '*1*4*5', pool);
+    padutils.warnDeprecated.disabledForTestingOnly = true;
+    try {
+      assem.appendOpWithText('-', 'test\ntest', '*3*4*5', pool);
+      assem.appendOpWithText('-', '\ntest\n', '*3*4*5', pool);
+      assem.appendOpWithText('-', '\ntest', '*1*4*5', pool);
+    } finally {
+      delete padutils.warnDeprecated.disabledForTestingOnly;
+    }
     assem.endDocument();
     expect(assem.toString()).to.equal('*3*4*5|3-f*1*4*5|1-1*1*4*5-4');
   });
--- a/src/tests/frontend/specs/easysync-mutations.js
+++ b/src/tests/frontend/specs/easysync-mutations.js
@@ -160,6 +160,25 @@ describe('easysync-mutations', function () {
     ['insert', 'z'],
   ], ['fuz']);
 
+  // #2836, #5214, #3560 regressions
+  runMutationTest(11, ['\n'], [
+    ['remove', 1, 1, '\n'],
+    ['insert', 'c', 0],
+  ], ['c']);
+
+  runMutationTest(12, ['\n'], [
+    ['remove', 1, 1, '\n'],
+    ['insert', 'a\n', 1],
+    ['insert', 'c'],
+  ], ['a\n', 'c']);
+
+  runMutationTest(13, ['\n', 'fun\n', '\n'], [
+    ['remove', 1, 1, '\n'],
+    ['skip', 4, 1, false],
+    ['remove', 1, 1, '\n'],
+    ['insert', 'c'],
+  ], ['fun\n', 'c']);
+
   it('mutatorHasMore', async function () {
     const lines = ['1\n', '2\n', '3\n', '4\n'];
     let mu;
--- a/src/tests/frontend/specs/easysync-mutations.js
+++ b/src/tests/frontend/specs/easysync-mutations.js
@@ -137,6 +137,29 @@ describe('easysync-mutations', function () {
     ['skip', 1, 1, true],
   ], ['banana\n', 'cabbage\n', 'duffle\n']);
 
+  runMutationTest(8, ['\n', 'fun\n', '\n'], [
+    ['remove', 1, 1, '\n'],
+    ['skip', 3, 0, false],
+    ['remove', 2, 2, '\n\n'],
+    ['insert', 'c'],
+  ], ['func']);
+
+  runMutationTest(9, ['\n', 'fun\n', '\n'], [
+    ['remove', 1, 1, '\n'],
+    ['skip', 3, 0, false],
+    ['remove', 2, 2, '\n\n'],
+    ['insert', 'c'],
+    ['insert', 'a\n', 1],
+    ['insert', 'c'],
+  ], ['funca\n', 'c']);
+
+  runMutationTest(10, ['\n', 'fun\n', '\n'], [
+    ['remove', 1, 1, '\n'],
+    ['skip', 2, 0, false],
+    ['remove', 3, 2, 'n\n\n'],
+    ['insert', 'z'],
+  ], ['fuz']);
+
   it('mutatorHasMore', async function () {
     const lines = ['1\n', '2\n', '3\n', '4\n'];
     let mu;
--- a/src/tests/frontend/specs/easysync-mutations.js
+++ b/src/tests/frontend/specs/easysync-mutations.js
@@ -179,6 +179,11 @@ describe('easysync-mutations', function () {
     ['insert', 'c'],
   ], ['fun\n', 'c']);
 
+  runMutationTest(14, ['\n'], [
+    ['remove', 1, 1, '\n'],
+    ['insert', 'a'],
+    ['insert', 'c\n', 1],
+  ], ['ac\n']);
   it('mutatorHasMore', async function () {
     const lines = ['1\n', '2\n', '3\n', '4\n'];
     let mu;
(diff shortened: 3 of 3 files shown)


webzwo0i:
As a side note: The commits that fixed https://github.com/ether/etherpad-lite/issues/2802 should be re-evaluated.

webzwo0i:
This should be ready now.

src/tests/frontend/specs/easysync-helper.js
```diff
@@ -0,0 +1,226 @@
...
+exports.randomMultiline = randomMultiline;
+
+const randomInlineString = (len) => {
+  const assem = Changeset.stringAssembler();
+  for (let i = 0; i < len; i++) {
+    assem.append(String.fromCharCode(randInt(26) + 97));
+  }
+  return assem.toString();
+};
+exports.randomInlineString = randomInlineString;
```
rhansen, line 35: This isn't used outside this file.

```suggestion
```

src/tests/frontend/specs/easysync-helper.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+  const maxOrig = numCharsLeft - 1;
+  if ('remove' in result) {
+    result.remove = Math.min(result.remove, maxOrig);
+  } else if ('skip' in result) {
+    result.skip = Math.min(result.skip, maxOrig);
+  }
+  return result;
+};
+exports.randomStringOperation = randomStringOperation;
```
rhansen, line 194: This isn't used outside this file.

```suggestion
```

src/tests/frontend/specs/easysync-helper.js
```diff
@@ -0,0 +1,226 @@
...
+    } else {
+      return `*${Changeset.numToString(randInt(2) * 2)}`;
+    }
+  } else if (opcode === '+' || randInt(4) === 0) {
+    return '*1*3';
+  } else {
+    return ['*0*2', '*0*3', '*1*2'][randInt(3)];
+  }
+};
+exports.randomTwoPropAttribs = randomTwoPropAttribs;
```
rhansen, line 212: This isn't used outside this file.

```suggestion
```

rhansen (commented):
* [x] The "chore Changeset.js: use template strings" commit missed one: Line 1383 (in `outputMutOp()`).
  * [x] The "easysync tests: move to separate files" commit is difficult to review, mostly because stuff was reordered. Please split this commit into two: One that splits `easysync.js` into multiple files but preserves relative order, and one that reorders stuff.
  * [x] The "easysync tests: move to separate files" commit introduces a `//TODO: why isn't this |2+2*x|1+4 ?` comment to `src/tests/frontend/specs/easysync-assembler.js`. The comment doesn't belong in that commit.
  * [x] Please keep the first paragraph of each commit message short—ideally less than 50 chars but definitely less than 70. If you need to say more, put it in a separate paragraph (blank line between the paragraphs).
  * [ ] Please add the regression tests in the commit that fixes the bug.

webzwo0i:
Thanks for the input @rhansen! Will clean up.

Regarding putting the test into the same commit as the fix, I thought we also put the test in a separate commit (before all other commits) as stated somewhere in CONTRIBUTING.md. But in this case, as it's fixing multiple bugs in one PR, I'll definitely include them into the correct commit.

rhansen:
Rebased onto latest `develop`.

rhansen:
> Regarding putting the test into the same commit as the fix, I thought we also put the test in a separate commit (before all other commits) as stated somewhere in CONTRIBUTING.md.

It's OK to add an `xit()` test before the commit that fixes the bug, then change the `xit()` to `it()` in the commit that fixes the bug. I generally don't like that approach because Mocha doesn't attempt `xit()` tests. Other test frameworks support the concept of "expected to fail" tests that are still run but don't result in overall failure unless they actually pass (in which case the "expected to fail" should be converted to the usual "expected to pass").

rhansen:
I rebased onto latest `develop` and made some changes to address my feedback. The "easysync tests: move to separate files" commit was split into 5 different commits to make it easier to review. I think the only thing left is to add the regression tests to the commits that fix the associated bugs.

webzwo0i:
Thanks! I added the tests to the appropriate commits, added `poolOrArray` to `easysync-assembler` and also refined the commit messages of the last three commits.

src/static/js/Changeset.js
```diff
@@ -903,20 +907,24 @@ class TextLinesMutator {
     let removed = '';
     if (this._isCurLineInSplice()) {
```
rhansen, line 908: What if `this._curCol` is greater than or equal to the current line's length? In that case, shouldn't we set `this._curCol` to 0 and increment `this._curLine` before this check? Or is that case not possible due to invariants?

If invariants make that case impossible, I think it would be worthwhile to add an assert statement.

src/static/js/Changeset.js
```diff
@@ -903,20 +907,24 @@ class TextLinesMutator {
...
       if (this._curCol === 0) {
```
rhansen, line 909: What about line markers? If the first character is a line marker and `this._curCol === 1`, shouldn't it be treated the same as if `this._curCol === 0`? Or is that not possible due to invariants? Unfortunately, I don't think it's possible to add an assert here because TextLinesMutator doesn't have access to the attributes.

src/static/js/Changeset.js
```diff
@@ -903,20 +907,24 @@ class TextLinesMutator {
...
         removed = nextKLinesText(L - 1);
         this._curSplice[1] += L - 1;
         const sline = this._curSplice.length - 1;
         removed = this._curSplice[sline].substring(this._curCol) + removed;
-        this._curSplice[sline] = this._curSplice[sline].substring(0, this._curCol) +
-            this._linesGet(this._curSplice[0] + this._curSplice[1]);
-        this._curSplice[1] += 1;
+        // Is there a line left?
+        const remaining = this._linesGet(this._curSplice[0] + this._curSplice[1]) || '';
+        this._curSplice[sline] = this._curSplice[sline].substring(0, this._curCol) + remaining;
```
rhansen, line 923: If `remaining` is the empty string, then `this._curSplice[sline]` will not end with a newline. Don't all entries in `this._lines` need to end with a newline character?

```diff
diff --git a/src/static/js/Changeset.js b/src/static/js/Changeset.js
index 5e0bc34fc..defc08852 100644
--- a/src/static/js/Changeset.js
+++ b/src/static/js/Changeset.js
@@ -918,11 +918,23 @@ class TextLinesMutator {
         removed = nextKLinesText(L - 1);
         this._curSplice[1] += L - 1;
         const sline = this._curSplice.length - 1;
-        removed = this._curSplice[sline].substring(this._curCol) + removed;
-        // Is there a line left?
-        const remaining = this._linesGet(this._curSplice[0] + this._curSplice[1]) || '';
-        this._curSplice[sline] = this._curSplice[sline].substring(0, this._curCol) + remaining;
-        this._curSplice[1] += remaining ? 1 : 0;
+        removed = this._curSplice[sline].slice(this._curCol) + removed;
+        this._curSplice[sline] = this._curSplice[sline].slice(0, this._curCol);
+        // The current line's newline was deleted, which means that the line after the splice (if
+        // there is one) should be joined with what remains of the current line.
+        const n = this._curSplice[0] + this._curSplice[1];
+        if (n < this._linesLength()) {
+          // There is a line that can be joined with the current line.
+          // TODO: What if the joined line starts with a line marker?
+          // TODO: What if the joined line has line attributes that differ from the current line's
+          // line attributes?
+          this._curSplice[sline] += this._linesGet(n);
+          this._curSplice[1] += 1;
+        } else {
+          // There is no line that can be joined with the current line. The current line's newline
+          // was removed, but all lines must end with a newline so add it back.
+          this._curSplice[sline] += '\n';
+        }
       }
     } else {
       // Nothing that is removed is in splice. Implies curCol === 0.
```

src/static/js/Changeset.js
```diff
@@ -968,18 +976,24 @@ class TextLinesMutator {
...
+        this._curCol = 0;
       } else {
         this._curSplice.push(...newLines);
         this._curLine += newLines.length;
       }
+    } else if (!this.hasMore()) {
+      // There are no additional lines. Although the line is put into splice, curLine is not
+      // increased because there may be more chars in the line (newline is not reached). We are
+      // inserting at the end of lines. curCol is 0 as curLine is not in splice.
+      this._curSplice.push(text);
```
rhansen, line 990: I don't think this is right. If my understanding is correct, each entry in `this._lines` represents a line of the document and must end with a newline character. By pushing to `this._curSplice`, this code creates a new line at the end of the document, one that line doesn't end with a newline. This code should instead insert the text after column `this._curCol` (which should be before the line's newline) in line `this._curLine`.

I think the actual bug is inside `this._putCurLineInSplice()`: That method should not attempt to read past the end of `this._lines` or delete lines that don't exist in `this._lines`. Something like this (untested):

```diff
diff --git a/src/static/js/Changeset.js b/src/static/js/Changeset.js
index 5e0bc34fc..930a5e187 100644
--- a/src/static/js/Changeset.js
+++ b/src/static/js/Changeset.js
@@ -823,11 +823,18 @@ class TextLinesMutator {
    * @returns {number} the index of the added line in curSplice
    */
   _putCurLineInSplice() {
-    if (!this._isCurLineInSplice()) {
-      this._curSplice.push(this._linesGet(this._curSplice[0] + this._curSplice[1]));
-      this._curSplice[1]++;
+    assert(this._inSplice, 'not in splice');
+    assert(this._curLine >= this._curSplice[0], 'curLine is before splice start');
+    while (this._curSplice[0] + this._curSplice.length - 2 < this._curLine + 1) {
+      const n = this._curSplice[0] + this._curSplice[1];
+      if (n < this._linesLength()) {
+        this._curSplice.push(this._linesGet(n));
+        this._curSplice[1]++;
+      } else {
+        // No more lines available to add to the splice. Create a new line from scratch.
+        this._curSplice.push('\n');
+      }
     }
-    // TODO should be the same as this._curSplice.length - 1
     return 2 + this._curLine - this._curSplice[0];
   }
 
```

src/static/js/Changeset.js
```diff
@@ -968,18 +976,24 @@ class TextLinesMutator {
         this._curLine += newLines.length;
         // insert the remaining chars from the "old" line (e.g. the line we were in
         // when we started to insert new lines)
-        this._curSplice.push(theLine.substring(lineCol));
-        this._curCol = 0; // TODO(doc) why is this not set to the length of last line?
+        const remaining = theLine.substring(lineCol);
+        if (remaining !== '') this._curSplice.push(remaining);
```
rhansen, line 980: If `remaining === ''`, then that means that `this._curCol` points just past the line's newline, correct? If so, I think the better fix would be to set `this._curCol` to 0 and increment `this._curLine` before doing anything else.

Also, `this._curCol` pointing just past the newline feels like a violation of an invariant. Maybe instead of checking if `remaining !== ''`, it should assert that and we should fix whatever logic is causing `this._curCol` to point past the newline.

src/tests/frontend/specs/easysync-assembler.js
```diff
@@ -18,6 +19,171 @@ describe('easysync-assembler', function () {
...
+      'attr3,3',
+      'attr4,4',
+      'attr5,5',
+    ]);
+
+    assem.appendOpWithText('-', 'test', '*3*4*5', pool);
+    assem.appendOpWithText('-', 'test', '*3*4*5', pool);
+    assem.appendOpWithText('-', 'test', '*1*4*5', pool);
+    assem.endDocument();
+    expect(assem.toString()).to.equal('*3*4*5-8*1*4*5-4');
```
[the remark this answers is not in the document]
webzwo0i: Thanks. I'll move this to another issue/PR so we can isolate it from the other changes. Not sure if it's explicitly forbidden (ie if it can cause any harm), but I can't imagine any real world scenario where remove ops will contain attributes.

src/tests/frontend/specs/easysync-assembler.js
```diff
@@ -18,6 +19,171 @@ describe('easysync-assembler', function () {
...
+      'attr3,3',
+      'attr4,4',
+      'attr5,5',
+    ]);
+
+    assem.appendOpWithText('-', 'test\ntest', '*3*4*5', pool);
+    assem.appendOpWithText('-', '\ntest\n', '*3*4*5', pool);
+    assem.appendOpWithText('-', '\ntest', '*1*4*5', pool);
+    assem.endDocument();
+    expect(assem.toString()).to.equal('*3*4*5|3-f*1*4*5|1-1*1*4*5-4');
```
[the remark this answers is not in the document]
webzwo0i: Will remove these two tests from this PR for now

src/static/js/Changeset.js
webzwo0i: skip:
- if L present, curCol is set to 0 in all branches (in skipLines)
- if L not present, curCol will be +=N (number of skipped chars) [0]

insert:
- if L present, curCol is set to 0 after applying L, because of the "multiline end with newline" constraint
- if L is present, but curLine was not in spliceXXX, curCol isn't touched
- if L is not present and we are processing the last line of lines array, curCol is increased by N chars [1]
- if L is not present and we are not processing the last line, curCol is increased by N chars [2]

remove:
- no changes to curCol

In cases [1] and [2] curCol calculation is safe, as a text string is given to the respective function (the term that's added to curCol is derived from from text).

Also after calling `close` or skipping over multiple L that don't have attributes, the splice is not active anymore and it's possible that curCol is > 0. So there are four possible "start" states:
- splice not active, curCol is 0 [X1]
- splice not active, curCol > 0 [X2]
- splice active, curCol is 0 [X3]
- splice active, curCol > 0 [X4]

So for all 4 "start" states, I'll add coverage for [0] exceeding the number of chars in the line (ie a text of "test\n" and skipping over 6 chars so we end in the middle of the next line), followed by a call to removeLines.

If we want to assert I think it's better to do it where curCol is changed.

src/static/js/Changeset.js
webzwo0i: Maybe I don't understand, but the ops that are applied to lines shouldn't care about lineMarker's special meaning in the editor. ie lineMarker is special in the editor, because it's not rendered and it changes the "meaning" of the whole line, but when it comes to Changeset.js whatever attributes apply to the lineMarker are only applied to this single, "real" character. So we should not treat `this._curCol === 1` as `this._curCol === 0`

src/static/js/Changeset.js
webzwo0i: There is not necessarly a newline at every line. Etherpad uses it with lines that always contain newlines (e.g. rep.lines), except for `mutateAttributionLines`, but this has other logic.
Also take a look at the test case, this is valid imo.

webzwo0i: I do understand what you wrote in the first paragraph, though. If there is a newline, we would never go "beyond" that. Let me test

src/static/js/Changeset.js
webzwo0i: Thanks for pointing to `_putCurLineInSplice`. Will take a look.

Imo, the "must end with newline" constraint isn't true. However, I'm not sure if the "documents must end with newline" constraint applies, though... I think as long as mutateAttributionLines keeps working without any change, I can ensure newlines are present on every line.

webzwo0i:
OK, I think I ensure that newlines are everywhere first. It's possible that I didn't saw missing newlines because it's dealing with edits at the end of the pad.

src/static/js/Changeset.js
rhansen: What should happen if an op removes the last line's newline but that op is immediately followed by another op that restores the newline? For example:

```javascript
const text = 'x\n';
const lines = [text];
const cs = '=1|1-1|1+1$\n'; // Keep a char, then remove a newline, then add a newline.
Changeset.mutateTextLines(cs, lines);
assert.equal(lines.join(''), 'x\n');
assert.equal(Changeset.applyToText(cs, text), 'x\n');
```

Similarly, what if an op adds text to the end of the document without a trailing newline, then the next op adds the newline:

```javascript
const text = 'x\n';
const lines = [text];
const cs = '|1=2+1|1+1$y\n'; // Keep existing text, add a char, then add a newline.
Changeset.mutateTextLines(cs, lines);
assert.equal(lines.join(''), 'x\ny\n');
assert.equal(Changeset.applyToText(cs, text), 'x\ny\n');
```

If we force newlines to always be present in each entry in `this._lines` then we're going to end up inserting extra newlines when we don't want them. I think the way to fix this is to strip the newlines from `this._lines` and add them back before returning.

rhansen:
Any update @webzwo0i?

webzwo0i:
Will have time in the next few days.

rhansen:
I rebased onto latest `develop` and pushed a couple of fixup commits.

LeaChemoul:
Any update on thoses fixup commits ?

JohnMcLear:
Bump @webzwo0i -- any progress?

JohnMcLear:
@dependabot rebase

SamTV12345:
@webzwo0i Do you have some time to fix this pr?

JohnMcLear:
Closing this PR as it's been open for over 4 years and is heavily conflicted with the current codebase. The underlying issues (#5213, #5214) may still be relevant — if so, a fresh PR against current develop would be the best path forward. Thank you @webzwo0i for the contribution\!

PR #5266, opened 2021-11-08
Update to the latest slim package

tommytran732:
The Nodejs 14 slim image has quite a few vulnerabilities, and I have tested the latest slim image. It works just fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 #
 # Author: muxator
 
-FROM node:14-buster-slim
+FROM node:lts-slim
 LABEL maintainer="Etherpad team, https://github.com/ether/etherpad-lite"
 
 # plugins to install while building the container. By default no plugins are
@@ -85,7 +85,7 @@ COPY --chown=etherpad:etherpad ./ ./
 # seems to confuse tools such as `npm outdated`, `npm update`, and some ESLint
 # rules.
 RUN { [ -z "${ETHERPAD_PLUGINS}" ] || \
-      npm install --no-save ${ETHERPAD_PLUGINS}; } && \
+      npm install --no-save --legacy-peer-deps ${ETHERPAD_PLUGINS}; } && \
     src/bin/installDeps.sh && \
     rm -rf ~/.npm
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,6 +60,7 @@ RUN mkdir -p "${EP_DIR}" && chown etherpad:etherpad "${EP_DIR}"
 RUN export DEBIAN_FRONTEND=noninteractive; \
     mkdir -p /usr/share/man/man1 && \
     apt-get -qq update && \
+    apt-get -qq dist-upgrade && \
     apt-get -qq --no-install-recommends install \
         ca-certificates \
         git \


JohnMcLear:
Wouldn't this still need sticking to a version?

Dockerfile
```diff
@@ -4,7 +4,7 @@
 #
 # Author: muxator
 
-FROM node:14-buster-slim
+FROM node:slim
```
rhansen, line 7: I would prefer to track the active LTS release because the [Node.js releases page](https://nodejs.org/en/about/releases/) says, "Production applications should only use Active LTS or Maintenance LTS releases."

```suggestion
FROM node:lts-slim
```


Also, Node.js v15 and later come with npm v7, which is a major change over npm v6. We've had some compatibility problems with npm v7 when installing plugins. I recommend changing line 88 to pass `--legacy-peer-deps` to `npm install`:

```diff
diff --git a/Dockerfile b/Dockerfile
--- a/Dockerfile
+++ b/Dockerfile
@@ -85,7 +85,7 @@ COPY --chown=etherpad:etherpad ./ ./
 # seems to confuse tools such as `npm outdated`, `npm update`, and some ESLint
 # rules.
 RUN { [ -z "${ETHERPAD_PLUGINS}" ] || \
-      npm install --no-save ${ETHERPAD_PLUGINS}; } && \
+      npm install --no-save --legacy-peer-deps ${ETHERPAD_PLUGINS}; } && \
     src/bin/installDeps.sh && \
     rm -rf ~/.npm
 
```

rhansen (commented):
> The Nodejs 14 slim image has quite a few vulnerabilities

Are you sure? Node.js v14 is still maintained, and gets regular security updates.

> Wouldn't this still need sticking to a version?

I don't think there's a need to pin to a specific release. If someone reports that a new Node.js release is incompatible, then we can easily change `Dockerfile` to pin to an older LTS release until the incompatibility is resolved.

Also, if we track the active LTS release instead of current, then we will have at least 6 months to address compatibility issues with a new release before it becomes a problem for the Docker image.

tommytran732:
> > The Nodejs 14 slim image has quite a few vulnerabilities
> 
> Are you sure? Node.js v14 is still maintained, and gets regular security updates.
> 
> > Wouldn't this still need sticking to a version?
> 
> I don't think there's a need to pin to a specific release. If someone reports that a new Node.js release is incompatible, then we can easily change `Dockerfile` to pin to an older LTS release until the incompatibility is resolved.
> 
> Also, if we track the active LTS release instead of current, then we will have at least 6 months to address compatibility issues with a new release before it becomes a problem for the Docker image.

Well yes, its not a Nodejs issue, but a Debian one. Quay.io uses clair and reported that there was 3 critical vulnerabilities + some ~100 others. The latest slim package still has a lot of vulberabilities, but there are fewer ones.

tommytran732:
The node-alpine container only has one small vulnerability with musl, and i tried building etherpad with it as well. It works, but then abiword was broken somehow. Thats why I made the PR for the slim image for now.

JohnMcLear:
Bit of a rant here but this is all so broken that we have to now learn about all this toolchain stuff that has nothing to do with our software at all.  Not constructive I appreciate....  /Rant

Dockerfile
tommytran732: I agree

tommytran732: Do you think we should add an apt update && apt upgrade -y in the Dockerfile as well?

rhansen: Updating the base image packages should be the responsibility of the Debian image maintainers, and the Node.js image maintainers should be regularly refreshing to get the latest Debian image. With that said, we're already running `apt-get update` so I'm OK with inserting an `apt-get -qq dist-upgrade` after that.

Don't forget to pass `--legacy-peer-deps` to `npm install` on line 88.

rhansen:
> It would be fantastic if we could ditch Debian for Alpine though...

I'm fine with that, assuming we can reasonably install LibreOffice and AbiWord on it. It would likely break downstream images, which I would rather not do. I don't know what best practice is for maintaining multiple variants.

rhansen:
@tommytran732 I rebased onto latest `develop` and modified your commit to pass `--legacy-peer-deps` to npm when installing plugins. I also pushed a new commit that upgrades the Debian packages. Unless I hear any objections, I'll merge this PR in day or so.

PR #5269, opened 2021-11-09
Put regression tests with the bugfix commit, mention bug in PR

rhansen:


JohnMcLear:
Looks good to me.  +1.  Approved.

webzwo0i (approved):

webzwo0i:
I don't have a real opinion here. Separate commit makes it easier to watch the test runners, but in the long-term, for doing auto-bisect to find (new and old) regressions, my plan is to have a copy of the test suites in a separate place. (Nothing will change for Etherpad core when it comes to testing/running the workflows, but it would make it possible to run the most up-to-date test suites against old Etherpad versions, including keeping a list of changes that need to be applied to the test suites so they can be run against older versions. There are limits to this approach, though, and I don't want to spend days on backporting tests, but I'm curious how far I can get.)

TLDR; if you feel like it saves you some time/make reviews easier - you are doing the majority of code work & reviews so I support anything that you think is helpful!

Regarding "Fix issue id" in commit messages, I kind of find this useful. How painful are the updates in an issue for you? Regarding problem with fork we could add a check that's parsing the commit messages and warns if it's not an absolute URL?

Anyway, feel free to merge this at any time with or without any changes.

rhansen:
> in the long-term, for doing auto-bisect to find (new and old) regressions, my plan is to have a copy of the test suites in a separate place.

Note that you can do `git checkout $COMMIT_ID -- src/tests` after a `git bisect start`, `git bisect good`, or `git bisect bad` command to precisely control the tests to run when bisecting.

> Regarding "Fix issue id" in commit messages, I kind of find this useful. How painful are the updates in an issue for you?

They're only mildly annoying; I softened the language a bit.

> Regarding problem with fork we could add a check that's parsing the commit messages and warns if it's not an absolute URL?

I don't think that's worth the effort.

> Anyway, feel free to merge this at any time with or without any changes.

Thanks for the feedback!

PR #5276, opened 2021-11-12
Delete session after corresponding group2session and author2session entries

timonegk:
Currently, when a session is deleted, the session itself is removed before the session entries are removed from `group2sessions` and `author2sessions`. This leads to an unclean state when the application is killed or an error occurs after the session was deleted but before all references are deleted because sessions that do not exist are still referenced. This pull request changes the order to delete the references first.

### DIFF
--- a/src/node/db/GroupManager.js
+++ b/src/node/db/GroupManager.js
@@ -55,9 +55,8 @@ exports.deleteGroup = async (groupID) => {
   // loop through all sessions and delete them (in parallel)
   await Promise.all(Object.keys(sessions).map((session) => sessionManager.deleteSession(session)));
 
-  // remove group and group2sessions entry
+  // remove group2sessions entry
   await db.remove(`group2sessions:${groupID}`);
-  await db.remove(`group:${groupID}`);
 
   // unlist the group
   let groups = await exports.listAllGroups();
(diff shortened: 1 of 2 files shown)


webzwo0i:
These kind of problems may exist in other places, too.

timonegk:
I found a similar situation in `deleteGroup`. There, the `group2sessions` entry is deleted before the group, but after deleting the group, it is removed from the list of all groups. I guess it would be better to change the order, I added a commit doing that. Overall, I think that this situation is less relevant than the session deletion as group deletion definitely happens less often than session deletion.
At first glance, the only other place where a similar problem could exist is in `Pad.remove`, however `Promise.all` is used there to perform the deletion of all references to the pad simultaneously.

rhansen:
There are also some insidious race conditions: multiple `createSession()` and `deleteSession()` calls can interleave, resulting in corruption in the `author2sessions` and/or `group2sessions` records. For example:

1. `deleteSession()` reads `group2sessions` or `author2sessions` from the database and puts the value in a local variable.
2. While `deleteSession()` is `awaiting` something, `createSession()` adds entries to the `group2sessions` and `author2sessions` records and saves to the database.
3. `deleteSession()` resumes execution after the `await`, modifies the local variable, and writes it back to the database. This effectively undoes the changes written in step 2.

We need to ensure that the whole read-modify-write sequence is atomic.

rhansen:
@timonegk I pushed a couple of commits to your branch that should fix the races. Would you mind testing them?

timonegk:
I tested the changes locally and everything worked as expected. I didn't test the race conditions explicitly, though. (So I didn't try to stop the server while it is deleting stuff or something like that.) But also from the code everything you changed looks alright in my opinion (and cleaner than before).
By the way, I noticed that `mapper2group` relations don't get cleaned up when a group is deleted.

PR #5288, opened 2021-11-20
New attributes processing library

rhansen:
Multiple commits; please do not squash.

The existing code to manipulate attributes was awkward. This PR introduces a new AttributeMap class that subclasses Map and has easy conversion to/from attribute strings.

I think we should bump the Etherpad minor version to 1.9.0 so that plugins can easily depend on the new class.

This PR is some yak shaving required to move import processing to a worker thread. (cc @packardone)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@
   (low-level API) and `ep_etherpad-lite/static/js/AttributeMap` (high-level
   API).
 
+### Compatibility changes
+
+#### For plugin authors
+
+* Changes to the `src/static/js/Changeset.js` library:
+  * The following attribute processing functions are deprecated (use the new
+    attribute APIs instead):
+    * `attribsAttributeValue()`
+    * `eachAttribNumber()`
+    * `makeAttribsString()`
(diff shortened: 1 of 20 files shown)


JohnMcLear:
I guess one thing worth doing is testing this against plugins such as ep_headings2 and ep_comments because we need to make sure backwards compat isn't broken for those plugins.  This is a huge improvement in implementation and very welcome!

Also worth noting there may be additional docs somewhere (possibly in the Wiki) that might need modifying once this lands.

+1 minor version release.

rhansen:
> I guess one thing worth doing is testing this against plugins such as ep_headings2 and ep_comments because we need to make sure backwards compat isn't broken for those plugins.

Those are included in the "Frontend tests powered by Sauce Labs / with plugins" tests. I also run ep_headings2 on my own instance and it works fine with this change.

> Also worth noting there may be additional docs somewhere (possibly in the Wiki) that might need modifying once this lands.

I looked at the wiki docs and they're quite old. Probably best to just replace them with links to the documentation comments in the source code.

PR #5233, opened 2021-10-17
WIP: Changeset: Refactor to improve API, and other cleanups

rhansen:
DO NOT MERGE: This will be split into multiple smaller pull requests to make things easier to review.

The code in `Changeset.js` had lots of readability problems so I spent some time trying to clean it up. There shouldn't be any behavior change.

cc @webzwo0i

PR #5295, opened 2021-11-22
Changeset: Replace `appendATextToAssembler()` with a generator

rhansen:
This is the next commit from PR #5233.


JohnMcLear:
approved

PR #5298, opened 2021-11-22
Changeset: `class`-ify some objects to improve readability

rhansen:
Turn some functions that return objects into real ES6 `class`es to improve readability.

This is the next batch of commits from PR #5233.


JohnMcLear:
Approved

PR #5305, opened 2021-11-28
GroupManager: Clean up any mappings when deleting a group

rhansen:
Thanks @timonegk for pointing out the oversight in https://github.com/ether/etherpad-lite/pull/5276#issuecomment-975600565.


JohnMcLear:
Thanks!

PR #5326, opened 2021-12-12
Pad: Don't create no-op revisions

rhansen:
Also:

* Accept `USER_CHANGES` retransmissions. (Currently the client doesn't ever retransmit, but accepting retransmissions removes an awkward exception in the server code and opens the door to future protocol improvements.)
* Fix race conditions in the `setText`, `appendText`, `restoreRevision` HTTP API functions.

### DIFF
--- a/src/node/db/Pad.js
+++ b/src/node/db/Pad.js
@@ -82,6 +82,9 @@ Pad.prototype.appendRevision = async function (aChangeset, author) {
   }
 
   const newAText = Changeset.applyToAText(aChangeset, this.atext, this.pool);
+  if (newAText.text === this.atext.text && newAText.attribs === this.atext.attribs) {
+    return this.head;
+  }
   Changeset.copyAText(newAText, this.atext);
 
   const newRev = ++this.head;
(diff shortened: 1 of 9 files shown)


src/node/db/Pad.js
```diff
@@ -82,6 +82,7 @@ Pad.prototype.appendRevision = async function (aChangeset, author) {
   }
 
   const newAText = Changeset.applyToAText(aChangeset, this.atext, this.pool);
+  if (newAText.text === this.atext.text && newAText.attribs === this.atext.attribs) return;
```
webzwo0i, line 85: Maybe better
`if (Changeset.isIdentity(aChangeset)) return;` instead? When there are no attribs and there are no ops, the atext won't change.

And also, is this code also run when submitting changesets (before distributing it to connected clients)? If yes, is it save to not accept the changeset using `return`, or will this lead to an endless loop because the client will try to submit the changeset over and over again and we should better throw an exception?

rhansen: `Changeset.isIdentity()` won't catch cases where the changeset isn't identity but doesn't actually change anything (e.g., delete some chars, then add them back with the same attributes).

I'm not 100% certain that clients that send a no-op changeset won't keep retrying or chronically wait for a response that won't come back. I'll do some more testing.

rhansen: I changed it to send an `ACCEPT_COMMIT` even if there is no net change, and I added some tests.

PR #5356, opened 2022-01-14
Uses ssl by default

mikesol:
This is a small patch I've used in my instance for a couple months now that makes connection to certain mysql instances possible. Thanks in advance for your consideration.

### DIFF
--- a/settings.json.docker
+++ b/settings.json.docker
@@ -206,6 +206,7 @@
    */
 
   "dbType": "${DB_TYPE:dirty}",
+  "dbUseSSL": "${DB_TYPE:false}",
   "dbSettings": {
     "host":     "${DB_HOST:undefined}",
     "port":     "${DB_PORT:undefined}",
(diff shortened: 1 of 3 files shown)


rhansen:
It looks like the `ssl` object is specific to the `mysql` driver, so it doesn't really make sense to add it this way.

If the default CA certificates are acceptable, I think users can use an empty object in their `settings.json` like this:

```json
  "dbType": "mysql",
  "dbSettings": {
    "ssl": {},
    "user": "etherpad",
    "host" "localhost",
    "database": "etherpad"
  },
```

If support for custom CA certificates is required, then it's probably best to add support in the [`mysql` ueberdb2 driver](https://github.com/ether/ueberDB/blob/4feff0c64966b7aa0537bf7336c2c230645a5d3a/databases/mysql_db.js#L23).

mikesol:
That makes sense - I'll close this PR and add it there when I have time!

PR #5405, opened 2022-02-05
Update http_api.md

Alienmaster:
The return Values of createGroupPad for success and error are not correct
grafik

Information about public status of pads is extended


rhansen:
Thank you!

PR #5429, opened 2022-02-23
PadMessageHandler: Don't trust user-provided `padId`

rhansen:

PR #5430, opened 2022-02-23
PadMessageHandler: prohibit reading of message.data.padId

webzwo0i:
see https://github.com/ether/etherpad-lite/pull/5429

### DIFF
--- a/src/node/handler/PadMessageHandler.js
+++ b/src/node/handler/PadMessageHandler.js
@@ -250,6 +250,12 @@ exports.handleMessage = async (socket, message) => {
     throw new Error('message.padId must not be accessed (for security reasons)');
   }});
 
+  if (message.data) {
+    Object.defineProperty(message.data, 'padId', {get: () => {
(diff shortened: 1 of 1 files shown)


webzwo0i:
This will break plugins, so we should add a note. Is a warning in CHANGELOG appropriated? Or do we want to switch to checking if CLIENT_READY's padId/what's stored in sessionInfo matches subsequent padIds until plugins are updated?
A third option could be to introduce a new setting that would allow reading, because at least public instances (without any read/write/create restrictions; those that use pad ids as secret) are safe.

rhansen:
As far as I can tell, there is no core API that sets `message.data.padId`. Is that correct? If so, I'm hesitant to put something like this in Etherpad core. Plugins can introduce all sorts of security vulnerabilities, and it's not really feasible for Etherpad core to defend against them all.

I think it would be better to pass the pad ID(s) in the hook context so that plugin authors aren't tempted to pass the pad ID in a message.

PR #5474, opened 2022-03-19
Docker: use buildx to build amd64 and arm64 images

Firefishy:
Add github action tasks to build for `amd64` and `arm64` platform.

`arm64` is now widely used: Raspberry PI, Docker for Mac (on Apple Silicon) and AWS Graviton.

Tested a docker building locally using the Dockerfile on an `arm64` system without issue.
Don't expect any issues with the same on Github Actions

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -16,6 +16,10 @@ jobs:
         name: Check out
         uses: actions/checkout@v3
       -
+        name: Set up QEMU
+        if: github.event_name == 'push'
+        uses: docker/setup-qemu-action@v1
+      -
         name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
       -
@@ -78,6 +82,7 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           context: .
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}


Firefishy:
I will look into the build failure.

.github/workflows/docker.yml
```diff
@@ -15,14 +15,21 @@ jobs:
       -
         name: Check out
         uses: actions/checkout@v3
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
       -
         name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
```
rhansen, line 23:
```suggestion
        uses: docker/setup-buildx-action@v1
        id: buildx
```

.github/workflows/docker.yml
```diff
@@ -15,14 +15,21 @@ jobs:
...
+      -
+        name: Available platforms
+        run: echo ${{ steps.buildx.outputs.platforms }}
       -
         name: Build and export to Docker
         uses: docker/build-push-action@v2
         with:
           context: .
+          platforms: linux/amd64,linux/arm64
```
rhansen, line 32: Build fails with:

>  Error: buildx failed with: error: docker exporter does not currently support exporting manifest lists

rhansen: See https://github.com/docker/buildx/issues/59. Looks like this can be fixed by doing the following:

  1. Build without `platforms` and with `load`.
  2. Test.
  3. Set up QEMU.
  4. Build again, this time with `platforms`, without `load`, and with `push`.

In other words, remove the first `platforms` line and move QEMU setup to just before the second build.

.github/workflows/docker.yml
Firefishy: I have removed the output block, as the QEMU step has the output already.

.github/workflows/docker.yml
Firefishy: Done ;-)

.github/workflows/docker.yml
```diff
@@ -23,6 +26,7 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           context: .
+          platforms: linux/amd64
```
rhansen, line 29: Nit: Remove this line so that we build and test for the runner's platform (which could theoretically change from amd64 to arm64 or something else in the future).

```suggestion
```

.github/workflows/docker.yml
```diff
@@ -15,6 +15,9 @@ jobs:
       -
         name: Check out
         uses: actions/checkout@v3
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
```
rhansen, line 20: Nit: I don't know how expensive this step is, so I'd prefer to avoid running it if tests are going to fail. (Please move it down to just before the second build step so that it doesn't run if tests fail.)

rhansen: Also add:

```yaml
        if: github.event_name == 'push'
```

because there's no point in setting up QEMU if it's not going to be used.

Firefishy:
@rhansen changes made as requested.

rhansen:
Thanks!

PR #5501, opened 2022-04-08
allow option to make pad names case-insensitive

### LINKED ISSUE #3844: Feature Request: case-insensitive pad names
It would be great if a setting would **allow the use of case-insensitive pad names** (`overview` -> `oVErvieW`). Having etherpad deployed in a German environment where nouns are written capitalized, this lead to some confusion of users (some capitalize the pad name, others dont).

Currently, i work around that with some nasty nginx-rule (because not all request should be lowercased, some javascript assets use capital letters).

I am not a JavaScript-person and do not know the EP codebase but would assume that this is a rather low-hanging fruit (but then I don't know if the code is structured such that plugins etc just use the API which would respect the lowercase-setting).

Probably related:
https://github.com/ether/etherpad-lite/issues/2557

DanielHabenicht:
fixes #3844

Adds Option to:
```
lowerCasePadIds = true // default is false
```

### DIFF
--- a/settings.json.docker
+++ b/settings.json.docker
@@ -639,5 +639,5 @@
   /*
    * Enable/Disable case-insensitive pad names.
    */
-  "enforceLowerCasePadIds": "${ENFORCE_LOWER_CASE_PAD_IDS:false}"
+  "lowerCasePadIds": "${LOWER_CASE_PAD_IDS:false}"
 }
--- a/settings.json.docker
+++ b/settings.json.docker
@@ -634,5 +634,10 @@
   "customLocaleStrings": {},
 
   /* Disable Admin UI tests */
-  "enableAdminUITests": false
+  "enableAdminUITests": false,
+
+  /*
+   * Enable/Disable case-insensitive pad names.
+   */
+  "enforceLowerCasePadIds": "${ENFORCE_LOWER_CASE_PAD_IDS:false}"
 }
--- a/src/node/db/PadManager.js
+++ b/src/node/db/PadManager.js
@@ -22,6 +22,7 @@
 const CustomError = require('../utils/customError');
 const Pad = require('../db/Pad');
 const db = require('./DB');
+const settings = require('../utils/Settings');
 
 /**
  * A cache of all loaded Pads.
@@ -169,6 +170,8 @@ exports.sanitizePadId = async (padId) => {
 
     padId = padId.replace(from, to);
   }
+
+  if (settings.enforceLowerCasePadIds) padId = padId.toLowerCase();
 
   // we're out of possible transformations, so just return it
   return padId;
--- a/src/node/db/PadManager.js
+++ b/src/node/db/PadManager.js
@@ -171,7 +171,7 @@ exports.sanitizePadId = async (padId) => {
     padId = padId.replace(from, to);
   }
 
-  if (settings.enforceLowerCasePadIds) padId = padId.toLowerCase();
+  if (settings.lowerCasePadIds) padId = padId.toLowerCase();
 
   // we're out of possible transformations, so just return it
   return padId;
--- a/src/node/utils/Settings.js
+++ b/src/node/utils/Settings.js
@@ -434,7 +434,7 @@ exports.enableAdminUITests = false;
  * Enable auto conversion of pad Ids to lowercase.
  * e.g. /p/EtHeRpAd to /p/etherpad
  */
-exports.enforceLowerCasePadIds = false;
+exports.lowerCasePadIds = false;
 
 // checks if abiword is avaiable
 exports.abiwordAvailable = () => {
--- a/src/node/utils/Settings.js
+++ b/src/node/utils/Settings.js
@@ -430,6 +430,11 @@ exports.importMaxFileSize = 50 * 1024 * 1024;
  */
 exports.enableAdminUITests = false;
 
+/*
+ * Enable auto conversion of pad Ids to lowercase.
+ * e.g. /p/EtHeRpAd to /p/ehtherpad
+ */
+exports.enforceLowerCasePadIds = false;
 
 // checks if abiword is avaiable
 exports.abiwordAvailable = () => {
--- a/src/node/utils/Settings.js
+++ b/src/node/utils/Settings.js
@@ -432,7 +432,7 @@ exports.enableAdminUITests = false;
 
 /*
  * Enable auto conversion of pad Ids to lowercase.
- * e.g. /p/EtHeRpAd to /p/ehtherpad
+ * e.g. /p/EtHeRpAd to /p/etherpad
  */
 exports.enforceLowerCasePadIds = false;
 
--- a/src/tests/backend/specs/enforceLowerCasePadIds.js
+++ b/src/tests/backend/specs/enforceLowerCasePadIds.js
@@ -6,7 +6,6 @@ const padManager = require('../../../node/db/PadManager');
 const settings = require('../../../node/utils/Settings');
 
 describe(__filename, function () {
-  this.timeout(30000);
   let agent;
   const cleanUpPads = async () => {
     const padIds = ['UPPERCASEpad', 'uppercasepad', 'ALREADYexistingPad', 'alreadyexistingpad'];
--- a/src/tests/backend/specs/enforceLowerCasePadIds.js
+++ b/src/tests/backend/specs/enforceLowerCasePadIds.js
@@ -1,0 +1,69 @@
+'use strict';
+
+const assert = require('assert').strict;
+const common = require('../common');
+const padManager = require('../../../node/db/PadManager');
+const settings = require('../../../node/utils/Settings');
+
+describe(__filename, function () {
+  this.timeout(30000);
+  let agent;
+  const cleanUpPads = async () => {
+    const padIds = ['UPPERCASEpad', 'uppercasepad', 'ALREADYexistingPad', 'alreadyexistingpad'];
+    await Promise.all(padIds.map(async (padId) => {
+      if (await padManager.doesPadExist(padId)) {
+        const pad = await padManager.getPad(padId);
+        await pad.remove();
+      }
+    }));
+  };
+
+  before(async function () { agent = await common.init(); });
+  beforeEach(async function () {
+    await cleanUpPads();
+  });
+  afterEach(async function () {
+    await cleanUpPads();
+  });
+
+  describe('not activated', function () {
+    Object.assign(settings, {
+      enforceLowerCasePadIds: false,
+    });
+    it('- do nothing', async function () {
+      const res = await agent.get('/p/UPPERCASEpad');
+      assert.equal(res.status, 200);
+    });
+
+    it('- do nothingg', async function () {
+      await agent.get('/p/UPPERCASEpad')
+          .expect(200);
+    });
+  });
+
+  describe('activated', function () {
+    it('- lowercase pad ids', async function () {
+      Object.assign(settings, {
+        enforceLowerCasePadIds: true,
+      });
+      await agent.get('/p/UPPERCASEpad')
+          .expect(302)
+          .expect('location', 'uppercasepad');
+    });
+
+    it('- keeps old pads accessible', async function () {
+      Object.assign(settings, {
+        enforceLowerCasePadIds: false,
+      });
+      const pad = await padManager.getPad('ALREADYexistingPad', 'alreadyexistingpad');
+      await padManager.getPad('ALREADYexistingPad', 'bla');
+      assert.equal(pad.text(), 'alreadyexistingpad\n');
+      Object.assign(settings, {
+        enforceLowerCasePadIds: true,
+      });
+
+      const newpad = await padManager.getPad('alreadyexistingpad', 'testcontent');
+      assert.equal(newpad.text(), 'testcontent\n');
+    });
+  });
+});
--- a/settings.json.template
+++ b/settings.json.template
@@ -640,5 +640,5 @@
   /*
    * Enable/Disable case-insensitive pad names.
    */
-  "enforceLowerCasePadIds": false
+  "lowerCasePadIds": false
 }
--- a/settings.json.template
+++ b/settings.json.template
@@ -635,5 +635,10 @@
   "customLocaleStrings": {},
 
   /* Disable Admin UI tests */
-  "enableAdminUITests": false
+  "enableAdminUITests": false,
+
+  /*
+   * Enable/Disable case-insensitive pad names.
+   */
+  "enforceLowerCasePadIds": false
 }
--- a/src/node/handler/PadMessageHandler.js
+++ b/src/node/handler/PadMessageHandler.js
@@ -236,6 +236,11 @@ exports.handleMessage = async (socket, message) => {
       padID: message.padId,
       token: message.token,
     };
+
+    // Pad does not exist, so we need to sanitize the id
+    if (!(await padManager.doesPadExist(thisSession.auth.padID))) {
+      thisSession.auth.padID = await padManager.sanitizePadId(thisSession.auth.padID);
+    }
     const padIds = await readOnlyManager.getIds(thisSession.auth.padID);
     thisSession.padId = padIds.padId;
     thisSession.readOnlyPadId = padIds.readOnlyPadId;
--- a/src/package-lock.json
+++ b/src/package-lock.json
@@ -4,6 +4,12 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@aashutoshrathi/word-wrap": {
+      "version": "1.2.6",
+      "resolved": "https://registry.npmjs.org/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz",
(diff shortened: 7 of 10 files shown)


rhansen:
I hesitate to merge this because it makes it impossible to access existing pads that have upper-case letters. What do you think about checking if the pad exists with the name as-is, and if not then redirect the user to the lower-case name?

Also, is the default Unicode case conversion algorithm sufficient? Or should this use `.toLocaleLowerCase()`?

DanielHabenicht:
> I hesitate to merge this because it makes it impossible to access existing pads that have upper-case letters. What do you think about checking if the pad exists with the name as-is, and if not then redirect the user to the lower-case name?


I could implement such a check if you want, but I would argue against it. As it would introduce inconsistency from the user and admin perspective. Because whoever choose to use lowercase only should be aware that turning it on won't be backwards compatible anyways. (This is also why I made it configurable)
I will add a comment highlighting this. 

> Also, is the default Unicode case conversion algorithm sufficient? Or should this use `.toLocaleLowerCase()`?

The whole code base [does not use `toLocaleLowerCase`](https://cs.github.com/ether/etherpad-lite?q=toLocaleLowerCase), but I am happy to change it, as it should not have any drawbacks.

DanielHabenicht:
Do the Tests normally fail? Or should I add anything to make them work?

rhansen:
> Scenario: Two pads that were created before (`EtHeRpaD` and `etherpad`). Which one should we choose?

The idea is that if the user visits `EtHeRpaD` then they will get `EtHeRpaD` because it already exists. If they use any different capitalization, they will be redirected to `etherpad`.

> Also one of them will be hidden forever. The only solution to it is migrating it before.

With your change as-is `EtHeRpaD` is hidden forever. With my suggested proposal users will still be able to visit `EtHeRpaD` as long as they capitalize it exactly.

In addition to hiding existing pads with upper-case letters, I'm worried about finding and adjusting all of the other code paths that might not hit the new `.toLowerCase()` call. For example, without additional changes it would still be possible for the [`createPad` HTTP API](https://etherpad.org/doc/v1.8.17/#index_createpadpadid--text) to create a pad with upper-case letters. Plugins would also be able to create pads with upper-case letters. Keeping Etherpad case-sensitive but redirecting to the lower-case name when creating a new pad avoids unintended issues.

> Do the Tests normally fail? Or should I add anything to make them work?

The tests are failing because `settings.json.template` is simply renamed to `settings.json` before the tests run. The template has `"caseInsensitivePadNames": true`, so any tests with upper-case letters will fail. Change it to `false` and it should be OK.

DanielHabenicht:
Ok, I will update this pr with your proposed changes.

DanielHabenicht:
Is there a way to writing tests with specific settings turned on? (e.g. the caseInsensitivePadeNames)?

DanielHabenicht:
@rhansen read from my side, except for the question above.

rhansen:
I rebased your PR onto latest `develop` and squashed the commits together.

> Is there a way to writing tests with specific settings turned on? (e.g. the caseInsensitivePadeNames)?

Yes; see [here](https://github.com/ether/etherpad-lite/blob/088bb1279766e94e10c6cdbbf770caa99353a62c/src/tests/backend/specs/socketio.js#L34) for an example. In order for test changes to have an effect you'll need to change:

```javascript
const {caseInsensitivePadNames} = require('../utils/Settings');
// ...
  if (caseInsensitivePadNames) {
```

to:

```javascript
const settings = require('../utils/Settings');
// ...
  if (settings.caseInsensitivePadNames) {
```

src/node/db/PadManager.js
```diff
@@ -170,6 +171,11 @@ exports.sanitizePadId = async (padId) => {
     padId = padId.replace(from, to);
   }
 
+  // If caseInsensitivePadNames is enabled, we need to make sure the padId is lowercase
```
rhansen, line 174: This comment is unnecessary; the code is easy enough to read.

```suggestion
```

src/node/db/PadManager.js
```diff
@@ -170,6 +171,11 @@ exports.sanitizePadId = async (padId) => {
...
+  if (caseInsensitivePadNames) {
+    padId = padId.toLowerCase();
+  }
```
rhansen, line 177: style nit (I know that most of the other code doesn't follow this suggested style, but I'm working on changing it bit by bit):

```suggestion
  if (caseInsensitivePadNames) padId = padId.toLowerCase();
```

src/node/utils/Settings.js
```diff
@@ -430,6 +430,11 @@ exports.importMaxFileSize = 50 * 1024 * 1024;
  */
 exports.enableAdminUITests = false;
 
+/*
+ * Enable case-insensitive pad names.
+ */
+exports.caseInsensitivePadNames = false;
+
```
rhansen, line 437: Superfluous newline:

```suggestion
```

src/node/utils/Settings.js
```diff
@@ -430,6 +430,11 @@ exports.importMaxFileSize = 50 * 1024 * 1024;
  */
 exports.enableAdminUITests = false;
 
+/*
+ * Enable case-insensitive pad names.
+ */
```
rhansen, line 435: I don't think these comments are all that useful; to me they just add noise. (We don't generate documentation from them or anything like that.)

```suggestion
```

src/node/utils/Settings.js
```diff
@@ -430,6 +430,11 @@ exports.importMaxFileSize = 50 * 1024 * 1024;
...
+exports.caseInsensitivePadNames = false;
```
rhansen, line 436: I wonder if renaming this to something like `lowerCasePadNames` would be better. Technically it's not case-insensitive—if pad `FooBar` exists and the user visits `Foobar` then they'll create a new `foobar` pad, not visit the existing `FooBar` pad.

rhansen (commented):
Let me know if you would like help adding tests.

src/node/utils/Settings.js
DanielHabenicht: I also thought about that, I will update it accordingly!

rhansen:
Also: I ran the front-end tests locally (http://localhost:9001/tests/frontend/) with the new option turned on and some of the tests failed. Can you look into that? (The front-end tests aren't run in a GitHub pull request workflow because the automated tests require Sauce Labs credentials.)

src/node/utils/Settings.js
rhansen: Actually, I wonder if `lowerCasePadIds` is better than `lowerCasePadNames`. I think we use "pad ID" more than "pad name" in both user-visible stuff and in source code.

src/node/utils/Settings.js
DanielHabenicht: Hmm, all other things have comments and I think it makes it easier to grasp what the settings is doing. I also added an example, that's better? Otherwise those settings are completely undocumented.

src/node/db/PadManager.js
DanielHabenicht: applied in 68fa4f19c45f23c01c8f6f0c54793f78cd1ce930

src/node/db/PadManager.js
DanielHabenicht: applied in 68fa4f19c45f23c01c8f6f0c54793f78cd1ce930

src/node/utils/Settings.js
DanielHabenicht: applied in 68fa4f19c45f23c01c8f6f0c54793f78cd1ce930

DanielHabenicht:
The failing tests mostly have to do with these lines: (note e.g. `FRONTED_TESTS_OxmGelclDdwtqszpderu`)
https://github.com/ether/etherpad-lite/blob/64757d1636299c743840ef9d1f5a8d507cb385f1/src/tests/frontend/helper.js#L85-L100
as with the setting on they are converted to lowercase.

settings.json.docker
```diff
@@ -634,5 +634,10 @@
   "customLocaleStrings": {},
 
   /* Disable Admin UI tests */
-  "enableAdminUITests": false
+  "enableAdminUITests": false,
+
+  /*
+   * Enable/Disable case-insensitive pad names.
+   */
+  "enforceLowerCasePadIds": "${ENFORCE_LOWER_CASE_PAD_IDS:false}"
```
rhansen, line 642: nit: The word "enforce" might be too strong—it implies that it is impossible to have upper-case characters.

(I wish that naming things wasn't one the hardest problems in software engineering. :slightly_smiling_face:)

src/node/utils/Settings.js
```diff
@@ -430,6 +430,11 @@ exports.importMaxFileSize = 50 * 1024 * 1024;
  */
 exports.enableAdminUITests = false;
 
+/*
+ * Enable auto conversion of pad Ids to lowercase.
+ * e.g. /p/EtHeRpAd to /p/ehtherpad
```
rhansen, line 435: Typo:

```suggestion
 * e.g. /p/EtHeRpAd to /p/etherpad
```

src/tests/backend/specs/enforceLowerCasePadIds.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+const assert = require('assert').strict;
+const common = require('../common');
+const padManager = require('../../../node/db/PadManager');
+const settings = require('../../../node/utils/Settings');
+
+describe(__filename, function () {
+  this.timeout(30000);
```
rhansen, line 9: Do the tests time out if this is removed? If not, I prefer to leave out the timeout call. If so, I'd prefer to move the timeout to the specific test case or hook that causes the timeout.

src/tests/backend/specs/enforceLowerCasePadIds.js
```diff
@@ -0,0 +1,69 @@
...
+    await cleanUpPads();
+  });
+  afterEach(async function () {
+    await cleanUpPads();
+  });
+
+  describe('not activated', function () {
+    Object.assign(settings, {
+      enforceLowerCasePadIds: false,
+    });
```
rhansen, line 32: This must happen inside a `before` or `beforeEach` hook, and the original value should be restored in an `after` or `afterEach` hook:

```suggestion
    let backup;

    before(async () => {
      backup = settings.enforceLowerCasePadIds;
      settings.enforceLowerCasePadIds = false;
    });

    after(async () => {
      settings.enforceLowerCasePadIds = backup;
    });

```

src/tests/backend/specs/enforceLowerCasePadIds.js
```diff
@@ -0,0 +1,69 @@
...
+      await agent.get('/p/UPPERCASEpad')
+          .expect(200);
+    });
+  });
+
+  describe('activated', function () {
+    it('- lowercase pad ids', async function () {
+      Object.assign(settings, {
+        enforceLowerCasePadIds: true,
+      });
```
rhansen, line 48: Move to a `before` or `beforeEach` hook, and restore the setting in an `after` or `afterEach` hook.

src/tests/backend/specs/enforceLowerCasePadIds.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+const assert = require('assert').strict;
+const common = require('../common');
+const padManager = require('../../../node/db/PadManager');
+const settings = require('../../../node/utils/Settings');
+
+describe(__filename, function () {
```
rhansen, line 8: More tests that I would like to see added:

  * Users should be unable to manually connect via socket.io and create a pad with upper-case letters. If they can, then a malicious user would be able to hijack a URL containing upper-case letters. (For example: User A creates a pad at `https://example.com/p/foobar` but shares `https://example.com/p/FooBar` in an announcement because it's easier to read. User B uses socket.io to create a pad named `FooBar` with malicious content, thus hijacking users that use the `FooBar` spelling.
  * Users should be able to connect (via socket.io) to an existing pad with upper-case letters.

See:

  * [`common.connect()`](https://github.com/ether/etherpad-lite/blob/074482afea74d7788228fa50fca0662e4d9962fe/src/tests/backend/common.js#L132-L167)
  * [`common.handshake()`](https://github.com/ether/etherpad-lite/blob/074482afea74d7788228fa50fca0662e4d9962fe/src/tests/backend/common.js#L169-L189)
  * [example use](https://github.com/ether/etherpad-lite/blob/074482afea74d7788228fa50fca0662e4d9962fe/src/tests/backend/specs/socketio.js#L60-L62)
  * [`securityManager.checkAccess()`](https://github.com/ether/etherpad-lite/blob/074482afea74d7788228fa50fca0662e4d9962fe/src/node/db/SecurityManager.js#L53)

src/tests/backend/specs/enforceLowerCasePadIds.js
```diff
@@ -0,0 +1,69 @@
...
+    });
+    it('- do nothing', async function () {
+      const res = await agent.get('/p/UPPERCASEpad');
+      assert.equal(res.status, 200);
+    });
+
+    it('- do nothingg', async function () {
+      await agent.get('/p/UPPERCASEpad')
+          .expect(200);
+    });
```
rhansen, line 41: These two are equivalent; delete one of them.

```suggestion
    it('- do nothing', async function () {
      await agent.get('/p/UPPERCASEpad')
          .expect(200);
    });
```

rhansen (commented):
Thanks for adding tests! I rebased onto latest `develop`, squashed the commits, and pushed some additional fixes.

settings.json.docker
DanielHabenicht: yes, lets just go with `lowerCasePadIds`

DanielHabenicht: see f91b8dbfcb7c00199fc6e94fd98a1d6a043bd505

src/tests/backend/specs/enforceLowerCasePadIds.js
DanielHabenicht: no this seems to be a copy and paste error, see 21d3a1508a57be98cf26767d5b30ca25d9b18115

src/tests/backend/specs/enforceLowerCasePadIds.js
DanielHabenicht: applied: e3a9eb75e7d4093ed0f8083201f2068f789f6b16

src/tests/backend/specs/enforceLowerCasePadIds.js
DanielHabenicht: applied: e3a9eb75e7d4093ed0f8083201f2068f789f6b16

src/tests/backend/specs/enforceLowerCasePadIds.js
DanielHabenicht: applied: e3a9eb75e7d4093ed0f8083201f2068f789f6b16

src/tests/backend/specs/enforceLowerCasePadIds.js
DanielHabenicht: > Users should be unable to manually connect via socket.io and create a pad with upper-case letters. If they can, then a malicious user would be able to hijack a URL containing upper-case letters. (For example: User A creates a pad at https://example.com/p/foobar but shares https://example.com/p/FooBar in an announcement because it's easier to read. User B uses socket.io to create a pad named FooBar with malicious content, thus hijacking users that use the FooBar spelling.

That's a new feature I suppose I should implement?

DanielHabenicht: The question being: Why is the socket padId not being santized?

DanielHabenicht: hi! Just a friendly reminder @rhansen

rhansen: Apologies for taking so long to get back to you, I got caught up with other stuff.

> The question being: Why is the socket padId not being santized?

Under normal circumstances, the pad ID sent over websocket is the pad ID seen by the browser, which is the sanitized ID (because the browser is redirected to the sanitized ID before it connects over websocket). However, due to an oversight, the server doesn't enforce sanitized IDs—it assumes the client behaves properly, which is dangerous programming practice.

Without this feature, it doesn't really matter whether the server enforces sanitized IDs because it's relatively difficult for a malicious client to use the permissiveness to its advantage. This feature changes the balance—the missing check means it is now easy for a malicious client to hijack any URL with a capital letter.

DanielHabenicht: hi @rhansen I finally got time to finish this PR. I implemented the behaviour in https://github.com/ether/etherpad-lite/pull/5501/commits/8ceb66f3eab0e0ff7a8f2d74c49da7d28bd6575e

DanielHabenicht:
@rhansen Any updates on this PR? Hit me up if something is still missing.

DanielHabenicht:
Nice that the bot added the `stale` label, but maybe somebody can review this PR? @rhansen thanks!

JohnMcLear:
@dependabot rebase

SamTV12345:
I just rebased your commits to the latest branch and tested the feature. It's working as expected. Older pads are even accessible even if I enable the option. Newer pads on the other hand are relying on this restriction and test123 and TEST123 are redirected to the same pad. Thanks for the contribution. I'll include it in the next release as it is an optional feature that even if activated doesn't break any logic.

PR #5589, opened 2022-08-26
deps: ueberDB 2.2.4 -> 4.0.1

webzwo0i:
deps: rebuild package-lock.json

PR so it runs the frontend tests which are disabled for dependabot


webzwo0i:
we also need to bump node's minimal version to v14.15.0

JohnMcLear:
@dependabot rebase

PR #5599, opened 2022-09-21
Fix document not scrolling in iOS (all browsers)

ZenulAbidin:
This patches a bug where users cannot scroll a document on iOS, because iOS does not allow iframes to be scrolled.

See https://davidwalsh.name/scroll-iframes-ios for details.

This is purely a cosmetic change, so a frontend test is not applicable for it.

Tested on iOS 12

### DIFF
--- a/src/static/skins/colibris/src/layout.css
+++ b/src/static/skins/colibris/src/layout.css
@@ -1,3 +1,18 @@
+.scroll-wrapper { 
+  display: flex;
+  min-width: 100vw;
+}
+
+@supports (-webkit-touch-callout: none)  {
+  .scroll-wrapper {
+    -webkit-overflow-scrolling: touch;
+    overflow-y: scroll;
+  }                      
+                       
+  .scroll-wrapper iframe {}
+                      
+}
+
 #outerdocbody {
   margin: 0 auto;
   padding-top: 20px;
--- a/src/templates/pad.html
+++ b/src/templates/pad.html
@@ -80,7 +80,9 @@
 
   <% e.begin_block("afterEditbar"); %><% e.end_block(); %>
 
-  <div id="editorcontainerbox" class="flex-layout">
+      <div class="scroll-wrapper">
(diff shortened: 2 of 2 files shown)


webzwo0i:
Thank you!
Did you test if the problem is also present in the timeslider?

ZenulAbidin:
> Thank you! Did you test if the problem is also present in the timeslider?

The issue doesn't appear to be present in the timeslider on any browser.

JohnMcLear:
I imagine this will break some plugins..

Afaik some plugins reference #editorcontainer > #editorcontainerbox and wrapping this will cause that selector to no longer function.

I am not 100% certain but I imagine this PR is going to cause issues so I wont merge until I'm sure.

JohnMcLear:
@dependabot rebase

SamTV12345:
I'm not sure if Dependabot also rebases non Dependabot pull requests 😅

ZenulAbidin:
> I'm not sure if Dependabot also rebases non Dependabot pull requests 😅

Why would it need to rebase this? It's only one commit.

SamTV12345:
> > I'm not sure if Dependabot also rebases non Dependabot pull requests 😅
> 
> Why would it need to rebase this? It's only one commit.

There is a good explanation [here](https://docs.github.com/de/get-started/using-git/about-git-rebase) about rebasing. In essence it puts your commit on top of the git commits. We want to test your commits against the latest develop branch so we can make sure that your changes work with the latest version and not with the version that you forked from 1/2 year ago. I'll let the pipelines run and after that I'll take a look if any plugin might be broken with this change. @JohnMcLear Could you point me in a direction. I'm not quite up to date which plugins there are and which might be broken after this.

JohnMcLear:
Thisbadds an element as a parent before the editorcontainer.

Any CSS reference to foo > bar > editorcontainer could be broken by this afaik but this I'm only 50% confident.

PR #5606, opened 2022-09-24
CI: add --omit=dev and --only=prod flags to npm to work around bug with

webzwo0i:
hanging `git ls-remote`
A better solution is needed. The bug only affects node v12 and v14


webzwo0i:
this does not work. the solution is probably https://github.com/ether/etherpad-lite/pull/5589

PR #5736, opened 2023-06-21
bumping ueberdb to 4

JohnMcLear:


SamTV12345:
I just tested it. Seems fine. No backend test failures and only 21 frontend failures in Firefox. So no new failures. Didn't expect that such a bump in dependencies would go that smooth.

PR #5737, opened 2023-06-21
github action fix for windows build

JohnMcLear:


SamTV12345:
Nice seems like you were faster xD.

PR #5763, opened 2023-06-24
deps: testing npm 9.5.1

JohnMcLear:


SamTV12345:
Seems fine. With that we will get most updates that are incompatible with npm@6.x.x.

PR #5764, opened 2023-06-24
deps: socketio v4.7.0

JohnMcLear:

### DIFF
--- a/settings.json.docker
+++ b/settings.json.docker
@@ -527,7 +527,7 @@
   /*
    * Restrict socket.io transport methods
    */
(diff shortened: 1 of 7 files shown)


SamTV12345:
This pr should resolve the legacy peer option from npm that I have to do in the other branch.

SamTV12345:
See: https://socket.io/docs/v4/client-installation/

SamTV12345:
https://socket.io/docs/v4/server-api/

SamTV12345:
It is weird. On v3 it says io not found. So somehow the variable is not injected in the global scope. On v4 it says nothing but it also doesn't call the document.ready function. Very weird.

PR #5766, opened 2023-06-25
Fix/log4js

SamTV12345:
This pull request updates log4js to the latest version. During testing I found out that if I bump typescript to 5.1.3 and npm enters legacy-peer-deps mode it somehow gets rid of all the old dependencies like request which I removed in this pull request too. 

I don't know why there are some tests that fail I'll look into that later.

### DIFF
--- a/.github/workflows/lint-package-lock.yml
+++ b/.github/workflows/lint-package-lock.yml
@@ -29,7 +29,7 @@ jobs:
             src/bin/doc/package-lock.json
       -
         name: Install lockfile-lint
-        run: npm install --no-save lockfile-lint
+        run: npm install --no-save lockfile-lint --legacy-peer-deps
       -
         name: Run lockfile-lint on package-lock.json
         run: >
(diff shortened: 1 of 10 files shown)


SamTV12345:
Because async await on top level is not supported. I would suggest supporting node 16>. Seems like the tests are also failing on the develop branch. So it might seems fine 😄 .

SamTV12345:
I'll rebase once again when the node branch is merged into develop.

JohnMcLear:
> Because async await on top level is not supported. I would suggest supporting node 16>. Seems like the tests are also failing on the develop branch. So it might seems fine 😄 .

we're now 16+

JohnMcLear:
This should be split up.

1. Axios
2. Log4JS

Also we need to talk about how this actually impacts Etherpad.  Afaik **this will break logging configs.**

SamTV12345:
@JohnMcLear I seem to have hit the same problem again. All pipelines fail and I didn't succeed in just cloning the project again.

SamTV12345:
Dependency review is probably not happy because I had some time the latest npm in it which doesn't work out of the box. If that pull request will crash instances I would postpone this pull request to milestone 2.0 @JohnMcLear

JohnMcLear:
Split it down to minor changes IE axios and see what works/doesn't then go from there : )

SamTV12345:
Axios has been merged in #5776 . This contains only the log4js changes necessary to migrate to v6

PR #5776, opened 2023-06-27
Feature/axios

SamTV12345:
Removes request package in favor of axios.


webzwo0i:
Won't this install all dev dependencies in production (ie not dev) setups?

SamTV12345:
Thanks for checking. I just opened another branch that should have fixed that. Seems like we need the optional dependencies for testing. #5781 removes the dev dependencies and it fails. That would mean that we include the dev dependencies by default. The newer npm versions warn you to choose between include dev or no dev. That is why I added it.

PR #5784, opened 2023-06-27
Update/socket.io to latest 2.x version

SamTV12345:


SamTV12345:
Don't know why but we have some really flaky tests in the frontend admin tests. Sometimes 16 and 18 fail or just 16.

PR #5790, opened 2023-06-28
npm i --no-optional --omit=optional --include=dev --lockfile-version 1

JohnMcLear:

### DIFF
--- a/src/package-lock.json
+++ b/src/package-lock.json
@@ -4,6 +4,12 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@aashutoshrathi/word-wrap": {
+      "version": "1.2.6",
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Mostly playing w/ dep commands, I'll keep experimenting

SamTV12345:
Yeah. There is also a problem with "Could not identity Etherpad base path in this linux installation in "/mnt/c/Users/user/WebstormProjects/etherpad-lite/src/node_modules/ep_etherpad-lite"" which is displayed at startup - with the message "/mnt/c/Users/user/WebstormProjects/etherpad-lite/src/node_modules/ep_etherpad-lite does not end with "src

SamTV12345:
Now that we have pnpm in place. We can skip the migration to the latest npm version as pnpm is more modern, stable and less prone to errors.

PR #5805, opened 2023-07-03
fix appendRevision

webzwo0i:


SamTV12345:
The error messages say: before each" hook for "Attempt to Install a plugin" Error: waitFor condition never became true ()=>helper.admin$&&helper.admin$(".menu").find("li").length>=3
Isn't this a new issue? It was before only error when uninstalling a plugin.

JohnMcLear:
> The error messages say: before each" hook for "Attempt to Install a plugin" Error: waitFor condition never became true ()=>helper.admin$&&helper.admin$(".menu").find("li").length>=3 Isn't this a new issue? It was before only error when uninstalling a plugin.

I have seen that error before.

SamTV12345:
Ok. So we can add that to the develop branch.

webzwo0i:
needs a test, but I can add it soon

webzwo0i:
Some feedback regarding the test. It was in the wrong file (not related to chat), so I moved it. Also it used an arrow function, which is discouraged in mocha, see https://mochajs.org/#arrow-functions We don't use mocha's context here, so it's technically okay, but it triggered an eslint rule.

But most importantly the second createPad call was never executed. As there was no check for the actual pad text and the call to deletePad successfully deleted the pad that was created with the first createPad call, everything looked okay although the call never hit the API.

SamTV12345:
Oh sorry. That was my fault. Thanks for fixing that. Is there a reason why you add async to every test? I think I haven't touched that much. 

If it is touching an eslint rule it is best to go with that. I found out that you can also use the arrow function with timeout but you need to append that after the last closing bracket of the arrow function. 

If you want a review I could do one in two hours.

webzwo0i:
The async migration is only a precaution so this kind of error does not reappear, e.g. if someone else would add tests in chat.js and add multiple API calls in a single test case, it could happen again. I don't know what's causing this behavior. When you call something like `expect(200, done)` the request would be made, but when you call `done()` outside the request chain, it won't work.

The docs say you need to call .then, .end or await, but we did not explicitly call any of the former two. I'm not sure why calling `done` - which is from mocha - makes it work. As you cannot call `done` multiple times, it would only work with a single request per test case.

Thanks for offering a review! I think it's okay if I just merge it, as I essentially only added the call to `getText` between create/delete

PR #5839, opened 2023-07-19
Npm v test

JohnMcLear:
Just wanted to have a little play w/ npm versions and testing..  Let's see what happens.

cc @webzwo0i @SamTV12345


SamTV12345:
Interesting so the bug is still not fixed? The security bug doesn't seem to have been triggered so far.

SamTV12345:
Btw I don't know what changed in the GitHub runners. The windows build is failing again.

PR #5866, opened 2023-08-07
Call exports.start() for Phusion Passenger

### LINKED ISSUE #5683: Etherpad-lite 1.8.18 doesn't run with Passenger
Etherpad-lite doesn't run with Passenger but working fine from commandline (src/bin/run.sh)
Hangs with:
```
[2023-01-28 19:51:17.940] [INFO] plugins - npm --version: 6.14.15
```
Tested on
Passenger 6
Node.js v14 and v16

I fixed it by editing src/node/server.js last lines:
```
//if (require.main === module) exports.start();
exports.start();
```
It seems that exports.start() is never called while running with Passenger.
I guess we need to do a better fix than that ;)

avkarenow:
Now `exports.start()` is called only for cli: `if (require.main === module) exports.start();`
To run etherpad-lite also with Passenger it is needed to call `exports.start();` too.

fixes #5683

### DIFF
--- a/src/node/server.js
+++ b/src/node/server.js
@@ -276,3 +276,4 @@ exports.exit = async (err = null) => {
 };
 
 if (require.main === module) exports.start();
+if (typeof(PhusionPassenger) !== 'undefined') exports.start();


SamTV12345:
Thanks. I'll let the tests run. As far as I can see this shouldn't change anything in our flow - it just adds something if the application server comes into play.

SamTV12345:
Nice that worked. Thanks for the contribution. It will be included in the next release.

Issue #3392, opened 2018-05-12
favicon setting issues: external custom favicon is not respected on /admin pages, but should be.


Wikinaut:
Favicon setting is disregarded on /admin pages: the standard EPL favicon is shown

When you define a favicon.ico in the settings.json, this is not respected on /admin pages, but should be.

Wikinaut:
There is also another major problem with the other favicon settings

I have used this since years (works, but not an /admin)
```
 "favicon": "https://another-server/favicon.ico",
 "faviconPad": "https://www.another-server/favicon.ico",
 "faviconTimeslider": "https://www.another-server/favicon.ico",
```
but when I read https://github.com/ether/etherpad-lite/blob/6dd172d6b76cb219f0aa095ea09e43ef7f259d3d/src/node/utils/Settings.js#L44 nd the following lines in /src/node/utils/settings.js, a single custom favicon ico definition like in the following should also be working one all pad pages, but does not:
```
 "favicon": "https://another-server/favicon.ico",
// "faviconPad": "https://www.another-server/favicon.ico",
// "faviconTimeslider": "https://www.another-server/favicon.ico",
```

JohnMcLear:
Confirmed.  Easy one for someone to pick up and solve.

KSSaiTeja:
i will pick this, can you assign me this?

JohnMcLear:
Great thanks @KSSaiTeja

KSSaiTeja:
My pleasure :D @JohnMcLear

KSSaiTeja:
Unfortunately i cant figure out the issue, as i am a beginner please help me with the file which i work on to. and please give some guidance to work upon this..

SamTV12345:
Looks like someone fixed it in the meantime between 2018 and today. It is working flawlessly. When I download a favicon and set it to ./favicon.ico in the settings.json it is also displayed on the admin pages after login.

KSSaiTeja:
ok thanks for confirming! i will work on other issues

SamTV12345:
Alright👍🏻. Maybe get yourself familiar with the codebase first.

KSSaiTeja:
yeah i will

q2apro:
Hello from Dec 2023.

Unfortunately, this is not fixed.

In `/admin/settings` the example setting:

`"favicon": "https://www.myothersite.com/favicon.ico",`

does have NO effect. Still the defaul ../favicon.ico can be seen in the HTML.

How to fix this?

[...]

PR #5877, opened 2023-08-16
Fix: favicon setting issues, external custom favicon is not respected…

KSSaiTeja:
… on /admin pages

the favicon is fixed (path fixed)

fixes: #3392

### DIFF
--- a/settings.json.docker
+++ b/settings.json.docker
@@ -110,7 +110,7 @@
    * is used. If this is a relative path it is interpreted as relative to the
    * Etherpad root directory.
    */
-  "favicon": "${FAVICON:null}",
+  "favicon": "tests/backend/specs/favicon-test-custom.png",
 
   /*
    * Skin name.
--- a/settings.json.template
+++ b/settings.json.template
@@ -101,7 +101,7 @@
    * is used. If this is a relative path it is interpreted as relative to the
    * Etherpad root directory.
    */
-  "favicon": null,
+  "favicon": "tests/backend/specs/favicon-test-custom.png",
 
   /*
    * Skin name.


KSSaiTeja:
please tell me if it was ok or not, if not please tell me how can i reproduce the issue and how to fix it. As i am a newbie to opensource i am a bit confused to change things in code.
Hope you understand and help me out

SamTV12345:
Hi thanks for the pull request. 

Unfortunately this does not fix this error. ${FAVICON:null} means that if there is no favicon take null which does not show a favicon. The favicon can be configured like you did it through the settings.json. The problem is that when visiting the admin panel the favicon is not shown. So to test this you need to go to http(s)://your-etherpad.instance.com/admin and check if the favicon is there. 

So I think something in the code needs to be changed. The settings.json is the source where the administrator can change settings of etherpad.

But thanks for your try. Feel free to continue your pull request.

KSSaiTeja:
thanks for the reply and suggestion. i will work on the issue you mentioned. I will start fixing it right now

KSSaiTeja:
i’ll let you know with the new commit

KSSaiTeja:
Right now i am checking the settings.json file unfortunately i cant figure the issue.. can you please tell me which lines of code are the cause of this issue. and mention the file names as links.. dont take it as dumb request, as i am beginner, it would be more helpful to me to find the issue by your help and start working out

please understand and try to give the code lines to work on... :D

SamTV12345:
I just had a look it is fixed. If I change the favicon it is also changed on the admin menu.

KSSaiTeja:
i will close the pull request then, thanks for confirming

PR #5907, opened 2023-09-04
fix abiword not working in container

ppom0:
cc @rdelaage
on alpine, plugin command (used in src/node/utils/Abiword.js at line 31) is not installed by defaut. Add this plugin when installing abiword.


SamTV12345:
Thanks for the pull request and creating the issue. I can confirm that Abiword is now working. Sorry for my overseight on this one 😅 .

PR #5943, opened 2023-09-21
Don't access jQuery through `parent.parent`

dhalenok:
When Etherpad is hosted on a different domain and the iframe is violating the same-origin policy, trying to close Import/Export window via Escape key results in an error:

Screenshot 2023-09-21 at 10 27 09

This PR fixes that by removing the reference to `parent.parent` when accessing the jQuery ($).

### DIFF
--- a/src/static/js/pad_editbar.js
+++ b/src/static/js/pad_editbar.js
@@ -318,7 +318,7 @@ exports.padeditbar = new class {
         }
       } else {
         // Focus on the editbar :)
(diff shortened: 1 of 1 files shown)


SamTV12345:
I'll get back to you this evening. Seems like all tests are fine. I need to run the backend tests manually for this branch.

SamTV12345:
Thanks for this pull request.

PR #5944, opened 2023-09-22
Docker: Add ability to specify configuration file as a build arg

dhalenok:
This PR adds ability to specify configuration file as a docker build argument for custom build.

Usage example:
```sh
docker build --build-arg SETTINGS=./settings.json.docker.development .
```

This is useful when there's a need to have different configuration files, for example to specify plugin settings for different environments.


SamTV12345:
Thanks. Great idea to make the docker build more flexible.

PR #5945, opened 2023-09-24
Improve update check

Gared:
This change will change the user-agent for the requests to get the last version information from "axios/1.5.0" to "Etherpad/1.9.3".
I also added a caching for one hour to reduce those calls.

### DIFF
--- a/src/node/utils/UpdateCheck.js
+++ b/src/node/utils/UpdateCheck.js
@@ -3,7 +3,7 @@ const semver = require('semver');
 const settings = require('./Settings');
 const axios = require('axios');
 const headers = {
-  'User-Agent': 'Etherpad ' + settings.getEpVersion(),
+  'User-Agent': 'Etherpad/' + settings.getEpVersion(),
 }
 
 const updateInterval = 60 * 60 * 1000; // 1 hour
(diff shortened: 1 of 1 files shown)


src/node/utils/UpdateCheck.js
```diff
@@ -2,23 +2,35 @@
...
-              return
-            }
-            return await Promise.resolve(infos);
-          }
-          catch (err) {
-            return   await Promise.reject(err);
-          }
-        })
+  return axios.get('https://static.etherpad.org/info.json', {headers: headers})
+  .then(async resp => {
```
SamTV12345, line 19: axios has a separate method for handling error. Everything that is not 2xx like 404, 500, etc. are handled in .catch after .then. There you can put the error handling measures.

Gared: I reformatted the code and have not changed the exception handling, but you're right. I improved the exception handling. Please have a look again 🙂

Thanks, but don't expect too much. Just wanted to improve a little bit how my server is getting called 😅

SamTV12345: Nice. I'll take a look asap and check if it's working. 

Ah okay. Thought I might get somebody back to the project 😂.

SamTV12345 (approved):
LGTM and works without a problem.

PR #5974, opened 2023-10-10
[Legacy][Tech Debt][Fix] Modernized some HTML and CSS

blackmambaza:
- Added default font to the font-family in CSS files
- Added lang attributes to HTML tags

This changes improves the compatibility with modern web browsers and ensures that the code meets HTML standards


SamTV12345:
Thanks for the pull request. Why were some properties deleted? Can you make it so that l10n dictates the language attribute value?

Issue #6094, opened 2024-01-06
Change language strings (DE)


q2apro:
Regarding file: https://github.com/ether/etherpad-lite/blob/develop/src/locales/de.json

In German the correct word for a chat is not `Unterhaltung` but `Chat`. 

To be changed:

```
	"pad.settings.stickychat": "Chat immer anzeigen",
	"pad.settings.chatandusers": "Chat und Benutzer anzeigen",
	"pad.chat": "Chat",
```

SamTV12345:
True. Unterhaltung sounds a bit dated. Chat is the better term.

SamTV12345:
Added in translatewiki

PR #6095, opened 2024-01-06
Fix: #6094 Change language strings (DE)

bhavani-shankar:
Change language strings (DE)
from 'Unterhaltung' to 'Chat'


SamTV12345:
Hi @bhavani-shankar . I agree with your changes and thanks for taking the opportunity to improve the translations. Unfortunately language changes should be submitted via https://translatewiki.net/, so please change the respective strings there.

PR #6112, opened 2024-01-20
Added ts node approach

SamTV12345:


src/package.json
```diff
@@ -103,8 +104,10 @@
   },
   "scripts": {
     "lint": "eslint .",
-    "test": "mocha --timeout 120000 --recursive tests/backend/specs ../node_modules/ep_*/static/tests/backend/specs",
-    "test-container": "mocha --timeout 5000 tests/container/specs/api"
+    "test": "mocha --require ts-node/register --timeout 120000 --recursive tests/backend/specs ../node_modules/ep_*/static/tests/backend/specs",
+    "test-container": "mocha --require ts-node/register --timeout 5000 tests/container/specs/api",
+    "dev": "ts-node node/server.ts",
```
AugustinMauroy, line 109: you can run ts file whit `tsx` like that `node import="tsx" ./myScritp.ts`

Ref:
- [TSX](https://www.npmjs.com/package/tsx?activeTab=readme)
- [article about loader on nodejs](https://dev.to/_staticvoid/how-to-run-typescript-natively-in-nodejs-with-tsx-3a0c)

SamTV12345: Is tsx better than ts-node?

AugustinMauroy: with ts-node you call a script that runs the typescriopt script, whereas with tsx in loader node it's node that runs the script directly. Normally tsx is supposed to perform better, but I'm not 100% sure.

src/tsconfig.json
```diff
@@ -0,0 +1,109 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
```
AugustinMauroy, line 3: I thinks we can remove all commented key to have something more clean. What do you thinks ?

SamTV12345: Good suggestion. I did that.

AugustinMauroy: 👍

src/package.json
```diff
@@ -108,7 +108,8 @@
     "test": "mocha --import=tsx --timeout 120000 --recursive tests/backend/specs ../node_modules/ep_*/static/tests/backend/specs",
     "test-container": "mocha --import=tsx --timeout 5000 tests/container/specs/api",
     "dev": "node --import tsx node/server.ts",
-    "prod": "node --import tsx node/server.ts"
+    "prod": "node --import tsx node/server.ts",
+    "ts-check": "tsc --noEmit --watch"
```
AugustinMauroy, line 112:
```suggestion
    "ts-check": "tsc --noEmit",
    "ts-check:watch": "tsc --noEmit --watch"
```

it's allow us to use `ts-check` on CI and `ts-check:watch` for dev. What do you think ?

src/node/models/PromiseWithStd.ts
```diff
@@ -0,0 +1,8 @@
+import {Readable} from "node:stream";
+import {ChildProcess} from "node:child_process";
```
AugustinMauroy, line 2: If you import something as type use this syntax:

```suggestion
import type {Readable} from "node:stream";
import type {ChildProcess} from "node:child_process";
```

this avoids confusion for junior devs. 

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

SamTV12345: Thanks for the suggestion. I fixed that.

AugustinMauroy:
is it a good idea to call the directory `model`? I'd go for `types` or maybe you've also got people using `@types` but I'm not a big fan because it creates confusion with DefinitelyTyped

SamTV12345:
> is it a good idea to call the directory `model`? I'd go for `types` or maybe you've also got people using `@types` but I'm not a big fan because it creates confusion with DefinitelyTyped

Maybe not. I renamed it to types.

.github/workflows/windows.yml
```diff
@@ -116,7 +116,8 @@ jobs:
       -
         name: Install Cypress
         run: cd etherpad && cd src && npm install cypress --legacy-peer-deps
-      - name: Install npm@6
+      - 
+        name: Install npm@6
         run: npm install npm@6 -g
```
AugustinMauroy, line 121: why install `npm` because it's already ship with node

SamTV12345: The problem is that Etherpad is one of the first projects that started to use node and npm. The only problem is that to run Etherpad you require this old npm version. More recent versions of npm will upgrade the package-lock.json file which breaks the Etherpad installation. There is a branch available that tries to do the npm update but there is unfortunately currently no one looking at it.

AugustinMauroy: I see. In short, the infrastructure is "old". We can start with typescript. 
I'm willing to open up an issue for the roadmap of how to modernise the infrastructure if you (the ether contributors) are interested.

SamTV12345: Yes that would be great. Thanks. I guess the best starting point is currently Typescript. After that current npm version and then move to ES instead of CJS.

src/node/hooks/express.ts
```diff
@@ -160,10 +165,10 @@ exports.restartServer = async () => {
   }
 
   // Measure response time
-  app.use((req, res, next) => {
+  app.use((req:any, res:any, next:Function) => {
```
AugustinMauroy, line 168: What if `req` doesn't have any specifique type ? And what if we type `req: Request`

[REF](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.request.html)

src/node/hooks/express.ts
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
+import {Socket} from "node:net";
+import {MapArrayType} from "../types/MapType";
+
 const _ = require('underscore');
```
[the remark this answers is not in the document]
SamTV12345: Does not work unfortunately. That throws an error in NodeJS. It only works for types because they are ignored by jsx.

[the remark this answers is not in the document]
AugustinMauroy: Could you give exact error. Because it's should work.
Unless the `type` in the `package.json` is set to `commonjs`.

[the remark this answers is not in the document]
SamTV12345: When I go to express.ts and change the import of SecretRotater to `import {SecretRotater} from '../security/SecretRotater'` it says TypeError import_SecretRotater.SecretRotater is not a constructor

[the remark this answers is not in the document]
SamTV12345: Oh. So everytime I use import I also have to make sure that the script to be imported uses the ES syntax. Otherwise I get weird errors.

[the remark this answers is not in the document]
AugustinMauroy: Ok, I thinks it's because the source file/package isn't use ES syntax .

src/node/stats.ts
```diff
@@ -0,0 +1,10 @@
+'use strict';
```
AugustinMauroy, line 1: `use strict` is still needed with typescript ?

AugustinMauroy:
why all CI didn't pass

AugustinMauroy:
ok

PR #6164, opened 2024-02-17
Improve plugins docker build

Gared:
Install plugins directly during docker build with live-plugin-manager (#6153)


Gared:
@SamTV12345 Can you resolve the merge conflicts please?

SamTV12345:
@Gared I rebased the changes. We moved the bin folder up one directory. That way we don't always have to switch between src and back. I added your script to the workspace package.json. That way you can call the script from Etherpad root without worrying being in the correct directory. Besides that I moved the backend of Etherpad to Typescript.

PR #6170, opened 2024-02-19
Refactor project structure extract server

### LINKED ISSUE #6166: Backend refactoring proposal: Step 1. multipackage monorepo
**Context**: As proposed by @SamTV12345  I had a look at https://github.com/ether/etherpad-lite/pull/5812 . To avoid creating a new massive PR that gets stuck I would propose to take small steps instead. As a start, I looked at the plugin architecture and especially this quote from the mentioned PR:

> get rid of the symbolic link ep_etherpad-lite; instead use npm link ep_etherpad-lite in the root directory to build a package.json file and install all the dependencies in node_modules, incl. the link to ./src

**Status quo**: The node_modules folder in the root directory essentially works as a plugin directory but is not a "real" npm node_modules folder (as no package.json exists to control it). The src folder is sym linked (already hardcoded but otherwise also done in src/bin/installDeps.sh) and further plugins source code can be dropped in there. Otherwise plugins can also be installed during the build, when given in ETHERPAD_PLUGINS env variable. The folder is then scanned within the plugins.js file for content.

**My opinion**: Despite being obviously a problem for the upgrade (as mentioned in the PR) I see multiple issues with the current architecture:

- First, the root level has no package.json file, so having only a nodule_modules folder is confusing and also results in a couple of hacks. For instance, installing the build time plugins inside the Dockerfile (line 112) is done using `—no-save` and the plugin detection scan has a long comment why certain parameters have to be used (see src/static/js/pluginfw/installer.js, line 45)
- Kind of a circular dependency: The current project is itself a plugin (ep_etherpad-lite) and is basically requiring itself.  

**Refactoring proposal**: I would suggest to split the project basically in two parts but keeping it in one repository (multipackage monorepo). The root project simply holds the `bin` folder and installs the ep_etherpad-lite plugin (through package.json) which is located in the src folder (could be renamed to ep_etherpad-lite). This can be achieved easily using pnpm workspaces:

```
- package.json # holds workspace reference to ep_etherpad-lite
- bin
- ...
- ep_etherpad-lite (formerly src, we can keep also the name src for now)
    - package.json
    - …
```

At a later point in time, one could optionally also extract at least the sever scaffold structure (server.ts, express, ...) and the plugin system.

This results in a cleaner setup and gets rid of the sym link (well at least doing it by hand as pnpm/npm workspaces also use sym links under the hood). Also, new plugins can be easily added either via the root package.json or by using npm link, if the source code is available locally (e.g. because a plugin is tested locally).

I started with a small prototype that leverages workspaces and roughly shows how the setup works: #6170 

If the direction is agreeable, I am happy to create a PR.

JannikStreek:
This PR is a lightweight proof of concept to introduce pnpm workspaces and to get rid of the the symlink workaround the root node_modules folder. All other script commands are simply proxied to the existing package.json file.

Plugins can now be simply installed via npm package manager in the root file (as they will appear in the root module folder) or linked with `pnpm link`.

closes #6166


package.json
```diff
@@ -0,0 +1,38 @@
...
+    "lint": "pnpm --filter ep_etherpad-lite run lint",
+    "test": "pnpm --filter ep_etherpad-lite run test",
+    "test-container": "pnpm --filter ep_etherpad-lite run test-container",
+    "dev": "pnpm --filter ep_etherpad-lite run dev",
+    "prod": "pnpm --filter ep_etherpad-lite run prod",
+    "ts-check": "pnpm --filter ep_etherpad-lite run ts-check",
+    "ts-check:watch": "pnpm --filter ep_etherpad-lite run ts-check:watch"
+  },
+  "dependencies": {
+    "ep_etherpad-lite": "workspace:./src"
```
JannikStreek, line 25: it would be cleaner to rename the folder. however, as src is used quite a lot this will result in a bigger refactoring

SamTV12345: We could leave it as is for now but this looks very promising.

SamTV12345:
> This PR is a lightweight proof of concept to introduce pnpm workspaces and to get rid of the the symlink workaround the root node_modules folder. All other script commands are simply proxied to the existing package.json file.
> 
> Plugins can now be simply installed via npm package manager in the root file (as they will appear in the root module folder) or linked with `pnpm link`.
> 
> closes #6166

Will the live plugin manager still work with this change? If the location is the same it should still migrate them into src.

JannikStreek:
> > This PR is a lightweight proof of concept to introduce pnpm workspaces and to get rid of the the symlink workaround the root node_modules folder. All other script commands are simply proxied to the existing package.json file.
> > Plugins can now be simply installed via npm package manager in the root file (as they will appear in the root module folder) or linked with `pnpm link`.
> > closes #6166
> 
> Will the live plugin manager still work with this change? If the location is the same it should still migrate them into src.

imho it should as the structure stayed the same. the ep_etherpad-lite plugin is still where it was (sym linked in node_modules) as are plugins installed from the start when the image was built. But I will double check that. Unfortunately, web sockets are currently broken for me since the update which are needed in the admin view 😅

SamTV12345:
> > > This PR is a lightweight proof of concept to introduce pnpm workspaces and to get rid of the the symlink workaround the root node_modules folder. All other script commands are simply proxied to the existing package.json file.
> > > Plugins can now be simply installed via npm package manager in the root file (as they will appear in the root module folder) or linked with `pnpm link`.
> > > closes #6166
> > 
> > 
> > Will the live plugin manager still work with this change? If the location is the same it should still migrate them into src.
> 
> imho it should as the structure stayed the same. the ep_etherpad-lite plugin is still where it was (sym linked in node_modules) as are plugins installed from the start when the image was built. But I will double check that. Unfortunately, web sockets are currently broken for me since the update which are needed in the admin view 😅

Yip 😨 . I just also realized that it is broken. But it's definitely a great addition. Now I don't have to change to src to start Etherpad.

SamTV12345:
Seems like the first step is done. Should we merge this or do you want to include anything else in this pull request?

JannikStreek:
> Seems like the first step is done. Should we merge this or do you want to include anything else in this pull request?

I think I want to test the live plugin manager before (just to make sure). And optional: I can move the bin folder to the root - I think thats shouldn't be too much refactoring. But thats optional, I am also fine to leave it there for now and do that in another PR. I like smaller pull requests anyway :)

SamTV12345:
> > Seems like the first step is done. Should we merge this or do you want to include anything else in this pull request?
> 
> I think I want to test the live plugin manager before (just to make sure). And optional: I can move the bin folder to the root - I think that shouldn't be too much refactoring. But thats optional, I am also fine to leave it there for now and do that in another PR. I prefer smaller pull requests anyway :)

I think moving bin one up sounds great. It's also symlinked in my local installation. Don't think bin suits well in terms of source of etherpad. They are more like scripts to start Etherpad. You can test the admin panel by navigating to adminsettings.ts and commenting out `if(!isAdmin) return`

JannikStreek:
> > > Seems like the first step is done. Should we merge this or do you want to include anything else in this pull request?
> > 
> > 
> > I think I want to test the live plugin manager before (just to make sure). And optional: I can move the bin folder to the root - I think that shouldn't be too much refactoring. But thats optional, I am also fine to leave it there for now and do that in another PR. I prefer smaller pull requests anyway :)
> 
> I think moving bin one up sounds great. It's also symlinked in my local installation. Don't think bin suits well in terms of source of etherpad. They are more like scripts to start Etherpad. You can test the admin panel by navigating to adminsettings.ts and commenting out `if(!isAdmin) return`

live package manager seems to work from what I can see.

yes I agree, bin folder has nothing to do with the core plugin and should be moved up. However, I would propose to do this in a different PR. Just had a look at all these scripts...every script uses different shell location writings and has to be essentially tested manually. So this is bigger than I thought. I would propose to extract this in another issue/PR. If thats fine with you we can merge this when the tests pass.

SamTV12345:
> > > > Seems like the first step is done. Should we merge this or do you want to include anything else in this pull request?
> > > 
> > > 
> > > I think I want to test the live plugin manager before (just to make sure). And optional: I can move the bin folder to the root - I think that shouldn't be too much refactoring. But thats optional, I am also fine to leave it there for now and do that in another PR. I prefer smaller pull requests anyway :)
> > 
> > 
> > I think moving bin one up sounds great. It's also symlinked in my local installation. Don't think bin suits well in terms of source of etherpad. They are more like scripts to start Etherpad. You can test the admin panel by navigating to adminsettings.ts and commenting out `if(!isAdmin) return`
> 
> live package manager seems to work from what I can see.
> 
> yes I agree, bin folder has nothing to do with the core plugin and should be moved up. However, I would propose to do this in a different PR. Just had a look at all these scripts...every script uses different shell location writings and has to be essentially tested manually. So this is bigger than I thought. I would propose to extract this in another issue/PR. If thats fine with you we can merge this when the tests pass.

Thanks for testing. Yes there are a lot of scripts. I agree it's better to split things up into multiple pull requests.

SamTV12345:
Perfect. Everything works flawlessly and I also found the bug why the admin panel isn't loading.

PR #6174, opened 2024-02-20
cleanup after workspace refactoring

### LINKED ISSUE #6171: Cleanup after workspace refactoring
After https://github.com/ether/etherpad-lite/issues/6166 some tasks remain open:

- [ ] Cleanup comments, e.g. in docker build file, as these are now outdated and refer to old commands
- [ ] node_modules directory in the root folder should be mapped to a value as pnpm hoists dependencies there 
- [ ] move bin folder one level up. it doesn't make sense anymore to have it in the src folder

JannikStreek:
closes #6171

**In essence**:
- I moved the bin folder to root directory
- Fixed the "upgrade from latest release" workflow as far as I could. The problem was that pnpm was used within the old master that just was using npm files - thats not really what the test is for. It should replicate an older setup and the migration to the new code base. I adjusted that but now a couple of backend tests further down the line are failing in regards to the new web socket upgrade @SamTV12345 any idea why?
- Optimizations in docker. Imo we don't need to install plugins first and then install the core plugin. thats also counter intuitive and makes no sense since we switched to pnpm and workspaces (same is true for the worklfow order of installation).
- removed mounts from docker compose for node_modules as these overwrite build time installed plugins.  

**Open**

- [x] Test all scripts to see if these still work as expected after bin move. Help wanted here...
- [x] Should I add the root pnpm-lock.yaml ? It basically only includes the hoisted versions of the src/pnpm-lock.yaml and you wouldn't usually edit it anyway.


SamTV12345 (approved):

SamTV12345:
Great work. I couldn't find an issue with your commit. Thanks a lot.

PR #6186, opened 2024-02-24
Added nextjs as a preview option for a new admin area

SamTV12345:
This pull request introduces nextjs to Etherpad. The purpose is to slowly build/rebuild the features that Etherpad introduced with jQuery and introduce a plugin mechanism that works with the old plugins as well.


src/postcss.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  plugins: {
+    tailwindcss: {},
+    autoprefixer: {},
+  },
+};
```
AugustinMauroy, line 6:
```suggestion
/** @type {import('postcss-load-config').Config} */
const config = {
  plugins: {
    'tailwindcss/nesting': {},
    tailwindcss: {},
    autoprefixer: {},
  },
};

module.exports = config;
```

also use `.cjs` on extension to be strict on cjs/esm

src/tailwind.config.ts
```diff
@@ -0,0 +1,20 @@
...
+            backgroundImage: {
+                "gradient-radial": "radial-gradient(var(--tw-gradient-stops))",
+                "gradient-conic":
+                    "conic-gradient(from 180deg at 50% 50%, var(--tw-gradient-stops))",
+            },
+        },
+    },
+    plugins: [],
+};
+export default config;
```
AugustinMauroy, line 20:
```suggestion
import type { Config } from 'tailwindcss';

export default {
  content: ['./**/*.tsx'],
} as Config;
```

src/pages/admin/admin.tsx
```diff
@@ -0,0 +1,32 @@
+import 'tailwindcss/tailwind.css'
```
AugustinMauroy, line 1: this should be done like that that bad idea.

we should have a `global.css`
and import it on `layout.tsx`

PR #6201, opened 2024-03-04
Feature: Docker build from behind a proxy

d-led:
## Problem

custom image cannot be built from behind a proxy

## Solution

- optionally set `http_proxy`, `https_proxy` and `no_proxy` as Docker build arguments to make the used package managers (`apk`, `npm`) work
- when the options are not provided, it's expected that the build works as before
- lowercase variables used as these work for both `apk` and `npm`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,15 @@
 FROM node:alpine as build
 LABEL maintainer="Etherpad team, https://github.com/ether/etherpad-lite"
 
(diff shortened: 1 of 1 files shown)


SamTV12345:
@d-led Thanks for the pull request :). It did work without a problem.

d-led:
thanks a lot, great to hear @SamTV12345 !

PR #6243, opened 2024-03-19
Allow providing local plugins to Docker image

dhalenok:
Due to new changes in the version 2.0, it's no longer possible to specify local plugins when building a Docker image.

In my testing, executing `pnpm run install-plugins` (one for remote and one for local plugins via `--path`) made it so only the plugins specified in the latest command were loaded.

This PR changes that, so you can specify both remote and local plugins in one command: `pnpm run install-plugins ep_openid_connect --path ../ep_my_local_plugin`. It should still be possible to specify only remote or local plugins.

In addition, this PR introduces a new Dockerfile argument called `ETHERPAD_LOCAL_PLUGINS`, value of which gets passed to `install-plugins` script.

There is, however, one thing left to do, which is copying the plugins to the docker image. IIRC, before v.2.0 the root directory of the repo was copied over, but it's no longer a case and I'm not too sure how to tackle it.
One idea was that local plugins could be stored in a directory called `plugins`or `local_plugins` and copied over to `EP_DIR`in Docker image:

```
COPY --chown=etherpad:etherpad ./plugins/ ./
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,14 @@ ARG SETTINGS=./settings.json.docker
 #   ETHERPAD_PLUGINS="ep_codepad ep_author_neat"
 ARG ETHERPAD_PLUGINS=
 
+# local plugins to install while building the container. By default no plugins are
+# installed.
+# If given a value, it has to be a space-separated, quoted list of plugin names.
+#
+# EXAMPLE:
+#   ETHERPAD_LOCAL_PLUGINS="../ep_my_plugin ../ep_another_plugin"
+ARG ETHERPAD_LOCAL_PLUGINS=
+
 # Control whether abiword will be installed, enabling exports to DOC/PDF/ODT formats.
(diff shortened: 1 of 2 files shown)


SamTV12345:
Thanks for the suggestions.

dhalenok:
@SamTV12345 

Sorry for reviving this PR, but what do you think about adding
```sh
COPY --chown=etherpad:etherpad ./plugins/ ./
```
to the Dockerfile?

Or maybe you have some better ideas on how to copy local plugins into the docker image?

SamTV12345:
No worries. The problem is that the plugin folder won't be present on every computer. We could add the folder to git. Then it would work.

dhalenok:
I've been able to circumvent the issue by storing local plugins in `src` directory on the host, and specifying build arg as `ETHERPAD_LOCAL_PLUGINS="../src/plugins/ep_my_plugin"`.

PR #6254, opened 2024-03-22
Improve dockerfile

Gared:
This PR will reduce the size of the docker image from 425 MB to 252 MB


SamTV12345:
Awesome. Thanks for discovering that. I was also wondering why it suddenly increased in size that much.

PR #6274, opened 2024-03-24
Added querying and removing plugins.

### LINKED ISSUE #6272: How to remove plugins ?
I know that you now need to install plugins with `pnpm run install-plugins ep_plugin-name` with v2.0.1.

How do you remove a plugin now from command line?

SamTV12345:
This currently does not work correctly as the array is always empty. Is this racy?

The commands are now more unix style.

You can do `pnpm run plugins i ep_headings2` to install ep_headings2 plugin.
You can do `pnpm run plugins ls` to list the plugins`
You can `pnpm run plugins rm ep_headings2` to install the ep_headings2 plugin

Closes #6272


hamoid:
`pnpm run` shows

```
 remove-plugins
    pnpm --filter bin run remove-plugins
 ```

Which doesn't work for me. 

But `pnpm run plugins rm ep_xxx` works fine.

PR #6281, opened 2024-03-25
Feat/oauth2

SamTV12345:


webzwo0i:
Why not make use of https://github.com/ether/ep_openid_connect ?

SamTV12345:
> Why not make use of https://github.com/ether/ep_openid_connect ?

Welcome back @webzwo0i ☺️. Great to read you again. It is because Etherpad is now the OAuth 2 provider.

webzwo0i:
I see :-D

Can you please add docs like for https://github.com/ether/etherpad-lite/issues/6320 Tests would be great too :)

People use the API a lot with various tools and it was very straight forward - although not best practice with an unencrypted file, so I'd really appreciate getting rid of the unencrypted file, but this seems to break those tools

PR #6299, opened 2024-04-02
LoginScreen.tsx -> Passwort -> Password

kekePower:
Found a typo in LoginScreen where is says Passwort instead of Password.


SamTV12345:
Thanks for the fix. That is indeed a typo :)

PR #6347, opened 2024-04-20
Added editorconfig configuration

SamTV12345:


Gared:
Do we need for batch files something like this?
```
[start.bat]
end_of_line = crlf
```

SamTV12345:
@Gared Good idea I added that for all bat files

PR #6382, opened 2024-05-14
Readded support for apikey

SamTV12345:

### DIFF
--- a/settings.json.docker
+++ b/settings.json.docker
@@ -172,6 +172,13 @@
   "showSettingsInAdminPage": "${SHOW_SETTINGS_IN_ADMIN_PAGE:true}",
 
   /*
+    The authentication method used by the server.
+    The default value is sso
+    If you want to use the old authentication system, change this to apikey
+   */
+  "authenticationMethod": "${AUTHENTICATION_METHOD:sso}",
+
+  /*
    * Node native SSL support
    *
    * This is disabled by default.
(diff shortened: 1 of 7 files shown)


Loki-Afro:
thank you @SamTV12345 
i guess i have to supply now an additional env var `AUTHENTICATION_METHOD: "apikey"` with the next release :)

webzwo0i:
Thank you! Do you think we need test coverage for the old APIKEY.txt way? I'd prefer a deprecation warning and if possible a migration guide, so that we can drop the old method in the future and not hit the same problems again

SamTV12345:
> Thank you! Do you think we need test coverage for the old APIKEY.txt way? I'd prefer a deprecation warning and if possible a migration guide, so that we can drop the old method in the future and not hit the same problems again

Yes. That would be the best to move away from the APIKey method. We could support it for some time as it is now opt-in.

elmatusov:
Thanks a lot, SamTV12345, for adding the possibility to generate and use APIKEY.txt in the newest releases. As you pointed out, I just replaced _"authenticationMethod": "${AUTHENTICATION_METHOD:**sso**}",_ with  _"authenticationMethod": "${AUTHENTICATION_METHOD:**apikey**},_ in the setting.json file. It works really well! I'm using APIKEY to link Etherpad and Moodle.

PR #6388, opened 2024-05-18
Fixed proxying Etherpad with nginx.

### LINKED ISSUE #6369: Can't retrive admin panel behind Nginx reverse proxy under subdirectory (possible upstream experimental)
**Describe the bug**
Set up an instance behind Nginx reverse proxy under the subdirectory with methods from [wiki](https://github.com/ether/etherpad-lite/wiki/How-to-put-Etherpad-Lite-behind-a-reverse-Proxy#from-subdirectory-url). The main project works fine, but the admin panel failed to load. 
Upon further investigation, found that the assets are loaded relative to the webroot(`<host>/admin/assets/...`), but not to the subdirectory, which results in a failed rewrite catch for nginx.
Checking for code, `admin` seems like a separate Vite project. Its build option [`base`](https://vitejs.dev/config/shared-options.html#base) specified absolute one level to root, which would not work aforementioned. Other options are not available either, `./` would just result in resolving to webroot without `admin` level, `./admin` is not supported. Seems like an experimental method described in [Vite project document](https://vitejs.dev/guide/build.html#advanced-base-options) would work but with tweaking.
Recommendation:
- Make a fix to the path problem using the provided feature (experimental);
    - Should be the best to do, still using the current method to treat subdirectory build, but still experimental;
- Provide a configuration to declare the use of subdirectory, rewriting the base path on the fly, for the admin panel only (Just adding a workaround instruction calling people to manually write would be helpful);
    - Quickest to do but not really recommended, more quirk to do between the main project and the admin panel;
- Other possible fixes?

**To Reproduce**
1. Manual install and start the instance as described in README;
2. Configure reverse proxy with the provided config in the wiki;
3. Open the admin panel.

**Expected behavior**
Admin panel should load.

**Screenshots**
Incorrect loading when using subdirectory deployment

 - Etherpad version: v2.0.3
 - OS: CentOS 9 Stream
 - Node.js version (`node --version`): v18.19.1
 - npm version (`npm --version`): 10.2.0
 - Is the server free of plugins: It's free of plugins.

Not relevant but here you go:
 - OS: Fedora
 - Browser [e.g. chrome, safari]: Mozilla Firefox
 - Version [e.g. 22]: 125.0.3

**Smartphone (please complete the following information):**
Not relevant.

**Additional context**

SamTV12345:
This fixes the subdirectory hosting of Etherpad. The proxy hasn't been respected thus loading files from wrong directories. In this case also the socket io connection did fail. This is fixed with this pr and the updated documentation

Closes #6369

### DIFF
--- a/admin/package.json
+++ b/admin/package.json
@@ -7,6 +7,7 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "build-copy": "tsc && vite build --outDir ../src/templates/admin --emptyOutDir",
     "preview": "vite preview"
   },
   "dependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
 
 settings:
   autoInstallPeers: false
@@ -172,7 +172,7 @@ importers:
         version: 7.2.0(express@4.19.2)
       express-session:
         specifier: npm:@etherpad/express-session@^1.18.2
-        version: /@etherpad/express-session@1.18.2
+        version: '@etherpad/express-session@1.18.2'
       fast-deep-equal:
         specifier: ^3.1.3
         version: 3.1.3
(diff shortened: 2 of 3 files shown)


webzwo0i:
Is the change in the lock file intentionally?

SamTV12345:
I guess this change is because from v6 to v9. Seems like a lot has changed.

webzwo0i:
Would be great to keep those changes separate if it's not related to the PR :-)

PR #6409, opened 2024-05-27
Increase the size of maxHttpBufferSize

### LINKED ISSUE #6407: Settings aren't saved in admin UI
**Describe the bug**
In Admin UI it's possible to edit and save `settings.json` and also to restart etherpad.
When doing changes and save these, the changes to `settings.json` are not saved

**To Reproduce**
1. Go to 'Admin Panel (/admin/plugins)'
2. Click on 'settings'
3. Change something in settings.json
4. Click on 'save settings' 
5. Click on 'restart etherpad'
6. changes are not saved

**Expected behavior**
changes to settings.json should be saved when done in Admin UI

 - Etherpad version: 2.0.3

heldersepu:
Fixes #6407 

The low `maxHttpBufferSize` is causing problems with the `saveSettings` the current settings file is larger than the buffer and nothing was getting saved


webzwo0i:
Thanks for the PR and debugging the root cause of this! It's always good to split individual changes into separate PRs, e.g. avoiding `console.log` can be another PR

iirc the value for maxHttpBufferSize was choosen very carefully, as it's possible to craft artificial socket.io messages, that use a huge amount of memory/CPU during deserialization. A small difference in the size could have a huge impact.

However, maybe we can have a different value for the admin connection (used for storing settings.json) vs. all the unauth'ed client connections, as opening the admin connection is limited to privileged users?
@HMarzban Maybe you have got an idea if this is possible with socket.io?

Or is it even possible now to split large messages into multiple ones, so we can avoid hitting a limit?

heldersepu:
@webzwo0i  as requested I moved my change into a separate PR:
https://github.com/ether/etherpad-lite/pull/6412

The idea of separate buffer size for the admin VS unauth'ed clients, I like very much, will look into that
But we are storming ideas to reduce size of that payload on the Issue: https://github.com/ether/etherpad-lite/issues/6407#issuecomment-2133627864

SamTV12345:
> @JohnMcLear Yep I get a feeling the 10K was just missing `00` or just an extreme abundance of caution...
> 
> I looked into the option of two `maxHttpBufferSize` one for admin other for unauthenticated users, but do not see a way
> 
> on the frontend, _(if we don't want to touch the existing Buffer Size)_ the setting input could be a lot smaller, easiest is to remove all the comments and we already have a function for that, see: https://github.com/ether/etherpad-lite/blob/develop/admin/src/utils/utils.ts#L1 we can also add some instructions there to point admin users to the full template with full comments



> @JohnMcLear Yep I get a feeling the 10K was just missing `00` or just an extreme abundance of caution...
> 
> I looked into the option of two `maxHttpBufferSize` one for admin other for unauthenticated users, but do not see a way
> 
> on the frontend, _(if we don't want to touch the existing Buffer Size)_ the setting input could be a lot smaller, easiest is to remove all the comments and we already have a function for that, see: https://github.com/ether/etherpad-lite/blob/develop/admin/src/utils/utils.ts#L1 we can also add some instructions there to point admin users to the full template with full comments

Yes that would be the thing. We could also add the hints in the frontend. Instead of a file to edit it is a mapping of keys to values. Like any normal json object

SamTV12345:
I'll merge this so our next release still works with socket io and saving the settings. Besides that every contribution towards incremental loading is welcome.

webzwo0i:
50k is okay.

I tested with https://github.com/bcaller/kill-engine-io and 1MB and on a i7-7700 it blocks the CPU for quite a few seconds! I did not use our latest release and an older Node version, but this probably does not change much.

So definitely be careful with this setting.

SamTV12345:
So we definitely need another solution in order to transfer large chunks via socket io. I merged that because 50kB seems to be a lot far from the 1 MB danger zone. Unfortunately I couldn't really think of a good way how to do the chunking. At some point there must be a buffer that holds the state.

PR #6420, opened 2024-06-01
Clean comments

heldersepu:
## Description

Add new function to remove the comments from the settings file.
It reduces the size of the payload when we save on the admin page `.../admin/settings`

____

I was considering changing the font to `font-family: monospace;` see the before and after, 
I think it looks better that way, if we all agree we can change it.

Screenshot from 2024-06-01 18-40-29

Screenshot from 2024-06-01 18-39-15

### DIFF
--- a/admin/src/index.css
+++ b/admin/src/index.css
@@ -263,6 +263,7 @@ td, th {
     outline: none;
     width: 100%;
     resize: none;
+    font-family: monospace;
 }
 
 #response {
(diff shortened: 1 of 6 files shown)


SamTV12345:
> ## Description
> Add new function to remove the comments from the settings file. It reduces the size of the payload when we save on the admin page `.../admin/settings`
> 
> I was considering changing the font to `font-family: monospace;` see the before and after, I think it looks better that way, if we all agree we can change it.
> 
> Screenshot from 2024-06-01 18-40-29
> 
> Screenshot from 2024-06-01 18-39-15

I agree monospace is probably a better suite for that. Feel free to add that to this pr and then we can merge it :)

heldersepu:
@SamTV12345 ... Done, pushed my change adding the monospace to the `.settings`

SamTV12345:
> @SamTV12345 ... Done, pushed my change adding the monospace to the `.settings`

Thank you :).

SamTV12345 (approved):

PR #6441, opened 2024-06-09
Move away from using toast in the plugin search

### LINKED ISSUE #6440: Searching for plugins shows: Error retrieving plugins
**Describe the bug**
An error is shown searching for plugin when there is no match

**To Reproduce**
1. Go to the `admin/plugins`
2. Click on the search box
3. Search for something that does not exists
4. An error is shown: "Error retrieving plugins"

**Expected behavior**
A clear and concise message that there is no plugin matching that search instead of an error

**Screenshots**
image

heldersepu:
Fixes #6440

Refactor the plugin search behavior in the `HomePage.tsx` instead of using the toast with the message:
`Error retrieving plugins`

we will now show a much softer message in the table letting the user that no plugins are found, it also uses the correct language messages instead of the hardcoded english in the toast  

https://github.com/ether/etherpad-lite/assets/545474/c166d9f2-e690-4454-b55a-9ac372e9534e

____

Add 'results:searcherror' event to truly show errors in the search
Also add logger to the adminPlugins  

Screenshot from 2024-06-09 11-44-03

### DIFF
--- a/admin/src/pages/HomePage.tsx
+++ b/admin/src/pages/HomePage.tsx
@@ -93,25 +93,20 @@ export const HomePage = () => {
         if (!pluginsSocket) {
             return
         }
-
         pluginsSocket?.emit('search', searchParams)
-
-
         pluginsSocket!.on('results:search', (data: {
             results: PluginDef[]
         }) => {
-            if (Array.isArray(data.results) && data.results.length > 0) {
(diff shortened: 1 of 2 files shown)


SamTV12345:
Thanks for the fix. This is now a lot better and even localized :)

Issue #6194, opened 2024-02-27
Limit number of versions of a pad or delete them in Pad Settings


FredMa01:
I don't see the possibility of limiting the number of versions of a pad or cleaning up old versions.
The database grows without limit over time and exporting pads becomes long and heavy with numerous revisions .
Would it be possible to have this revision deletion option in Pad Settings?

Fred

peter-rs:
To add on to this, it should be possible to disable the revisions feature entirely (at least for read only pads), either via a plugin or through etherpad settings.
For documents that have been published using a read-only link, anyone can go back and view the author's writing process which results in: Best case (a few embarrassing spelling mistakes) or Worst Case (personally identifying information, sensitive content)

q2apro:
I absolutely need this feature too ... or at least "delete revisions older than 30 days" or alike.

I used Etherpad (via Docker) for two months now. About hundred users.

The postgresql database is zipped 400 MB now (!) This is insane.

1. We require an option to clean old revisions older than 30 days.
2. Optional an option to keep the last x revisions.
3. Definitely an option to **disable revisioning**.

Point 3 is actually the most helpful to save storage.

q2apro:
In each revision there is an option:

![image](https://github.com/ether/etherpad-lite/assets/6764548/9d7b27f8-810b-42e9-9d91-35d2fc3372f3)

Does this disable the revisions for this pad?

If so, this option we would need globally. Disable all revisions for all pads.

SamTV12345:
The next version of Etherpad will feature a built in way to manage pads and also delete them. You can also sort by revision number and can then delete pads with a lot of revision numbers.

![image](https://github.com/ether/etherpad-lite/assets/40429738/5753a729-eb67-4b31-9b71-02ad69f7c7ca)

q2apro:
That is nice to see, however, admins do not want to manually spend 20 min a day to go through all posts... also considering data privacy/protection, this would not even be allowed.

Again, we need:

1. option so revisions older than 30 days are removed
2. option to keep the last x revisions
3. option to disable revisioning

**Disabling revisioning** is the most important one.

Buesra24:
Highly agree with this issue.

[...]

PR #6442, opened 2024-06-09
Add code for revision cleanup

Gared:
This will add two methods for revision cleanup:
* deleteAllRevisions (copies only the latest version and removes the old database entries)
* deleteRevisions (deletes all revisions, but keeps the last x revisions of a pad and rewrites the first revision)

Relates to #6194

There are still some TODOs:
- [x] Remove duplicated code
- [x] Add config to change behaviour of cleanup (and enable/disable)
- [x] Test with a lot of revisions (mainly the key revisions)


src/node/utils/Cleanup.ts
```diff
@@ -0,0 +1,158 @@
...
+  if (pad.head <= keepRevisions) {
+    logger.debug('Pad has not enough revisions')
+    return false
+  }
+
+  padMessageHandler.kickSessionsFromPad(padId)
+
+  const cleanupUntilRevision = pad.head - keepRevisions
+  logger.debug('Composing changesets: ', cleanupUntilRevision)
+  const changeset = await padMessageHandler.composePadChangesets(pad, 0, cleanupUntilRevision + 1)
```
Gared, line 60: use this changeset instead of iterating over all changesets for "aText"

src/node/utils/Cleanup.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  const revisions: Revision[] = [];
+
+  for (let rev = 0; rev <= pad.head; ++rev) {
```
Gared, line 64: parallelize this block

SamTV12345 (approved):

PR #6457, opened 2024-06-15
next-page-disabled-if-zero-page

### LINKED ISSUE #6455: Paginator should be disabled with no pads
**Describe the bug**
A picture is worth 1000 words:
image

**To Reproduce**

0. On a fresh clone of the repo
1. Go to http://0.0.0.0:9001/admin/pads
2. Click on Next Page link
3. See that the page counter increases

**Expected behavior**
Next Page link should be disable if there are 0 pages

samyakj023:
fixes #6455 

Screenshot (152)

### DIFF
--- a/admin/src/pages/PadPage.tsx
+++ b/admin/src/pages/PadPage.tsx
@@ -26,7 +26,7 @@ export const PadPage = ()=>{
     const [padToDelete, setPadToDelete] = useState<string>('')
     const pages = useMemo(()=>{
         if(!pads){
-            return [0]
+            return 0;
         }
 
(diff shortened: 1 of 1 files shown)


heldersepu (approved):

heldersepu:
A bit more comments on the PR will be great! 
... maybe explain what the problem was and what are you fixing

samyakj023:
earlier problem was that in case of zero page next page is not disabled then on clicking next page counter increases on every click  
now if there is no next page next page will be disabled

SamTV12345:
Thanks for the fix :)

SamTV12345 (approved):

PR #6469, opened 2024-06-23
Feat/frontend vitest

SamTV12345:


SamTV12345:
Hi @AugustinMauroy maybe you have an idea why the vitest tests don't work. I ported settings.spec.ts to the new directory for tests but it is complaining about not finding a module. Do you know why?

AugustinMauroy:
Heyyy !
how dit you call test, I mean which command did you use ?

SamTV12345:
> Heyyy !
> 
> how dit you call test, I mean which command did you use ?

Hey :). Thanks for helping me out. It's npx vitest --config vitest.config.ts

AugustinMauroy:
Ok I got it !
it's because vitest cannot make esm and cjs in same time so when you still have require you I'll need to use import

SamTV12345:
> Ok I got it ! it's because vitest cannot make esm and cjs in same time so when you still have require you I'll need to use import

Thanks. I'll check if import works :).

PR #6496, opened 2024-07-07
fix #6495 failing Docker build when checked out as git submodule

### LINKED ISSUE #6495: `docker build` does not work when repo is a git submodule
When the repository is checked out as a submodule of another git repository, then these commands in the `Dockerfile` fail:

```
COPY --chown=etherpad:etherpad ./.git/HEAD ./.git/HEAD
COPY --chown=etherpad:etherpad ./.git/refs ./.git/refs
```

because these files obviously don't reside there, but at the other repositories' `.git` folder.

neurolabs:
When checked out as a git submodule, `.git` is not a folder containing `refs` and `HEAD`, but a file pointing to the parent repositories' git metadata.

This commit fixes the failing Docker build by making the file references copied from the `.git` folder glob patterns, which don't fail when no source file can be found.

I could not resist to fix some linter errors in the `Dockerfile` along the way.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 #
 # Author: muxator
 
-FROM node:alpine as adminBuild
+FROM node:alpine AS adminbuild
 
 WORKDIR /opt/etherpad-lite
 COPY ./ ./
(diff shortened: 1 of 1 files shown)


SamTV12345:
> When checked out as a git submodule, `.git` is not a folder containing `refs` and `HEAD`, but a file pointing to the parent repositories' git metadata.
> 
> This commit fixes the failing Docker build by making the file references copied from the `.git` folder glob patterns, which don't fail when no source file can be found.
> 
> I could not resist to fix some linter errors in the `Dockerfile` along the way.

Thanks for the fix and the linting 😅.

PR #6576, opened 2024-08-12
fix: correct the transport protocol check

eternalphane:
lt seems that the check to transport protocol introduced in #6497 does not match its description.


SamTV12345:
Thanks for the fix. Yes that seems to be an upsi by me :)

PR #6584, opened 2024-08-15
fix github plugin installation

### LINKED ISSUE #6245: installing etherpad plugins from GitHub repos is broken
## Context
So far I have installed many of our etherpad plugins from GitHub repos, e.g. `JannikStreek/ep_123` (sometimes also leveraging branch names, some plugins being forks). Since the adjustment of using the live plugin manager for installing also build time given plugins, the process is bugged.

## Problem
There are currently two problems, which make the installation process of such plugins quite confusing:

1) Plugins are correctly placed in root node_modules folder after installation. However, when the server starts and `migratePluginsFromNodeModules` is called, the plugins are *NOT* taken from the node_modules folder, but are installed from the npm registry with the same name. However, as I haven't published my plugins, it's using just *any* npm package with the same name. As many of my plugins are forks, it basically uses the parental project instead of my plugin (which is published). Thats quite confusing and took me some time to debug.

2) Adding plugins at a later point in time via `pnpm install ep_123` does add the plugin again to the node_modules folder. But restarting the server does not copy those plugins to the correct location, as this is dependent on the file `var/installed_plugins.json`. If this file is already existing (created during first server start) the second run completely ignores later installed plugins. Again quite confusing.

## Suggested solution

At least for 1) as a quick fix, I made a small change to `installer.ts` method `migratePluginsFromNodeModules` to exclude GitHub links:

```
if (!_info.resolved || _info.resolved.includes('github')) {
          // Install from node_modules directory
          await linkInstaller.installFromPath(`${findEtherpadRoot()}/node_modules/${pkg}`);
        } else {
          await linkInstaller.installPlugin(pkg);
        }
```

Maybe it should be the other way around, only use `installPlugin`when a npm url is given?

For 2) I would suggest to maybe recreate this file on startup? Not sure as I do not know the exact architectural decisions behind this.

JannikStreek:
This PR adds support for another env variable: "ETHERPAD_GITHUB_PLUGINS". I am not sure why, but adding GitHub plugins to the existing "ETHERPAD_PLUGINS" wasn't working for me anymore, the plugin manager tried to install those plugins from npm and failed as it couldn't find them. Therefore, I added support for this variable.

Also, I fixed a crash for the development docker image, as the plugin manager source wasn't copied to the docker file (the src folder wasn't in general). Therefore, I got this error which I fixed:

```
   const err = new Error(message);
11.11               ^
11.11 
11.11 Error: Cannot find module 'ep_etherpad-lite/static/js/pluginfw/installer'
11.11 Require stack:
11.11 - /opt/etherpad-lite/bin/plugins.ts
```

However, I realized that installing plugins during build time doesn't work anyway as these plugins get overwritten during the mount of the bin and src folders.

closes #6245


SamTV12345:
Hi thanks for taking over the GitHub plugin issue. I couldn't get that to work. So far it worked for me. Can you add test cases for the regexs?

JannikStreek:
> Hi thanks for taking over the GitHub plugin issue. I couldn't get that to work. So far it worked for me. Can you add test cases for the regexs?

will do. yes it also worked for me before, the live plugin manager install command should work with GitHub repos. But somehow now I get 405 errors from npm...not sure what changed.

JannikStreek:
> Hi thanks for taking over the GitHub plugin issue. I couldn't get that to work. So far it worked for me. Can you add test cases for the regexs?

I would be happy to provide proper test cases, however, so far the bin workspace does not include any test runner and neither test cases as far as I see. So this would boil down in adding the complete testing setup for the bin workspace (which will make this PR way bigger than the idea was). Was that your intention? Do you have test runners in mind you want to use or shall I just pick what I find best?

SamTV12345:
> > Hi thanks for taking over the GitHub plugin issue. I couldn't get that to work. So far it worked for me. Can you add test cases for the regexs?
> 
> 
> 
> I would be happy to provide proper test cases, however, so far the bin workspace does not include any test runner and neither test cases as far as I see. So this would boil down to adding the complete testing infrastructure for the bin workspace (which will make this PR way bigger than I planned it). Was that your intention? Do you have test runners in mind you want to use or shall I just pick what I find best?

You're right. That would be too much for a single PR and its just a modification in the bin folder.

PR #6654, opened 2024-09-11
Update docker-compose.dev.yml

h4km4k:
If path is "/var/lib/postgresql/data/pgdata" docker creates a unnamed volume.


SamTV12345:
Thanks for the pr. I checked and it also seems that the complete data path is mapped into the docker container everywhere :)

PR #6655, opened 2024-09-11
Update docker-compose.yml

h4km4k:
If path is "/var/lib/postgresql/data/pgdata" docker creates a unnamed volume.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -42,7 +42,7 @@ services:
     # ports:
     #   - "5432:5432"
     volumes:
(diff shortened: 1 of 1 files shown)


SamTV12345:
@JannikStreek Do you have a suggestion what is better? I'm a bit clueless

h4km4k:
For me it works without any problems with "/var/lib/postgresql/data" and i can not see why it shouldn't be this path.

In "/var/lib/postgresql/data" is only the "pgdata" folder, maybe postgres wants to create this folder by itsself.

PR #6705, opened 2024-10-09
fix(admin): Fixed updating plugins

SamTV12345:
This pull request addresses the issue that plugins can't be updated. This is because at first the setInterval function is not triggered resulting in nothing happening. You'd have to wait 60 minutes until the update button is shown.

### DIFF
--- a/admin/src/pages/HomePage.tsx
+++ b/admin/src/pages/HomePage.tsx
@@ -4,7 +4,7 @@ import {InstalledPlugin, PluginDef, SearchParams} from "./Plugin.ts";
 import {useDebounce} from "../utils/useDebounce.ts";
 import {Trans, useTranslation} from "react-i18next";
 import {SearchField} from "../components/SearchField.tsx";
-import {Download, Trash} from "lucide-react";
+import {ArrowUpFromDot, Download, Trash} from "lucide-react";
 import {IconButton} from "../components/IconButton.tsx";
 import {determineSorting} from "../utils/sorting.ts";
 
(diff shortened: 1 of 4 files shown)


admin/src/pages/HomePage.tsx
```diff
@@ -111,6 +111,7 @@ export const HomePage = () => {
 
         // check for updates every 5mins
         const interval = setInterval(() => {
+          console.log("Checking for updates")
```
Gared, line 114: Just for Debugging, right? Maybe remove?

admin/src/pages/HomePage.tsx
```diff
@@ -159,6 +160,9 @@ export const HomePage = () => {
         })
     }, 500, [searchTerm])
 
+
+  console.log("Installed plugins", installedPlugins)
```
Gared, line 164: Same Here. Only for Debugging?

admin/src/pages/HomePage.tsx
```diff
@@ -4,15 +4,16 @@ import {InstalledPlugin, PluginDef, SearchParams} from "./Plugin.ts";
...
 import {IconButton} from "../components/IconButton.tsx";
 import {determineSorting} from "../utils/sorting.ts";
 
 
 export const HomePage = () => {
     const pluginsSocket = useStore(state=>state.pluginsSocket)
     const [plugins,setPlugins] = useState<PluginDef[]>([])
-    const [installedPlugins, setInstalledPlugins] = useState<InstalledPlugin[]>([])
+  const installedPlugins = useStore(state=>state.installedPlugins)
+  const setInstalledPlugins = useStore(state=>state.setInstalledPlugins)
```
Gared, line 16: I think you know what you do here, but can you explain why it is needed to have installedPlugins and setInstalledPlugins?

Gared (approved):

admin/src/pages/HomePage.tsx
SamTV12345: Yeah. So I had it fixed that the plugins are updated but somehow everytime I visited the page it showed for an instant the list of installed plugins and then it disappeared. The problem is that the useEffect does not contain the installedPlugins in its dependency array. That way the useEffect always contained an empty list of installed plugins. You could fix that by adding installedPlugins into the dependency array but that would cause a circle. The installedPlugins will be updated but the installedPlugins cause another rerender. Later in the code there is this useStore().getState() call which extracts the current installedPlugins from the store. This does not cause a rerender because we simply call a function that returns a value. That way I broke the cycle and the installedPlugins are correctly displayed with their updatable state.

Issue #6701, opened 2024-10-09
I think etherpad-lite is not GDPR compliant


devnull4242:
I think Etherpad Lite is not GDPR compliant because of:

**Individuals have an absolute right to have their data deleted (right to be forgotten)** 

Please add the possibility to delete all old entries of an etherpad. Thanks.

SamTV12345:
You can delete Etherpads via the admin panel for this. There is also the possibility to move a pad to one revision. Is there anything else we could do to improve the GDPR compliance?

devnull4242:
I think the user must have the possibilty itself. I think that is not possible.

SamTV12345:
> I think the user must have the possibilty itself. I think that is not possible.
> 
> I also don't know why such a basic function is not simply built in. When I have finished working on my Etherpad with a few people, the result and the way there is nobody's business. And if I really need it, I can also export it. And if I don't want to delete it, I don't need to delete it.

So you'd like to delete the Etherpad yourself. Who should be able to do that? The one that created the pad? If we allow anybody you could have trolls on the public instances that just delete all pads over and over again.

SamTV12345:
> For GDPR-compliance it's not necessary that the user is allowed to delete their own data. [Art. 17](https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:32016R0679&from=EN#d1e2606-1-1) provides you the right to demand the deletion of your (personal) data, but doesn't require that you're able to do it yourself.

So as long as the user hands the pads he worked on to the pad admin they can be safely deleted.

devnull4242:
I don't think anyone can guess Etherpad names unless you take "test" or "12345". Conversely, I once created an Etherpad https://etherpad.wikimedia.org/p/6701. Of course, only those who know it write there. However, I think it is completely unrealistic to contact Wikimedia to delete this Etherpad in this case. It doesn't even have an e-mail address.

dcht00:
First of all, I am not a lawyer, so everything I write here is only my understanding.

This seems false:

> **Individuals have an absolute right to have their data deleted (right to be forgotten)**

Simply put "personal data" in GDPR is not data *by* person, but *about* a person. Furthermore, this data needs to be processed in some way to fall under GDPR.

https://gdpr.eu/eu-gdpr-personal-data/

GDPR does not at all regulate Bob's right to have his (let alone others'!) contributions to some generic pad deleted.

dcht00:
[...]

PR #6730, opened 2024-10-28
feat(pad-settings): added possibility to delete pad by the creator

SamTV12345:
Added button to delete a pad. This can only be performed by the creator. 

A usecase would be a teacher creates an Etherpad and scribbles down some notes/questions pupils need to answer. After the lesson he/she can delete the pad.


SamTV12345:
Closes #6701

PR #6743, opened 2024-11-05
Use temporary directory for esbuild

erdnaxe:
This patch proposes to move `<root>/var/js/` to a temporary directory.

This simplifies the setup for system administrators as they no longer have to give special permissions to the user running Etherpad on `var/js/` folder. It is also required for packaging Etherpad in NixPkgs: https://github.com/NixOS/nixpkgs/pull/353814.


SamTV12345:
Thanks for the pull request :) . I improved it a bit by ditching writing to disk. Our JavaScript files are like a few kilobytes so writing to disk isn't necessary and we can easily regenerate them at startup.

erdnaxe:
Thank you for the review! 

I noticed you reverted `resolveDir: path.join(settings.root, 'var','js'),` in your [commit](https://github.com/ether/etherpad-lite/pull/6743/commits/b267a72e1fe413d14fb312ddaa5178b0f901deb1#diff-98e312e1d475ff696ae934e218557d9bd9edfa39467e53cb33d5847b19674946R92), which means that ./var/js/ folder is used but never written into. Couldn't we `resolveDir:  settings.root` in esbuild configuration? Am I missing something?

PR #6930, opened 2025-04-07
settings: use correct syntax for default values containing "\n" in settings.json.template

muxator:
79406051fa66 added support for including an escaped `\n` in the default value of an interpolated setting, but the example in `settings.json.template` and `settings.json.docker` contained a slight syntax error.
This fixes it.

No functional changes.


SamTV12345:
Thank you for the fixes. Looks fine now. Good that this is fixed after 5 years :)

Issue #6893, opened 2025-03-06
Allow access to padbootsrap-XXX.min.js without authorization


n-goncalves:
**Is your feature request related to a problem? Please describe.**

Some resources can be accessed without authorization/authentication like those in the  `/static` folder or the API endpoint.
However padboostrap-XXX-min.js isn't, this can be a problem if your Etherpad instance has both `requireAuthentication` and `requireAuthorization` set to true. 
In that case, you need to add a plugin with a `preAuthorize` hook to specifically allow access to this file, without it the pad fails to load. This is done in this plugin: [ep_auth_citizenos](https://github.com/citizenos/ep_auth_citizenos/blob/32fdefc66d5000610b667e1ab23e759102344780/serverHooks.js#L296)

**Describe the solution you'd like**

If the resource is supposed to be available without authorization, it shouldn't trigger authorization hooks.

**Describe alternatives you've considered**

If I understood wrong and this is intended, it might be good to document somewhere so people know how it support. 

Thank you for reading this.

SamTV12345:
Fixed on develop

PR #6931, opened 2025-04-07
fix(frontend): add timesliderBootstrap to static resources

n-goncalves:
Basically the same thing as I reported in https://github.com/ether/etherpad-lite/issues/6893, except it's the timeslider. Since it's not loaded when accessing a pad, I missed it in my initial report.

Also I'm not sure but looking at the `src/node/hook/express/specialpages.ts` maybe a entry should be added for indexBootstrap ? I wasn't able to find where it is used so that's why I didn't include it in this PR.


SamTV12345:
Thanks for the pr. Yes we also need to exclude the timeslider.

PR #6934, opened 2025-04-07
Update importexport.ts - fixes 4805

### LINKED ISSUE #4805: replace doc export with docx format
* * *

name: Feature request
title: ''
labels: enhancement
assignees: 

* * *

DOC format is old, DOCX format is more widely supported

Replace the DOC export with DOCX export, both for abiword and libreoffice

Current DOC export works, this is not a pressing issue, just an idea.

What do you think about this? Is this a simple change, or is it more complicated?

lonesomewalker:
this changes the export from doc to docx.
This fixes #4805
A lot of other projects (like BBB) rely on etherpad lite, but nobody seemed to care about :-)

### DIFF
--- a/src/node/hooks/express/importexport.ts
+++ b/src/node/hooks/express/importexport.ts
@@ -28,7 +28,7 @@ exports.expressCreateServer = (hookName:string, args:ArgsExpressType, cb:Functio
   args.app.use('/p/:pad/:rev?/export/:type', limiter);
   args.app.get('/p/:pad/:rev?/export/:type', (req:any, res:any, next:Function) => {
     (async () => {
-      const types = ['pdf', 'doc', 'txt', 'html', 'odt', 'etherpad'];
+      const types = ['pdf', 'docx', 'txt', 'html', 'odt', 'etherpad'];
       // send a 404 if we don't support this filetype
       if (types.indexOf(req.params.type) === -1) {
         return next();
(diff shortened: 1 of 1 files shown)


SamTV12345:
Thanks for the pr. Well the problem is that I think Libreoffice only supports the --convert-to doc instead of docx. You can't simply change the file ending and call it a day. Doc and docx have very different file structuring. Docx is the modern and open standard while doc is sealed behind doors.

lonesomewalker:
This is true, but:
as stated in the issue, many organizations and public offices do block doc in a hard way, but let docx pass, because it is guaranteed without macros.
Word itself can then easily recognize it, also Libreoffice does so.
As i have seen, the PR is not complete... :-/ in src/node/utils/LibreOffice.js there is also some work to do, right?

lonesomewalker:
I see that there is MUCH more work to do, because i forgot Abiword.
This means, there also needs to be a if/else, otherwise it breaks this -> is a bad idea.

To be honest, i am not that familiar with etherpad-lite.
All filters can be seen here:
https://git.libreoffice.org/core/+/refs/heads/master/filter/source/config/fragments/filters

SamTV12345:
Wikipedia states that abiword can do docx https://gitlab.gnome.org/World/AbiWord That's their gitlab but at least for me their documentation page does not work. Nothing happens.

PR #6993, opened 2025-05-17
Consider device dark mode setting

Gared:
This will consider the dark mode setting of the device to switch automatically to the dark-mode theme of etherpad

### DIFF
--- a/src/static/js/skin_variants.ts
+++ b/src/static/js/skin_variants.ts
@@ -1,38 +1,43 @@
 // @ts-nocheck
 'use strict';
 
+const containers = ['editor', 'background', 'toolbar'];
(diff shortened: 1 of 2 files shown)


src/static/js/skin_variants.ts
```diff
@@ -1,38 +1,43 @@
...
+
+  colors.forEach((color) => {
+    containers.forEach((container) => {
+      domsToUpdate.forEach((el) => { el.removeClass(`${color}-${container}`); });
+    });
+  });
+
+  domsToUpdate.forEach((el) => { el.removeClass('full-width-editor'); });
+
+  domsToUpdate.forEach((el) => { el.addClass(newClasses.join(' ')); });
```
SamTV12345, line 23: Can we combine these two calls?

PR #7051, opened 2025-07-28
Fix interaction URL and diagnostics for subdirectory setup

### LINKED ISSUE #4191: Connection diagnostics don't work

AbhishekG-Codes:
What problem does this solve?

This fixes an issue where Etherpad's SSO interaction endpoint and diagnostic routes (e.g. `/ep/pad/connection-diagnostic-info`, `/jserror`, `/api`) fail to function when Etherpad is served from a subdirectory (like `/pad`).

Without this fix, form submissions and client-side error reporting do not reach the server when `settings.subdirectory` is set.

Current vs Desired Behavior

Current (broken):
  When deployed under a subdirectory, form submissions like `/pad/interaction/:sessionId` fail because the server expects the root path.

Desired (fixed):
  All client-side form actions and backend routes respect the `subdirectory` setting and continue working.

Screenshot / Visuals

Not applicable, routing logic fix.

- No breaking changes.
- Routes are conditionally prefixed using `settings.subdirectory || ''`.
- Frontend dynamically builds the form URL using `window.location.pathname`.

closes #4191

### DIFF
--- a/src/node/hooks/express/apicalls.ts
+++ b/src/node/hooks/express/apicalls.ts
@@ -5,10 +5,13 @@ const clientLogger = log4js.getLogger('client');
 const {Formidable} = require('formidable');
 const apiHandler = require('../../handler/APIHandler');
 const util = require('util');
+const settings = require('ep_etherpad-lite/node/utils/Settings');
+
+const subdir = settings.subdirectory || '';
 
 exports.expressPreSession = async (hookName:string, {app}:any) => {
   // The Etherpad client side sends information about how a disconnect happened
-  app.post('/ep/pad/connection-diagnostic-info', async (req:any, res:any) => {
+  app.post(`${subdir}/ep/pad/connection-diagnostic-info`, async (req:any, res:any) => {
     const [fields, files] = await (new Formidable({})).parse(req);
     clientLogger.info(`DIAGNOSTIC-INFO: ${fields.diagnosticInfo}`);
     res.end('OK');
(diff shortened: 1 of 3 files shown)


AbhishekG-Codes:
> Hi. Thanks for the pull request. Are you sure this is necessary? Normally with nginx you can e.g. strip the pathprefix of your subdirectory and then the backend only gets the normal path without subdirectory.


Thanks for the comment, @SamTV12345 

You're absolutely right that in many production setups, reverse proxies like nginx can be configured to strip subdirectory prefixes before forwarding to Etherpad.

But the issue raised in #4191 highlights that Etherpad’s client-generated routes (e.g. for connection diagnostics) assume root-relative paths. This causes broken URLs like "/p/ep/pad/connection-diagnostic-info" when Etherpad is deployed under a subdirectory like /pad.

While nginx rewrites can fix this, it puts the burden on every operator to know and correctly configure all internal paths some of which aren't obvious (like /jserror, /api, etc.). This PR aims to make Etherpad natively aware of settings.subdirectory, avoiding this problem entirely and making deployments more portable and reliable.

That said, if you'd prefer this be handled via documentation or config options, happy to adjust the approach.

PR #7072, opened 2025-08-17
After delete the user will now redirected to home page

prashant1177:


prashant1177:
It was my pleasure to contribute

PR #7093, opened 2025-08-25
fix "pnpm run plugins install ep_xxx"

prigaux:
With previous code, "pnpm run plugins i ep_xxx" is working correctly, but "pnpm run plugins install ep_xxx" does not work correctly (since var `registryPlugins` is empty)

Either "pnpm run plugins install ep_xxx" should be disallowed (by removing `case "install"`) or this fix should be applied.


SamTV12345:
Thanks for the fix. Let's better apply your fix :)

PR #7104, opened 2025-08-28
correcting plugin install doc

dcht00:
correcting plugin install doc


SamTV12345:
Thanks for the update :)

PR #7141, opened 2025-09-27
plugins.ts - absence of case "remove"

WC2-Foundation:


SamTV12345:
Thanks for the merge request. Good idea 😊

PR #7277, opened 2025-12-23
Rename token transfer cookies to Etherpad-specific names to avoid conflicts

GauravWaghmare23:
This change updates the cookies set by the /tokenTransfer/:token endpoint to use Etherpad‑specific names:

token → epToken

prefsHttp → epPrefsHttp

Using generic cookie names like token can easily collide with existing session or auth cookies used by apps that embed Etherpad under the same parent domain (for example, sessionID, token, etc.). Renaming these cookies to Etherpad‑specific names improves interoperability and avoids unexpected overwrites while keeping the current behavior intact.

Implementation details:

The persistence format in the database (tokenTransfer:<id>) and the response payload remain unchanged.

Cookie options (path: "/" and one‑year maxAge) are preserved so existing clients can migrate by simply reading the new cookie names.

The change is limited to the token transfer Express hook and does not modify any other cookies or session logic.

### DIFF
--- a/src/node/hooks/express/tokenTransfer.ts
+++ b/src/node/hooks/express/tokenTransfer.ts
@@ -1,45 +1,51 @@
 import {ArgsExpressType} from "../../types/ArgsExpressType";
-const db = require('../../db/DB');
-import crypto from 'crypto'
-
+const db = require("../../db/DB");
+import crypto from "crypto";
 
 type TokenTransferRequest = {
   token: string;
-  prefsHttp: string,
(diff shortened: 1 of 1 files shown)


SamTV12345:
Hi @GauravWaghmare23 I think that change is not easy to change. Etherpad uses the cookie token to identify a session everywhere.  Changing the name of the cookie would render this to be incompatible if the cookie is not renamed. We could rename the cookie but then it would destroy all open sessions. 
image

PR #7316, opened 2026-02-06
feat: add logout and exit buttons to settings popup (fixes #5633)

### LINKED ISSUE #5633: Logout / Exit button(s)
* * *

name: Logout / Exit button(s)
about: 
title: ''
labels: enhancement
assignees: 

* * *

Currently there is no way to forget the logged-in user, except from manually remove cookies. It would be useful to have a "Logout" button that clears cookies and returns to the home page. Maybe this is a feature for plugins like ep_hash_auth.

Another problem is that there is no way to easily return to the main "home" page. The user must manually change the url. An "Exit" button would be useful.

deepshekhardas:


SamTV12345:
Hi @deepshekhardas thanks for the pull request. There are some logs inside. Would be great if you could remove them. What would be the upside now that we have a home button on the top right of the pad and one in the settings popup

src/static/js/pad_editor.ts
```diff
@@ -86,21 +86,35 @@ const padeditor = (() => {
...
-          pad.collabClient.sendMessage({type: 'PAD_DELETE', data:{padId: pad.getPadId()}});
+          pad.collabClient.sendMessage({ type: 'PAD_DELETE', data: { padId: pad.getPadId() } });
           // redirect to home page after deletion
           window.location.href = '/';
         }
       })
 
+      // exit button - redirect to home without clearing cookies
+      $('#exit-pad').on('click', () => {
+        window.location.href = '/';
```
SamTV12345, line 97: This needs to be a relative path

PR #7327, opened 2026-02-13
Remove advertising of Wikimedia etherpad instance

pppery:
Per https://phabricator.wikimedia.org/T371591 they don't want their etherpad instance advertised for unrelated uses.


SamTV12345:
Thanks for the notice. We'll remove the notice

PR #7369, opened 2026-03-15
fix: improve accessibility with ARIA labels and i18n support

deepshekhardas:
Adds ARIA labels with internationalization support to improve accessibility in the editor.

### DIFF
--- a/settings.json.template
+++ b/settings.json.template
@@ -261,7 +261,9 @@
     "rtl":              false,
     "alwaysShowChat":   false,
     "chatAndUsers":     false,
-    "lang":             null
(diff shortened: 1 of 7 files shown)


JohnMcLear:
@deepshekhardas your pull requests are still not inline with requirements.  Your IDE or whatever LLM is not following the agents rule...

SamTV12345:
> @deepshekhardas your pull requests are still not inline with requirements. Your IDE or whatever LLM is not following the agents rule...

Still funny that someone actually wastes his energy with creating pull requests :)

PR #7375, opened 2026-03-16
fix: catch all errors that can occur during plugin installation and log the error

### LINKED ISSUE #7365: Uncaught exception causing crash during startup when not connected to the Internet
**Describe the bug**
Line 127 of src/static/js/pluginfw/installer.js calls linkInstaller.installPlugin without using a try/catch block. If a plugin is installed but the external plugin source is unreachable (such as from an isolated network), etherpad fails to start rather than just continuing and using the plugin from disk. 

I believe this could be handled with a simple try/catch block. Inside the catch, you could log a warning if the plugin is already installed and then just continue 

**To Reproduce**
1. Install etherpad with a plug-in 
2. Disable outbound traffic from the container (except to your workstation)
3. Restart the etherpad container

**Expected behavior**
Etherpad should continue operating with already installed plugins while operating in an isolated network

 - Etherpad version:2.6.1
 - OS: Ubuntu 22.04
 - Node.js version (`node --version`): 20.19.2
 - npm version (`npm --version`): 11.5.1
 - Is the server free of plugins: no

 - OS: macos
 - Browser: chrome

SamTV12345:

### DIFF
--- a/src/.npmrc
+++ b/src/.npmrc
@@ -1,1 +1,0 @@
-auto-install-peers=false
--- a/src/static/js/pluginfw/installer.ts
+++ b/src/static/js/pluginfw/installer.ts
@@ -124,7 +124,11 @@ export const checkForMigration = async () => {
 
   for (const plugin of installedPlugins.plugins) {
     if (plugin.name.startsWith(plugins.prefix) && plugin.name !== 'ep_etherpad-lite') {
-      await linkInstaller.installPlugin(plugin.name, plugin.version);
+      try {
+        await linkInstaller.installPlugin(plugin.name, plugin.version);
+      } catch (e) {
+        logger.error(`Error installing plugin ${plugin.name} with version ${plugin.version}: ${e}`);
+      }
     }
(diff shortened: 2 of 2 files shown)


JohnMcLear:
You delete .npmrc for some reason?  Is your AI haunted?

SamTV12345:
> You delete .npmrc for some reason? Is your AI haunted?

No. 😂 I used my AI in my brain. We don't need that file anymore because we don't have any peer dependencies anymore.

JohnMcLear:
For best practice should that be two pull requests?  I'm pretty flexible with this stuff so I'm happy to just be chill about it :p

SamTV12345:
> For best practice should that be two pull requests? I'm pretty flexible with this stuff so I'm happy to just be chill about it :p

As long as you can review the files in a timely manner I am okay with that. At work we have the boyscout rule to always leave the place in your code better than before. :) As long as you don't put two things with hundreds of files changed I am okay with it too :p

PR #7408, opened 2026-03-31
Fix webkit frontend tests silently passing when they fail

### LINKED ISSUE #7405: tests are completely broken
Pass when they fail..

https://github.com/ether/etherpad-lite/actions/runs/23790269623

Not sure how long it's been like this ;z

JohnMcLear:
## Summary
- Remove `|| true` from the webkit Playwright test step in `frontend-tests.yml`
- This was swallowing non-zero exit codes, causing the workflow to always report success regardless of actual test results

## Test plan
- [ ] Verify the webkit frontend test job now fails when tests fail
- [ ] Confirm chrome and firefox test jobs are unaffected

Fixes #7405

### DIFF
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -192,7 +192,7 @@ jobs:
           done
           cd src
           gnpm exec playwright install webkit --with-deps
-          gnpm run test-ui --project=webkit || true
+          gnpm run test-ui --project=webkit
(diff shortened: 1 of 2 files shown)


JohnMcLear:
This should fail on testing because tests were failing but the test runner was ignoring failed tests

😮‍💨 

I'll work on the bugs that make the tests fail next...

PR #7420, opened 2026-04-01
Add CJS compatibility for Settings and ReadOnlyManager

JohnMcLear:
## Summary
- Add `module.exports` shim to Settings.ts and ReadOnlyManager.ts
- Both modules use `export default` which breaks plugins that `require()` them
- Plugins get `{ default: ... }` instead of the module directly
- This affects 28+ plugins that `require('ep_etherpad-lite/node/utils/Settings')`
- And ep_comments_page which requires ReadOnlyManager

## Test plan
- [ ] Verify `require('ep_etherpad-lite/node/utils/Settings')` returns the settings object directly
- [ ] Verify `import settings from 'ep_etherpad-lite/node/utils/Settings'` still works
- [ ] Run etherpad-lite backend tests

### DIFF
--- a/src/node/db/ReadOnlyManager.ts
+++ b/src/node/db/ReadOnlyManager.ts
@@ -74,7 +74,7 @@ const getIds = async (id:string) => {
   return {readOnlyPadId, padId, readonly};
 };
 
-export default {
+const readOnlyManager = {
   isReadOnlyId,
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Superseded by #7421 which included the Settings CJS compatibility fix.

PR #7431, opened 2026-04-01
fix: prevent crash when pasting bare <li> without parent list element

### LINKED ISSUE #6665: Pad crashes on pasting html containing <li>
Env: Ubuntu + Firefox

Reproduce:
`echo '<li>this crashes</li>' | xclip -selection clipboard -t text/html
`
(or just copy any html with a list item)
... and paste into a pad

Result
```
TypeError: b.lineAttributes.list is undefined
at https://#####/padbootstrap-sxsSWavtLWA.min.js at line 20
ErrorId: vYIhQJ3X8o7VeqsTlv2t
Uncaught exception
URL: https://#####/p/test-wtfpaste
UserAgent: Mozilla/5.0 (X11; Linux x86_64; rv:129.0) Gecko/20100101 Firefox/129.0
```

Discovered by trying to paste a ChatGPT response including a list of items into a pad.

JohnMcLear:
## Summary

- **Fix** (`contentcollector.ts`): When a bare `<li>` element (no parent `<ul>`/`<ol>`) is encountered during content collection, default it to `bullet1` list type instead of crashing. The list state is properly cleaned up via `oldListTypeOrNull` so subsequent non-list elements aren't affected.
- **Guard** added on `state.lineAttributes.list.indexOf()` call to prevent the TypeError even if the list attribute is somehow undefined.
- **Tests**: 3 new backend contentcollector tests covering bare `<li>`, multiple bare `<li>` elements, and mixed bare `<li>` with normal content.

## Root Cause

`contentcollector.ts:509` calls `state.lineAttributes.list.indexOf('number')` when processing an `<li>` element. This assumes `state.lineAttributes.list` was set by a parent `<ul>`/`<ol>` via `_enterList()`. When pasting HTML with bare `<li>` elements (common when copying from ChatGPT or malformed HTML), no parent list element exists, so `.list` is `undefined` and `.indexOf()` throws:

- **Firefox:** `TypeError: b.lineAttributes.list is undefined`
- **Chrome:** `Cannot read properties of undefined (reading 'indexOf')`

## Test plan

- [x] Backend: `contentcollector.ts` — 108 tests passing (3 new, 105 existing), 0 failing
- [x] Backend: full suite — 744 passing, 0 failing
- [ ] Manual: Copy `<li>this should not crash</li>` as HTML to clipboard and paste into a pad

Fixes https://github.com/ether/etherpad-lite/issues/6665

### DIFF
--- a/src/static/js/contentcollector.ts
+++ b/src/static/js/contentcollector.ts
@@ -504,9 +504,14 @@ const makeContentCollector = (collectStyles, abrowser, apool, className2Author) 
             // See https://github.com/ether/etherpad-lite/issues/2412 for reasoning
             if (!abrowser.chrome) oldListTypeOrNull = (_enterList(state, undefined) || 'none');
           } else if (tname === 'li') {
+            // If the <li> has no parent <ul>/<ol> (e.g., pasted bare HTML), default to bullet list.
+            // See https://github.com/ether/etherpad-lite/issues/6665
+            if (!state.lineAttributes.list) {
+              oldListTypeOrNull = (_enterList(state, 'bullet1') || 'none');
+            }
             state.lineAttributes.start = state.start || 0;
             _recalcAttribString(state);
-            if (state.lineAttributes.list.indexOf('number') !== -1) {
+            if (state.lineAttributes.list && state.lineAttributes.list.indexOf('number') !== -1) {
               /*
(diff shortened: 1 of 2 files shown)


SamTV12345 (approved):
Great merge request. Nothing to complain about bud :) . I'll review the other two in the evening/night.

PR #7465, opened 2026-04-05
Update versions in markdown files

Gared:
Update static version information in our markdown documentation


JohnMcLear:
What's the strat RE updating this automatically?!

PR #7478, opened 2026-04-06
fix: preserve non-breaking spaces instead of converting to regular spaces

### LINKED ISSUE #3037: Text input not allowing non-breaking space

JohnMcLear:
## Summary

- Three locations stripped U+00A0 (non-breaking space) characters via `.replace(/\xa0/g, ' ')`
- Removed the conversion from `contentcollector.ts`, `ace2_inner.ts`, and `Pad.ts` cleanText
- Updated contentcollector tests to expect preserved nbsp characters
- This fixes French typography and other languages that use non-breaking spaces

## Test plan

- [x] Type check passes
- [x] All 752 backend tests pass (7 test expectations updated)
- [ ] Manual: type a non-breaking space (e.g., Ctrl+Shift+Space), verify it's preserved

Fixes #3037


JohnMcLear:
Closing this PR — Qodo correctly identified that removing the `\xa0` → space conversion breaks the rendering round-trip.

**The problem**: Etherpad's `domline.processSpaces()` intentionally converts regular spaces to `&nbsp;` for HTML display (to prevent whitespace collapsing). The content collector's `textify()` was converting them back to regular spaces — this round-trip is by design. Removing the conversion causes every regular space in the pad to silently become a non-breaking space in the stored document model after any edit.

**To properly fix #3037**, the rendering layer needs to stop using `&nbsp;` for normal space display (e.g., by switching to CSS `white-space: pre-wrap` or `break-spaces`). That's a larger architectural change that needs careful testing. Tracking as a separate effort.

PR #7515, opened 2026-04-14
Fix Docker: remove stale lockfile after workspace override

JohnMcLear:
## Summary

Follow-up to #7514. The previous fix overwrote `pnpm-workspace.yaml` but the lockfile still referenced the removed workspace packages (admin, doc, ui), causing pnpm to warn and produce incomplete installs.

Removing the lockfile lets pnpm resolve fresh for the production-only workspace.

## Test plan

- [ ] Docker CI container starts and passes health check


JohnMcLear:
Superseded by #7516 which bumps ueberdb2 to 5.0.43 with proper production dependencies.

PR #7523, opened 2026-04-16
fix: explicitly ignore scarf build scripts to fix bin install CI

JohnMcLear:
## Summary
- Fixes the `Update Plugins` workflow failure on develop caused by my earlier commit (2f632e5a) that removed `@scarf/scarf` from `onlyBuiltDependencies`
- pnpm 10.7+ requires every encountered build script to be either allow-listed or explicitly deny-listed, otherwise it errors with `ERR_PNPM_IGNORED_BUILDS`
- Add `@scarf/scarf` to a new `ignoredBuiltDependencies` list so pnpm silently skips its install-time telemetry script instead of failing

## Context
`scarf` is pulled in as a transitive dep via `swagger-ui-dist`. It's pure install-time telemetry (phones home on every install) — we don't want it running, but we need pnpm to know that so it doesn't error out. Using `ignoredBuiltDependencies` is the clean way to express "block this, don't fail."

## Test plan
- [x] Verified `pnpm install` in `./bin/` no longer fails locally
- [x] Verified root `pnpm install` still succeeds
- [ ] Confirm `Update Plugins` CI now passes on develop

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -6,3 +6,8 @@ packages:
   - ui
 onlyBuiltDependencies:
   - esbuild
+# Explicitly ignore build scripts we don't want to run. Listing them here
+# stops pnpm from failing with ERR_PNPM_IGNORED_BUILDS when they're
+# encountered as transitive deps (e.g. scarf pulled in via swagger-ui-dist).
+ignoredBuiltDependencies:
+  - '@scarf/scarf'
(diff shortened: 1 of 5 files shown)


JohnMcLear:
Updated to also include admin-spec test serialization fix (3134860d). Two changes now:

1. **Fix bin install CI breakage** (`pnpm-workspace.yaml`) — add `@scarf/scarf` to `ignoredBuiltDependencies` so pnpm silently skips its telemetry script instead of failing with `ERR_PNPM_IGNORED_BUILDS`
2. **Serialize admin-spec tests** — admin tests mutate global server state (install/uninstall plugins, restart etherpad); running them in parallel across browsers caused ~13% flake rate on develop
   - Drop firefox from `test-admin` (admin UI is browser-agnostic, chromium alone is sufficient)
   - Add `test.describe.configure({ mode: 'serial' })` to each admin spec file as a safety net

PR #7535, opened 2026-04-17
checkPlugin: flag absolute /static/plugins/ asset paths in templates (#5203)

### LINKED ISSUE #5203: Extend checkPlugins.js to check for static paths
Details @ https://github.com/ether/ep_embedmedia/pull/4/files

Checking for "/static" would be ideal and also could be autofixed with a ``/s/static/..\/static``

Also maybe this should be in esliint @rhansen ??

JohnMcLear:
Closes #5203.

Plugin templates that reference assets with absolute paths like
\`/static/plugins/ep_foo/x.css\` silently break any Etherpad instance
hosted behind a reverse proxy at a sub-path (e.g. `https://example.com/etherpad/pad\`) — the browser resolves the path against the domain root instead of the proxy prefix and the asset 404s. The correct form is \`../static/plugins/...\` (relative). \[ep_embedmedia#4\](https://github.com/ether/ep_embedmedia/pull/4/files) fixed one plugin manually; #5203 asked for a mechanical check.

## Summary
- Walk each plugin's \`templates/\` and \`static/\` for \`*.ejs\` / \`*.html\` files.
- Scan for \`/static/plugins/\` occurrences that are **not** preceded by a URL scheme (\`:\`), a dot, a slash, or a word char — so `\\"https://host/static/plugins/...\\"` and already-correct \`../static/plugins/...\` don't match.
- Emit a \`WARN\` line describing the bad path and how to fix it.
- In \`autofix\` mode, rewrite them in place to \`../static/plugins/\`.

## Test plan
- [x] Regex verified against six cases (absolute bad, absolute in URL, already-relative, embedded in src/href, /static/ outside /plugins/, bare path).
- [x] \`pnpm run ts-check\` clean.

### DIFF
--- a/bin/plugins/checkPlugin.ts
+++ b/bin/plugins/checkPlugin.ts
@@ -374,6 +374,45 @@ log4js.configure({
                  'Translation files help with Etherpad accessibility.');
   }
 
+  // Check template files for absolute `/static/plugins/...` asset paths.
+  // Those break any Etherpad instance hosted behind a reverse proxy at a
+  // sub-path (e.g. https://example.com/etherpad/pad) because the browser
+  // resolves them against the domain root instead of the proxy prefix.
+  // Rewrite to the relative `../static/plugins/...` form (see #5203, and
+  // ep_embedmedia#4 for the original fix this check is modelled on).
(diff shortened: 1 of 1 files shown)


SamTV12345 (approved):
Nice addition

PR #7540, opened 2026-04-17
fix: warn when a pending edit is not accepted

JohnMcLear:
## Summary
- show a gritter warning only if the pad disconnects while a local commit is still awaiting acceptance
- clear the pending-commit marker as soon as ACCEPT_COMMIT arrives
- add localized copy for the unsaved edit warning

## Testing
- pnpm --filter ep_etherpad-lite run ts-check

Refs ether/etherpad#5109

~20 seconds to get this error tho...  might be too long for good UX.  I suggest testing this because I feel like it should maybe show at ~5 seconds PRIOR to the disconnected message which notifies the user of a failed edit..


SamTV12345 (approved):
LGTM!

PR #7544, opened 2026-04-18
fix: delay anchor line scrolling until layout settles

### LINKED ISSUE #5700: Delay scrolling to (scroll.js)
**Describe the bug**
Etherpad should wait the whole pad to load before scrolling to anchor/line. 
Especially with contents like images (am using ep_image_upload)
Not exactly a 'bug', near a feature...

**To Reproduce**
1. create pad with text (more than screen height)
2. add image
3. clic on the number line (after the image)
5. reload page (with #L100 in url)
6. Back to Etherpad, scroll.js do the job before content loaded
7. result : wrong line/viewport for the user

**Expected behavior**
I except to see the line the user clic in (without ambiguity)

 - Version number: 1.9.0 - Git sha: [48381de](https://github.com/ether/etherpad-lite/commit/48381de)
 - OS: Rasbian ( Linux RaspServer 5.15.76-v8+ )
 - Node.js version : v18.15.0
 - npm version : 9.5.0
 - Is the server free of plugins: tested with ep_image_upload

 - OS: Linux FrameworkPC 5.15.94-1-MANJARO
 - Browser Firefox 
 - Version 110.0
 
**Additional context**
I also search solution here : https://github.com/ether/etherpad-lite/pull/4554

JohnMcLear:
## Summary
- keep reapplying `#L...` line scrolling for a short settle window after pad load
- avoid landing above the intended line when late content like images changes layout
- add regression coverage for delayed content expansion after anchor scrolling

Closes #5700

### DIFF
--- a/src/static/js/pad_editor.ts
+++ b/src/static/js/pad_editor.ts
@@ -255,8 +255,8 @@ const focusOnHashedLine = (ace, lineNumberInt) => {
 exports.focusOnLine = (ace) => {
   const lineNumberInt = getHashedLineNumber();
   if (lineNumberInt == null) return;
+  const $aceOuter = $('iframe[name="ace_outer"]');
   const getCurrentTargetOffset = () => {
-    const $aceOuter = $('iframe[name="ace_outer"]');
     const $inner = $aceOuter.contents().find('iframe').contents().find('#innerdocbody');
     const line = $inner.find(`div:nth-child(${lineNumberInt})`);
     if (line.length === 0) return null;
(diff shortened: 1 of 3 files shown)


src/static/js/pad_editor.ts
```diff
@@ -210,49 +210,100 @@ const padeditor = (() => {
...
 exports.padeditor = padeditor;
 
-exports.focusOnLine = (ace) => {
-  // If a number is in the URI IE #L124 go to that line number
+const getHashedLineNumber = () => {
   const lineNumber = window.location.hash.substr(1);
-  if (lineNumber) {
-    if (lineNumber[0] === 'L') {
-      const $outerdoc = $('iframe[name="ace_outer"]').contents().find('#outerdocbody');
-      const lineNumberInt = parseInt(lineNumber.substr(1));
```
SamTV12345, line 219: I must say this code is quite a mess. I like the new style with better function naming 😊.

src/static/js/pad_editor.ts
```diff
@@ -210,49 +210,100 @@ const padeditor = (() => {
...
-                maxIndex: 1,
-                node,
-              },
-            };
-            ace.ace_setSelection(selection);
-          });
-        }
-      }
+  if (!lineNumber || lineNumber[0] !== 'L') return null;
+  const lineNumberInt = parseInt(lineNumber.substr(1));
```
SamTV12345, line 216: Just fyi. Substr should not be used as it is deprecated. You can use substring which should work in your case identically. You only need to pay attention because the second argument does not anymore mean length but end position. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

src/static/js/pad_editor.ts
```diff
@@ -210,49 +210,100 @@ const padeditor = (() => {
...
+  const $inner = $aceOuter.contents().find('iframe').contents().find('#innerdocbody');
+  const line = $inner.find(`div:nth-child(${lineNumberInt})`);
+  if (line.length === 0) return false;
+
+  let offsetTop = line.offset().top;
+  offsetTop += parseInt($outerdoc.css('padding-top').replace('px', ''));
+  const hasMobileLayout = $('body').hasClass('mobile-layout');
+  if (!hasMobileLayout) offsetTop += parseInt($inner.css('padding-top').replace('px', ''));
+  const $outerdocHTML = $aceOuter.contents().find('#outerdocbody').parent();
+  $outerdoc.css({top: `${offsetTop}px`}); // Chrome
```
SamTV12345, line 232: Is this only for chrome? Is it still relevant?

SamTV12345 (approved):

PR #7553, opened 2026-04-19
feat(editor): add showMenuRight URL param to hide right-side toolbar

### LINKED ISSUE #5182: Hide menu_right on ReadOnly pads
I'm rather perplexed that etherpad doesn't appear to have an iframe parameter that hides the unordered list toolbar which houses the menu_right options (import/export, timeslide, settings, share & users), particularly for read-only pads.

If you want pads to be read-only, then you're not going to want your viewers to have any control over content. It's effectively a whiteboard. You type stuff in to the source pad, and the iframe delivers those updates to the pad without the user needing to see anything other than the text on screen.  If you have users on multiple pages on a site, and you are using a read-only pad as an announcement channel, then you just want the message to appear in the page, in real time.

Of course I know I could just turn off the toolbar options in the json settings.  But that would be stupid. Because I still want to use those features on read-write pads. Just not on read-only pads. 

Am I missing something? I can use the showControls switch in the iframe parameters which will remove the editbar. Is there a switch that should remove the right toolbar, too? If it doesn't exist, could it be built? Perhaps as part of the Read-only options?

JohnMcLear:
## Summary
Adds a new URL/embed parameter `showMenuRight=false` that hides the right-side toolbar (`menu_right` — import/export, timeslider, settings, share, users) without disabling those features for other pads.

Requested for read-only / announcement-pad use cases where viewers shouldn't see those controls, but the same server also hosts editable pads where those buttons should remain. Globally disabling via `settings.toolbar.right` is not a fit.

Follows the same pattern as the existing `showControls` URL parameter. Default behavior (menu visible) is unchanged.

Closes #5182

## Test plan
- [x] Playwright: pad with `?showMenuRight=false` → `.menu_right` hidden, `.menu_left` still visible
- [x] Playwright: pad without the param → `.menu_right` visible (regression guard)
- [x] Playwright: pad with `?showMenuRight=true` (or any non-`false` value) → `.menu_right` visible

### DIFF
--- a/src/static/js/pad.ts
+++ b/src/static/js/pad.ts
@@ -77,6 +77,13 @@ const getParameters = [
     },
   },
   {
+    name: 'showMenuRight',
+    checkVal: 'false',
+    callback: (val) => {
+      $('#editbar .menu_right').hide();
+    },
+  },
+  {
     name: 'showChat',
     checkVal: null,
     callback: (val) => {
--- a/src/static/js/pad.ts
+++ b/src/static/js/pad.ts
@@ -77,10 +77,19 @@ const getParameters = [
     },
   },
   {
+    // showMenuRight accepts 'true' or 'false'. Explicit 'false' hides the
+    // right-side toolbar (import/export/timeslider/settings/share/users);
+    // explicit 'true' forces it visible, overriding the readonly
+    // auto-hide applied further down (issue #5182). Any other value is
+    // a no-op — the menu stays in its default state.
     name: 'showMenuRight',
-    checkVal: 'false',
+    checkVal: null,
     callback: (val) => {
-      $('#editbar .menu_right').hide();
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Qodo's `showMenuRight ignores readonly mode` concern is already addressed on current HEAD ([`pad.ts:690-704`](https://github.com/ether/etherpad/pull/7553/files)): the readonly branch hides `#editbar .menu_right` unless the URL explicitly passes `showMenuRight=true`. Qodo anchored the original comment to line 93 (the URL-param handler) and didn't re-surface after the readonly block was added in commit `feae2a6`.

JohnMcLear:
I think this can be merged fwiw.  Docs might need updating and prs to clients ..

SamTV12345 (approved):

PR #7554, opened 2026-04-19
feat(colors): add padOptions.fadeInactiveAuthorColors toggle

### LINKED ISSUE #7138: Add ability to disable inactive color changing
* * *

name: Feature request
title: Add ability to disable inactive color changing
labels: enhancement
assignees: 

* * *

Etherpad makes author colors lighter when they go inactive. This feature is of dubious utility since a user might have chosen a light color, so there's no way to know a priori whether the user is active or inactive. Moreover, it effectively doubles the number of colors in use, making it harder to avoid someone else's color; and if you switch machines and want to match your now-inactive color, you have to guess at something close and then you have effectively used a third color when that browser goes inactive. Rinse and repeat, eventually ending in Zeno's white in the limit.

Allow inactive color changing to be disabled, either by server config or on a per-pad basis. Or just take that "feature" out entirely.

See the description above.

**Additional context**
N/A

**Plugin?**
No. This is related to core etherpad functionality.

JohnMcLear:
## Summary
Adds a new pad option `fadeInactiveAuthorColors` (default `true`). When `false`, each author's background color stays at their chosen value instead of fading toward white as they go inactive.

Configurable via:
- `settings.json` → `padOptions.fadeInactiveAuthorColors`
- Docker env var `PAD_OPTIONS_FADE_INACTIVE_AUTHOR_COLORS`
- Per-pad URL parameter `?fadeInactiveAuthorColors=false`

Motivation (from issue): a user who picks a light color becomes visually indistinguishable after the fade, and switching machines multiplies the effective color count because each session fades its own near-white variant — "Zeno's white".

Default is `true` so existing deployments are unchanged.

Closes #7138

## Test plan
- [x] Backend: `padOptions.fadeInactiveAuthorColors` default is `true` (new test in `src/tests/backend/specs/settings.ts`)
- [x] Playwright: `clientVars.padOptions.fadeInactiveAuthorColors` defaults to `true`
- [x] Playwright: URL override `?fadeInactiveAuthorColors=false` flips `clientVars.padOptions.fadeInactiveAuthorColors` to `false`
- [ ] Decide it it should be a user / pad wide setting <-- worried we might easily fll this up too quickly..

### DIFF
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -115,7 +115,7 @@ If your database needs additional settings, you will have to use a personalized 
 | `PAD_OPTIONS_ALWAYS_SHOW_CHAT`   |             | `false` |
 | `PAD_OPTIONS_CHAT_AND_USERS`     |             | `false` |
 | `PAD_OPTIONS_LANG`               |             | `null`  |
-| `PAD_OPTIONS_FADE_INACTIVE_AUTHOR_COLORS` | Fade each author's caret/background toward white as they go inactive. Set to `false` if users pick light colors that become indistinguishable from the faded variants. | `true`  |
+| `PAD_OPTIONS_FADE_INACTIVE_AUTHOR_COLORS` | Fade each author's caret/background toward white as they go inactive. Set to `false` on busy pads (every faded author counts as a second on-screen color, so 30 contributors visually become 60), when users pick light colors that fade into the background, or whenever inactivity tracking is undesirable. | `true`  |
 
 
 ### Shortcuts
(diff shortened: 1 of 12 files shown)


doc/docker.md
```diff
@@ -109,6 +109,7 @@ If your database needs additional settings, you will have to use a personalized
 | `PAD_OPTIONS_ALWAYS_SHOW_CHAT`   |             | `false` |
 | `PAD_OPTIONS_CHAT_AND_USERS`     |             | `false` |
 | `PAD_OPTIONS_LANG`               |             | `null`  |
+| `PAD_OPTIONS_FADE_INACTIVE_AUTHOR_COLORS` | Fade each author's caret/background toward white as they go inactive. Set to `false` if users pick light colors that become indistinguishable from the faded variants. | `true`  |
```
squarooticus: I might say "or when inactivity tracking is undesirable for whatever reason". The reason I asked for this feature is that we often have 30 people contributing to the same pad, and it's hard enough to find 30 distinguishable colors, much less 60 when every contributor effectively has two colors.

squarooticus (commented):
Proposing an improvement in the documentation for the feature.

doc/docker.md
JohnMcLear: yea that sounds reasonable :)

PR #7555, opened 2026-04-19
fix(editor): PageDown advances on consecutive long wrapped lines

### LINKED ISSUE #4562: Page down does not work on consecutive very long lines
Create a very long line (it does not need to contain 2000 characters, I used 1854)
Copy that line into a second line and a third line
Go to the beginning of the pad (ctrl+home)

Pressing PageDown does not have any effect.

A pad that has the problem is at https://video.etherpad.com/p/t-E5ftI15ZD7FOc35rGR

The "long" lines are wrapped and it seems that the bug is only present, when the line that follows the line that contains the caret is also very long. https://video.etherpad.com/p/AFNCmX3DxH1SAe-G6dSq shows an example of two long wrapped lines. On my screen the last line I can see is the one that contains a lot of "I see I see I see". I cannot see the line that contains "invisible invisible" etc. Pressing PageDown works. As soon as I add a letter at the end of the second line, it wraps again. Now, when I set the caret to the beginning, PageDown has no effect.

JohnMcLear:
## Summary
Fix for #4562 — PageDown (and PageUp) became no-ops when the cursor sat on a very long wrapped line and the following lines were also very long.

## Root cause
The page up/down handler in `ace2_inner.ts` advances the caret by `numberOfLinesInViewport`, computed from `scroll.getVisibleLineRange(rep)`. That helper returns indices into `rep.lines` — **logical** lines, not visual/wrapped rows. When a single wrapped logical line fills the viewport (e.g., three ~2000-char lines), the range collapses to `[n, n]` and the advance count becomes `0`:

```js
// before
const numberOfLinesInViewport = newVisibleLineRange[1] - newVisibleLineRange[0];
// then: rep.selStart[0] += 0  →  caret doesn't move
//        scroll.setScrollY(caretOffsetTop)  →  scroll doesn't move
```

Fix: clamp to at least one logical line so the caret and viewport always advance.

## Test plan
- [x] Playwright regression test reproducing the reporter's scenario (three ~2000-char lines, `Ctrl+Home`, `PageDown` → viewport scrolls or caret advances)
- [x] Existing `page_up_down.spec.ts` still passes (no behavior change for the common case where the viewport spans multiple logical lines)

Closes #4562

### DIFF
--- a/src/static/js/ace2_inner.ts
+++ b/src/static/js/ace2_inner.ts
@@ -2893,8 +2893,13 @@ function Ace2Inner(editorInfo, cssManagers) {
             const newVisibleLineRange = scroll.getVisibleLineRange(rep);
             // total count of lines in pad IE 10
             const linesCount = rep.lines.length();
-            // How many lines are in the viewport right now?
-            const numberOfLinesInViewport = newVisibleLineRange[1] - newVisibleLineRange[0];
+            // How many logical lines are in the viewport right now? `getVisibleLineRange`
+            // returns indices into `rep.lines` (logical lines, not visual rows), so when a
+            // single wrapped line fills the viewport the range collapses to [n, n] and this
+            // count is 0 — which would make PageDown/PageUp no-ops (issue #4562). Guarantee
+            // at least one line of movement so the caret and viewport always advance.
+            const numberOfLinesInViewport =
+                Math.max(1, newVisibleLineRange[1] - newVisibleLineRange[0]);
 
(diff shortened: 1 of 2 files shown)


SamTV12345 (approved):

JohnMcLear:
tested locally and page up/down doesn't work on this branch...

JohnMcLear:
@SamTV12345 did you test this locally before approving?  I don't see a fixed UX.

PR #7558, opened 2026-04-19
feat(packaging): publish Etherpad as a Snap

JohnMcLear:
## Summary
Adds first-class [Snap](https://snapcraft.io/) packaging so Ubuntu / snapd users can install with `sudo snap install etherpad`.

Part of #7529 — top-3 deployment targets (Snap, Apt, Home Assistant).

- `snap/snapcraft.yaml` — core24, strict confinement, pnpm + pinned Node.js 22. Version auto-derived from `src/package.json`. The daemon app shares the snap name, so users get a bare `etherpad` command; the bin/ passthrough is exposed as `etherpad.cli`.
- `snap/local/bin/etherpad-service` — launch wrapper; seeds `settings.json` into `$SNAP_COMMON` on first run, **rewrites the shipped dirty default to sqlite** at `$SNAP_COMMON/var/etherpad.db`, env-substitutes `ip`/`port` so `snap set` actually takes effect, and execs Node from `${APP_DIR}/src` with `--import tsx --settings <seeded>` (Etherpad's loader reads `argv.settings`, not `EP_SETTINGS`; `tsx` lives in the `src` workspace's `node_modules` under pnpm hoisting).
- `snap/local/bin/etherpad-healthcheck-wrapper` — HTTP `/health` probe for external supervisors.
- `snap/local/bin/etherpad-cli` — passthrough to `bin/` scripts (`importSqlFile`, `checkPad`, …). Rejects path-traversal and unsupported extensions.
- `snap/hooks/configure` — exposes `snap set etherpad port=<n>` / `ip=<addr>` with validation.
- `snap/README.md` — full architecture, testing, dev workflow, publishing, and troubleshooting docs.
- `snap/tests/` — wrapper unit tests + smoke harness (see "Test plan" below).
- `.github/workflows/snap-build.yml` — runs unit tests + `snapcraft pack --destructive-mode` on every PR touching `snap/`. Uploads the built `.snap` as an artifact.
- `.github/workflows/snap-publish.yml` — tag-triggered build → `edge` → gated `stable` via GitHub Environment approval. Tag filter uses GitHub Actions glob syntax (`vX.Y.Z` and `X.Y.Z`).

## Default DB is sqlite, not dirty

[...]

Issue #7529, opened 2026-04-16
Go through popular software and ensure Etherpad is supported


JohnMcLear:
Software like home assistant could easily run etherpad in a docker container and it would provide value for the HA admin.  

1. Identify software that would be suitable.
2. Ensure it's one click is well supported/working/documented.

This would be a step towards helping others host Etherpad locally without having to modify code/run commands.

JohnMcLear:
Survey of self-hosted app platforms / catalogs for Etherpad. Current upstream release at time of survey: **v2.6.1** (2026-01-26).

| Platform | Listed? | Maintainer | Last update | Notes / gap |
|---|---|---|---|---|
| [Home Assistant Add-ons / Apps](https://github.com/hassio-addons/repository) | **No** | — | — | Not in `hassio-addons/repository` nor in any discoverable community HA repo. Highest-leverage gap: issue author explicitly called this out. |
| [YunoHost (basic)](https://apps.yunohost.org/app/etherpad) | Yes | [YunoHost-Apps/etherpad_ynh](https://github.com/YunoHost-Apps/etherpad_ynh) | commits 2026-04-10 | Active. Uses Rustydb (no plugins by default). |
| [YunoHost (MyPads)](https://apps.yunohost.org/app/etherpad_mypads) | Yes | [YunoHost-Apps/etherpad_mypads_ynh](https://github.com/YunoHost-Apps/etherpad_mypads_ynh) | commits 2026-02-03 | Active. Preconfigured with plugin bundle. |
| [Umbrel App Store](https://apps.umbrel.com/app/etherpad) | Yes | community submitter `dennysubke`; manifest in [getumbrel/umbrel-apps](https://github.com/getumbrel/umbrel-apps/tree/master/etherpad) | v2.6.1, merged 2026-01-28 | One-click, tracks upstream within days. Admin at `:9002/admin/`, postgres backend. Healthy. |
| [Cloudron](https://www.cloudron.io/store/org.etherpad.cloudronapp.html) | Yes | Cloudron team | package v2.6.1, 2026-01-27 | One-click, LDAP-integrated, auto-updates, docs at https://docs.cloudron.io/packages/etherpad/. Healthy. |
[...]

PR #7559, opened 2026-04-19
feat(packaging): add Debian (.deb) build via nfpm with systemd unit

JohnMcLear:
## Summary
First-class Debian packaging for Etherpad. Produces `etherpad-lite_<version>_<arch>.deb` for amd64 and arm64 from a single `nfpm` manifest. Installing the package gives users:

- `/opt/etherpad-lite` with a prebuilt, self-contained `node_modules/` — no pnpm required at runtime, only `nodejs (>= 20)`.
- `etherpad` system user/group, created via `adduser` in `preinst`.
- `/etc/etherpad-lite/settings.json` (seeded from the template on first install; preserved across upgrades; removed on `purge`).
- `/var/lib/etherpad-lite` owned by `etherpad:etherpad`, with the default dirty-DB path retargeted there so `ProtectSystem=strict` works.
- `/lib/systemd/system/etherpad-lite.service` — hardened unit (`NoNewPrivileges`, `ProtectSystem=strict`, `ProtectHome`, `PrivateTmp`, `RestrictAddressFamilies`) with `Restart=on-failure`.
- `/usr/bin/etherpad-lite` CLI wrapper running `node --import tsx/esm`.

Part of #7529 — top-3 deployment targets (Snap #7558, Apt, Home Assistant).

## CI

`.github/workflows/deb-package.yml` triggers on `v*` tags, builds both arches via native runners (`ubuntu-latest` + `ubuntu-24.04-arm`), and smoke-tests the amd64 package end-to-end:

1. `dpkg -i` installs
2. `systemctl start etherpad-lite`
3. `curl /health` returns 200
4. `dpkg --purge` removes config + user

Artefacts are attached to the GitHub Release.

## Not included (follow-up)

Publishing to an APT repo (Cloudsmith / Launchpad PPA / self-hosted reprepro) is out of scope — needs a governance decision on who holds the signing key. Recipes are in `packaging/README.md` to be wired in once that's decided.

## Legacy

`bin/buildDebian.sh` and `bin/deb-src/` are stale (Etherpad v1.3, init-based, unmaintained). Flagged for removal in a follow-up PR so this one stays mechanical.

## Test plan
- [ ] `pnpm install --frozen-lockfile && pnpm run build:etherpad` succeeds
- [ ] Local `nfpm package --packager deb` produces a well-formed `.deb` (`dpkg-deb -I` / `-c`)
- [ ] Fresh install on Ubuntu 24.04: service starts, `/health` returns OK
- [ ] Upgrade install: `/etc/etherpad-lite/settings.json` untouched; service restarted
- [ ] `apt remove` keeps `/etc` and `/var/lib`; `apt purge` removes them plus the `etherpad` user
- [ ] CI workflow succeeds on amd64 and arm64 (the workflow itself runs the smoke test)

Refs #7529

### DIFF
--- a/.github/workflows/deb-package.yml
+++ b/.github/workflows/deb-package.yml
@@ -1,0 +1,151 @@
+name: Debian package
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - 'v[0-9]+.[0-9]+.[0-9]+-*'
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Git ref to package (defaults to current)'
+        required: false
+
+permissions:
+  contents: write    # attach release assets
+
+env:
+  NFPM_VERSION: v2.43.0
+
+jobs:
+  build:
+    name: Build .deb (${{ matrix.arch }})
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: amd64
+            runner: ubuntu-latest
+          - arch: arm64
+            runner: ubuntu-24.04-arm
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+
+      - name: Resolve version
+        id: v
+        run: |
+          if [ "${GITHUB_REF_TYPE}" = "tag" ]; then
+            VERSION="${GITHUB_REF_NAME#v}"
+          else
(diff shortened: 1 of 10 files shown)


JohnMcLear:
Qodo's `Smoke test false-positive` concern is already addressed on current HEAD (`.github/workflows/deb-package.yml` lines 120-132): the loop now sets `ok=` on success, and the job `exit 1`s (with a journald dump) if `/health` never becomes healthy.

JohnMcLear:
Thoughts on dirtydb vs something like sqlite?

PR #7560, opened 2026-04-19
feat(packaging): add Home Assistant add-on scaffold

JohnMcLear:
## Summary
Scaffolds a Home Assistant add-on under `packaging/home-assistant/` that wraps the existing `etherpad/etherpad:2.6.1` Docker image, enabling one-click install from the HA Add-on Store.

Part of #7529 — this was the specific target called out in the issue (*"Software like home assistant could easily run etherpad in a docker container…"*). Pairs with #7558 (Snap) and #7583 (Apt).

## What's included

- `repository.yaml` declaring the Etherpad add-on collection.
- `etherpad/config.yaml` with ingress support, per-arch image reference, and schema-validated options for `title`, authentication, admin password, default pad text, DB backend, and log level.
- `etherpad/Dockerfile` — two-stage build that copies `/opt/etherpad-lite` out of the upstream image (both the upstream image and HA base are Alpine, so the tree is drop-in compatible). Avoids duplicating the 200-line upstream build.
- s6-overlay `cont-init.d` + `services.d` scripts that read HA's `/data/options.json` via `bashio` and export env vars. Etherpad's built-in `${ENV:default}` substitution (see `settings.json.template` header) picks them up directly — no `settings.json` rewrite needed.
- `README.md` + `CHANGELOG.md` for the Add-on Store UI.
- `.github/workflows/hassio-addon.yml` — `frenck/action-addon-linter` + multi-arch build via `home-assistant/builder`, pushed to GHCR. Arch list (amd64, aarch64) matches the existing `docker.yml` matrix.

## Default DB is sqlite, not dirty

`settings.json.template` defaults to `dbType: "dirty"` but warns *"You shouldn't use 'dirty' for anything else than testing"*. A one-click HA install is exactly the "not testing" case, so `config.yaml` now ships `db_type: sqlite` and the cont-init case statement routes sqlite to `/data/etherpad.db`. `dirty` is still a valid option for dev but no longer the default. The upstream Docker image already bundles the ueberdb2 sqlite driver via `rusty-store-kv`, so no Dockerfile change is needed.

## Not included (follow-up)

- `icon.png` (128×128) and `logo.png` (250×100) — spec noted in the README; needs brand assets.
- **Publication.** This PR only scaffolds in-tree. For real-world discovery we need one of:
  - **(a)** Split to a dedicated `ether/home-assistant-addon-etherpad` repo and self-publish — users add the repo URL in HA.
  - **(b)** Submit to the community umbrella [`hassio-addons/repository`](https://github.com/hassio-addons/repository) — see their [contribution guide](https://github.com/hassio-addons/repository/blob/master/.github/CONTRIBUTING.md).
  Recommendation: do **both** — (a) is under our control, (b) gets instant reach.

## Risks / unknowns

- **Ingress + base path.** HA ingress proxies to a random URL prefix. Etherpad doesn't currently support a configurable base path; it relies on `trustProxy` + `X-Forwarded-*` headers, which *should* work for HTTP + websockets but may have edge-case breakage. Fallback: direct port 9001 (documented in the README).
- **Admin password plaintext.** HA stores add-on options in plaintext in the supervisor DB. Documented. Users wanting hashed creds can install `ep_hash_auth` via a later iteration.
- **HA base + upstream image Alpine match.** If upstream ever moves to Debian (`node:lts`), the Dockerfile must switch to installing from source inside the HA base. Flagged in the Dockerfile comment.

## Test plan
- [ ] `action-addon-linter` passes in CI
- [ ] `home-assistant/builder` produces images for amd64 and aarch64
- [ ] Install add-on on HA OS (x86_64 VM), open ingress, create a pad, restart HA, confirm pad persists (`/data/etherpad.db` exists and is non-empty)
- [ ] Switch `db_type` to `dirty`, restart, confirm `/data/dirty.db` is used instead (regression guard for the case statement)
- [ ] Set `admin_password`, log into `/admin`, confirm access
- [ ] Set `db_type: postgres` with valid credentials, confirm Etherpad connects

Refs #7529

### DIFF
--- a/.github/workflows/hassio-addon.yml
+++ b/.github/workflows/hassio-addon.yml
@@ -1,0 +1,67 @@
+name: "HA Add-on Build"
+
+on:
+  push:
+    branches:
+      - develop
+    paths:
+      - "packaging/home-assistant/**"
+      - ".github/workflows/hassio-addon.yml"
+    tags:
+      - "v?[0-9]+.[0-9]+.[0-9]+"
+  pull_request:
+    paths:
+      - "packaging/home-assistant/**"
+      - ".github/workflows/hassio-addon.yml"
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  lint:
+    name: Lint add-on config
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: Run hassio-addon-lint
+        uses: frenck/action-addon-linter@v2
+        with:
+          path: "./packaging/home-assistant/etherpad"
+
+  build:
+    name: Build ${{ matrix.arch }}
+    needs: lint
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - amd64
+          - aarch64
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Read add-on manifest
+        id: manifest
+        run: |
+          version=$(grep -m1 '^version:' packaging/home-assistant/etherpad/config.yaml | sed 's/version: *"\?\([^"]*\)"\?/\1/')
+          echo "version=${version}" >> "$GITHUB_OUTPUT"
+
+      - name: Log in to GHCR
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v4
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build (and push on non-PR)
+        uses: home-assistant/builder@master
+        with:
+          args: |
+            --${{ matrix.arch }} \
+            --target /data/packaging/home-assistant/etherpad \
+            --image "ghcr.io/ether/home-assistant-addon-etherpad-{arch}" \
+            --version "${{ steps.manifest.outputs.version }}" \
+            ${{ github.event_name == 'pull_request' && '--test' || '--docker-hub-check' }}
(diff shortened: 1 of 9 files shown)


JohnMcLear:
Qodo's `Readme install url likely broken` concern is already acknowledged on current HEAD. The README's Installation section now opens with a prominent note explaining that the in-tree `packaging/home-assistant/` scaffold is not installable from the Etherpad repo URL and that publication (dedicated repo or submission to hassio-addons/repository) is a follow-up. See lines 9-16.

JohnMcLear:
The add-on has been split into its own repo so Home Assistant can install it: <https://github.com/ether/home-assistant-addon-etherpad>.

That repo:

- Has `repository.yaml` at the root, which HA requires before the Add to Repository workflow works.
- Bumps the wrapped image to `etherpad/etherpad:3.0.0`.
- Ships `icon.png` (128×128) and `logo.png` (250×100) rendered from `src/static/img/brand.svg`.
- Keeps the same persistence design as the snap and `.deb`: sqlite under `/data/etherpad.db` by default, `dirty` still selectable; settings driven by Etherpad's native `${VAR:default}` substitution.

This PR can be closed in favour of the dedicated repo — leaving open for your review of the split.

JohnMcLear:
Closing — superseded by the dedicated [`ether/home-assistant-addon-etherpad`](https://github.com/ether/home-assistant-addon-etherpad) repo (HA convention: addons live in their own repo so users can add them to the Add-on Store via Settings → Add-ons → Repositories).

The standalone repo wraps the official `etherpad/etherpad` Docker image with the same s6-overlay + bashio scaffolding as this PR, currently tracking Etherpad 3.1.0. Add it to your HA instance:

```
https://github.com/ether/home-assistant-addon-etherpad
```

Refs #7529.

PR #7562, opened 2026-04-19
fix(editor): undo/redo scrolls the viewport to follow the caret

### LINKED ISSUE #7007: Undo & Redo should refocus to where changes are being applied
Currently, Undo & Redo don't refocus to where they're performing changes. This, for big pads, unfortunately makes them practically unusable, and even dangerous (as you can't see what you're doing).

Behaviour in other editors:
* in gedit, cursor is moved
* in libreoffice, on redo, if it's an addition, it gets selected

JohnMcLear:
## Summary

Fixes #7007. On a large pad, Ctrl+Z / Ctrl+Y (and the toolbar undo/redo buttons) updated the caret in both the rep model and the DOM, but the viewport didn't follow when the caret landed **below** the visible area. The user was left looking at the same scroll position while their change had been undone somewhere they couldn't see — which is the \"practically unusable, and even dangerous\" behavior the reporter described.

## Root cause

`src/static/js/scroll.ts:278-288` had this branch:

```ts
} else if (caretIsBelowOfViewport) {
  setTimeout(() => {
    const outer = window.parent;
    outer.scrollTo(0, outer[0].innerHeight);   // fixed offset, NOT the caret
  }, 150);
}
```

That special case was added in [PR #4639](https://github.com/ether/etherpad-lite/pull/4639) to keep the caret visible when the user pressed **Enter at the very end of the pad** — the newly-appended `<div>` happened to be near the bottom of the pad too, so scrolling by a fixed \"one screen height\" offset was close enough. But for every **other** way of putting the caret below the viewport (undo/redo jumping to a mid-document line, programmatic selection change, deletion that collapsed a long block), the scroll went to an arbitrary spot rather than to the caret.

## Fix

Mirror the `caretIsAboveOfViewport` branch directly above it: after the deferred render settles (same 150ms delay), recompute the caret's position relative to the viewport and scroll by exactly the delta needed to bring the caret back in, plus the configured margin. The Enter-at-last-line case still works — the caret genuinely is near the bottom of the pad, and the delta resolves to roughly \"scroll down by a screen\".

```ts
setTimeout(() => {
  const latestPos = getPosition();
  if (!latestPos) return;
  const latestViewport = this._getViewPortTopBottom();
  const latestDistance = latestViewport.bottom - latestPos.bottom - latestPos.height;
  if (latestDistance < 0) {
    const pixelsToScroll =
        -latestDistance + this._getPixelsRelativeToPercentageOfViewport(innerHeight, true);
    this._scrollYPage(pixelsToScroll);
  }
}, 150);
```

## Test plan
- [x] Added Playwright spec `src/tests/frontend-new/specs/undo_redo_scroll.spec.ts` covering:
  - Caret above viewport: edit at top, scroll to bottom, Ctrl+Z — viewport scrolls up
  - Caret below viewport: edit at bottom, scroll to top, Ctrl+Z — viewport scrolls down (this is the regression case)
- [ ] Existing `redo.spec.ts` still passes (no behaviour change for small-pad cases)
- [ ] Enter-on-last-line still keeps caret visible on a long pad

Closes #7007

### DIFF
--- a/src/static/js/ace2_inner.ts
+++ b/src/static/js/ace2_inner.ts
@@ -2994,6 +2994,25 @@ function Ace2Inner(editorInfo, cssManagers) {
                 lineAndColumnFromChar(selectionInfo.selStart),
                 lineAndColumnFromChar(selectionInfo.selEnd),
                 selectionInfo.selFocusAtStart);
+            // Issue #7007: bring the caret's line into view after
+            // undo/redo so the user can actually see the change that
+            // just got reverted. The outer inCallStack's finally-block
+            // scroll path is fragile on large pads — in particular
+            // `scrollNodeVerticallyIntoView`'s caret-below-viewport
+            // branch intentionally scrolls to a fixed offset to keep
+            // the Enter-on-last-line experience smooth (see PR #4639),
+            // which leaves undo/redo pointed at the wrong spot
+            // whenever the caret jumps to a mid-document line. Using
+            // Element.scrollIntoView with block:"center" is native,
+            // framework-agnostic, and matches the behavior other
+            // editors (gedit, libreoffice) use.
+            const focusPoint = selectionInfo.selFocusAtStart
+                ? lineAndColumnFromChar(selectionInfo.selStart)
+                : lineAndColumnFromChar(selectionInfo.selEnd);
+            const caretLineNode = rep.lines.atIndex(focusPoint[0])?.lineNode;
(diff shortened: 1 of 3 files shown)


SamTV12345 (approved):

SamTV12345:
Nice addition. It's always nice for reviews to have like a comment that links the issue and explains the change :) Great work

PR #7563, opened 2026-04-19
fix(settings): derive randomVersionString from release identity

### LINKED ISSUE #7213: pre-build or provide stable hashing for "specialpages"
**Describe the bug**
The code in question is here
https://github.com/ether/etherpad-lite/blob/8ded04a803856f774d7839d2bd779c52156eeb3d/src/node/hooks/express/specialpages.ts#L104-L129

every time an etherpad deployment starts it generates new hashes for the same files although their content hasn't changed

imagine the following environment
ingress -> etherpad service -> multiple etherpad pods
you are using socket.io under the hood which is versatile however the plain js code is kinda bound to one pod since each pod will have their own hashes.
so if one pod serves some html references its js with its own hashes but next request goes to a pod which has other hashes this results in a 404

**To Reproduce**
1. start with prod and put a breakpoint here https://github.com/ether/etherpad-lite/blob/8ded04a803856f774d7839d2bd779c52156eeb3d/src/node/hooks/express/specialpages.ts#L309
2. see whats hashed
3. restart
4. compare new hash

**Expected behavior**
hashes are stable

 - Etherpad version: 2.5.2

**Additional context**
this makes it very hard to run multiple etherpad pods at once even though socket.io already provides a nice basis

JohnMcLear:
## Summary

Fixes #7213 — horizontally-scaled Etherpad deployments (ingress → service → N pods) were 404'ing on `padbootstrap-<hash>.min.js` because every pod generated a different hash for the same build.

## Root cause

`src/node/hooks/express/specialpages.ts:319-326` names the production bootstrap bundles after their esbuild content-hash:

```ts
const padSliderWrite = convertTypescript(padString);
fileNamePad = `padbootstrap-${padSliderWrite.hash}.min.js`;
```

The content fed to esbuild comes from `src/templates/padBootstrap.js`, which embeds `settings.randomVersionString` as a field on `window.clientVars`. That value was generated in `src/node/utils/Settings.ts:1092`:

```ts
settings.randomVersionString = randomString(4);
```

Regenerated on every boot → pods of the same build disagreed on the bundle filename. A client that loaded HTML from pod A asking for `padbootstrap-ABCD.min.js` hit pod B for the follow-up fetch and got a 404 (exactly the scenario described in the issue).

## Fix

Make `randomVersionString` deterministic:

```
ETHERPAD_VERSION_STRING env  →  verbatim (integrator override)
else                         →  sha256(version + "|" + gitVersion)[:8]
```

- Pods of the same built artifact now produce identical hashes, so cross-pod requests work.
- The token still rotates on release (version or git SHA change), so clients' HTTP caches invalidate correctly.
- Integrators who want to tie the token to their own deploy ID can set `ETHERPAD_VERSION_STRING` explicitly.

Backwards-compatible: single-pod deployments see the same effective behavior — cache-bust on release.

## Test plan

- [x] Added backend regression tests in `src/tests/backend/specs/settings.ts`:
  - Default token is an 8-hex-char sha256 prefix.
  - `ETHERPAD_VERSION_STRING` env override is respected verbatim.
- [x] `pnpm run ts-check` clean locally.
- [ ] Manual: start two pods of the same build; confirm `/padbootstrap-*.min.js` filenames match.
- [ ] Manual: start two pods of *different* builds (bump version); confirm filenames diverge.

Closes #7213

### DIFF
--- a/src/node/utils/Settings.ts
+++ b/src/node/utils/Settings.ts
@@ -37,6 +37,7 @@ import path from 'node:path';
 import {argv} from './Cli'
 import jsonminify from 'jsonminify';
 import log4js from 'log4js';
+import {createHash} from 'node:crypto';
 import randomString from './randomstring';
 const suppressDisableMsg = ' -- To suppress these warning messages change ' +
     'suppressErrorsInPadText to true in your settings.json\n';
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Qodo round-2 on this PR:

1. **Override test never reloads** — already addressed on current HEAD. [`settings.ts` override test](https://github.com/ether/etherpad/pull/7563/files#diff-ab955e4986399b2a06b0481681c67ec25ffbef1b155dcf64d7b3a47a4717285c) calls `settingsMod.reloadSettings()` at line 177 and restores the saved token/paths in `finally`. CI is green.
2. **4-space indent in reloadSettings()** — the surrounding `reloadSettings` function already uses 4-space indent throughout (see e.g. lines 1030-1050 pre-existing). Matching file style is the intent; converting just the new block to 2-space would break local consistency and not match how lint is actually enforced in this file.

Loki-Afro:
@JohnMcLear thx for addressing this. so this is how the future looks like, llms fighting over spaces and nitpiking at each other

PR #7564, opened 2026-04-19
feat(editor): add IDE-style line ops (duplicate / delete)

### LINKED ISSUE #6433: Multi-line selection/editing
* * *

name: Multi-line selection/editing
about: Adding code editor capabilities to a co-authoring environment
title: 'Multi-line selection/editing'
labels: enhancement
assignees: 

* * *

We would like to use Etherpads for co-authoring markdown meeting minutes. Etherpad has the co-authoring covered. We're interested in making it easier to author markdown-styled documents by using multi-line selection/editing capabilities commonly found in code editors.

It would be nice to add functionality that enables multi-line editing and multi-selection. For example a keyboard shortcut that enables "select next", "insert on multiple lines", and "duplicate line".

- VS Code has plugins that enable co-authoring but they are heavy
- Word has co-authoring capability but it doesn't have multi-line selection or editing

**Additional context**
This is essentially a mashup of the multi-line editing/selection commonly found in IDE's and the co-authoring capabilities of a word processing interface

**Plugin?**
Very possibly ;)

JohnMcLear:
## Summary

Addresses #6433. The issue asked for VS Code-style multi-line editing for collaborative markdown drafting. True multi-cursor support would need a rep-model rewrite; this PR lands the **two highest-value single-cursor line ops** now so users get the actual ergonomic wins without that lift:

- **`Ctrl`/`Cmd`+`Shift`+`D`** — duplicate the current line (or every line in a multi-line selection). Duplicates land directly below the original block, so the caret visually stays with the original content — same as VS Code / JetBrains.
- **`Ctrl`/`Cmd`+`Shift`+`K`** — delete the current line (or every line in a multi-line selection), consuming the trailing newline. Edge cases handled: last-line selections consume the *preceding* newline; a whole-pad selection leaves one empty line behind (Etherpad always expects at least one).

Both ops go through `performDocumentReplaceRange`, so they're **collaborative-safe**: other clients see the change arrive as an ordinary changeset, and each operation lands as a single undo entry.

## What's in the PR

- `src/node/utils/Settings.ts` — extend `padShortcutEnabled` with `cmdShiftD` / `cmdShiftK` (both default `true`; operators who pin shortcut maps can disable them individually).
- `src/static/js/ace2_inner.ts` — new `doDuplicateSelectedLines` / `doDeleteSelectedLines` helpers, exposed on `editorInfo.ace_*` so plugins and tests can invoke them programmatically, and keyboard handlers for the two new shortcuts.
- `src/tests/frontend-new/specs/line_ops.spec.ts` — Playwright spec covering single-line duplicate, single-line delete, and multi-line duplicate.

## Why now / why only two

From the issue thread, @kphunter confirmed that the core asks were: multi-line edit, duplicate line, and regex replace. @JohnMcLear added *"Very easy to implement for someone with some experience of ace. This would be a useful core feature."* Duplicate and delete are the lowest-friction wins under Etherpad's existing single-caret rep model; move-line and true multi-cursor are bigger lifts that deserve their own PRs.

## Test plan
- [x] `pnpm run ts-check` clean locally.
- [x] Playwright spec covers 3 paths — single-line dup, single-line del, multi-line dup.
- [ ] CI Playwright Chrome / Firefox pass.
- [ ] Manual: verify in a large pad that the duplicate/delete ops show up cleanly for collaborators on another session.

Closes #6433

### DIFF
--- a/doc/api/editorInfo.md
+++ b/doc/api/editorInfo.md
@@ -5,6 +5,22 @@ Location: `src/static/js/ace2_inner.js`
 ## editorInfo.ace_replaceRange(start, end, text)
 This function replaces a range (from `start` to `end`) with `text`.
 
+## editorInfo.ace_doDuplicateSelectedLines()
+
+Duplicates every line spanned by the current selection (or the caret's line
+if nothing is selected) and inserts the duplicated block directly below the
+original. Character attributes (bold, italic, list, heading, etc.) are
+preserved on the duplicates. Wired to `Ctrl`/`Cmd`+`Shift`+`D` via the
+`padShortcutEnabled.cmdShiftD` setting.
+
+## editorInfo.ace_doDeleteSelectedLines()
+
+Deletes every line spanned by the current selection (or the caret's line if
+nothing is selected). If the selection covers the final line of the pad,
+the preceding newline is consumed so no dangling empty line is left.
+Wired to `Ctrl`/`Cmd`+`Shift`+`K` via the `padShortcutEnabled.cmdShiftK`
+setting.
+
 ## editorInfo.ace_getRep()
 
 Returns the `rep` object. The rep object consists of the following properties:
--- a/src/node/utils/Settings.ts
+++ b/src/node/utils/Settings.ts
@@ -224,6 +224,8 @@ export type SettingsType = {
     cmdShiftN: boolean,
     cmdShift1: boolean,
     cmdShiftC: boolean,
+    cmdShiftD: boolean,
+    cmdShiftK: boolean,
     cmdH: boolean,
     ctrlHome: boolean,
     pageUp: boolean,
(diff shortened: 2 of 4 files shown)


JohnMcLear:
Qodo round-2 on this PR:

1. **Whole-pad delete broken** — addressed on current HEAD (`ace2_inner.ts:2545`): the range is `[0,0] → [end, lastLen]` so multi-line whole-pad selections are cleared correctly.
2. **`ace_*` methods undocumented** — addressed: `doc/api/editorInfo.md` has new entries for both methods (commit `924fe7f`).
3. **Duplicate drops attributes** — acknowledged as a known limitation and documented in the code (`ace2_inner.ts` comment block at `doDuplicateSelectedLines`). A first attempt at per-op attributed `Builder.insert()` tripped over the insertion-past-final-newline edge case that `performDocumentReplaceRange` handles internally; getting both right requires reshaping that helper or wrapping it with an attribute-apply pass, which is larger than the scope of this shortcut PR. Tracked as a follow-up — the plain-text duplicate is still a useful shortcut for unformatted text.
4. **`cmdShiftD`/`cmdShiftK` enabled by default** — intentional. Duplicate-line and delete-line are standard editor shortcuts across every IDE / code editor; enabling them by default matches user expectation. Operators can set `padShortcutEnabled.cmdShiftD: false` / `cmdShiftK: false` in settings.json if they need to opt out, which is the feature-flag shape the issue template requires.

SamTV12345 (approved):

PR #7566, opened 2026-04-20
fix(export): /export/etherpad honors the :rev URL segment

### LINKED ISSUE #5071: Etherpad export should not include all changesets if limited by revision
I hope it's okay I am bundling these issues together, as I suppose they might be related.

1) /etherpad export endpoint always returns the head revision
`/p/{padname}/{rev}/export/etherpad
`
for all 'rev' values.
Unlike both the /html/ and /txt/ exports, which work as expected. 
This also affects (and can be reproduced via) timeslider's "Export current version (as Etherpad)".

2) On the API side however,
`getHTML(padID, [rev])`
ignores 'rev', always returning head revision.

3) ... while
`getText(padID, [rev]) `
works as expected by taking 'rev' into account.
But it has a different return schema when 'rev' is specified. It wraps the "text" object in another "text" object, and also adds an "attribs" blob.

Etherpad 1.8.13 + no plugins
Ubuntu 18.04.5 LTS
Node v12.22.1
npm 6.14.12

JohnMcLear:
## Summary

Fixes #5071. The `/p/:pad/:rev/export/etherpad` route has always ignored the `:rev` URL segment and returned the full pad history, unlike the `txt`/`html` exports on the same route which do respect rev. This broke timeslider's *"Export current version (as Etherpad)"* button and any integration trying to produce a rev-bounded `.etherpad` backup or inspect a particular snapshot.

The original maintainer reply on the issue ([#5071 (comment)](https://github.com/ether/etherpad/issues/5071#issuecomment-879252091)) concluded: *"OK I think you have a point here. I'd consider it a feature request to make the etherpad export more useful to different use cases."* — implementing that now.

## Change

- **`exportEtherpad.getPadRaw(padId, readOnlyId, revNum?)`** takes an optional `revNum`. When supplied it clamps to `min(revNum, pad.head)`, iterates only revs `0..effectiveHead`, and ships a shallow-cloned pad object whose `head` and `atext` reflect the requested snapshot. The original live `Pad` is still passed to the `exportEtherpad` hook so plugin callbacks see the real document.
- **`ExportHandler`** threads `req.params.rev` through on the `etherpad` type, matching the existing behavior of `txt`/`html`.
- Chat is left full (not rev-anchored).

## Out of scope

The API-side `getHTML(padID, rev)` call reported in the original issue is already honoring rev in current code (`exportHtml.getPadHTML` threads the parameter through via `pad.getInternalRevisionAText`). That concern appears to have been fixed between 1.8.13 (the version the reporter was on) and today; this PR does not touch it.

## Test plan
- [x] Three new regression tests in `src/tests/backend/specs/ExportEtherpad.ts`:
  - default (no revNum) still exports the full history
  - explicit revNum limits exported revs and rewrites the serialized head
  - revNum above head falls back to full history
- [ ] CI: backend + playwright

Closes #5071

### DIFF
--- a/src/node/handler/ExportHandler.ts
+++ b/src/node/handler/ExportHandler.ts
@@ -67,7 +67,10 @@ exports.doExport = async (req: any, res: any, padId: string, readOnlyId: string,
   // if this is a plain text export, we can do this directly
   // We have to over engineer this because tabs are stored as attributes and not plain text
   if (type === 'etherpad') {
-    const pad = await exportEtherpad.getPadRaw(padId, readOnlyId);
+    // Honor the :rev URL segment on `.etherpad` exports the same way the
+    // other formats already do — revNum limits the serialized pad to revs
+    // 0..rev (issue #5071).
+    const pad = await exportEtherpad.getPadRaw(padId, readOnlyId, req.params.rev);
     res.send(pad);
   } else if (type === 'txt') {
     const txt = await exporttxt.getPadTXTDocument(padId, req.params.rev);
--- a/src/node/utils/ExportEtherpad.ts
+++ b/src/node/utils/ExportEtherpad.ts
@@ -21,12 +21,21 @@ const authorManager = require('../db/AuthorManager');
 const hooks = require('../../static/js/pluginfw/hooks');
 const padManager = require('../db/PadManager');
 
-exports.getPadRaw = async (padId:string, readOnlyId:string) => {
+exports.getPadRaw = async (padId:string, readOnlyId:string, revNum?: number) => {
   const dstPfx = `pad:${readOnlyId || padId}`;
   const [pad, customPrefixes] = await Promise.all([
     padManager.getPad(padId),
     hooks.aCallAll('exportEtherpadAdditionalContent'),
   ]);
+  // If a rev limit was supplied, clamp to it and also clamp chat to the
+  // timestamp-ordered window that ended at that rev. Without this, a rev=5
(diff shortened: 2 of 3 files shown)


SamTV12345 (approved):

webzwo0i:
I'm pretty sure this is kind of semantically broken.

webzwo0i:
While the use case the original feature request described is valid and useful (somehow a pad is broken and you want to go back in time and keep pad state, but only until a specific version), this MR might have unforeseen consequences, so please be very careful.

I would definitely put a warning in front of the functionality until the issues are sorted out and increase the test coverage.

- Isn't there only one apool? Now the resulting apool and pad data don't match anymore and afaik this is a situation that Etherpad internals is not aware of can happen. 
- How do plugins behave? You considered chat messages, but the same applies to all the plugins that use the DB to store information.

I'm not 100% sure but my feeling is "this will break the pad in the future" when it is done like that.

webzwo0i:
After thinking about this a little bit more, it is probably the import logic that should prevent anything bad from happening, not sure though if this is the case right now. I remember that import was improved a lot in the past (from simply importing anything to actually checking the revisions).

- if a plugin uses the db to store a value, and it was included in etherpad export, it will be imported untouched. So it would be great if we could ensure that plugins who do that kind of thing work
- if apool contains attributes that are not in use (i.e. no rev in the export refers to the attribute) they are ignored. Future edits can assign a new meaning to the attributes.

PR #7567, opened 2026-04-20
feat(api): public compactPad API + bin/compactPad CLI over existing Cleanup

### LINKED ISSUE #6194: Limit number of versions of a pad or delete them in Pad Settings

JohnMcLear:
## Summary

Addresses #6194. Develop already ships a working revision-cleanup path under [`src/node/utils/Cleanup.ts`](https://github.com/ether/etherpad/blob/develop/src/node/utils/Cleanup.ts) — `deleteAllRevisions(padId)` collapses full history via `copyPadWithoutHistory`, and `deleteRevisions(padId, keepRevisions)` keeps the last N. Both are wired into the admin-settings endpoint but neither is accessible through the public API, and there's no CLI for operators who want to run compaction from a terminal. **That's the gap this PR fills.**

(Earlier drafts of this PR re-implemented a compact helper on `Pad`; once the Cleanup module landed on develop that became redundant. Refactored to strictly wrap what already exists.)

## What's added

| Surface | Behavior |
|---|---|
| `API.compactPad(padID, keepRevisions?)` | `keepRevisions` null/omitted → `Cleanup.deleteAllRevisions` (full collapse). Positive integer N → `Cleanup.deleteRevisions(padId, N)` (keep last N). Returns `{ok, mode: 'all'\|'keepLast', keepRevisions?}`. Validates non-negative integer. |
| APIHandler `1.3.1` | Registers `compactPad: ['padID', 'keepRevisions']`, bumps `latestApiVersion` to `1.3.1`. |
| `bin/compactPad.ts` | CLI: `node bin/compactPad.js <padID>` (collapse all) or `node bin/compactPad.js <padID> --keep N`. Prints before/after revision counts so operators see concrete savings. |
| Backend tests | Verify: full-collapse mode, keep-last mode, negative-keep rejection, non-numeric-keep rejection, text preservation in both paths. |

## Why wrap over re-implement

- `Cleanup.deleteAllRevisions` / `deleteRevisions` are tested in place and already used in production via the admin UI.
- A second implementation of the same primitive (what my earlier iteration did) risks diverging behavior between the admin-UI path and the CLI/API path.
- Operators who want finer control already have `--keep N`; hitting the full admin UI isn't a realistic workflow for scripted maintenance.

## Test plan
- [x] `pnpm run ts-check` clean locally after rebase onto develop
- [x] Backend tests: collapse mode, keep-last mode, validation errors
- [ ] CI green
- [ ] Manual: run `node bin/compactPad.js <padID>` against a pad with many revisions; observe the before/after count

Closes #6194


JohnMcLear:
This provides an API endpoint but not anything in /admin - is that okay?  
Where are the API docs?
Does this method provide a setting to allow X data of Y staleness to be compacted??

JohnMcLear:
**1. /admin coverage** — there's already a per-pad cleanup button in `/admin/pads` (the `FileStack` icon on each row, wired to `cleanupPadRevisions` in `adminsettings.ts`). It's the keep-last mode only and is gated on `settings.cleanup.enabled` + uses `settings.cleanup.keepRevisions`. So this PR's surfaces complement what's already there:

| Surface | Mode | Gate |
|---|---|---|
| /admin/pads button (already on develop) | keep-last only, uses `cleanup.keepRevisions` from settings | `cleanup.enabled` |
| `compactPad` API (this PR) | both — full-collapse or per-call `keepRevisions` | API auth |
| `bin/compactPad` CLI (this PR) | wraps the API; same modes | runs against the local API key |

This is per pad, not per instance

**2. API docs** — addressed in 26e12ff7. `compactPad` is now in both `doc/api/http_api.md` and `http_api.adoc`, and the documented "latest version" bumps from 1.3.0 to 1.3.1 to match `latestApiVersion`.

**3. Staleness gating** — not in this PR. The current knobs are:
- omit `keepRevisions` → full collapse
- `keepRevisions=N` → keep last N

"Compact pads where last edit was older than Y days" is a different shape — it's a *targeting* concern, not a *compaction* concern, so it composes naturally on top of this rather than living inside `compactPad` itself. Two natural follow-ups, both small:
- a `bin/compactStalePads.ts` that calls `listAllPads` → filters by `getLastEdited` → calls `compactPad` per match. ~30 lines.
- a settings-driven nightly job (e.g. `cleanup.compactOlderThanDays`) that runs the same loop on a timer.

Happy to do either as a follow-up PR. Which (if either) would you actually use? That'll tell me whether the right move is the script (operator pulls it on demand) or the cron (turn-it-on-and-forget).

Also: while reviewing for your comment I caught a real bug — the 1.3.1 entry registered `['padID', 'authorId']` but the function takes `keepRevisions`. The dispatcher in `APIHandler.handle` maps URL params by name, so over HTTP the keep-last branch was silently never reached. The direct-call backend tests didn't catch it because they bypass the handler. Fixed in 26e12ff7 with a new HTTP-level test that exercises the dispatch.

JohnMcLear:
Done — pushed in 59455af1:

- `bin/compactAllPads.ts` walks every pad on the instance and compacts each. Full-collapse by default, `--keep N` for keep-last, `--dry-run` to list pads + rev counts without writing.
- Per-pad failures don't abort the run; they're logged + counted, and the script exits 1 if anything failed.
- Prints `before → after` per pad and a totals line at the end.
- Registered as `pnpm --filter bin compactAllPads` (same with `compactPad`, which had been missing a script alias).

Deliberately did **not** add a `compactAllPads` HTTP endpoint — bulk compaction over a single HTTP request would be one giant response and a long-held connection. Operators running this locally get progress output and can `^C` cleanly.

Staleness gating is now tracked as #7642 with a concrete proposal (`bin/compactStalePads.ts` + an optional `cleanup.compactOlderThanDays` setting) so it doesn't get lost.

JohnMcLear:
**CI status on 75a08a13:**

| Check | Result |
|---|---|
| Backend tests (Linux ×2, with-plugins ×2) | ✅ |
| Frontend admin tests | ✅ |
| Type check | ✅ |
| build-test | ✅ |
| CodeQL / Analyze | ✅ |
| dependency-review | ✅ |
| Upgrade-from-latest-release | ✅ |
| rate limit | ✅ |
| Playwright Chrome | ❌ (pre-existing) |
| Playwright Firefox / with-plugins | still running |

The two Playwright Chrome failures (`enter.spec.ts:33` "enter is always visible after event" and `indentation.spec.ts:9` "indent text with keypress") are pre-existing flakes on develop — verified on develop's latest workflow run 25214868650 (commit 5e8704f8) which fails on the exact same two specs. Nothing I changed in this PR touches the editor frontend.

The 16 backend tests for `compactPad` and the new `runCompactAll` loop all pass.

JohnMcLear:
**Final CI:** all 4 Playwright variants now finished and all 4 failed on the same pre-existing specs:

- `enter.spec.ts:33` "enter is always visible after event" — fails on **all 4 variants** (chromium, chromium+plugins, firefox, firefox+plugins) **and on develop** (run 25214868650 on commit 5e8704f8)
- `indentation.spec.ts:9` "indent text with keypress" — same pattern, all 4 variants, also failing on develop
- `undo_redo_scroll.spec.ts:31` "Ctrl+Z scrolls viewport up when caret lands above view" — chromium+plugins only
- `ep_markdown` Firefox-with-plugins flake — third-party plugin spec, unrelated to etherpad core

None of these touch anything in this PR (CLI, `compactPad` API, docs). Everything I changed is green:
- Backend tests (Linux + plugin matrix ×2): ✅
- Type check, build-test, CodeQL, dependency-review: ✅
- 16 backend specs for `compactPad` + `runCompactAll`: ✅

Marking this ready for review when you're happy with the answers above.

JohnMcLear:
Pushed 5fb8039b7 addressing the Qodo review.

**Fixed (3):**
1. **cleanup.enabled gate** — the admin/Cleanup path checks this before doing anything destructive. The API was bypassing it. Now mirrors the admin check, returns a clear apierror when disabled. Test: `refuses to run when cleanup.enabled is false`.
2. **Number.isFinite → Number.isInteger** — `2.5` slipped through the old check and `Cleanup.deleteRevisions` assumes integer math. Reject at the API boundary now. Test: `rejects fractional keepRevisions`.
3. **SSL-aware baseURL** in `bin/compactPad.ts` and `bin/compactAllPads.ts` — uses `settings.ssl ? 'https' : 'http'` like the rest of the codebase. (Other bin scripts have the same hardcode but are out of scope here.)

API docs (`http_api.md` + `http_api.adoc`) updated to document the new gate and error message.

**Skipped Qodo items (intentionally):**
- "Wrong compactPad parameters" — already fixed in 26e12ff7 with `['padID', 'keepRevisions']`. Qodo's persistent review was scanning an older commit.
- "Unbounded revision deletions" / "No session eviction" / "Changeset base length mismatch" / "padCreate hook emitted" — all four targeted the earlier on-`Pad` `compactHistory()` implementation that was refactored away in commit 54913412b. The current code wraps `Cleanup.deleteAllRevisions` and `Cleanup.deleteRevisions`, which already handle concurrency (`timesLimit`), session eviction (via the temp-pad swap path), correctness (changeset built from real `oldAText`), and hook semantics.

**On Playwright failures:** 5e8704f8 ("feat(gdpr): pad deletion controls") landed on develop and broke `enter.spec.ts:33` and `indentation.spec.ts:9` — these specs pass on the prior develop commit (7357871ac) and fail on the GDPR commit. My PR just inherits that failure since it's rebased on develop. None of my changes touch the editor frontend; verifiable by diffing `git log develop..HEAD` against the failing specs' code paths.

20 backend specs passing locally for `compactPad` + `runCompactAll`.

PR #7568, opened 2026-04-20
feat(export): native DOCX export via html-to-docx (opt-in)

### LINKED ISSUE #7538: Support docx/pdf import/export natively
docx and pdfkit could potentially give us local support.

It would mean no requirement for abiword/libreoffice however would req. ~200MB (afaik in puppeteer) extra bloat on initial download so it's a trade off...

Also Abiword is dead so it's being removed.

JohnMcLear:
## Summary

Closes #7538. Replaces the original flag-gated DOCX-only path with a complete soffice-free import/export story:

- **Export:** With `settings.soffice = null`, pads can be exported as `html`, `txt`, `etherpad`, `docx`, **`pdf`** — all in-process, no subprocess, no native binaries.
- **Import:** `.html`, `.txt`, `.etherpad`, **`.docx`** — all in-process.
- **soffice configured:** behavior is unchanged bit-for-bit. There is no opt-in flag.

## Selection model

A single cascade in `ExportHandler.ts` and `ImportHandler.ts`:

- `sofficeAvailable() === 'yes'` → existing soffice path
- `'withoutPDF'` (Windows) → soffice for everything except `pdf`, which goes native
- `'no'` (soffice null) → native DOCX/PDF (export) and native DOCX (import); ODT/DOC/RTF (and PDF import) remain blocked with a clear message

No fallback chain — if a native converter throws, the response is 5xx with the error logged. Mirrors the spec's "soffice if installed, native otherwise, fail clearly" decision.

## Native converters

| Format | Library | Approach |
|---|---|---|
| DOCX export | `html-to-docx` | already in PR |
| PDF export | `pdfkit` + `htmlparser2` | small walker (~170 LOC) — pure JS, no jsdom |
| DOCX import | `mammoth` | mammoth → HTML → existing setPadHTML pipeline |

Total runtime install added: ~10–12 MB. Compared with ~500 MB for LibreOffice or ~200 MB for puppeteer, this is the right tradeoff for the structural-fidelity bar #7538 calls out.

## Hardening

New `stripRemoteImages` sanitizer removes any `` whose `src` is not `data:` or relative. Both DOCX and PDF export branches run plugin-modified HTML through it before conversion, closing Qodo's SSRF finding (#4) on the existing html-to-docx path *and* preventing the equivalent issue on PDF.

## Out of scope (follow-ups)

- Native ODT export — no maintained pure-JS writer in the ecosystem.
- Native PDF / ODT / DOC / RTF import — no mature pure-JS readers.
- Memory/timeout caps on conversion — add when production signal warrants.

## Test plan
- [x] `pnpm run ts-check` clean
- [x] Backend tests pass on a clean checkout: 1012 / 1012
- [x] Unit + integration coverage for sanitizer, walker, mammoth wrapper, DOCX/PDF export endpoints, DOCX import endpoint, ODT negative cases
- [ ] CI green
- [ ] Manual: with `SOFFICE=null`, export DOCX and PDF; both produce valid files
- [ ] Manual: with `SOFFICE=null`, import the fixture .docx and verify pad content

Closes #7538


JohnMcLear:
I feel like we can't drop pdf so need to have a conversation here...

JohnMcLear:
Qodo follow-up — addressed all four findings, plus expanded scope to close #7538 properly:

1. **Requirement gap (DOCX still needs soffice)** — fixed. Removed the `nativeDocxExport` flag entirely. Selection is now purely soffice-presence-driven: soffice configured → soffice; soffice null → native (html-to-docx for DOCX, pdfkit for PDF). No fallback chain.
2. **DOCX blocked without soffice** — fixed. Tightened the route guard to `['odt','doc']` only when `exportAvailable() === 'no'`; pdf/docx fall through to `ExportHandler`'s native dispatch. UI in `pad_impexp.ts` always shows DOCX + PDF links now.
3. **Native DOCX test bypass** — fixed. Tests use `settings.soffice = null` (was `'false'`) so they exercise the real no-soffice deployment shape.
4. **Unrestricted HTML-to-DOCX I/O** — fixed. New `stripRemoteImages` sanitizer drops non-`data:`/non-relative `` before either DOCX or PDF conversion. The PDF walker also rejects remote `` at its own boundary as defense-in-depth. No converter ever sees a remote URL.

Also added (per my "we can't drop pdf" comment):
- Native PDF export via `pdfkit` + `htmlparser2` walker (~170 LOC, well under the 500-LOC bail-out threshold defined in the spec)
- Native DOCX import via `mammoth` so a soffice-less deployment can also ingest .docx files

Design + plan committed to the branch:
- `docs/superpowers/specs/2026-05-08-native-docx-pdf-export-import-design.md`
- `docs/superpowers/plans/2026-05-08-native-docx-pdf-export-import.md`

src/node/handler/ExportHandler.ts
```diff
@@ -90,7 +90,72 @@ exports.doExport = async (req: any, res: any, padId: string, readOnlyId: string,
       return;
     }
 
-    // else write the html export to a file
+    // Soffice-first dispatch (issue #7538). When soffice is configured
+    // we keep the legacy convert-via-tempfile path; when it's not, we
+    // hand DOCX to html-to-docx and PDF to our pdfkit walker — both
+    // pure-JS, in-process. No fallback chain: native errors surface as
+    // 5xx so admins see real failures instead of silent shadowing.
+    const {sofficeAvailable} = require('../utils/Settings');
```
SamTV12345, line 98: At some point we should migrate to ESM. It's such a pain in editors.

SamTV12345 (approved):

src/node/handler/ExportHandler.ts
JohnMcLear: Noted, I'll add it to the list :)

PR #7580, opened 2026-04-21
feat: Remove paths from plugin packages

Gared:
This will remove realPath and path from plugin package before sending it to the client


SamTV12345:
Is there a better way to type that? I always try to avoid any.

PR #7583, opened 2026-04-22
feat(packaging): add Debian (.deb) build via nfpm with systemd unit (v2)

JohnMcLear:
## Summary

Reintroduces the Debian packaging from #7559 (reverted in #7582) with fixes for review feedback from #7559 and #7583:

1. **Rename `etherpad-lite` → `etherpad`** everywhere — package name, installed paths (`/opt/etherpad`, `/etc/etherpad`, `/var/lib/etherpad`, `/var/log/etherpad`), systemd unit (`etherpad.service`), CLI wrapper (`/usr/bin/etherpad`), `Documentation=` URL, and the `.deb` filename. Kept `replaces`/`conflicts`/`provides: etherpad-lite` so any dev builds upgrade cleanly.
2. **Default DB → `sqlite`** (was `dirty`). `postinstall` rewrites the seeded `/etc/etherpad/settings.json` to `dbType: "sqlite"` pointed at `/var/lib/etherpad/etherpad.db`.
3. **`/opt/etherpad/var` → `/var/lib/etherpad/var` symlink** in `postinstall`. Etherpad writes runtime files (`var/js`, `var/installed_plugins.json`, …) under `settings.root/var`, which resolves to `/opt/etherpad/var` in this layout — `ProtectSystem=strict` blocks that. The symlink keeps the existing `ReadWritePaths=/var/lib/etherpad` sufficient and avoids opening `/opt` for writes. (Fixes Qodo bug "Readonly /opt breaks startup".)
4. **Seed `installed_plugins.json`** in `postinstall` with `{"plugins":[{"name":"ep_etherpad-lite","version":"<v>"}]}`. Otherwise `checkForMigration()` falls back to `pnpm ls` on first boot, but `pnpm` is not a runtime dep — and the bundled `node_modules` already contains every shipped plugin, so no migration is needed. Also prevents accidental network plugin installs at first run. (Fixes Qodo bug "Plugin migration startup failure".)
5. **Verify nfpm `.deb` sha256** against upstream `checksums.txt` before `sudo dpkg -i` in CI. (Fixes Qodo bug "Unverified nfpm download".)
6. **CI runner Node 22 → 24** (current LTS, per @SamTV12345). The `Depends: nodejs (>= 20)` floor is unchanged — that mirrors Etherpad's `engines.node`.
7. **Stable `etherpad-latest_<arch>.deb` aliases** attached to each GitHub Release, so users can `curl …/releases/latest/download/etherpad-latest_amd64.deb` without knowing the version. README updated.
8. **README refresh**: drop stale `2.6.1` example, document the latest URL, point to `setup_lts.x` instead of pinning a NodeSource major.

## What you get installing the package

- `/opt/etherpad` with prebuilt, self-contained `node_modules/` — no pnpm at runtime, only `nodejs (>= 20)`.
- `etherpad` system user/group, created via `adduser` in `preinst`.
- `/etc/etherpad/settings.json` (seeded sqlite default; preserved across upgrades; removed on `purge`).
- `/var/lib/etherpad` writable under `ProtectSystem=strict`, with `var/` symlinked from `/opt/etherpad/var` so first-boot writes succeed.
- `/lib/systemd/system/etherpad.service` — hardened unit (`NoNewPrivileges`, `ProtectSystem=strict`, `ProtectHome`, `PrivateTmp`, `RestrictAddressFamilies`) with `Restart=on-failure`.
- `/usr/bin/etherpad` CLI wrapper running `node --import tsx/esm`.

## CI smoke test

`.github/workflows/deb-package.yml` triggers on `v*` tags, builds amd64 + arm64 via native runners (`ubuntu-latest` + `ubuntu-24.04-arm`), and on amd64:

1. Verifies nfpm checksum, then installs nfpm
2. `dpkg -i` installs the Etherpad `.deb`
3. Asserts `/opt/etherpad/var` symlinks to `/var/lib/etherpad/var`
4. Asserts `installed_plugins.json` exists and lists `ep_etherpad-lite`
5. Asserts `/etc/etherpad/settings.json` contains `"dbType": "sqlite"`
6. `systemctl start etherpad`
7. `curl /health` returns 200 (exits non-zero + dumps `journalctl` if not)
8. `dpkg --purge` removes config + user

The `release` job attaches both `etherpad_<ver>_<arch>.deb` and `etherpad-latest_<arch>.deb` to the GitHub Release.

## Not included (follow-up)

Publishing to an APT repo (Cloudsmith / Launchpad PPA / self-hosted reprepro) is out of scope — needs a governance decision on who holds the signing key. Recipes are in `packaging/README.md`.

## Legacy

`bin/buildDebian.sh` and `bin/deb-src/` are stale (Etherpad v1.3, init-based). Flagged for removal in a follow-up PR.

## Test plan
- [ ] `pnpm install --frozen-lockfile && pnpm run build:etherpad` succeeds
- [ ] Local `nfpm package --packager deb` produces a well-formed `.deb` (`dpkg-deb -I` / `-c`)
- [ ] Fresh install on Ubuntu 24.04: service starts, `/health` returns OK, `/etc/etherpad/settings.json` shows `"dbType": "sqlite"`, `/opt/etherpad/var → /var/lib/etherpad/var`, `installed_plugins.json` is seeded, `/var/lib/etherpad/etherpad.db` is created by the etherpad user
- [ ] Service starts cleanly without spawning `pnpm` (check `journalctl -u etherpad` for any pnpm references)
- [ ] Upgrade install: `/etc/etherpad/settings.json` and `installed_plugins.json` untouched; service restarted
- [ ] `apt remove` keeps `/etc` and `/var/lib`; `apt purge` removes them plus the `etherpad` user
- [ ] CI workflow succeeds on amd64 and arm64; `etherpad-latest_<arch>.deb` aliases attached to release

Refs #7529, #7559, #7582


.github/workflows/deb-package.yml
```diff
@@ -0,0 +1,164 @@
...
+          echo "Packaging version: ${VERSION}"
+
+      - uses: pnpm/action-setup@v6
+        with:
+          version: 10
+
+      - name: Setup Node
+        uses: actions/setup-node@v6
+        with:
+          node-version: '22'
```
SamTV12345: Current lts is 24

packaging/README.md
```diff
@@ -0,0 +1,95 @@
...
+  systemd/etherpad.service
+  systemd/etherpad.default
+  etc/settings.json.dist   # populated in CI from settings.json.template
+```
+
+Built artefacts land in `./dist/`.
+
+## Building locally
+
+Prereqs: Node 22, pnpm 10+, nfpm.
```
SamTV12345: Node 24 makes more sense I think

packaging/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+VERSION=$(node -p "require('./package.json').version") \
+ARCH=amd64 \
+    nfpm package --packager deb -f packaging/nfpm.yaml --target dist/
+```
+
+## Installing
+
+```sh
+sudo apt install ./dist/etherpad_2.6.1_amd64.deb
```
SamTV12345: Do we want to make this more generic? This is already out of date.

SamTV12345 (approved):

packaging/README.md
JohnMcLear: etherpad-latest.deb?

packaging/README.md
JohnMcLear: I follow Etherpad's own engine.node - if we want to bump node version we can, but this isn't the place to begin that bump IMHO.

Issue #7593, opened 2026-04-23
User Log out button overflows off-screen


JohnMcLear:
The **Log out** button in the admin interface overflows its container and extends off-screen, at least at common viewport widths.

### Reproduction

1. Drop User dialoge down
2. Observe the Log out button placement — it extends beyond the visible area.

### Expected

Log out button stays within the admin layout at all viewport sizes.

### Notes

This problem also exists when chat & users is permanently on the screen

Issue #7590, opened 2026-04-23
Chat panel (bottom-right) content stacks vertically / layout broken


JohnMcLear:
The chat panel that pops up from the bottom-right of the pad lays out its contents vertically stacked rather than in the intended layout. Controls (header buttons, message list, input) are no longer positioned correctly against each other.

### Reproduction

1. Open a pad.
2. Click the chat icon in the bottom-right.
3. Observe the panel: chat messages, header controls, and input are stacked in a way that doesn't match the intended layout.

### Notes

- Appears to be pre-existing (not caused by the open #7584 a11y PR). Suspicion: one of the flex/grid properties on `#chatbox` or `#chatbox-header` was broken by a previous CSS refactor.
- Discovered while manual-testing #7584. Listed as a separate concern because the a11y PR didn't touch the chat panel layout.

Willing to share screenshots on request.

JohnMcLear:
Attempting to shoe horn fix into #7584

PR #7597, opened 2026-04-23
fix(chat): icon click, disabled toggles, username layout (#7590, #7592, #7593)

### LINKED ISSUE #7592: Disable chat checkbox doesn't visually disable dependent 'Chat always on screen' control
In the pad Settings dialog, checking **Disable chat** disables the underlying behaviour but does not visually disable the dependent **Chat always on screen** control. The checkbox still looks enabled / clickable to the user, which is misleading UX.

### Reproduction

1. Open a pad.
2. Open Settings (toolbar cog icon).
3. Under "My View": tick the **Disable chat** checkbox.
4. Note that **Chat always on screen** is functionally disabled, but visually renders as if enabled (no greyed-out label, checkbox still looks interactive).

### Expected

When Disable chat is checked, the dependent "Chat always on screen" row should be visually disabled: greyed label, `disabled` attribute on the checkbox, etc.

### Notes

- Pre-existing (not related to the open #7584 a11y PR). The sibling-disable logic never existed; disabling is implicit in the JS but not reflected in the DOM/CSS.
- Small fix: toggle `disabled` attribute on `#options-chatandusers` in the `#options-disablechat` change handler, or wrap both in a fieldset that toggles `disabled` collectively.

Discovered while manual-testing #7584.

JohnMcLear:
Bundles three small chat-related bug fixes that all touch the same area of the editor window. Per review feedback.

## Closes

- #7590 — chat icon click sometimes did nothing (icon button click target was being eaten by the inner `.buttonicon` flex span after #7584)
- #7592 — "Chat always on screen" / "Show Chat and Users" rows didn't visibly disable when "Disable chat" was ticked, even though the underlying inputs were disabled
- #7593 — partial: revert of the original width-cap fix (no Log out button visible in vanilla etherpad-lite, and the cap made the username field unnecessarily narrow)

## Changes

**`src/static/css/pad/chat.css`**
- Reset flex / pointer behaviour on `#chaticon .buttonicon` so the inner glyph span doesn't intercept clicks meant for the `<button id="chaticon">`. Preferred over weakening the global `.buttonicon` rule, which the toolbar relies on for icon centring.
- Title-bar layout left as-is (the float-based original); a separate visual pass on `#titlecross` / `#titlesticky` positioning can come later.

**`src/static/js/chat.ts`**
- `chat.show()` now clears any inline `display: none` on `#chatbox` before adding the `.visible` class. Without this, doing **Disable Chat → Re-enable Chat → click chat icon** left the chatbox hidden by the inline style that `applyShowChat(false)` had set via jQuery `.hide()`.

**`src/static/skins/colibris/src/components/form.css`**
- Drop the `:checked` qualifier from the disabled-toggle visual rule so unchecked-but-disabled toggles also dim. Fixes #7592 in the colibris skin.

**`src/static/css/pad/popup.css`**
- Generic baseline for non-colibris skins: greyed label colour and `cursor: not-allowed` on disabled checkbox/radio rows in popups.

**`src/static/skins/colibris/src/components/users.css`**
- `#myusernameform` margin-left 35px → 10px to match the base CSS. The 35px value was sized for the chatAndUsers sticky layout but in the standalone Users popup it just opens a wide gap.

**`src/static/css/pad/popup_users.css`**
- Reverts the width:75px cap on `#myusernameform` and the forced `width: 100%` / `box-sizing: border-box` on `#myusernameedit` from the original PR commit. Vanilla etherpad-lite has no Log out button to overlap, and the cap makes the field too small.

## Test plan

Tested in browser with the colibris skin:
- [x] #7590 — clicking the corner chat icon opens the panel reliably; closing and re-opening still works
- [x] #7592 — ticking "Disable chat" greys both the labels and the colibris toggle visuals (opacity .4); untick restores
- [x] disable → enable → click chat icon → chat panel appears (regression fix from chat.ts change)
- [x] #7593 — username input renders at its natural width; no Log out overlap regression in environments that don't have one

### DIFF
--- a/src/static/css/pad/chat.css
+++ b/src/static/css/pad/chat.css
@@ -59,20 +59,27 @@
 }
 
 /* -- TITLE BAR -- */
+/* Single flex row, vertically centred:  [ CHAT          _  [] ]
+   - #titlelabel takes the remaining width so the controls sit at the
+     right edge.
+   - Source order is titlecross then titlesticky, which is also the
+     desired visual order (minus on the left, sticky on the right). */
 #titlebar {
   font-weight: bold;
   padding: 5px;
+  display: flex;
+  align-items: center;
+  gap: 8px;
 }
 #titlebar #titlelabel {
-  margin: 4px 0 0 4px;
-  display: inline;
+  margin: 0 0 0 4px;
   font-size: 1.4rem;
+  flex: 1;
 }
 #titlebar .stick-to-screen-btn,
 #titlebar .hide-reduce-btn {
   font-size: 25px;
   color: inherit;
-  float: right;
   text-align: right;
   text-decoration: none;
   cursor: pointer;
@@ -84,7 +91,6 @@
 }
 #titlebar .stick-to-screen-btn {
   font-size: 10px;
-  padding-top: 2px;
 }
 #titlebar .stick-to-screen-btn:focus-visible,
 #titlebar .hide-reduce-btn:focus-visible {
--- a/src/static/css/pad/chat.css
+++ b/src/static/css/pad/chat.css
@@ -62,17 +62,29 @@
 #titlebar {
   font-weight: bold;
   padding: 5px;
+  /* Lay the label and corner controls on a single row and bottom-align
+     them. The controls were previously `float: right` `<a>` tags, so the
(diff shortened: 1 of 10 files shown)


JohnMcLear:
waiting on reg tests to merge .  eta 15 mins

PR #7601, opened 2026-04-25
feat(updater): tier 1 — notify admin and pad users of available updates

JohnMcLear:
## Summary

Ships **Tier 1 (notify)** of the four-tier auto-update design at `docs/superpowers/specs/2026-04-25-auto-update-design.md`.

- Periodic poll of GitHub Releases (default 6h, ETag-cached) writes to `var/update-state.json`.
- Admin UI banner on every admin page + a dedicated `/admin/update` page (read-only) with version, install method, tier, changelog.
- Pad-side discreet footer badge — only renders when severely outdated or running a flagged-vulnerable version. Public endpoint `/api/version-status` never leaks the running version string.
- Optional escalating email nudges via new top-level `adminEmail` setting (weekly while vulnerable, monthly while severely outdated). Real SMTP transport deferred to a follow-up; PR 1 ships the cadence/dedupe machinery.
- New `updates.*` settings block with a default of `tier: "notify"`. Set `tier: "off"` to disable entirely.
- Tier 1 contains **no execution code**. PRs 2 (manual click), 3 (auto with grace window) and 4 (autonomous in maintenance window) are designed and will land in subsequent releases.

See `doc/admin/updates.md` for full configuration.

## Architecture

- `src/node/updater/` — pure modules: `types`, `versionCompare`, `state` (atomic write + schema validation), `InstallMethodDetector`, `UpdatePolicy`, `VersionChecker` (fetcher abstraction), `Notifier` (cadence decider) + boot wiring (`index.ts`).
- `src/node/hooks/express/updateStatus.ts` — public `/api/version-status` (cached 60s, no version leak) and admin `/admin/update/status`.
- Settings in `src/node/utils/Settings.ts` + `settings.json.template` + `settings.json.docker`.
- Admin UI: `admin/src/components/UpdateBanner.tsx`, `admin/src/pages/UpdatePage.tsx`, store slice, route, nav link, CSS.
- Pad UI: `src/static/js/pad_version_badge.ts`, `src/templates/pad.html`, `src/static/css/pad.css`.
- i18n: 13 keys in `src/locales/en.json`.
- Docs: `doc/admin/updates.md` (new), `CHANGELOG.md` (Unreleased section).

## Test plan

- [x] vitest unit tests for `versionCompare`, `state`, `InstallMethodDetector`, `UpdatePolicy`, `VersionChecker`, `Notifier`
- [x] mocha integration tests for `/admin/update/status` (admin auth, shape) and `/api/version-status` (no version leak, severe state)
- [x] Playwright admin spec — banner render, page render, nav link
- [x] Playwright pad spec — badge visibility on `null` / `severe` / `vulnerable`
- [x] Manual smoke (real boot, real curl, browser-tested admin login + banner + page + pad badge with seeded state)
- [x] `pnpm ts-check` clean, `pnpm run build:ui` succeeds

## Notes

- Email transport is intentionally not wired in PR 1 — Notifier logs `(would send email)`. Follow-up PR will add nodemailer or rely on a notification plugin.
- Pad badge strings are hardcoded English (TODO comment); switching to html10n is a separate PR (the pad uses html10n, the admin uses i18next).
- All 23 commits on this branch include corresponding tests for the changes they make. Four `fix(updater)` commits address bugs surfaced during code review (semver regex four-part rejection, null-email/array schema validation, ETag preservation on prerelease, tagChanged email cadence).


admin/src/components/UpdateBanner.tsx
```diff
@@ -0,0 +1,35 @@
...
+import {useStore} from '../store/store';
+
+export const UpdateBanner = () => {
+  const {t} = useTranslation();
+  const updateStatus = useStore((s) => s.updateStatus);
+  const setUpdateStatus = useStore((s) => s.setUpdateStatus);
+
+  useEffect(() => {
+    let cancelled = false;
+    fetch('/admin/update/status', {credentials: 'same-origin'})
```
SamTV12345, line 13: Best would be to use tanstack query. If you use fetch in a call it is mostly always wrong see https://react.dev/reference/react/useEffect#fetching-data-with-effects

admin/src/components/UpdateBanner.tsx
```diff
@@ -0,0 +1,35 @@
...
+      .then((r) => r.ok ? r.json() : null)
+      .then((data) => { if (data && !cancelled) setUpdateStatus(data); })
+      .catch(() => {});
+    return () => { cancelled = true; };
+  }, [setUpdateStatus]);
+
+  if (!updateStatus || !updateStatus.latest) return null;
```
SamTV12345, line 20: If you useSuspense Query you can save the manual return null and show with suspense boundary that the ui is loading e.g. with a loading spinner

admin/src/pages/UpdatePage.tsx
```diff
@@ -0,0 +1,103 @@
...
+  | {kind: 'ok'};
+
+export const UpdatePage = () => {
+  const {t} = useTranslation();
+  const us = useStore((s) => s.updateStatus);
+  const setUpdateStatus = useStore((s) => s.setUpdateStatus);
+  // Self-fetch so the page renders an explicit state even if UpdateBanner's
+  // best-effort fetch never landed (route returns 404 when tier=off, 401/403
+  // if requireAdminForStatus is set, or a transient network error).
+  const [fetchState, setFetchState] = useState<FetchState>(us ? {kind: 'ok'} : {kind: 'loading'});
```
SamTV12345, line 19: Same as this :)

SamTV12345 (commented):
I'd like to get rid of the manual data fetching. It is a lot of code and it seems to be the wrong one. Please use tanstack query. If you have the rest api in the swagger you can refer to openapi-typescript and openapi-react-query with openapi-fetch. It generates a typesafe client with basically no boilerplate.

JohnMcLear:
Okidoki will do

JohnMcLear:
Thanks @SamTV12345 — agree the `useEffect`+`fetch` pattern is the wrong shape long-term, and the openapi-react-query + openapi-fetch combo is genuinely nice (codegen from our existing Swagger spec means no hand-written endpoint types at all).

That said, TanStack Query isn't currently a dep in `admin/` and none of the existing admin pages use it — they all use raw `fetch`+`useEffect` (settings, plugins, users, etc.). If I migrate just `UpdateBanner` and `UpdatePage` here, the admin ends up with two competing patterns mid-codebase, and the next reviewer will reasonably ask why the rest of admin isn't migrated too.

I'd rather do this as a dedicated PR that adopts the codegen + TanStack Query stack across `admin/` coherently. Filed #7638 to track that, with the openapi-typescript / openapi-fetch / openapi-react-query approach you suggested as the centerpiece. Happy to take it on once this lands.

For this PR, can we keep the existing raw-fetch shape (matching what's already in admin) and address the migration in #7638?

SamTV12345:
> Thanks @SamTV12345 — agree the `useEffect`+`fetch` pattern is the wrong shape long-term, and the openapi-react-query + openapi-fetch combo is genuinely nice (codegen from our existing Swagger spec means no hand-written endpoint types at all).
> 
> That said, TanStack Query isn't currently a dep in `admin/` and none of the existing admin pages use it — they all use raw `fetch`+`useEffect` (settings, plugins, users, etc.). If I migrate just `UpdateBanner` and `UpdatePage` here, the admin ends up with two competing patterns mid-codebase, and the next reviewer will reasonably ask why the rest of admin isn't migrated too.
> 
> I'd rather do this as a dedicated PR that adopts the codegen + TanStack Query stack across `admin/` coherently. Filed #7638 to track that, with the openapi-typescript / openapi-fetch / openapi-react-query approach you suggested as the centerpiece. Happy to take it on once this lands.
> 
> For this PR, can we keep the existing raw-fetch shape (matching what's already in admin) and address the migration in #7638?

Yes I am okay with that. That way we can cleanly separate the feature and the tech change

JohnMcLear:
Great, thanks :)

PR #7605, opened 2026-04-26
Backend esm vitest

SamTV12345:


JohnMcLear:
?????

SamTV12345:
> ?????

So far that works. So we can convert to ESM and use modern test runners like vitest + have actual editor integrations. With require everything is any.

JohnMcLear:
Hey @SamTV12345 can you document what you are trying to achieve here?  I see a lot of AI fixes but I feel like this PR will never land in the state and it's just a waste of flops.

Like if you are trying to port to mode modern syntax then I'd suggest a phased approach (lots of iterative small PRs) for example like how we did the supervised updater in 4 phases.

Issue #7622, opened 2026-04-28
Revive plugin-own frontend tests (the legacy mocha+helper specs)


JohnMcLear:
Companion to #7608 / PR #7609.

PR #7609 wires up "run **core's** frontend tests with the /ether plugin set loaded" — catches *core changes that break plugin UX*. That's one direction.

The *other* direction — "run **a plugin's own** frontend tests against core" (catch *core changes that break a plugin's behavior*) — has been broken across the entire ecosystem since 2023.

**Change type:** N/A (tracking issue / scope discovery)

## What's broken

Every official plugin still ships a `static/tests/frontend/specs/test.js` written in the legacy mocha + `helper.padChrome$` jQuery-style API, e.g.:

```js
// ep_headings2/static/tests/frontend/specs/test.js
describe('ep_headings2 - Set Heading and ensure its removed properly', function () {
  beforeEach(function (cb) { helper.newPad(cb); });
  it('Heading select has aria-label for accessibility', async function () {
    const chrome$ = helper.padChrome$;
    const $select = chrome$('#heading-selection');
    expect($select.attr('aria-label')).to.not.be(undefined);
  });
});
```

That format requires the in-page jQuery test runner that lived at `static/index.html` (the old "Etherpad jQuery frontend tests" page). That runner — and the workflow that loaded it — was removed in `cc80db2d3` (2023-07-02) when the workflows were nuked en masse. Nothing has picked these specs up since.

Worse, the plugin's own CI **looks like** it runs them. `ep_headings2/.github/workflows/frontend-tests.yml` ends with:

```yaml
cd src
pnpm exec playwright install chromium  --with-deps
pnpm run test-ui --project=chromium
```

But `test-ui` resolves to `playwright test tests/frontend-new/specs --project=chromium`, which only scans **core's** `src/tests/frontend-new/`. The plugin's `static/tests/frontend/specs/test.js` is never executed. So the plugin's CI green badge is misleading — it tests core's specs with the plugin loaded (== what #7609 does), not the plugin's own assertions.

## Two paths to fix

[...]

PR #7625, opened 2026-04-28
test(playwright): use insertText so Firefox stops dropping keystrokes

JohnMcLear:
## Summary

`writeToPad` has been calling `page.keyboard.type`, which dispatches one keydown/keyup per character against the inner `#innerdocbody` contenteditable. Under WITH_PLUGINS load Firefox's input pipeline can't keep up while plugin hooks are still warming, and randomly swallows characters from the tail of the string — the pad ends up with e.g. \`aligned tex\` instead of \`aligned text\`. The dropped character is irrecoverable: there's no event for the helper to retry against.

Switch to `page.keyboard.insertText`, which dispatches a **single** input event per call. Etherpad's `incorporateUserChanges` loop reads the resulting DOM atomically, so the visible result is identical to typing — minus the per-key race.

`insertText` does not translate `\n` into a real Enter keystroke (it concatenates `"One\nTwo"` into `"OneTwo"`), so split on newlines and press Enter between segments to preserve multi-line input. This matches what the existing callers (`timeslider_line_numbers`, `page_up_down`, etc.) rely on.

**Change type:** patch (test infrastructure only; no production behaviour change).

## Why this matters

After #7622 began discovering plugin Playwright specs in CI, the Firefox-with-plugins job started failing on every run. The most visible casualty was `ep_align`'s `Alignment of Text` suite (all 4 cases × 5 retries = 20 failures per run). Tracing it down showed the `<center>` wrapper element never appeared because the click into the toolbar was never preceded by the text actually being typed — `keyboard.type` had silently dropped the trailing char(s), leaving `selectAllText` selecting nothing and the alignment click no-op.

The same root cause was producing intermittent failures across other plugin specs that exercise typing.

## Test plan

- [x] `Playwright Chrome with plugins` ep_align Alignment: 4/4 pass locally (was 4/4 before; no regression)
- [x] `Playwright Firefox with plugins` ep_align Alignment: 4/4 pass locally (was 0/4 with retries previously)
- [x] `Playwright Firefox` core italic.spec: 2/2 pass
- [x] `tests/frontend-new/specs/timeslider_line_numbers.spec.ts` (which writes \`One\nTwo\nThree\` and asserts three lines): passes — confirms the newline-splitting branch
- [ ] CI to verify across the full suite

## What this PR does **not** address

There are still independent failures in the Firefox-with-plugins suite that have **different** root causes (plugin spec asserts that don't match runtime DOM, missing/wrong selectors, etc.):

- \`ep_headings2\` "Option select is changed when heading is changed"
- \`ep_markdown\` "Bold section renders the markdown class on body when 'Show Markdown' is enabled"
- \`ep_spellcheck\` "Spellcheck is on by default when not disabled"
- core \`tests/frontend-new/specs/timeslider_identity_changeset.spec.ts\` "timeslider playback advances..."

Those need per-spec fixes — out of scope here. Filing a tracking issue separately.

### DIFF
--- a/src/tests/frontend-new/helper/padHelper.ts
+++ b/src/tests/frontend-new/helper/padHelper.ts
@@ -142,7 +142,18 @@ export const clearPadContent = async (page: Page) => {
 export const writeToPad = async (page: Page, text: string) => {
   const body = await getPadBody(page);
   await body.click();
-  await page.keyboard.type(text);
+  // Use insertText (single input event) instead of keyboard.type
+  // (one keydown/keyup per char). Firefox under WITH_PLUGINS load
+  // racily drops characters from per-key events; insertText delivers
+  // each chunk in one event, which Etherpad's incorporateUserChanges
+  // pipeline handles atomically. insertText does not translate \n
+  // into a real Enter keystroke, so split on newlines and press
+  // Enter between segments to preserve multi-line input.
+  const lines = text.split('\n');
+  for (let i = 0; i < lines.length; i++) {
+    if (lines[i]) await page.keyboard.insertText(lines[i]);
+    if (i < lines.length - 1) await page.keyboard.press('Enter');
+  }
 }
 
 export const clearAuthorship = async (page: Page) => {


JohnMcLear:
> No regression test for writeToPad()

The regression coverage already exists in three places that all broke before this fix and pass after it (verified locally on Firefox + WITH_PLUGINS):

- `tests/frontend-new/specs/italic.spec.ts` — types `Foo`, asserts the editor saw it. Was failing intermittently with a partial string in the body.
- `tests/frontend-new/specs/timeslider_line_numbers.spec.ts` — calls `writeToPad(page, 'One\nTwo\nThree')` and asserts three line numbers. Specifically exercises the newline-splitting path.
- The four `Alignment of Text` cases in `ep_align`'s spec — type `aligned text` then click an alignment button. Previously failed 0/4 (with retries) on Firefox; now passes 4/4. This was the original symptom that surfaced the bug.

Reverting `writeToPad` to `keyboard.type(text)` makes all three of these go red. Adding a fourth standalone test that re-asserts "writeToPad delivers exactly N characters" would be redundant — its only signal would duplicate what those three already detect.

Happy to add one anyway if there's a strong preference, but flagging that the coverage is real, not absent.

Issue #7611, opened 2026-04-27
Frontend specs that fail with /ether plugin set loaded


JohnMcLear:
Tracking issue for the small set of frontend specs that are skipped under `WITH_PLUGINS=1` in PR #7609 (the with-plugins frontend coverage) so the suite can stay green while we fix them individually.

**Change type:** N/A (tracking issue)

Each of these can be a separate small PR. The fix pattern is usually one of:
- Selector/index assumption (e.g. `.nice-select.nth(1)`) that drifts when plugins add UI before/after the targeted element. Fix: target by ID or scoped container. Example precedent: the language-spec fix in #7609 itself (used `#languagemenu + .nice-select`).
- Hardcoded `waitForTimeout(N)` that races slower pad boot when plugins load. Fix: replace with explicit selector waits or longer waits.
- Suite-only flake — passes in isolation, fails in the full run. Usually shared-state leakage between specs (clipboard, cookies, pad ID collision).
- Genuine plugin-vs-core behavior conflict (less common; needs plugin or core fix).

The with-plugins jobs run with `retries: 5` (vs. `retries: 2` for vanilla) to absorb the higher flake rate from plugins slowing pad boot — but a number of specs still don't recover within that budget.

## Skipped specs

Plugin-vs-core selector/timing issues:

- `src/tests/frontend-new/specs/enter.spec.ts:33` — `enter is always visible after event`.
- `src/tests/frontend-new/specs/indentation.spec.ts:56` — `indents text with spaces on enter if previous line ends with ':', '[', '(', or '{'`.
- `src/tests/frontend-new/specs/indentation.spec.ts:118` — `appends indentation to the indent of previous line if previous line ends with ':', '[', '(', or '{'`.
- `src/tests/frontend-new/specs/timeslider_follow.spec.ts:50` — `only to lines that exist in the pad view, regression test for #4389`.
- `src/tests/frontend-new/specs/undo_redo_scroll.spec.ts:26` — `Ctrl+Z scrolls viewport up when the caret lands above the view`.
- `src/tests/frontend-new/specs/undo_redo_scroll.spec.ts:71` — `Ctrl+Z scrolls viewport down when the caret lands below the view`.
- `src/tests/frontend-new/specs/list_wrap_indent.spec.ts` (entire describe) — `numbered list wrapped line indentation`.
- `src/tests/frontend-new/specs/clear_authorship_color.spec.ts:73` — `clears authorship when first line has line attributes`.
[...]

Issue #5718, opened 2023-06-16
Consecutive numbering fails


NeuerDennis:
**Describe the bug**
Instead of getting consecutive numbering in different lines (1. 2. 3.) it stays at "1." (1. 1. 1.).

**To Reproduce**
1. Go to '[...](https://yopad.eu/p)' or a similiar etherpad instance
2. Create a Standard Bulletpoint and write something in the line
3. Press Enter to get to the next line and press Tab once to create a sub-bullet point
4. Now enter the next line without pressing Enter and write something down
5. Mark this line and activate numbering

**Expected behavior**
If you now press enter for the next line the numbering stays at "1." instead of going to "2.". This continues to happen in every other line you create under this.

**Server (please complete the following information):**
 - Etherpad version:
 - OS: [e.g., Ubuntu 20.04]
 - Node.js version (`node --version`):
 - npm version (`npm --version`):
 - Is the server free of plugins: 

**Desktop (please complete the following information):**
 - OS: Windows 10 
 - 19041.1.amd64fre.vb_release.191206-1406 10.0.0.0.2.101
 - Browser Google Chrome, 115.0.5790.32, Google LLC, CHROME.EXE, CHROME.EXE

JohnMcLear:
Confirmed.  Needs test coverage and a fix.

arvind9140:
@NeuerDennis  and @JohnMcLear  I am interested to solve this  issue . Please assign me .

SamTV12345:
> @NeuerDennis and @JohnMcLear I am interested to solve this issue . Please assign me .

I assigned you to the issue.

arvind9140:
I am understand  this issue but  i am not able to find   location of issue file please share issue file location @SamTV12345

SamTV12345:
I'll send the relevant file today or tomorrow. Depends on how long I can stay awake.

arvind9140:
Ok.. I am Waiting.

SamTV12345:
It is done in here: https://github.com/ether/etherpad-lite/blob/f1fea377818ec966a8b6e3c2a11786bb9dfb9133/src/static/js/ace2_inner.js#L67

Mahas1234:
Is the issue still open @SamTV12345  I'm a new contributor can I fix it

SamTV12345:
> Is the issue still open @SamTV12345  I'm a new contributor can I fix it
> 
> 

Yes feel free to fix this :)

[...]

Issue #5160, opened 2021-08-24
Ordered list not incrementing when put directly after an unordered list


Vringe:
**Describe the bug**
If you put an ordered list directly after an unordered list, the numbers are not incrementing.
As soon as you put an empty line between, the numbers are correct.

**To Reproduce**
1. Go to https://video.etherpad.com/
2. Create a new pad
3. Create an unordered list with 2 list items
4. Create an ordered list directly after

**Expected behavior**
An ordered list should increment its number

**Screenshots**
<img width="222" alt="Screenshot 2021-08-24 at 10 04 52" src="https://user-images.githubusercontent.com/11569136/130580289-5d4334ed-5d8c-442e-890f-a40c8f0046d2.png">

**Server (please complete the following information):**
 - Etherpad version: 1.8.14
 - OS: Host runs Ubuntu 18.04 (But I'm using the official docker image)
 - Node.js version (`node --version`): v14.17.1
 - npm version (`npm --version`): 6.14.13

**Desktop (please complete the following information):**
 - OS: Mac OS 11.4
 - Browser Chrome & Firefox
 - Version 92.0.4515.159 (Chrome) & 91 (Firefox)

webzwo0i:
Thank you!

As a side note: importing HTML that was exported from a pad which is in a state as described above (ie all list numbers are "1") will also result in false list numbers. So I assume the bug is in contentcollector.js

JohnMcLear:
Confirmed.  Will work on this ASAP :)

JohnMcLear:
Test coverage is available, I intend to work on a fix once we have some bad weather and I'm stuck inside.

rhansen:
Possibly related to #5176.

PR #7630, opened 2026-04-29
ci(frontend-tests): exclude ep_cursortrace + un-flake 30 of 31 #7611 skips

JohnMcLear:
## Summary

Two changes that together un-flake the **`Playwright * with plugins`** CI matrix:

1. **Drop `ep_cursortrace` from `frontend-tests.yml`'s WITH_PLUGINS plugin set.** Bisected as the sole cause of the Firefox-with-plugins flakiness this PR was chasing. Its `aceEditEvent` hook fires per keyboard event and unconditionally sends a socket `cursorPosition` message — under the test harness's `writeToPad` bursts that stream saturates the editor's input pipeline in Firefox, dropping keystrokes and producing the entire class of #7611 symptoms. The plugin itself needs a debounce around the socket send before it can come back into the test set; that fix lives outside this PR.

2. **Remove 30 of 31 `test.skip(WITH_PLUGINS, '#7611')` markers** across the `src/tests/frontend-new/specs/` tree. With `ep_cursortrace` out of the plugin set, those tests pass under WITH_PLUGINS=1. The remaining 1 skip is `timeslider_follow #4389`, which builds ~120 sequential Enter keypresses — needs a different setup mechanism (REST-driven import or clipboard paste) regardless of which plugins are loaded.

**Change type:** patch (CI/test infrastructure only; no production behaviour change).

## Bisection (4 iterations on this branch)

| Iter | Plugin sub-set | Firefox+plugins result |
|---|---|---|
| 1 | A: `align, author_hover, cursortrace, font_size, headings2` | ❌ fails |
| 1 | B: `markdown, readonly_guest, set_title_on_pad, spellcheck, subscript_and_superscript, table_of_contents` | ✅ passes |
| 2 | A1: `align, author_hover` | ✅ passes |
| 2 | A2: `cursortrace, font_size, headings2` | ❌ fails |
| 3 | A2a: `cursortrace` | ❌ fails (1 plugin alone tips it) |
| 3 | A2b: `font_size, headings2` | ✅ passes |
| 4 | All 10 minus `cursortrace` | ✅ passes (×2 independent Firefox runs) |

## Helper changes (carry-overs from before the bisection landed)

These all stay because they're either independent improvements or precondition fixes once the cursortrace load is gone:

- `helper/padHelper.ts`: added `waitForEditorReady` used by `goToNewPad` / `goToPad`. Blocks until `#innerdocbody[contenteditable="true"]`. Without this, specs that started interacting immediately after `#editorcontainer.initialized` would race the ace static→editable flip.
- Per-spec swaps from `page.keyboard.type` to `keyboard.insertText` / `writeToPad` (continuation of #7625's logic). Reliable in Firefox under any non-cursortrace plugin load.
- Per-spec `force:true` on toolbar button clicks after `selectAllText` (matches the existing `clearAuthorship` pattern for the `#toolbar-overlay` interception).
- `enter.spec.ts`: replaced an unverified Enter-press loop with a per-iteration `expect(...).toHaveCount(...)` so the loop doesn't advance until the previous press has landed.
- `ordered_list.spec.ts`: switched the OL toolbar selector from `.buttonicon-insertorderedlist.first()` to `button[data-l10n-id='pad.toolbar.ol.title']` (per Qodo review — plugin-resilient, no `.first()` index assumption).

## Test plan

- [x] Bisection confirmed via 4 CI iterations.
- [x] Final 10-plugin set passes both Chrome+plugins and Firefox+plugins (also 2× independent Firefox confirmation runs).
- [x] No regressions on Chromium / no-plugins.
- [x] Locally verified all batches on Firefox + WITH_PLUGINS=1 (against the older test setup that included cursortrace; the patterns the helper/spec changes establish are still correct).

## Followups (not blockers for this PR)

- Add a `mousemove`/`keypress` debounce to `ep_cursortrace`'s `aceEditEvent` so it can return to the test plugin set.
- `timeslider_follow.spec.ts` `#4389` — only remaining `#7611` skip; needs a non-keyboard setup path.
- Several core specs (`Linux with Plugins (24)` `SessionStore.ts shutdown cancels timeouts`) hit a separate timing flake unrelated to this work; saw it fire ~3× across this session's CI runs.

### DIFF
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -198,11 +198,10 @@ jobs:
       - name: Install all dependencies and symlink for ep_etherpad-lite
         run: pnpm install --frozen-lockfile
       - name: Install Etherpad plugins
-        # BISECT-A2: other half of the 5-plugin set that failed
-        # iteration 1. Job still runs --project=firefox per probe.
+        # BISECT-A2b: Iteration 3 — the other half of the failing
+        # set without ep_cursortrace.
         run: >
           pnpm add -w
-          ep_cursortrace
           ep_font_size
           ep_headings2
       - name: Create settings.json
@@ -275,13 +274,11 @@ jobs:
       - name: Install all dependencies and symlink for ep_etherpad-lite
         run: pnpm install --frozen-lockfile
       - name: Install Etherpad plugins
-        # BISECT-A1: Iteration 2 — HALF A failed iteration 1, so
-        # culprit is in the 5-plugin set. Splitting into A1 (here)
-        # and A2 (Chrome-with-plugins job).
+        # BISECT-A2a: Iteration 3 — A2 (cursortrace+font_size+headings2)
+        # failed iter 2. Singling out ep_cursortrace.
         run: >
           pnpm add -w
-          ep_align
-          ep_author_hover
+          ep_cursortrace
       - name: Create settings.json
         run: cp ./src/tests/settings.json settings.json
       - name: Run the frontend tests
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -198,14 +198,10 @@ jobs:
       - name: Install all dependencies and symlink for ep_etherpad-lite
         run: pnpm install --frozen-lockfile
       - name: Install Etherpad plugins
-        # Same plugin set as backend-tests.yml's withpluginsLinux job.
+        # BISECT-B: HALF B of the standard plugin set. Sister job
+        # Firefox-with-plugins has HALF A. Revert before merge.
         run: >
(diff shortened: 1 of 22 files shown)


JohnMcLear:
Addressed in 7157af107: replaced `page.locator('.buttonicon-insertorderedlist').first()` with `page.locator("button[data-l10n-id='pad.toolbar.ol.title']")` — the localizationId declared in `src/node/utils/toolbar.ts` for the orderedlist button. Unique by construction, no `.first()` needed, plugin-resilient.

Left the class-based locator in `#5160`, `#5718`, and the indent/outdent sub-describes since they don't currently strict-mode-match more than one element. Happy to swap those too for consistency if you'd prefer.

PR #7636, opened 2026-04-30
feat(pad): add theme-color meta to match toolbar on mobile (#7606)

### LINKED ISSUE #7606: Add <meta name="theme-color"> html to pads.
At the moment, there is white space above the pads on mobile, independent of the pad theme.

I think it would be good to add the meta value "theme-color" to the pads to solve this problem, more details here: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/theme-color

The theme-color should be the same color as the toolbar.

JohnMcLear:
## Summary

Closes #7606.

Mobile browsers paint the address-bar / status-bar strip above the viewport using a system color when `<meta name="theme-color">` is absent, leaving a visible gap above the Etherpad toolbar. This PR renders `theme-color` server-side so the bar blends into the configured toolbar from the first paint.

- New `src/node/utils/SkinColors.ts` resolves the colibris `*-toolbar` skin variants to the `--bg-color` values declared in `colibris/src/pad-variants.css`. Iteration order matches the CSS source so the last applicable token wins, mirroring the cascade.
- The helper returns `null` for non-colibris skins (`no-skin`, third-party skins) since their toolbar colors aren't known server-side; the templates then omit the meta rather than emit a misleading value.
- `pad.html` and `timeslider.html` emit a single unconditional `<meta name="theme-color">` when the helper returns a color. No `prefers-color-scheme` media-query variants — the runtime dark-mode auto-switch in `pad.ts` is also gated on a localStorage white-mode override that a media query cannot express, so any media-query variant would risk painting a dark address bar over a still-light toolbar.
- Backend tests in `specialpages.ts` cover the rendered metas and the non-colibris skip path; vitest unit tests cover the helper.

## Semver

`patch` — additive HTML metadata, no behavior change for existing clients.

## Test plan

- [ ] Open a pad on a mobile device / Chromium devtools mobile emulator and confirm the address bar matches the toolbar.
- [ ] Configure `skinVariants` with `dark-toolbar`; confirm the address bar matches the dark toolbar.
- [ ] Configure `skinName: "no-skin"`; confirm the page renders with no `theme-color` meta.
- [ ] cd src && pnpm test -- tests/backend/specs/specialpages.ts
- [ ] cd src && pnpm test:vitest tests/backend-new/specs/SkinColors.ts

Generated with Claude Code

### DIFF
--- a/src/node/utils/SkinColors.ts
+++ b/src/node/utils/SkinColors.ts
@@ -1,0 +1,30 @@
+'use strict';
+
+// Toolbar background colors that the colibris skin variants resolve to.
+// Mirrors --bg-color in src/static/skins/colibris/src/pad-variants.css so
+// that <meta name="theme-color"> can match the toolbar on first paint
+// (before client-side JS runs).
+const TOOLBAR_COLORS: {[variant: string]: string} = {
+  'super-light-toolbar': '#ffffff',
+  'light-toolbar': '#f2f3f4',
+  'dark-toolbar': '#576273',
+  'super-dark-toolbar': '#485365',
+};
+
+const DEFAULT_LIGHT = '#ffffff';
+const DEFAULT_DARK = '#485365';
+
+export const toolbarThemeColors = (skinVariants: string | undefined | null) => {
+  const tokens = (skinVariants || '').split(/\s+/).filter(Boolean);
+  let light = DEFAULT_LIGHT;
+  let dark = DEFAULT_DARK;
+  for (const token of tokens) {
+    const color = TOOLBAR_COLORS[token];
+    if (!color) continue;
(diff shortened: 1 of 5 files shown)


JohnMcLear:
On Qodo finding #4 ("theme-color lacks feature flag"): leaving this as-is and not gating it.

`<meta name="theme-color">` is additive HTML metadata that fixes a mobile UX bug (#7606) where the browser address bar painted a system color above the toolbar. It's not a behavior change for any client, has no runtime cost, and falls back gracefully on browsers that ignore it. Adding a config flag would just add a setting nobody needs to toggle.

SamTV12345 (approved):

PR #7647, opened 2026-05-01
test(ci): widen timing windows for flaky SessionStore + socket.io tests

JohnMcLear:
## Summary

Two flaky test groups dominate develop CI failures (~30% of recent runs over the past week). Both are timing fragility, not logic bugs.

- **`tests/backend/specs/SessionStore.ts`** — `set(sess)` schedules a timeout based on the cookie's `expires`. Tests use 100ms expiries + 110ms waits. On slow runners (Windows / loaded Linux), the wall-clock between `set()` and the assertion can exceed 100ms; `_updateExpirations` then sees `exp.real <= now` and calls `_destroy()`, deleting the DB record. The next assertion reads `null` / `undefined` and fails. Tripled affected windows (100→300, 110→330, 200→600) — semantics preserved, ~+3s headroom.
- **`tests/backend/common.ts:waitForSocketEvent`** — 1s timeout. Socket.io handshake + auth + CLIENT_READY round-trip can exceed 1s on a slow runner. Bumped to 5s.

This addresses the specific test cases consistently appearing in failed runs:
- SessionStore: `get of record from previous run (not yet expired)`, `external expiration update is picked up`, `shutdown cancels timeouts`
- socketio: `!authn anonymous cookie /p/pad -> 200, ok`, `authn user /p/pad -> 200, ok`
- clientvar_rev_consistency: `CLIENT_VARS stays consistent under concurrent edits during handshake`

The latest develop commit happened to be green, but ~25–40% of recent develop runs failed on at least one of these. This PR removes that flake.

## Test plan
- [x] `mocha tests/backend/specs/SessionStore.ts` — 28 pass locally (5s)
- [x] `mocha tests/backend/specs/socketio.ts` — 33 pass locally (2s)
- [x] CI Backend tests green across all 8 OS×Node matrix combos
- [x] CI "Upgrade from latest release" green

### DIFF
--- a/src/tests/backend/common.ts
+++ b/src/tests/backend/common.ts
@@ -126,7 +126,7 @@ export const waitForSocketEvent = async (socket: any, event:string) => {
       const timeout = setTimeout(() => {
         reject(new Error(`timed out waiting for ${event} event`));
         cancelTimeout = () => {};
-      }, 1000);
+      }, 5000);
       cancelTimeout = () => {
         clearTimeout(timeout);
         resolve();
--- a/src/tests/backend/common.ts
+++ b/src/tests/backend/common.ts
@@ -111,7 +111,7 @@ export const init = async function () {
  * @param {string} event - The socket.io Socket event to listen for.
  * @returns The argument(s) passed to the event handler.
  */
-export const waitForSocketEvent = async (socket: any, event:string) => {
+export const waitForSocketEvent = async (socket: any, event:string, timeoutMs = 1000) => {
   const errorEvents = [
     'error',
     'connect_error',
@@ -126,7 +126,7 @@ export const waitForSocketEvent = async (socket: any, event:string) => {
       const timeout = setTimeout(() => {
         reject(new Error(`timed out waiting for ${event} event`));
         cancelTimeout = () => {};
-      }, 5000);
+      }, timeoutMs);
       cancelTimeout = () => {
         clearTimeout(timeout);
         resolve();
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Pushed f33999117 addressing both Qodo points:

**1. Hardcoded 5s socket wait** — `waitForSocketEvent(socket, event, timeoutMs?)` now accepts an optional timeout, default `1000ms` (matches pre-PR behaviour). Only the known-slow `connect()` and `handshake()` call sites pass `5000ms` — they're the ones that were timing out on loaded CI. Per-message waits in `messages.ts`, `clientvar_rev_consistency`, etc. keep the 1s default so failures surface fast with the descriptive helper error rather than getting masked by Mocha's per-test timeout.

**2. SessionStore waits still tight** — for the three "record should be gone after expiry" assertions (`set of session that expires`, `switch from non-expiring to expiring`, `get of record from previous run (not yet expired)`) replaced the fixed sleep-then-assert with a small `eventually()` poller (25ms / 2000ms max). These now pass immediately when cleanup completes on a fast runner and tolerate hundreds of ms of event-loop delay on a slow one. No fixed coupling between sleep duration and expiry duration.

For the inverse case `shutdown cancels timeouts` polling doesn't apply (we're verifying that a *cancelled* timer did **not** fire — requires waiting past the original expiry, can't poll for "still nothing happened"). Bumped expires 300→500ms and wait 330→700ms so setup (set+get+shutdown) has a 500ms cushion (vs 30ms previously) before the timer would have fired, and the 700ms wait is still past the original expiry.

All 28 SessionStore + 33 socketio tests pass locally.

PR #7688, opened 2026-05-07
fix(7686): username 'false' / 'malformed color: false' for legacy settings.json

### LINKED ISSUE #7686: Username is false and color is white after upgrading to 2.7.3
**Describe the bug**

When I open an existing or a new pad, my username is called "false", and the preselected color is white.

**To Reproduce**
Update to 2.7.3
Open a new pad, or an existing pad.

On the server, there is this output:

[2026-05-07T06:30:18.562] [ERROR] socket.io - Error handling pad message from u-k8z9M_Q-1oQpRfAAAB: Error: COLLABROOM: USERINFO_UPDATE: malformed color: false
    at Object.exports.handleMessage (/home/.../etherpad-lite/src/node/handler/PadMessageHandler.ts:545:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async <anonymous> (/home/.../etherpad-lite/src/node/handler/SocketIORouter.ts:85:14)

**Expected behavior**
The color and the name should not be "false".

 - Etherpad version: 2.7.3
 - OS: Debian Bookworm
 - Node.js version (`node --version`): v24.15.0
 - npm version (`npm --version`): 11.12.1

 - OS: Debian Trixie
 - Browser: Firefox

**Additional context**
I just upgraded to 2.7.3

JohnMcLear:
## Summary

Closes #7686.

Settings.json files generated before December 2021 still ship `"userName": false` / `"userColor": false` for `padOptions`. These two options are documented as strings, so the boolean was always meaningless — commit [8c857a85a](https://github.com/ether/etherpad-lite/commit/8c857a85ac9af08ebec1c94031ca7c54bed34c38) switched the template default to `null` and explicitly warned: *"this change has no effect due to a bug in how pad options are processed; that bug will be fixed in a future commit."* The follow-up never landed.

The cascade:

1. `clientVars.padOptions.userName === false` (legacy) reaches `pad.ts:getParams()`.
2. `if (serverValue == null) continue;` — `false != null`, doesn't skip.
3. `serverValue = serverValue.toString();` → `"false"`.
4. `settings.globalUserName = "false"`, `clientVars.userColor = "false"`.
5. `pad.myUserInfo.colorId = clientVars.userColor` → `"false"`.
6. `if (settings.globalUserName !== false)` → `"false" !== false` is `true` → `notifyChangeName("false")` ships USERINFO_UPDATE with `name="false"`, `colorId="false"`.
7. Server's hex regex throws `Error: COLLABROOM: USERINFO_UPDATE: malformed color: false`.

The user sees their name as `false` and a white swatch.

## Fix (defense in depth)

- **Server boundary** — `Settings.ts::reloadSettings()` coerces `padOptions.userName === false` and `padOptions.userColor === false` to `null` and warns once, matching the existing `disableIPlogging` shim pattern.
- **Client guard** — `pad.ts` `userName`/`userColor` callbacks reject empty / literal `"false"` so URL params like `?userName=false` are also a no-op.
- **Regression test** — `tests/backend/specs/padOptionsLegacyDefaults.ts` mirrors the shim and asserts it normalizes the legacy boolean, leaves explicit string/null values intact, doesn't touch `showChat: false`/`rtl: false`/etc., and does not coerce the string `"false"` (that's the client guard's job).

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm exec mocha tests/backend/specs/padOptionsLegacyDefaults.ts` — 8 passing
- [x] `pnpm exec mocha tests/backend/specs/settings.ts tests/backend/specs/ipLoggingSetting.ts tests/backend/specs/padOptionsLegacyDefaults.ts` — 48 passing
- [ ] CI runs the full backend + Playwright suite
- [ ] Reporter (@tpokorra) confirms the warning appears and the bug is gone after upgrading

### DIFF
--- a/src/node/utils/Settings.ts
+++ b/src/node/utils/Settings.ts
@@ -1094,14 +1094,20 @@ export const reloadSettings = () => {
     // applying it, which then propagates as the user's name and color and
     // triggers `malformed color: false` on the server (#7686). Normalize
     // legacy booleans to null at the boundary so downstream code sees the
-    // expected sentinel.
-    for (const key of ['userName', 'userColor'] as const) {
-      if ((settings.padOptions as any)[key] === false) {
-        logger.warn(
-            `padOptions.${key}=false is a legacy default (pre-2021) and is ` +
-            `now treated as null. Update settings.json to use null instead ` +
-            `to silence this warning.`);
-        (settings.padOptions as any)[key] = null;
+    // expected sentinel. Guard against a malformed padOptions (null, array,
+    // primitive) — storeSettings() will overwrite it raw if settings.json
+    // declares it as anything other than a plain object.
+    if (settings.padOptions != null
+        && typeof settings.padOptions === 'object'
+        && !Array.isArray(settings.padOptions)) {
+      for (const key of ['userName', 'userColor'] as const) {
+        if ((settings.padOptions as any)[key] === false) {
+          logger.warn(
+              `padOptions.${key}=false is a legacy default (pre-2021) and is ` +
+              `now treated as null. Update settings.json to use null instead ` +
+              `to silence this warning.`);
+          (settings.padOptions as any)[key] = null;
+        }
       }
     }
 
--- a/src/node/utils/Settings.ts
+++ b/src/node/utils/Settings.ts
@@ -1087,6 +1087,24 @@ export const reloadSettings = () => {
       settings.privacyBanner.dismissal = 'dismissible';
     }
 
+    // Settings.json files generated before December 2021 used `false` as the
+    // default for these string options. The client treats the boolean `false`
+    // as a sentinel meaning "no enforced value", but the dispatch in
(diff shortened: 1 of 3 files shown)


JohnMcLear:
@qodo-free-for-open-source-projects re: padOptions shim crash —

Verified. `storeSettings()` overwrites `settings.padOptions` raw if `settings.json` declares it as anything other than a plain object. (Existing call sites in `Pad.ts` already assume `padOptions` is an object, so a primitive `padOptions` would already kill Etherpad on the first pad open — the shim isn't a *new* failure mode, just an earlier one.)

Fixed in 2f875af9f: added a shape guard (`!= null && typeof === 'object' && !Array.isArray`) around the shim, and extended `padOptionsLegacyDefaults.ts` with three cases asserting the guard handles null / primitives / arrays without throwing. 11 specs passing.

PR #7690, opened 2026-05-07
fix(7606): sync theme-color meta with client-side dark-mode switch

JohnMcLear:
## Summary

- After #7636 shipped, [stffen reported on #7606](https://github.com/ether/etherpad/issues/7606#issuecomment-2883330517) that 2.7.3 still paints a white address bar above a dark pad.
- Root cause: the meta is computed server-side from `settings.skinVariants`, but the toolbar can be flipped to `super-dark-toolbar` *client-side* in three places — `pad.ts:712` auto-switch (`enableDarkMode` + `prefers-color-scheme: dark` + no localStorage white-mode override), the `#options-darkmode` toggle in `pad_editor.ts`, and the `#skinvariantsbuilder` admin tool. All three call `skinVariants.updateSkinVariantsClasses()`, which until now never touched the meta.
- Fix: push the lookup into `updateSkinVariantsClasses` so the meta tracks every class change. Mirrors the CSS-source-order table from `src/node/utils/SkinColors.ts` so the same "last matching `*-toolbar` token wins" rule the server uses also applies on the client. No-op when the server didn't emit a meta (non-colibris skins).

## Test plan

- [x] New Playwright spec `tests/frontend-new/specs/theme_color_dark_mode.spec.ts` covers both paths:
  - `colorScheme: 'light'` — manual toggle via `#options-darkmode` flips meta `#ffffff` → `#485365` → `#ffffff`.
  - `colorScheme: 'dark'` — auto-switch on dark-OS clients lands on `#485365` after init (the case stffen reported).
- [x] TDD verified: reverted the fix, both tests went red; restored, both green.

### DIFF
--- a/src/node/utils/SkinColors.ts
+++ b/src/node/utils/SkinColors.ts
@@ -1,34 +1,16 @@
 'use strict';
 
-// Toolbar background colors that the colibris skin variants resolve to.
-// Mirrors --bg-color in src/static/skins/colibris/src/pad-variants.css. Only
-// the colibris skin has a known mapping; for any other skin we cannot derive
-// the toolbar color server-side and emit no theme-color meta.
-//
-// Order matters: when skinVariants contains multiple *-toolbar tokens the
-// CSS cascade picks the rule defined last in pad-variants.css, so iterate in
-// source order and let the last matching token win.
-const TOOLBAR_COLORS_IN_CSS_ORDER: Array<[string, string]> = [
-  ['super-light-toolbar', '#ffffff'],
-  ['light-toolbar', '#f2f3f4'],
-  ['super-dark-toolbar', '#485365'],
-  ['dark-toolbar', '#576273'],
-];
-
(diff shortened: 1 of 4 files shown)


JohnMcLear:
Actioned Qodo's two findings in `1b724c5`:

1. **Duplicated table (bug, drift hazard)** — extracted the CSS-source-order toolbar→color map and the default-color constant into `src/static/js/skin_toolbar_colors.ts`. Both `src/node/utils/SkinColors.ts` (server/EJS helper) and `src/static/js/skin_variants.ts` (client runtime) now import from it. Server-side imports of `static/js/` modules are already established (`Changeset`, `AttributeMap`, `ImportHtml`, `hooks`), so no new boundary. Existing `configuredToolbarColor` unit tests + the `specialpages` server-render checks all still pass against the refactored helper, locking the abstraction.
2. **2-space indent (rule violation)** — re-indented the `themeColor` helper and the multiline `test(...)` call in the Playwright spec to match `.editorconfig`'s `indent_size = 2`.

Issue #7693, opened 2026-05-07
Document admin endpoints in the OpenAPI spec


JohnMcLear:
## Background

PR #7638 (admin typesafe API client + TanStack Query rails — implementation in JohnMcLear/etherpad#2) lays down a codegen toolchain that produces a typed client from `src/node/hooks/express/openapi.ts`. The generated `admin/src/api/schema.d.ts` currently covers only the public `/api/{version}/*` surface (48 paths).

The admin UI talks to `/admin-auth/*` and the upcoming `/admin/*` routes (#7601). None of those are documented in the OpenAPI hook today, so the typed client has no admin call sites to migrate yet.

## Proposal

Add OpenAPI definitions for the admin endpoints to `src/node/hooks/express/openapi.ts` (or a sibling `openapi-admin.ts`) so:

1. `admin/src/api/schema.d.ts` exposes typed paths the admin UI actually calls.
2. Existing admin `fetch()` sites can be migrated to `$api` / `fetchClient` with end-to-end type safety.
3. `#7601` admin REST API work can be developed against the same generated types.

## Scope

- Document `/admin-auth/*` (login flow currently consumed by the admin UI).
- Document `/admin/*` once #7601 lands — coordinate so that the new endpoints ship with their OpenAPI definitions in the same PR.
- No call-site migrations in this issue; that's a separate follow-up once the schema includes admin paths.

## Why split from #7638

The codegen rails are decoupled from the spec-coverage work so reviewers can land them independently. The rails are a no-op until this issue is resolved, which is documented in the PR description.

PR #7695, opened 2026-05-07
chore(admin): typesafe API client + TanStack Query rails (#7638)

### LINKED ISSUE #7638: chore(admin): generate typesafe API client + adopt TanStack Query
Spun out of #7601 review feedback from @SamTV12345.

## Problem

The admin UI currently hand-writes `fetch` calls in `useEffect` throughout (`UpdateBanner`, `UpdatePage`, plus the existing settings/plugins/users pages). Two issues with this:

1. **Pattern is officially discouraged** — see the React docs: https://react.dev/reference/react/useEffect#fetching-data-with-effects (no caching, no dedup, no race-condition handling, manual loading states).
2. **Endpoint shape is duplicated** — every fetch site re-types the request and response by hand. We already publish a Swagger/OpenAPI spec that describes these endpoints authoritatively, so any drift between the spec and the hand-written types is a silent bug.

## Proposal

Generate the admin's API client directly from the OpenAPI spec, and use TanStack Query as the runtime.

**Toolchain:**
- [`openapi-typescript`](https://github.com/openapi-ts/openapi-typescript) — generates a TS types file from `src/node/hooks/express/swagger.ts` output. Run as a build step; the result is checked in or generated pre-build.
- [`openapi-fetch`](https://github.com/openapi-ts/openapi-typescript/tree/main/packages/openapi-fetch) — typed `fetch` wrapper that consumes the generated types. Zero runtime overhead, ~3 KB gzipped.
- [`openapi-react-query`](https://github.com/openapi-ts/openapi-typescript/tree/main/packages/openapi-react-query) — thin TanStack Query bindings that derive query/mutation hooks from the same generated types. Caching, dedup, refetch-on-focus, devtools all come along for free.
- [`@tanstack/react-query`](https://tanstack.com/query) — the runtime.

Net result: every endpoint becomes one line — `useQuery({ ...client.GET("/admin/...") })` — with full request/response inference and no hand-written types.

## Scope

- Wire the OpenAPI generator into the admin build pipeline (or the root `pnpm build`).
- Add `@tanstack/react-query` + provider at the admin root, plus React Query Devtools in dev.
- Migrate every `useEffect`+`fetch` site in `admin/src/` to the generated hooks.
- Replace `if (!data) return null` patterns with Suspense boundaries (`useSuspenseQuery`).
- Document the codegen step in `admin/README` so contributors know how to regen after API changes.

## Out of scope

- Migrating the pad-side frontend (different bundle, different concerns).
- Backend changes — the OpenAPI spec is the contract; if anything moves, that moves first.

## Why now

Tracking out of #7601 specifically — that PR adds two new admin fetch sites, and bundling this migration into it would balloon the diff. Cleaner as a dedicated PR.

JohnMcLear:
## Summary

Lays down the rails for a typesafe, OpenAPI-derived admin API client backed by TanStack Query. Closes #7638.

- Codegen toolchain (`pnpm --filter admin gen:api`) producing `admin/src/api/schema.d.ts` and `admin/src/api/version.ts` from `src/node/hooks/express/openapi.ts`.
- Runtime client (`openapi-fetch` + `openapi-react-query`) at `admin/src/api/client.ts` with `baseUrl = \`/api/\${LATEST_API_VERSION}\`` so calls land at `/api/<version>/createGroup`.
- `<QueryProvider>` mounted at the admin root with dev-only React Query devtools (verified absent from the production bundle).
- CI freshness check on the generated files folded into `frontend-admin-tests.yml`.
- `admin/README.md` documenting the workflow.

**No call sites migrated.** Admin endpoints aren't in the OpenAPI spec yet — that gap is filed as #7693 and must land before any migration is useful.

## Why one squashed commit

This was developed iteratively on the fork (JohnMcLear/etherpad#2 — merged into the fork's `develop`) with 16 small fixup commits along the way (Windows shell, Dockerfile workspace yaml, lockfile overrides, baseUrl emit, etc.). Squashed for upstream so review is one cohesive change rather than the back-and-forth.

## Semver

Patch — build tooling + currently-unused runtime libs, no observable behavior change.

## Test plan

- [x] `pnpm --filter admin gen:api` runs clean (48 paths, OpenAPI 3.0.2, version 1.3.1)
- [x] `pnpm --filter admin run build` succeeds; production bundle excludes `ReactQueryDevtools`
- [x] `pnpm --filter admin test` passes (smoke test: 1 pass)
- [x] `pnpm ts-check` clean for admin/openapi files
- [x] Dev server boots and lazy-loads `@tanstack/react-query-devtools` (verified `import.meta.env.DEV === true` and the lazy import is rendered)


JohnMcLear:
Qodo's earlier rule violation on the JohnMcLear/etherpad fork PR ([rule 467291: "Exclude build artifacts and runtime-generated files from version control"](https://app.qodo.ai/rules/467291)) is now addressed in 40fe8a4c2:

- `admin/src/api/schema.d.ts` and `admin/src/api/version.ts` are gitignored and removed from VCS.
- Every admin script that needs the generated files (`dev`, `build`, `build-copy`, `test`) now chains `gen:api` first, so a fresh checkout produces them automatically.
- Dropped the now-redundant freshness-diff step from `frontend-admin-tests.yml` — with no committed file, the build's `gen:api` invocation is the only check needed.

SamTV12345 (approved):
This looks really good.

PR #7698, opened 2026-05-07
feat(padOptions): pass plugin-namespaced ep_* keys through applyPadSettings

JohnMcLear:
## Summary
- Native pad-wide settings ride a single `padOptions` object: server seeds `clientVars.initialOptions`, client mutates via `pad.changePadOption()`, and the existing `padoptions` COLLABROOM message broadcasts changes. Plugins can't use the same rail today because `applyPadSettings` (client) and `normalizePadSettings` (server) silently drop any key not in their hardcoded whitelist.
- Add a passthrough loop on both sides that preserves keys matching `/^ep_[a-z0-9_]+$/`. Plugins can now stash pad-wide values under their own namespace (e.g. `pad.padOptions.ep_table_of_contents = {enabled: true}`) and inherit broadcast, persistence, creator-only-write enforcement, and `enforceSettings` semantics for free.
- New `src/node/utils/PluginCapabilities` module exposes `padOptionsPluginPassthrough = true` so plugins can feature-detect via `require()` and fall back to per-user-only behavior on older cores.

## Why
First consumer is a new `padToggle` helper in `ep_plugin_helpers` (separate PR) that gives plugins parallel User Settings + Pad Wide Settings checkboxes — matching every native toggle. Most plugins today put their setting only in User Settings (cookie-only, per-user) because the helper they use can't do better; the result is plugin features behaving differently from native ones (no pad-wide default, no enforce). This patch unblocks the helper.

## Risk
- ~10 lines on each side, surgical. Only keys matching the `ep_` prefix are touched; existing whitelisted keys unchanged.
- `Pad.normalizePadSettings` already round-trips through `saveToDatabase()`, so persistence is automatic for the new keys.
- Capability flag is additive — older plugins continue to work, new ones can opt in via the same well-known require path.

## Test plan
- [x] 3 new backend tests cover preserve, drop-non-matching-prefix, and core-key-collision-isolation (`src/tests/backend/specs/Pad.ts`).
- [x] Existing `normalizePadSettings` tests untouched and passing (6/6 in this section).
- [x] Live verified end-to-end with ep_plugin_helpers padToggle + migrated ep_table_of_contents — both User Settings and Pad Wide Settings checkboxes render and the id pattern matches native (`options-toc` and `padsettings-options-toc`).
- [ ] Two-browser broadcast + enforce verification in Playwright once helper PR lands.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -238,6 +238,77 @@ operations in `templates/`, in files of type ".ejs", since Etherpad uses EJS for
 HTML templating. See the following link for more information about EJS:
 <https://github.com/visionmedia/ejs>.
 
+## Plugin-namespaced pad-wide options
+
+Plugins can ride the existing `padoptions` COLLABROOM rail to store
+pad-wide settings — broadcast to every connected client, persisted with the
+pad, and honored by `enforceSettings` — instead of inventing their own
+message type and storage. The model matches how `enablePadWideSettings`
+works for native toggles like sticky chat or line numbers.
+
+### Capability detection
+
+```js
+let padOptionsPluginPassthrough = false;
+try {
+  // The require throws on Etherpad versions that predate this capability;
+  // plugins should degrade gracefully (typically falling back to a per-user
+  // cookie toggle) when the flag is missing.
+  padOptionsPluginPassthrough =
+      require('ep_etherpad-lite/node/utils/PluginCapabilities')
+          .padOptionsPluginPassthrough === true;
+} catch (_e) { /* older core */ }
+```
+
+The flag means the core has the passthrough patch *available*. Whether it
+is actually *enabled* at runtime is a separate per-instance setting — see
+below.
+
+### Runtime flag
+
+The passthrough is gated by `settings.enablePluginPadOptions`, default
+`false`. Operators must opt in via `settings.json`:
+
+```json
+{
+  "enablePluginPadOptions": true
+}
+```
+
+When enabled, the server reflects the value to every client via
+`clientVars.enablePluginPadOptions` so plugins can detect both *capable*
+(static) and *active* (per-pad request) at the same point.
+
+### Key namespace
+
+Plugins must use keys matching `/^ep_[a-z0-9_]+$/`. The recommended pattern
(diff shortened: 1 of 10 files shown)


JohnMcLear:
## Addressed Qodo review

### 1. ✅ Feature flag (Rule violation, Security)
Added `settings.enablePluginPadOptions` (default **false** per AGENTS.MD §52). Both `normalizePadSettings` (server) and `applyPadSettings` (client) now gate the passthrough on this flag. Mirrored to clients via `clientVars.enablePluginPadOptions` so plugins can detect runtime state at the same moment they receive the pad.

\`\`\`json
{
  "enablePluginPadOptions": true
}
\`\`\`

### 2. ✅ Documentation (Rule violation, Maintainability)
New "Plugin-namespaced pad-wide options" section in `doc/plugins.md` covering:
- Capability detection via `require('ep_etherpad-lite/node/utils/PluginCapabilities')`
- The runtime flag and how plugins should consume it from `clientVars`
- The `/^ep_[a-z0-9_]+$/` key namespace
- Validation rules (JSON-safe, 64 KB per key, 256 KB total)

Also documented the new flag in `settings.json.template`.

### 3. ✅ Unbounded plugin option payload (Bug, Reliability)
`normalizePadSettings()` now validates every `ep_*` value before accepting it:
- Must round-trip through `JSON.stringify` — rejects functions, symbols, BigInt, circular references.
- Per-key serialized size capped at **64 KB**.
- Combined `ep_*` size capped at **256 KB** per pad.

Rejected values are dropped with a single `console.warn` line; the rest of the settings round-trip cleanly. New tests cover all three rejection paths plus a happy-path "all caps + flag respected" scenario.

### 4. ✅ PadOption type misses ep_* (Bug, Maintainability)
Added \`[k: \`ep_\${string}\`]: unknown\` index signature to `PadOption` in `src/static/js/types/SocketIOMessage.ts`. TS callers can now read plugin-namespaced keys without casts; the type matches runtime behavior.

### Test coverage
Backend test suite extended in `src/tests/backend/specs/Pad.ts`:
- `with enablePluginPadOptions = true` (5 cases including 3 validation/cap tests)
- `with enablePluginPadOptions = false` (1 case verifying gate)

cc @qodo-free-for-open-source-projects

PR #7703, opened 2026-05-07
fix(7696): scrollable settings popup on short viewports

### LINKED ISSUE #7696: Pad-wide Settings get cropped if UI too small
If UI shrinks to 1/3 screen or so then Pad-wide settings isn't tall enough to fit all the items in My View so starts cropping them off without a scroll bar to access them.

JohnMcLear:
## Summary
- Settings popup (especially with Pad-wide Settings enabled) currently crops items off-screen with no scrollbar when the viewport is short — the bug reported in #7696.
- Move `max-height` + `overflow-y:auto` from the mobile-only (`max-width:800px`) media query onto the base `.popup-content` rule so all popups stay scrollable at any viewport width. The chat panel (`#users`) keeps `overflow:visible` so the colour-picker can float outside it.
- Adds a Playwright regression test that opens Settings at 900×500, asserts the popup is a scroll container, and confirms the bottom-most "Delete pad" button is reachable via scroll.

Fixes #7696

## Test plan
- [x] `pnpm run ts-check` — clean for changed files (pre-existing errors are all under `plugin_packages/`)
- [ ] CI runs the new Playwright test `settings popup stays scrollable when the viewport is short`
- [x] Manual: shrink browser window to ~1/3 screen, open Settings → scrollbar appears, all items reachable, "Delete pad" still clickable

### DIFF
--- a/src/static/css/pad/form.css
+++ b/src/static/css/pad/form.css
@@ -141,6 +141,14 @@ select, .nice-select {
   top: auto;
   left: auto;
 }
+/* The default .reverse rule above sets bottom: calc(100% + 5px) so an
+   absolutely-positioned list opens upward inside its parent. Once the list is
+   position:fixed, that percentage resolves against the viewport instead and
+   would push the list off-screen, so we let JS place it via `top` only. */
+.toolbar .nice-select.reverse .list,
+.popup .nice-select.reverse .list {
+  bottom: auto;
+}
 .nice-select .list:hover .option:not(:hover) {
   background-color: transparent !important;
 }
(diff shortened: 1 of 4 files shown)


JohnMcLear:
Thanks @qodo-free-for-open-source-projects — both addressed in 26cebda.

1. **Dropdowns clipped in popups** — Mirrored the toolbar workaround for popup-contained nice-selects: the list switches to `position: fixed` (`form.css`) and `nice-select.ts` sets viewport-relative `top`/`left`/`min-width` from `getBoundingClientRect()`, picking the anchor side based on the existing `.reverse` class so upward-opening lists still work. Toolbar path is untouched.

2. **Brittle scrollHeight assertion** — Dropped `scrollHeight > clientHeight`. The test now asserts the popup declares `overflow-y: auto` and proves the bug is fixed by checking `#delete-pad` is initially **not** in viewport, then reachable after `scrollIntoViewIfNeeded()`.

Issue #7607, opened 2026-04-26
Clean update path


JohnMcLear:
As per docs in https://github.com/ether/etherpad/pull/7601/changes

We can provide a much cleaner path to upgrade for admins.

JohnMcLear:
I think with tier 3 landing we're pretty close now, I'll try wrap this up by the weekend 🏩

PR #7704, opened 2026-05-08
feat(updater): tier 2 — manual-click update from /admin/update (#7607)

JohnMcLear:
## Summary

Ships **Tier 2 (manual click)** of the four-tier auto-update design at `docs/superpowers/specs/2026-04-25-auto-update-design.md`. Builds on PR #7601 (Tier 1 — notify, merged 2026-05-01). Closes part of #7607.

- Admins on a git install can click **Apply update** at `/admin/update`. Etherpad runs a 60s session drain (with T-60 / T-30 / T-10 broadcasts to every connected pad), `git fetch / checkout / pnpm install --frozen-lockfile / pnpm run build:ui`, and exits with code 75 so a process supervisor restarts it on the new version.
- The next boot runs a 60s health-check; if `/health` doesn't come up the previous SHA + lockfile are restored automatically. **Crash-loop guard:** if the new version reboots more than twice (`bootCount > 2`) RollbackHandler forces a rollback regardless of the timer.
- Terminal `rollback-failed` state surfaces a strong red banner; admins click **Acknowledge** after manual recovery to clear the lock and re-allow Tier 2 attempts. `canAuto`/`canAutonomous` are denied while terminal; manual stays on because clicking Apply *is* the intervention.
- New settings under `updates.*`: `preApplyGraceMinutes`, `drainSeconds`, `rollbackHealthCheckSeconds`, `diskSpaceMinMB`, `requireSignature`, `trustedKeysPath` (all opt-in / sane defaults).
- Signature verification (`updates.requireSignature`) is opt-in and stub-friendly: `false` → log warning and pass; `true` → `git verify-tag <tag>` against the user keyring (or `trustedKeysPath` via `$GNUPGHOME`). Etherpad's release process does not yet sign tags consistently — turning on by default would block every Tier 2 update, so this is documented as follow-up.

A process supervisor (systemd / pm2 / docker `--restart=unless-stopped`) is **required** to apply updates. Without one, exit 75 leaves the instance down. The supervisor requirement and a sample systemd unit live in `doc/admin/updates.md`.

Tier 3 (auto with grace window) and Tier 4 (autonomous within maintenance window) are out of scope for this PR.

## Architecture

New atomic units under `src/node/updater/`:
- `lock.ts` — PID-based `var/update.lock` with stale-PID reaping.
- `trustedKeys.ts` — `verifyReleaseTag` (GPG via `git verify-tag`).
- `preflight.ts` — sequenced check pipeline with typed reasons.
- `UpdateExecutor.ts` — DI spawn pipeline (snapshot → fetch → checkout → install → build → exit 75).
- `RollbackHandler.ts` — boot-time health-check timer + crash-loop guard + restore-from-backup.
- `SessionDrainer.ts` — timed broadcasts + accept-flag.
- `updateLog.ts` — rolling `var/log/update.log` (10 MB × 5) + `tailLines(n)`.

New routes in `src/node/hooks/express/updateActions.ts`: `POST /admin/update/{apply,cancel,acknowledge}`, `GET /admin/update/log` — strict admin auth on all four.

`RollbackHandler.checkPendingVerification` wires into boot in `src/node/updater/index.ts`; `markBootHealthy` is called from `src/node/server.ts` once state hits `RUNNING` so `/health`-being-up is the implicit happy-path signal that cancels the rollback timer.

Admin UI (`admin/src/pages/UpdatePage.tsx`): renders Apply / Cancel / Acknowledge per `execution.status`, polls `/admin/update/log` while in flight, surfaces lastResult and policy denial copy. `UpdateBanner.tsx` adds a terminal-state alert variant.

Pad UI: existing shoutMessage pipeline (`src/static/js/pad.ts`) learns `messageKey + values` and routes through `html10n.get(key, values)` so `update.drain.t60` / `t30` / `t10` interpolate the seconds remaining and render the localised string. (Avoids the unbound `window._` bug documented in memory.)

## Test plan

- [x] `pnpm exec vitest run tests/backend-new/specs/updater` — **128 unit tests** across 13 files (lock, trustedKeys, preflight, updateLog, SessionDrainer, UpdateExecutor, RollbackHandler, drainer-handshake, UpdatePolicy, index-boot wiring, state validator backfill).
- [x] `pnpm run test --grep updater-integration` — **5 mocha integration tests** against a tmp git repo: happy path lands on pending-verification, install-fail rollback, build-fail rollback, crash-loop forced rollback, target-sha-doesn't-exist terminal `rollback-failed`.
- [x] `pnpm run test --grep updateActions` — **11 mocha API tests** for the four new endpoints (auth, policy, terminal-state acknowledge, content-type).
- [x] `pnpm run test --grep updateStatus` — **7 existing PR 1 API tests** still pass with the extended status payload.
- [x] Playwright admin spec at `src/tests/frontend-new/admin-spec/update-page-actions.spec.ts` — Apply button posts and re-fetches, install-method-not-writable hides Apply + shows denial copy, rollback-failed shows Acknowledge, lockHeld hides Apply.
- [x] `pnpm run ts-check` clean.
- [x] `pnpm --filter admin run build` clean.
- [ ] **Manual smoke runbook** at `docs/superpowers/specs/2026-04-25-auto-update-runbook.md` — pre-merge gate per design spec § "Phased rollout / PR 2". Cover happy path, both rollback variants, crash-loop guard, rollback-failed terminal, cancel-during-drain, drain announcement renders the localised string in a real pad. **Reviewer must run this on a disposable VM before merge.**

## Notes

- All 17 commits on this branch are independently revertable. The plan that drove them is at `docs/superpowers/plans/2026-05-08-auto-update-pr2-manual-click.md`.
- Tag signature verification is intentionally opt-in pending a separate "sign all releases going forward" project.
- Tier 1 state files written by an existing 2.7.3 install load cleanly — the state validator backfills the new `execution`/`bootCount`/`lastResult` fields, regression-tested.


JohnMcLear:
## Qodo review — resolution status

All 8 concerns from the original Qodo review have been addressed across two commits (`41d4a422a`, `a32256b63`). Qodo itself can't auto-re-review (free-tier monthly limit hit), so this comment is the audit trail.

| # | Concern | Resolution | Commit |
|---|---|---|---|
| 1 | `/admin/update/apply` exposed without flag (registered at `tier !== off`) | Endpoints now gate on `tier in {manual, auto, autonomous}`. Gate is per-request middleware so settings-reload takes effect without a restart, and integration tests can flip the tier dynamically. New regression test exercises 404 at `tier=notify` for all four endpoints. | `41d4a422a` |
| 2 | Cancel races preflight (cancel released the lock + set idle while apply was mid-preflight) | Cancel no longer touches the lock — apply's `finally` block owns it for the request's lifetime. Apply reloads state after preflight and aborts with `409 cancelled-during-preflight` if execution flipped under it. Prevents a second apply slipping in during a slow preflight. | `41d4a422a` |
| 3 | Drain wedges joins (acceptingConnections never restored if executor throws post-drain) | `acceptingConnections` is now restored at drain completion, not just on cancel. The lock + persisted `execution.status` are sufficient to keep a fresh apply from racing in. Unit test asserts the flag is restored after natural drain end. | `41d4a422a` |
| 4 | `socket.json.send` API misuse | Switched to `socket.emit('message', {disconnect: 'updateInProgress'})` for consistency with the other disconnect paths in `PadMessageHandler.ts` (lines ~221, ~569). | `41d4a422a` |
| 5 | Unhandled spawn `'error'` events (promise hangs forever on missing binary) | `'error'` listeners added to `runStep` in `UpdateExecutor`, `RollbackHandler`, and the gpg path in `trustedKeys`. `SpawnFn` type extended to allow `on('error', ...)`. Spawn errors resolve with `code: 1` + the error message in stderr so existing failure-detection branches fire normally. | `41d4a422a` |
| 6 | Executor leaves busy state on throw | `executeUpdate` body wrapped in try/catch. Any exception from `readSha`, `saveState`, `copyFile`, or any step persists `rolling-back` and returns `failed-checkout` so the route's post-executor rollback path picks it up. State can no longer wedge at `executing`. The catch's inner `saveState` is itself try-wrapped. | `41d4a422a` |
| 7 | `/admin/update/status` leaks diagnostics to unauthenticated callers | When `requireAdminForStatus=false` (default), unauth callers now see a sanitised payload: `execution.status` enum + `lastResult.outcome` enum are kept (the banner / pad-side badge need them) but `reason`, `fromSha`, and `targetTag` are stripped. Authed admin sessions still get the full diagnostic detail. Two new mocha tests cover both paths. | `a32256b63` |
| 8 | Wrong short-drain countdown (T-30 fired with `seconds: 30` even when only 5s actually remained) | `SessionDrainer` no longer schedules T-30 when `drainSeconds <= 30` and skips T-10 when `<= 10`. Admins who configure a short window get fewer announcements but each carries an accurate countdown. The opening announcement also now carries the configured `drainSeconds` rather than a hardcoded 60. | `a32256b63` |

Plus the CI Playwright strict-mode selector failure (banner + `lastResult` both contained "Manual intervention required") was fixed in `41d4a422a` by scoping the assertion to `p.last-result-rollback-failed`.

131 vitest unit + 26 mocha integration tests passing locally; full CI matrix green on the latest commit before this push.

JohnMcLear:
## Qodo follow-up review — resolution status (3 new concerns)

Qodo's persistent review surfaced **three additional concerns** after my first fix push (`41d4a422a`). All three are now addressed in `9340d279f`.

| New # | Concern | Resolution | Commit |
|---|---|---|---|
| 9 | Git tag option injection (security) — tag from GitHub `tag_name` could start with `-` and be parsed as a git option, bypassing `verify-tag` or changing `checkout` semantics. | Three-layer defence: (a) new `refSafety` helper (`isValidTag` / `assertValidTag` / `refsTagsForm`) enforces a strict subset of git's check-ref-format — refuses leading `-` / `.`, whitespace, control chars, and `~ ^ : ? * [ \` plus `..` sequence; (b) `VersionChecker` validates `tag_name` before persisting to state so a malformed value never lands on disk; (c) `UpdateExecutor` calls `assertValidTag` and uses `refs/tags/<tag>` form for git checkout, `trustedKeys` validates and adds `--` end-of-options marker to `git verify-tag`, `updateActions` does an up-front `isValidTag` check on `state.latest.tag` for a clean 409 instead of a 500 if state is corrupt. | `9340d279f` |
| 10 | Unhandled rollback rejections — `checkPendingVerification` did `void deps.saveState(...)` / `void performRollback(...)` without `.catch()`, so an fs error during boot's rollback path bubbled out as an unhandled rejection. | Both callsites now go through `fireSaveState` / `fireRollback` helpers that catch and log. Rollback rejections fall through to a best-effort terminal-state write + `exit(75)` so the supervisor restart can re-try with `bootCount++`. | `9340d279f` |
| 11 | Execution state under-validated — `isValidExecution` only checked `status` was a known enum, so a hand-edited file with `{execution: {status: 'pending-verification'}}` (missing `fromSha` / `targetTag` / `deadlineAt`) would pass validation and reach RollbackHandler with `undefined` refs. | Validator now consults a per-status required-fields map mirroring the `ExecutionStatus` union in `types.ts`, rejects missing or empty-string fields. Same tightening applied to `lastResult.outcome` (must match the allowed enum, not just `typeof === 'string'`). Six new unit tests cover hand-edited corruption variants. | `9340d279f` |

### Running tally — all 11 Qodo concerns resolved

| # | Concern | Commit |
|---|---|---|
| 1 | `/admin/update/apply` exposed without flag | `41d4a422a` |
| 2 | Cancel races preflight | `41d4a422a` |
| 3 | Drain can wedge joins | `41d4a422a` |
| 4 | `socket.json.send` API misuse | `41d4a422a` |
| 5 | Unhandled spawn `'error'` events | `41d4a422a` |
| 6 | Executor can leave busy state on throw | `41d4a422a` |
| 7 | `/admin/update/status` leaks diagnostics | `a32256b63` |
| 8 | Wrong short-drain countdown | `a32256b63` |
| 9 | Git tag option injection | `9340d279f` |
| 10 | Unhandled rollback rejections | `9340d279f` |
| 11 | Execution state under-validated | `9340d279f` |

**Test totals after all fixes:** 145 vitest unit tests + 26 mocha integration tests passing locally; ts-check clean. CI rebuilding now.

Free-tier Qodo can't auto-re-review again this month, so this comment serves as the audit trail for human reviewers.

src/node/hooks/express/updateActions.ts
```diff
@@ -0,0 +1,360 @@
...
+let drainer: Drainer | null = null;
+
+const requireAdmin = (req: any, res: any): boolean => {
+  const u = req.session?.user;
+  if (!u) { res.status(401).send('Authentication required'); return false; }
+  if (!u.is_admin) { res.status(403).send('Forbidden'); return false; }
+  return true;
+};
+
+const wrapAsync =
```
SamTV12345, line 38: That's a lot of any.  There are still these fetch calls everywhere in place. Is this setup for the next tier upgrade?

SamTV12345 (approved):
Besides the one comment. Nice addition.

src/node/hooks/express/updateActions.ts
JohnMcLear: Yea, its' all prep code towards auto update...  ~Trying to do things bit by bit..

PR #7708, opened 2026-05-09
feat(7642): bin/compactStalePads — staleness-gated bulk compaction

### LINKED ISSUE #7642: Staleness gating for compactPad: target pads not edited in N days
## Background

#7567 lands per-pad and per-instance compaction (`compactPad` API + `bin/compactPad` + `bin/compactAllPads`). Both knobs operate on *every targeted pad* — there's no way to say \"only compact pads that haven't been edited in a while.\"

For long-lived instances with a mix of hot and cold pads, that's the next missing piece. Compacting a recently-active pad is mostly wasted work — the history is what users are still navigating in timeslider. Compacting pads that haven't been touched in months is where the real reclaim is.

## Proposal

Stay out of the API surface — staleness is a *targeting* concern, not a *compaction* concern. Two complementary surfaces:

### 1. `bin/compactStalePads.ts` (operator-on-demand)

```
node bin/compactStalePads.js --older-than 90    # collapse all history on pads
                                                # not edited in the last 90 days
node bin/compactStalePads.js --older-than 90 --keep 50    # keep last 50 revisions
node bin/compactStalePads.js --older-than 90 --dry-run    # list, don't write
```

Composition: `listAllPads` → for each, `getLastEdited` → filter by age → `compactPad`. Mirrors `bin/compactAllPads.ts` and reuses the same progress / error-tolerance shape.

### 2. `cleanup.compactOlderThanDays` setting (turn-it-on-and-forget)

Optional integer in `settings.json`. When set, runs the same loop on a daily timer (lines up with the existing `cleanup.enabled` story). Off by default.

## Out of scope for this issue

- Adding staleness as a `compactPad` API param. Targeting *which* pads is a different concern from *how* to compact one — keeping them separate keeps both simple.
- Non-time staleness signals (revision count thresholds, byte-size thresholds). Easy follow-ons once the day-based filter is in.

## Acceptance

- [ ] `bin/compactStalePads.ts` lands with `--older-than`, `--keep`, `--dry-run`
- [ ] Decision (in this issue) on whether the daily-cron variant ships at the same time or as a separate ticket
- [ ] Docs updated (no new HTTP API expected, so just admin/CLI docs)

JohnMcLear:
Closes #7642.

## Summary
- Adds `bin/compactStalePads.ts` with `--older-than <days>`, `--keep N`, and `--dry-run`. Composes `listAllPads` → `getLastEdited` → `compactPad`, mirroring the per-pad error tolerance / dry-run / tally shape of `bin/compactAllPads`.
- Hot pads in active timeslider use are skipped; only pads not edited within the cutoff are compacted. Counts fresh-skipped vs. compacted vs. failed in the final report.
- A `getLastEdited` fault on one pad is counted as a failure but doesn't stop the run — same posture as a `getRevisionsCount` fault in `compactAllPads`.

## Design decisions
- **Staleness stays out of the API surface.** Per the issue, targeting *which* pads to compact is a CLI concern; the `compactPad` endpoint is unchanged. Keeps both surfaces simple and lets non-time staleness signals (rev count, byte size — issue calls these out as follow-ons) layer on later without further API churn.
- **Daily-cron variant deferred.** The issue asks for a decision on whether `cleanup.compactOlderThanDays` ships in the same PR. I've kept this PR scoped to the on-demand operator tool (the primary acceptance bullet) and split the cron variant out as a follow-up — easier to review, and the cron variant has its own settings/lifecycle questions (timer registration, log shape, interaction with `cleanup.enabled`) that deserve their own diff. Happy to fold it in here if reviewers prefer.

## Test plan
- [x] `pnpm ts-check` clean
- [x] `mocha tests/backend/specs/compactPad.ts` — 30 passing, including:
  - `parseArgs` accepts `--older-than` / `--keep` / `--dry-run`, rejects missing/negative/unknown
  - cutoff filters fresh pads from stale ones with injected `now`
  - `--keep N` is plumbed through to `compactPad`
  - `--dry-run` does not call `compactPad`
  - one stale pad failing to compact does not abort the run
  - a `getLastEdited` failure on one pad is counted as failure but doesn't stop iteration
  - `listAllPads` failure aborts cleanly with `failed=1`
  - empty instance / all-fresh instance / `--older-than 0` edge cases
  - end-to-end through real `/api/1.3.1/getLastEdited` + `compactPad`

## Semver
patch (CLI addition; no API surface change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
   - Tier 1 ships in this release. Tiers 2 (manual click), 3 (auto with grace window) and 4 (autonomous in maintenance window) are designed and will land in subsequent releases.
   - See `doc/admin/updates.md` for full configuration.
 - **Pad compaction.** New `compactPad` HTTP API plus `bin/compactPad` and `bin/compactAllPads` CLIs to reclaim database space on long-lived pads with heavy edit history (issue #6194). `--keep N` retains the last N revisions; `--dry-run` previews per-pad rev counts before writing. Per-pad failures don't stop the bulk run.
+  - `bin/compactStalePads` (issue #7642) targets only pads not edited in the last `--older-than N` days, so hot pads in active timeslider use are left alone. Same `--keep` / `--dry-run` shape as `bin/compactAllPads`. Targeting is deliberately a CLI concern — the `compactPad` API surface stays unchanged.
 - **New packaging targets.**
   - Etherpad is now published as a **Snap** package.
   - **Debian (.deb)** packages are built via nfpm with a systemd unit, and a signed apt repository is published to `etherpad.org/apt`.
--- a/bin/compactStalePads.ts
+++ b/bin/compactStalePads.ts
@@ -157,6 +157,27 @@ export const runCompactStale = async (
       continue;
     }
 
+    // Re-check staleness right before compacting. Without this the
+    // first-pass selection is a TOCTOU window: on a long bulk run a
+    // pad can become active between selection and compaction, and
+    // compactPad would then kick those sessions. Re-checking here
+    // shrinks the window to one round-trip and treats the pad as
+    // freshened (skipped, not failed).
+    let lastEditedNow: number;
+    try {
+      lastEditedNow = await api.getLastEdited(padId);
+    } catch (e: any) {
+      logger.error(`${idx} ${padId}: getLastEdited recheck failed: ${e.message ?? e}`);
+      report.failed++;
+      continue;
+    }
+    if (lastEditedNow > cutoff) {
+      logger.info(`${idx} ${padId}: edited during run — skipping (now fresh)`);
+      report.skippedFresh++;
+      report.stale--;
+      continue;
+    }
+
     try {
       await api.compactPad(padId, opts.keepRevisions);
     } catch (e: any) {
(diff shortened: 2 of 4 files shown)


JohnMcLear:
Addressed Qodo's findings in d19dc0ecd:

1. **No docs for compactStalePads** (requirement gap) — fixed. `doc/cli.md` now has a "Pad compaction" section covering all three CLIs (`compactPad`, `compactAllPads`, `compactStalePads`) with flags, examples, and the gating note. The doc gap actually pre-dates this PR — `compactPad`/`compactAllPads` from #6194 had no entry either — so this closes the whole hole.

2. **Staleness check race window** (bug) — fixed. Added a `getLastEdited` recheck immediately before every `compactPad` call. If the pad has been edited since first-pass selection, it's reclassified as `skippedFresh` (not `failed`) and we move on without kicking sessions. Tests cover both the recheck-skip path and a recheck-failure path.

3. **Duplicated CLI wiring logic** (advisory) — deferring to a follow-up. Extracting a shared `bin/_compactCliApi.ts` helper would also rewrite the already-merged `bin/compactPad.ts` and `bin/compactAllPads.ts` to consume it — a refactor that's worth doing, but is broader than this issue's scope and would muddy the diff for reviewers tracking the staleness feature. Happy to open a small follow-up PR right after this lands.

/review

SamTV12345 (approved):

Issue #15, opened 2011-05-23
Add an option to settings.json to specify an own css file


AlexanderWillner:
See title.

Issue #14, opened 2011-05-23
[WebSocket] SYNTAX_ERR: invalid url: wss://...


AlexanderWillner:
Might be related with #11 or an invalid server certificate. When opening a new/existing pad I receive the following error messages in the browser log and the GUI does not respond (Safari 5. and Chrome 13 are working fine):

Firefox 5:

<pre>
[WebSocket] SYNTAX_ERR: invalid url: wss://proxy:/socket.io/flashsocket
console.error(decodeURIComponent(message));
socket.io.js (line 2190)
Error calling method on NPObject!
[Break On This Error] self.__id, url, protocols,...ll, proxyPort || 0, headers || null);
socket.io.js (line 1930)
Error calling method on NPObject!
[Break On This Error] WebSocket.__flash.close(this.__id);
socket.io.js (line 1969)
</pre>

Opera 11.11:

<pre>
[23. Mai 2011 23:35:12] JavaScript - https://proxy/p/ET1tpyJRS4
Unknown thread
[WebSocket] SYNTAX_ERR: invalid url: wss://proxy:/socket.io/flashsocket
[23. Mai 2011 23:35:16] JavaScript - https://proxy/p/ET1tpyJRS4
Timeout thread: delay 5000 ms
Uncaught exception: Error: Error in Actionscript. Use a try/catch block to find error.
Error thrown at line 1969, column 4 in <anonymous function: WebSocket.prototype.close>() in https://proxy/socket.io/socket.io.js:
    WebSocket.__flash.close(this.__id);
called from line 794, column 21 in <anonymous function: WS.prototype.disconnect>() in https://proxy/socket.io/socket.io.js:
    this.socket.close();
called from line 1613, column 4 in <anonymous function: Socket.prototype.disconnect>(soft) in https://proxy/socket.io/socket.io.js:
    this.transport.disconnect();
called from line 1569, column 12 in <anonymous function>() in https://proxy/socket.io/socket.io.js:
    self.disconnect(true);
</pre>

Issue #13, opened 2011-05-23
Support subfolders


AlexanderWillner:
Similar to #11:

I've the following setup tested:
- Client <-> Server:9001

This is working quite well. For several reasons I've to use this setup:
- Client <-> proxy:443/etherpad <-> Server:9001

I can open the first page but there are several issues:
1. The "new pad" link on https://proxy/etherpad/ forwards to https://proxy/p/I8ZsAKR4Ug
2. When opening https://proxy/etherpad/p/I8ZsAKR4Ug I see parts of the GUI and in the browser console I see many 404 error codes.

Issue #12, opened 2011-05-23
Not able to find the subvalue colorId in null


AlexanderWillner:
Since I don't know how to reproduce it, feel free to close this issue again. The server stopped twice today with the following error:

<pre>
Client 14786210702732205 disconnected

/opt/node/lib/node_modules/ep-lite/node/MessageHandler.js:86
    if(err) throw err;
            ^
Not able to find the subvalue colorId in null
</pre>

PR #7709, opened 2026-05-09
admin: parsed JSONC settings editor (takes over #7666, closes #7603)

### LINKED ISSUE #7603: Parsing the settings.json in /admin could provide a better UX
Currently we just display a horricly messy blob (no idea why it lost line structure etc), we can relaly improve the admin ux if we just parse it and include inline comments for each thing from the comments within the json...

JohnMcLear:
Replaces the textarea on `/admin/settings` with a parsed, sectioned form: each top-level object/array is rendered as its own card with a documentation heading, and primitive keys collect into a 'General' card. Labels and help text come from settings.json comments first, falling back to `settings.json.template` documentation when the live file has none. Round-trip is byte-identical for untouched regions: edits go through `jsonc-parser`'s `modify()` so comments, key order, whitespace, and `${ENV:default}` placeholders all survive.

A "Raw" mode toggle keeps the existing textarea editor behind it for power users and structural edits.

Takes over #7666 (original author AWOL). Closes #7603.

## What it looks like
- **General** card: title, favicon, skinName, ip, port, etc. Two-column label + control rows with help text below.
- **Database settings**, **Pad options**, **SSO**, **Users**, etc.: one card per top-level group, with the comment's first sentence as the section heading and the rest as descriptive copy.
- Boolean values render as Etherpad-green Radix switches; numbers use a draft-buffer input that won't poison the file with NaN; strings get a normal text input; nulls show as a chip; `${VAR:default}` placeholders show as a read-only env pill (raw mode is the escape hatch for editing them).

## What's in the diff
- New dep: `jsonc-parser@^3.3.1` in `admin/`.
- New components under `admin/src/components/settings/`:
  - Pure helpers: `envPill.ts`, `comments.ts` (extract leading + trailing comments adjacent to a property node), `jsoncEdit.ts` (wraps `modify()` + `applyEdits`), `templateComments.ts` (build-time path → comment map from `settings.json.template`), `labels.ts` (first-sentence-as-label, humanized-key fallback).
  - Widgets: `StringInput`, `NumberInput`, `BooleanToggle`, `NullChip`, `EnvPill`.
  - Top-level: `JsoncNode` dispatcher, `FormView` (sectioned layout), `ParseErrorBanner`, `ModeToggle`.
- `SettingsPage` becomes a Form/Raw toggle shell. Save / Validate / Restart wiring unchanged.
- `IconButton` (earlier on this branch) forwards `ButtonHTMLAttributes` so `data-testid`/`aria-*` compile and render.
- Playwright helper switched from `.settingsButton.nth(1)` to `getByTestId('restart-etherpad-button')`.
- All new strings go through `react-i18next`; new keys in `src/locales/en.json`.
- `vite.config.ts` allows reading from the repo root so `settings.json.template` can be imported with `?raw`.

## Tests (`src/tests/frontend-new/admin-spec/adminsettings.spec.ts`)
All 10 specs pass locally (`cd src && npx playwright test admin-spec/adminsettings.spec.ts --reporter=line`):
- Existing: settings visible / save round-trip; comments preserved across save; validate toasts success+failure and blocks save on invalid JSON; restart works.
- New: form view splits comment into label + help text; template fallback supplies a label when the live file has no comment; editing `title` via the form input round-trips through save preserving structure; `requireAuthentication` boolean toggle round-trips; `${SSO_ISSUER:…}` renders as a read-only env pill (no `<input>` for that path); breaking raw JSON and toggling to form shows a parse-error banner with a working "Switch to raw" CTA.

## Spec / plan
- Design: `docs/superpowers/specs/2026-05-09-admin-settings-parsed-view-design.md`
- Plan: `docs/superpowers/plans/2026-05-09-admin-settings-parsed-view.md`

## Out of scope (follow-ups)
- Add/remove keys from the form (raw mode is the escape hatch).
- Inline editing of `\${VAR:default}` placeholders.
- Custom (per-section) Save buttons.
- The pre-existing `Are Settings visible…` Playwright test wrecks the dev settings.json on each run by writing then "restoring" from in-memory; worth a separate fix to snapshot from disk instead.

## Semver
patch — admin UI only, no API or settings-file format changes.


JohnMcLear:
@qodo addressing review on commits 5bcd3e3..bd84291. 12 of 14 issues fixed (#4 was already resolved by Qodo).

## Pushing back on
- **#1 "Form editor enabled by default"** and **#5 "Validate JSON lacks feature flag"**: declining. The whole point of #7603 is to *replace* the textarea with a parsed UI; gating the new editor behind a default-off flag would mean the issue isn't actually fixed for anyone who doesn't go find the flag, while still shipping the maintenance cost of two code paths. Compliance ID 8/10 doesn't apply when the new feature *is* the bug fix the issue is asking for. The Raw mode toggle (still one click away, with Tab/JSON validation/Restart all working there) is the preserved-prior-behaviour path. Validate is a small dry-run button next to Save, not a behaviour change worth a flag.

## Fixed
- **#2** Stable React keys derived from property name / byte offset (`5bcd3e377`)
- **#3** `onEdit` reads latest text from `useStore.getState()` instead of closing over stale render-time text (`8f94f7134`)
- **#6** Disambiguated `.settings` collision: App.css now scopes to `textarea.settings`, removed the duplicated block in `index.css` (`3971e686f`)
- **#8** While `settings === undefined` (socket hasn't delivered yet) FormView returns an `aria-busy` placeholder instead of flashing the parse-error banner (`e6e92ab72`)
- **#9** `NumberInput` syncs `draft` from the `value` prop when not focused so a server round-trip canonicalisation no longer diverges (`4bbac8b3e`)
- **#10** `aria-describedby` issue already gone after the form redesign — no element references a missing id; documented (`a57fd2e0c`)
- **#11** Added explanatory comment on `handleKeyDown` covering Tab/rAF rationale (`2a76a20be`)
- **#12** Server now emits `saveprogress: 'saved'` only on successful `writeFile`, `'error'` (with message) on failure; client toast moved into the `saveprogress` listener so the success toast only fires on confirmed write (`c1d360bfb`)
- **#13** `saveSettings` Playwright helper waits for `.ToastRootSuccess[data-state="open"]` and first detaches any previous open toast (`05ad93f98`)
- **#14** `Section` titles use `t('admin_settings.section.general')`, `ModeToggle` uses `t('admin_settings.mode.aria_label')`. Parser-error tokens (`Invalid symbol` etc.) deliberately left in English with a comment explaining why — they're jsonc-parser enum labels, not user prose (`4dd16c2d4`)
- **#15** `IconButton` defaults `type="button"`, callers can still override (`237200dd9`)
- **#7** Wiki links are now explicit `https://github.com/...` (`bd84291da`)

`cd admin && npx tsc --noEmit` clean. `cd src && npx playwright test admin-spec/adminsettings.spec.ts` 10/10 passing.

/review

SamTV12345:
image

So this is just for visualisation not for editing? I'm a bit puzzle how I would edit this.

JohnMcLear:
Looks like maybe your browser cache wasn't refreshed?  Do you see an input for site name?

JohnMcLear:
ahhh it because you are using docker, good point..  On it :)  Thanks for spotting, adding tests and rules too

JohnMcLear:
@SamTV12345 not a cache issue — diagnosed with your screenshot.

Your `settings.json` has nearly every value wrapped in `${VAR:default}` placeholders (likely `settings.json.docker`, which has 122 of them vs ~28 in `settings.json.template` I was testing against). The form view renders every `${VAR:default}` string as a read-only `EnvPill`, so all your visible rows ended up labels-only with no `<input>`. That was originally listed as out-of-scope ("inline editing of `${VAR:default}` placeholders") but per discussion this is a must-fix — the form is the dominant Docker deploy's only editor, and degrading UX silently based on deployment type isn't acceptable.

Pushed [845440e](https://github.com/JohnMcLear/etherpad/commit/845440efff113b1d1c6eb336f092e50d332e28df) on top of the merge from develop ([424a067](https://github.com/JohnMcLear/etherpad/commit/424a06732)):

- `EnvPill` is now an inline `<input>` editing the placeholder's default value. Edits emit `"${VAR:newDefault}"` via `jsoncEdit.modify()` so surrounding placeholder syntax, key order, and comments are preserved.
- Existing env-pill test flipped from a read-only assertion (which was silently encoding the bug) into an editing round-trip.
- New regression test seeds a docker-shaped `settings.json` where every value is `"${VAR:default}"` and asserts every visible env control is an editable `<input>` + editing one round-trips through save.

**Why Docker CI didn't catch this**: `.github/workflows/docker.yml` is container-up healthcheck only — it never runs admin-spec against a Docker config. `frontend-admin-tests.yml` runs admin-spec against `settings.json.template`, where `title`, `requireAuthentication`, `sso.issuer` still rendered editable widgets. The previous `env placeholder renders as read-only pill (no input)` test (line 235) actively asserted the bug as expected behaviour. New test closes that gap.

Could you retry on your Docker setup once CI is green?

SamTV12345 (approved):

SamTV12345:
Works also for me

JohnMcLear:
Yay I'll get this merged.

PR #7710, opened 2026-05-09
feat(pad): scrub history in-place on the pad URL (#7659)

### LINKED ISSUE #7659: Timeslider and pad is somewhat disconnected
I think we can probably just have the timeslider be in the pad page for example on the same url but clear you are travelling back in time..

JohnMcLear:
### BIG PR - Proceed with caution!

Closes #7659.

Click the timeslider toolbar button now keeps the user on `/p/:pad` and
toggles a hash-based history mode (`#rev/N`) instead of navigating to a
separate `/timeslider` page. The pad shell — chat, users panel,
settings, plugin chrome — stays mounted across the transition. A sticky
banner plus a sepia tint on the toolbar make it unmistakable that what
is visible is historical, not live.

## Summary
- New `PadModeController` (`src/static/js/pad_mode.ts`) owns
  enter/exit, the URL hash, browser back/forward, Esc to exit, and a
  mutation-observer bridge that copies the inner timeslider's revision
  label/date into the outer banner.
- `pad.html` grows a banner element and an iframe mount slot. The live
  ACE iframe stays mounted (just hidden), so the live socket survives
  the round-trip and the user snaps back to the current state without a
  reconnect on exit.
- `/p/:pad/timeslider` 302-redirects to the pad page for direct visits
  (legacy bookmarks; the browser preserves any `#NN` hash through the
  redirect, and a client-side shim translates it to `#rev/NN`). Iframe
  consumers pass `?embed=1` and still get the timeslider HTML; the
  embedded variant hides the redundant title and return-to-pad button
  via CSS, but the slider, settings, and export controls stay reachable.

## Visuals

| Live | History mode |
|---|---|
| Pad as today | Banner: *Viewing history • Version N • Saved …* + Return-to-live button. Outer toolbar dimmed (sepia + saturate); editor area hosts the embedded timeslider iframe. |

## Test plan
- [x] `pnpm run ts-check` clean for changed files
- [x] Backend: `tests/backend/specs/timesliderRedirect.ts` covers 302
      direct visits, pad-name preservation, and `?embed=1` HTML response
- [x] Playwright: new `padmode.spec.ts` (toolbar entry, return-to-live,
      browser back, legacy-URL redirect; asserts the rendered localized
      banner string, not just element presence)
- [x] Playwright: existing `timeslider*.spec.ts` updated to pass
      `?embed=1` for direct-iframe access; all 14 timeslider + padmode
      tests green
- [x] Adjacent UI specs (`editbar`, `embed_value`, `a11y_dialogs`)
      re-run with no regressions
- [x] Manual browser walkthrough on a live pad: enter via toolbar →
      banner appears, sepia tint applied, slider works inside iframe;
      Return-to-live cleanly restores live editing without reconnect.


JohnMcLear:
Pushed `f276bb8` addressing the Qodo review:

**Fixed**
- `?embed=` truthy bypass — now strict `req.query.embed !== '1'`.
- `#rev/latest` jumping to revision 0 — outer hash parser yields `-1`, but the iframe sync handler now resolves "latest" to `BroadcastSlider.getSliderLength()` instead of clamping to 0.
- Doc update — `doc/skins.md` + `.adoc` now describe the redirect and the `?embed=1` URL.
- Backed out a stray `data-theme="editorial"` and Google Fonts `<link>` tags that leaked into the branch from an unrelated working-tree change. Those were not part of this feature.
- (CI) Three pre-existing backend tests (`socialMeta`, `specialpages`) hit `/p/:pad/timeslider` directly; updated them to pass `?embed=1` like the others.

**Dismissed**
- Feature flag — Etherpad doesn't gate UX changes behind feature flags (and the existing toolbar button is already the only way users reach this code path; legacy URLs still redirect cleanly). Happy to add one if a maintainer wants it, but matching repo convention here.

SamTV12345 (approved):

PR #7726, opened 2026-05-11
ci(rate-limit): wait for etherpad readiness before running test

JohnMcLear:
## Summary
- Adds a readiness poll on \`http://127.0.0.1:8081/\` between \`run docker images\` and the test step in \`.github/workflows/rate-limit.yml\`.
- Eliminates the race where the etherpad container started with \`&\` isn't yet listening when the test fires, causing nginx to return 502 and the workflow to fail.

## Why
Developments three most recent runs of this workflow — all on README-only commits (#7723, #7724, #7725) — failed with the same 502 from nginx upstream. PR #7721 also tripped it three reruns in a row. With a warm pnpm-store, \`pnpm install --frozen-lockfile\` finishes faster than etherpad can boot inside the container, so the test starts before the upstream is responsive. Polling until nginx stops returning 5xx makes the run deterministic.

## Test plan
- [x] CI's \`rate limit\` job passes on this PR (the failure mode it fixes).
- [x] Polling step output shows "etherpad is ready behind nginx" within ~30s on a normal run.
- [x] If etherpad fails to come up, the new step prints \`docker logs etherpad-docker\` so future failures are easier to triage.

### DIFF
--- a/.github/workflows/rate-limit.yml
+++ b/.github/workflows/rate-limit.yml
@@ -58,7 +58,7 @@ jobs:
         name: run docker images
         run: |
           docker run --name etherpad-docker -p 9000:9001 --rm --network ep_net --ip 172.23.42.2 -e 'TRUST_PROXY=true' epl-debian-slim &
-          docker run -p 8081:80 --rm --network ep_net --ip 172.23.42.1 -d nginx-latest
+          docker run --name nginx-docker -p 8081:80 --rm --network ep_net --ip 172.23.42.1 -d nginx-latest
           docker run --rm --network ep_net --ip 172.23.42.3 --name anotherip -dt anotherip
       -
         name: install dependencies and create symlink for ep_etherpad-lite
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Addressed Qodo's review in c38e5ea:

- **Missing nginx logs on timeout** — nginx container is now named (\`--name nginx-docker\`) and its logs are dumped on readiness timeout, along with \`docker ps -a\` for container-state visibility.
- **Wait can reach 4m** — tightened to 30 iterations × (1s curl timeout + 1s sleep) ≈ 60s budget. Still well above observed cold-start time, but fails-fast when something is genuinely wrong.
- **No regression test** — pushing back on this one. The failure mode is "test runs before container is ready," which is inherently a CI-orchestration race — there is no realistic deterministic regression test for it short of running this very workflow. The workflow itself, which was failing on three consecutive develop commits prior to this PR, is the regression signal. Adding a synthetic guard that races a sleep against the same poll would just be testing the test.

Issue #7718, opened 2026-05-10
ep_aa_file_menu_toolbar crashes the app


kimllee:
Hello,

Looks like another plugin (ep_aa_file_menu_toolbar) crashes the app (v2.7.3)
Thanks

```
[2026-05-10T19:46:10.216] [ERROR] settings - Failed to load hook function "/opt/etherpad-lite/src/plugin_packages/ep_aa_file_menu_toolbar/eejs:eejsBlock_body" for plugin "ep_file_menu_toolbar" part "aa_file_menu_toolbar" hook set "hooks" hook "eejsBlock_body": Error: Cannot find module '/opt/etherpad-lite/src/plugin_packages/ep_aa_file_menu_toolbar/eejs'
Require stack:
- /opt/etherpad-lite/src/static/js/pluginfw/shared.ts
- /opt/etherpad-lite/src/static/js/pluginfw/plugins.ts
- /opt/etherpad-lite/src/static/js/pluginfw/installer.ts
- /opt/etherpad-lite/src/node/server.ts
    at node:internal/modules/cjs/loader:1383:15
    at nextResolveSimple (/opt/etherpad-lite/node_modules/.pnpm/tsx@4.21.0/node_modules/tsx/dist/register-D46fvsV_.cjs:4:1004)
    at /opt/etherpad-lite/node_modules/.pnpm/tsx@4.21.0/node_modules/tsx/dist/register-D46fvsV_.cjs:3:2630
    at /opt/etherpad-lite/node_modules/.pnpm/tsx@4.21.0/node_modules/tsx/dist/register-D46fvsV_.cjs:3:1542
    at resolveTsPaths (/opt/etherpad-lite/node_modules/.pnpm/tsx@4.21.0/node_modules/tsx/dist/register-D46fvsV_.cjs:4:760)
    at Function._resolveFilename (/opt/etherpad-lite/node_modules/.pnpm/tsx@4.21.0/node_modules/tsx/dist/register-D46fvsV_.cjs:4:1102)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1025:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1030:22)
    at Function._load (node:internal/modules/cjs/loader:1192:37)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
```

JohnMcLear:
Can you provide steps to replicate please?  Installed w/ CLI or admin page?  Any other plugins?

[...]

Issue #7687, opened 2026-05-07
Dockerfile issue with corepack & pnpm


ferrlen:
### Describe the bug
When running the Dockerfile, right before installing local plugins, corepack seems to try to install pnpm@latest (which is v 10.33.4), even though it had `corepack prepare pnpm@11.0.7 --activate`. Up till now, it seems pnpm was running with v 11.0.7, but at this point of the Dockerfile I think pnpm isn't found and corepack tries to redownload it.

This is the error I get when trying to install my local plugin:

```
! Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-10.33.4.tgz
10.33.4
[6/6] STEP 16/22: RUN bash -c ./bin/installLocalPlugins.sh
Installing plugin: 'ep_myplugin'
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

Your pnpm version is incompatible with "/opt/etherpad-lite/src".

Expected version: >=11.0.0
Got: 10.33.4

This is happening because the package's manifest has an engines.pnpm field specified.
To fix this issue, install the required pnpm version globally.

To install the latest version of pnpm, run "pnpm i -g pnpm".
```

### My fix
I added `corepack use pnpm@11.0.7` to:

```RUN  \
    mkdir -p /usr/share/man/man1 && \
	npm install -g corepack@latest && \
    corepack enable && corepack prepare pnpm@${PnpmVersion} --activate && \
    corepack use pnpm@11.0.7 && \
...
``` 

This made corepack download the right version, 11.0.7, at that point. I don't know enough about corepack, pnpm, or even Docker to correctly pinpoint the issue, but I think it's got something to do with pnpm workspace not finding the pnpm installation (which makes corepack redownload it).  

### To Reproduce
1. Run Dockerfile with a local plugin installation
2. See the above error when trying to install the plugin

PR #7727, opened 2026-05-11
fix(docker): bypass pnpm at runtime to avoid spurious deps-status reinstall (#7718)

JohnMcLear:
## Summary

- pnpm 11's `runDepsStatusCheck` fires before every `pnpm run …` and spuriously decides `node_modules` is out of sync on container first start under the named-volume layout in `docker-compose.yml` (mounting `src/plugin_packages`).
- It then spawns `pnpm install --production`, which either prompts to wipe `node_modules` (`tty: true`) or aborts with `ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY` (no tty) — both manifest to the user as a crashed container.
- Reproduced by @kimllee in #7718 on `etherpad/etherpad:latest` (arm64, RPi5).
- Fix: run `node` directly in `CMD` instead of going through `pnpm run prod`. `node_modules` was already verified during image build, so the runtime check adds no value. `sh -c 'cd src && exec node …'` keeps `WORKDIR` consistent for `docker exec` users while making node PID 1 for clean SIGTERM handling.

## Root cause (from stack trace)

```
[ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY] Aborted removal of modules directory due to no TTY
[ERROR] Command failed with exit code 1: …/pnpm.mjs install --production
    at runDepsStatusCheck (…/pnpm.mjs:210115:7)
```

`runDepsStatusCheck` is pnpm 11's automatic pre-run hook on `pnpm run`. The check compares `node_modules` to `pnpm-lock.yaml`; when the named volume on `src/plugin_packages` initialises, pnpm's heuristic perturbs and pnpm tries to reinstall — entirely unwanted at runtime.

## Test plan

- [x] CI `build-test` job runs the image, hits the health endpoint, and runs `pnpm run test-container` — must stay green
- [x] CI `build-test-local-plugin` regression test (#7687) must stay green
- [x] CI `build-test-db-drivers` MySQL + Postgres smoke must stay green
- [ ] After merge + image republish, ask @kimllee to confirm the next `etherpad/etherpad:latest` boots cleanly with their `docker-compose.yml` (named volumes on `plugin_packages` + `var`, no `tty:`)


JohnMcLear:
Addressed Qodo review (0ad5d01c6):

Added a regression step to `build-test` in `.github/workflows/docker.yml` that reproduces the #7718 production-docker-compose layout — named volume on `src/plugin_packages`, no allocated TTY — and waits for the container to go healthy. If the Dockerfile `CMD` ever reverts to `pnpm run prod`, this step times out and fails CI.

This matches kimllee's reported conditions exactly (named volume + non-TTY mode); the failure mode from the stack trace in #7718 (`ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY` → container exits before health endpoint comes up → step times out) is what the assertion catches.

PR #7741, opened 2026-05-14
Support Node 25 with corepack-managed pnpm

JohnMcLear:
## Summary
- route root scripts and startup plugin pnpm calls through Corepack so fresh Node 25 installs use the pinned pnpm version
- bump the Corepack pin to pnpm 11.1.2 and keep the explicit pnpm engine requirement
- move fixed-version CI jobs and PR-default workflow runs to Node 25 while keeping the wider 22/24/25 matrix on push

## Testing
- corepack pnpm test-utils
- clean-copy install with corepack pnpm install --frozen-lockfile
- clean-copy startup on Node 25 and HTTP probe on a free port


JohnMcLear:
❯ why in this PR is tehre so much pnpm shimming/dancing?

● Because Node 25 wasn’t really the hard part; pnpm resolution was. Etherpad already mostly runs on Node 25. The failures came from different places picking
  up different pnpm binaries:

   - local/global pnpm could be older than the repo expects
   - CI actions were installing one pnpm version while the repo pinned another
   - runtime paths like plugin migration, updater, and shell scripts were bypassing the pinned packageManager version entirely

  So the “shimmy” is mostly making all those entrypoints converge on the repo-pinned pnpm via Corepack instead of “whatever pnpm happens to be on PATH”.

  Without that, a fresh clone on Node 25 could:

   - install with one pnpm,
   - run scripts with another,
   - and hit runtime/update/plugin code with a third.

  That’s why the PR touches more than package.json: the point was to make install, CI, startup, plugins, updater, and Docker all use the same pnpm contract.

  Short version: the PR is not working around Node 25 so much as removing hidden assumptions that “bare pnpm on PATH is fine.”

JohnMcLear:
Corepack is dropped in 25.  Need to check that.

JohnMcLear:
Closing — superseded by #7747 (clean Node 25 scope, no corepack shimming), which has merged as 99f8258811d7405387a5e50765e6a147b756b832.

PR #7755, opened 2026-05-15
security: allow integrator sessionID cookie to be HttpOnly (#7045)

### LINKED ISSUE #7045: Could the sessionId cookie be made httpOnly to avoid potential XSS?
Currently the sessionId can not be made httpOnly

sessionId to be made httpOnly

going the oauth route might not be worth it and might be affected too?

documentation I found https://docs.etherpad.org/cookies.html

JohnMcLear:
## Summary

Closes #7045.

- Etherpad's client JS used to read the integrator-set `sessionID` cookie via `document.cookie` and forward it in the socket.io `CLIENT_READY` payload, which forced the cookie to be non-`HttpOnly` and exposed it to XSS.
- Mirror the GDPR PR3 author-token migration: `PadMessageHandler.handleClientReady` now reads `sessionID` from the socket.io handshake's `Cookie` header, with a one-time deprecation warning per socket if a legacy client still forwards it on the message.
- Drop `Cookies.get('sessionID')` in `pad.ts` and `timeslider.ts`. Current clients no longer send the field; old clients keep working via the legacy fallback.
- Update `doc/cookies.md` to recommend `HttpOnly; Secure; SameSite=Lax` for the integrator-set cookie.

## Compatibility

Non-breaking. Integrators can now *optionally* set `HttpOnly`; existing setups without `HttpOnly` keep working.

## Test plan

- [x] New backend test (`src/tests/backend/specs/sessionIdCookie.ts`):
  - reads `sessionID` from the handshake `Cookie` header
  - honours the configured cookie prefix
  - falls back to `message.sessionID` for legacy clients
  - prefers the cookie over the legacy message field
  - records `null` when no `sessionID` is provided
- [x] `pnpm run ts-check`
- [x] Existing `socketio.ts`, `authorTokenCookie.ts`, `ensureAuthorTokenCookie.ts` suites still pass (the only socketio.ts failures observed are a pre-existing live-reload race in `specialpages.ts` unrelated to this change).


JohnMcLear:
fixed in 4c36d5c. `readCookie` now catches `URIError` from `decodeURIComponent` and treats the value as absent so a malformed cookie like `sessionID=%ZZ` no longer aborts CLIENT_READY. Added a regression test in `src/tests/backend/specs/sessionIdCookie.ts` covering that case.

JohnMcLear:
I just want to be 100% sure on this that this doesn't negatively impact embedded Etherpad instances..  It's not behind a feature flag/setting so we have to be careful...

SamTV12345 (approved):

Issue #7255, opened 2025-12-06
Inaccessibility to screenreaders


StrangeGirlMurph:
I am shocked to find out that etherpad isn't accessible to screenreaders at all. How did no one think about that? It's truly horrible. For me it just cycles through the line numbers. No sight of a skip to content link or anything. I mean there is no point in even listing out everything that's bad. See for yourself. 

Yes making it truly accessible to screenreaders with all the edit functionality is hard but allowing people to actually read the content is a must.

Please work on that.

SamTV12345:
Hi @StrangeGirlMurph Could you ellaborate on where Etherpad is not accessible? Would be great to find out so we can fix it.

StrangeGirlMurph:
Hey @SamTV12345, sadly basically everywhere. I don't feel like giving an accessibility in web dev 101 talk and explain the absolute basics. I would hope someone on the team would have at least some knowledge. There is also: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility
I highly recommend you also just try it out yourself! I find tablets are a great way to test screen reader support. 

Basically is etherpad right now just not navigable for screen readers. Some points:
- To get to the main content you have to go over all editing tools in the header. Thats 18 swipes. Thats what skip to content buttons are there for!
- On etherpad.wikimedia.org you can't individually cycle through the paragraphs and links. You have know where they are located in order to click them. It just treats all the text as one big block.
- After that it individually cycles through every single line number and just reads them out loud. Of course a blind person would have no clue that the numbers read are line numbers.

These are just some things. Someone really needs to actually think this through. Your LLM of choice can also offer some directions.

StrangeGirlMurph:
Okay I took the time to make a small screen recording of me just swiping through. People who know what to look for will find many problems from that. Note that doesn't even cover all the problems with actually interacting with the buttons. The chat for example in the end is just not available. 

https://github.com/user-attachments/assets/f94097fd-addd-417f-9888-cd1df0ae604b

JohnMcLear:
We put a lot of work into accessibility with aria..  Is aria not working or something?

StrangeGirlMurph:
Sorry for the delay. Was busy with an exam.
Now: The navigation and using divs and spans for text are the problem! The aria tags seems fine from a quick look.
[...]

PR #7758, opened 2026-05-15
fix(a11y): skip-to-content link + hide line numbers from screen readers (#7255)

JohnMcLear:
## Summary

Follow-up to PR #7451 — addresses the two screen-reader complaints from issue #7255 that PR didn't cover:

- **Skip link (WCAG 2.4.1 Bypass Blocks)** — adds a visually-hidden `<a id="skip-to-content">` as the first child of `<body>`. Keyboard focus reveals it; activating routes focus into the editor via `padeditor.ace.focus()` (a plain `href="#editorcontainer"` only scrolls — the real focus target sits inside two nested iframes).
- **`aria-hidden="true"` on `#sidediv`** — line numbers are visual scaffolding only. Without this, screen readers iterate "1, 2, 3, …" past every line. The reporter explicitly asked for them to be hidden until line-start-announcement is properly designed.

## Bugs fixed along the way

- **Keyboard hint was being wiped.** PR #7451 added `#editor-keyboard-hint` to the inner iframe's `<body>`, but `Ace2Inner.init()` clears body children to set up the line model — so the hint never actually existed at runtime, leaving the body's `aria-describedby` dangling. Moved the hint to `<head>` of the inner document (the ARIA spec lets aria-describedby resolve to any ID in the same document).
- **Hardcoded English in `ace.ts:304`.** The hint text now uses `html10n.get('pad.editor.keyboardHint')`.

## New locale keys

- `pad.editor.skipToContent` — "Skip to editor"
- `pad.editor.keyboardHint` — "Press Escape to exit the editor. Press Alt+F9 to access the toolbar."

## What this does NOT address from #7255

Paragraph-level navigation (Murphy's complaint that text is `<div>`/`<span>` and screen readers treat it as one block). As discussed in the issue thread, that's an architectural change to the line model in `domline.ts` and warrants its own PR.

## Test plan

- [x] `pnpm ts-check` passes
- [x] `pnpm test` passes (1121/1121, 8 pending)
- [x] Local smoke test against dev server: skip link href, localized text, `aria-hidden` on sidediv, hint text in inner `<head>`, Enter on skip link transfers focus into iframe — all pass
- [x] Two new Playwright assertions in `a11y_dialogs.spec.ts`
- [x] Manual: NVDA/VoiceOver: Tab from address bar → skip link visible → Enter jumps into editor
- [x] Manual: NVDA: navigate the pad, confirm line numbers no longer read out

Refs #7255

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -360,6 +360,8 @@
   "pad.historyMode.settings.playbackSpeed": "Playback speed:",
   "pad.historyMode.chat.replayHeader": "Chat as of {{time}}",
   "pad.historyMode.users.authorsHeader": "Authors at this revision",
+  "pad.editor.skipToContent": "Skip to editor",
+  "pad.editor.keyboardHint": "Press Escape to exit the editor. Press Alt+F9 to access the toolbar.",
   "timeslider.toolbar.authors": "Authors:",
   "timeslider.toolbar.authorsList": "No Authors",
   "timeslider.toolbar.exportlink.title": "Export",
--- a/src/static/css/pad.css
+++ b/src/static/css/pad.css
@@ -26,6 +26,28 @@
   border: 0;
 }
 
+/*
+ * Skip link — hidden until keyboard focus reveals it at the top-left of the
+ * viewport, giving screen-reader and keyboard-only users an immediate jump
+ * past the toolbar to the editor (WCAG 2.4.1 Bypass Blocks).
+ */
+.skip-link {
+  position: absolute;
+  top: -100px;
+  left: 8px;
+  z-index: 10000;
+  padding: 8px 12px;
+  background: #2c3e50;
+  color: #fff;
+  text-decoration: none;
+  border-radius: 0 0 4px 4px;
+}
+.skip-link:focus {
+  top: 0;
+  outline: 2px solid #fff;
+  outline-offset: 2px;
+}
+
 html {
   font-size: 15px;
   color: #3e3e3e;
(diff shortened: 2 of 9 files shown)


JohnMcLear:
I'll keep working on this, I have a test environment setup now.

JohnMcLear:
Qodo review actioned on commit `089f08e`:

- **Bug #1 (`hidden=true` defeats `aria-describedby`)** — removed the `hidden` attribute. Hint lives in the inner document's `<head>` where it doesn't render anyway, so there's no visual cost; screen readers can now resolve the description.
- **Bug #2 (async `html10n.get` race)** — seeded with a hardcoded English fallback so the hint is always non-empty, and bound to `html10n.bind('localized', ...)` so the fallback is replaced once translations load.
- **Bug #3 (no relocalization on language change)** — same `'localized'` event binding handles this — runtime `applyLanguage()` calls fire `localized` again, refreshing the hint.
- **Bug #5 (brittle `BODY` precondition in the Tab test)** — replaced the precondition with an explicit `blur()` of whatever happens to be focused. The assertion is now purely about tab order, not about the post-`goToNewPad` starting state.
- **Rule violation: missing regression test for the hint** — added `editor-keyboard-hint exists in the editor iframe with localized text` in `a11y_dialogs.spec.ts`. Asserts the hint exists post-init, is non-empty, isn't the literal string `'undefined'`, and contains `'Escape'`.

**Declining: Rule violation #4 (skip link not feature-flagged).** A WCAG 2.4.1 compliance fix shouldn't be opt-in — a flagged-off skip link wouldn't help any user, and "preserve pre-existing behavior" isn't a useful default when the pre-existing behavior is the bug. The change is also visually inert (the link is offscreen until keyboard focus reveals it), so there's no risk of regressing the sighted-user UX.

PR #7765, opened 2026-05-15
docs: scaling dive 2026-05 (closes Phase 2 of #7756)

JohnMcLear:
## Summary

Numbers-backed answer to #7756: "how many editors can be on one pad, and where is the bottleneck?" Every claim links to a CI run whose `report.json` is downloadable for re-analysis.

This PR description is intentionally brief; the **doc itself (`docs/scaling-dive-2026-05.md`) is the source of truth** and contains the full methodology, every lever scored, the negative results, and the honest re-evaluations.

## Headline findings

1. **Real ceiling on a 4-vCPU GitHub runner: ~350-400 authors per pad.** Beyond that, p95 cliffs from ~80 ms to ~2000 ms within one step. CPU consumed 7+ cores of work per wall-second — OS preemption, not slow code paths.

2. **The "250-author cliff" earlier reports kept hitting was a measurement artefact.** Per-IP `commitRateLimiting` + harness colocation (all simulated authors share `127.0.0.1`). Fixed in [load-test#105](https://github.com/ether/etherpad-load-test/pull/105). Production deployments with many client IPs are not affected.

3. **Runner noise is significant.** Each GHA matrix entry runs on a potentially different physical runner. The same `develop` baseline produced p95 between 38 ms and 122 ms at step 350 across three identical runs. **All single-run lever conclusions sit inside the noise envelope** below the cliff. The doc adopts N=3-per-lever methodology and reports min/median/max.

4. **Two changes hold up under N=3 scoring:**
   - **[#7774](https://github.com/ether/etherpad/pull/7774)** — engine.io socket flush deferral. Tightens the tail at mid-load (step 350 max: 122 → 110 ms; step 300 max: 71 → 58 ms). Wire-compatible, server-side only. **The only N=3-confirmed perf win in this program.**
   - **[#7768](https://github.com/ether/etherpad/pull/7768)** — per-socket fan-out serialization. No measurable perf benefit (the earlier "70% drop" was single-run noise), but a real correctness fix: the previous code was racy under concurrent commits and could lose revisions on emit failure. Recommend merging for correctness, not perf.

5. **Five directions did not pan out, documented in full:**
   - WebSocket-only transport — reliably worse at cliff (N=3 confirms).
   - `--max-old-space-size` heap bump — no effect.
[...]

Issue #7756, opened 2026-05-15
Performance: Look deeper into scaling N+ contributors per pad.


JohnMcLear:
We have scale.etherpad.org and etherpad-load-test as tooling to document & validate how many editors can be on a pad but I'm not sure we ever seriously dived deep into alternatives to get the numbers of editors on a pad beyond what is available on scale.etherpad.org

My thoughts are..

1. Transport layer IE try websocket over socketio
2. Node ENV vars IE allow more memory to node (afaik our bottleneck is Network IO tho).
3. Message reduction size IE providing a buffer within the changeset messages for when there are thousands of edits landing a second its not 1000 tiny packets but 100 larger ones...

PR #7766, opened 2026-05-15
feat(scaling): fan-out debounce (#7756 lever 3 prototype)

JohnMcLear:
## Summary

The scaling-dive doc ([#7765](https://github.com/ether/etherpad/pull/7765)) identifies fan-out as the dominant cost at high concurrency:

- \`etherpad_changeset_apply_duration_seconds\` stays under 5 ms even at 200 authors — server-side apply is not the bottleneck.
- \`etherpad_socket_emits_total{type=NEW_CHANGES}\` scales O(N²): 1 160 emits/dwell at 20 authors becomes 66 032 at 200 authors.
- Baseline cliffs at ~250 authors (errorRate flag fires).

This is the prototype for spec lever 3 — coalesce per-pad fan-out within a debounce window. With \`fanoutDebounceMs: 50\` (or any value > 0), rapid commits batch into a single \`updatePadClients\` invocation per pad per window. With \`fanoutDebounceMs: 0\` (the default) the legacy behaviour is preserved byte-for-byte.

## Implementation

- New setting \`fanoutDebounceMs: 0\` (default 0). Production deployments are not affected.
- New module \`src/node/handler/FanoutScheduler.ts\` (~30 lines) holding the per-pad \`Map<padId, Timeout>\` and the public \`scheduleFanout(padId, callback)\`. Lives in its own file so it can be unit-tested without standing up the full pad/DB/socket.io stack — extracting it avoided a module-load cycle when the test imports \`PadMessageHandler\` directly.
- \`handleUserChanges\` replaces \`await exports.updatePadClients(pad)\` with \`scheduleFanout(pad.id, runFanout)\`. Errors from the deferred fan-out route through \`messageLogger\` via \`setErrorHandler\`.
- Initial-join catch-up at line ~1377 keeps the synchronous call — a joining client must not miss any revisions, no batching there.

## Tests

\`src/tests/backend-new/specs/fanout-debounce.test.ts\` — 5 passing tests:

- debounce=0 passthrough fires every call
- debounce>0 coalesces N rapid calls into one
- per-pad independence (one pad's window doesn't gate another)
- after-flush rescheduling starts a fresh window
- error routing through the configurable handler

## Scoring this

The dive workflow doesn't have a \`fanout-debounce\` matrix entry yet — that's a small follow-up in [ether/etherpad-load-test](https://github.com/ether/etherpad-load-test). For now, the lever can be scored manually:

\`\`\`
gh workflow run "Scaling dive" --repo ether/etherpad-load-test \\
  --ref main \\
  -f core_ref=feat/fanout-debounce \\
  -f sweep='authors=20..200:step=20:dwell=10s:warmup=2s'
\`\`\`

(Plus a sed step in the workflow to inject \`fanoutDebounceMs: 50\` — I'll send that as a separate ether/etherpad-load-test PR.)

## Test plan

- [x] tsc clean
- [x] 5/5 fanout-debounce tests pass
- [x] 3/3 prom-instruments tests still pass (no regression in #7762 work)
- [ ] CI green
- [ ] After merge, score with the dive workflow against \`develop\` post-merge — expect \`emits_new_changes\` to drop and client p95 to improve at step ≥ 100

### DIFF
--- a/settings.json.template
+++ b/settings.json.template
@@ -733,6 +733,14 @@
    */
   "loadTest": false,
 
+  /*
+   * Coalesce NEW_CHANGES fan-out within a debounce window of N milliseconds
+   * (#7756 lever 3). 0 (default) preserves legacy synchronous behaviour.
+   * Increase (typical: 25-100) to trade a little latency for a quadratic
+   * reduction in socket.io emit volume under high pad concurrency.
+   */
+  "fanoutDebounceMs": 0,
+
   /**
   * Disable dump of objects preventing a clean exit
   */
--- a/src/node/handler/FanoutScheduler.ts
+++ b/src/node/handler/FanoutScheduler.ts
@@ -1,8 +1,14 @@
 // Per-pad fan-out scheduler — debounce wrapper for #7756 lever 3.
 //
-// When `settings.fanoutDebounceMs <= 0` (default), fan-out fires immediately
-// — legacy behaviour. When > 0, rapid scheduleFanout(pad) calls coalesce
-// into a single fan-out per pad per debounce window.
+// When `settings.fanoutDebounceMs <= 0` the scheduler is bypassed entirely;
+// PadMessageHandler.handleUserChanges awaits updatePadClients directly. When
+// > 0, rapid scheduleFanout(padId, fn) calls coalesce into a single fanout
+// invocation per pad per debounce window.
+//
+// Concurrency contract: a single pad has at most one in-flight fan-out at a
+// time. If a new commit arrives while a fan-out is running, a "dirty" flag
+// is set; when the running fan-out finishes it triggers a follow-up schedule
+// so the late commit isn't missed.
 //
 // Lives in its own module rather than inside PadMessageHandler so the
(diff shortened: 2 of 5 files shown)


JohnMcLear:
Closing. Scored against the dive workflow on run [25940112728](https://github.com/ether/etherpad-load-test/actions/runs/25940112728) — the lever does **not** meaningfully reduce emit volume.

Why: this PR debounces *calls* to \`updatePadClients\` but the per-socket \`while (sessioninfo.rev < head)\` loop still fires one emit per revision per socket regardless of how many revs are pending. So a debounced call fans out *more* revs per socket but the total emit count stays the same. Comparison at 200 authors:

| | p95 | emits |
|---|---:|---:|
| baseline | 22 ms | 66 006 |
| this PR (fanoutDebounceMs=50) | 20 ms | 66 014 |

The real fix is a NEW_CHANGES_BATCH message type that carries an array of revisions in a single emit per socket. That requires client-side handling and is a bigger change — to be opened as its own PR.

\`FanoutScheduler.ts\` and the test scaffolding are still useful as primitives but I don't want to ship a setting that operators would set without benefit. Closing without merging.

Issue #7767, opened 2026-05-15
Investigate why dropping socket.io polling fallback degrades p95 under load (#7756 follow-up)


JohnMcLear:
From the [2026-05 scaling dive](https://github.com/ether/etherpad/pull/7765), setting \`socketTransportProtocols: [\"websocket\"]\` (i.e. dropping the polling fallback) consistently *worsens* client p95 latency under high concurrency:

| Authors | baseline p95 | websocket-only p95 | apply_mean baseline | apply_mean ws-only |
|---:|---:|---:|---:|---:|
| 100 | 11 ms | 18 ms | 4.16 ms | 5.13 ms |
| 140 | 14 ms | 25 ms | 4.02 ms | 6.09 ms |
| 180 | 16 ms | **68 ms** | 4.48 ms | 9.81 ms |
| 200 | 22 ms | **82 ms** | 4.95 ms | 13.33 ms |

Same harness, same runner, same sweep — only difference is the transport setting.

## Hypotheses worth checking

1. WS-only forces clients that can't establish WebSocket within socket.io's handshake timeout to retry-loop rather than fall back to polling, producing reconnect storms that drive up server CPU.
2. The WS-only path changes socket.io's handshake protocol routing in a way that interacts badly with load balancers / proxies.
3. Per-message-WebSocket framing overhead becomes significant when emits/sec is high (66k/dwell at 200 authors).
4. The polling fallback acts as a *natural* coalescer (multiple events per HTTP poll) that we lose when forcing pure-WS.

Hypothesis 4 is particularly interesting because it would mean polling-as-batching is doing real work for us today.

## Reproducing

\`\`\`
gh workflow run \"Scaling dive\" --repo ether/etherpad-load-test --ref main \\
  -f core_ref=develop \\
  -f sweep='authors=20..200:step=20:dwell=10s:warmup=2s'
\`\`\`

Compare \`scaling-dive-baseline\` vs \`scaling-dive-websocket-only\` artifacts. Run [25940112728](https://github.com/ether/etherpad-load-test/actions/runs/25940112728) is the reference.

## Why this matters

If we know *why* the polling fallback helps so much, we can either preserve that property explicitly (via batching) or stop carrying socket.io's polling-fallback code path.

JohnMcLear:
## Diagnosis

Confirmed hypothesis 4 by reading the engine.io 6.6.5 transport code shipped with etherpad core. The polling fallback is a natural coalescer; WebSocket isn't.

[...]

PR #7769, opened 2026-05-16
feat(pad): cache historicalAuthorData (remove join-time thundering herd)

JohnMcLear:
## Summary

Per-pad cache for the `{authorId → {name, colorId}}` map that `handleClientReady` builds for every CLIENT_READY message. Eliminates a thundering-herd condition on join when multiple users connect at once to a long-lived pad.

## Origin: the dive's 250-author cliff turned out to be two stacked artifacts

I built this PR thinking it would move the dive's 250-author cliff (#7756). It didn't, and the data is interesting:

1. **The 250-author cliff was a measurement artifact.** `NODE_ENV=production` enables Etherpad's per-IP `commitRateLimiting`. `handleMessage` runs the rate limiter on every message including `CLIENT_READY`. With the dive harness and SUT colocated on one runner, all simulated authors share `127.0.0.1` = one rate-limit bucket = trips at N × edit-rate. Fixed in [etherpad-load-test#105](https://github.com/ether/etherpad-load-test/pull/105).

2. **The real cliff on a 4-vCPU GitHub runner is 350-400 authors**, with or without this PR. Both develop and this branch hit the `p95` flag at step 400 with `p95 ≈ 2000 ms` and CPU at 54-65 CPU-seconds per 8-second dwell. Steady-state fan-out CPU saturation, not join-path cost.

So: this PR doesn't move the dive's cliff. **It does fix a real thundering-herd condition that only appears when many users join one pad at the same time**, which is exactly the case it's worth caching for — just not what limits concurrency at the saturation boundary.

## Change

Each CLIENT_READY previously did:

```ts
const authors = pad.getAllAuthors();
await Promise.all(authors.map(async (id) => {
  const a = await authorManager.getAuthor(id);
  historicalAuthorData[id] = {name: a.name, colorId: a.colorId};
}));
```

At 200 existing authors × 50 simultaneous joiners that's **10 000 ueberdb cache lookups + async/await bookkeeping** for a result the joiners share. New per-pad cache (5-second TTL) collapses that to one shared computation per burst.

- `src/node/db/HistoricalAuthorDataCache.ts` — pure helper, no DB or Pad dependency. Takes the author-id list and the per-id fetcher as constructor args.
- `Pad.getHistoricalAuthorData()` — lazy-inits a cache per pad.
- `PadMessageHandler.handleClientReady` — replaces the inline Promise.all loop with a single await.

## Tests (6/6)

- Returns one entry per author with `{name, colorId}`.
- **50 simultaneous `get()` calls produce 1 fetch per author** — the win.
- Refetches once the TTL expires.
- Omits authors the fetcher returns falsy for.
- Explicit `invalidate()` works (hookable later for author-add events).
- Failed fetch leaves the cache clearable, next call retries.

## When this matters

Production-significant when **many users join the same pad in a short window**: classroom-style "everyone open the link now", live-event onboarding bursts, integration tests, etc. The cost it removes is CPU-bound on the event loop right at the moment new users are waiting on `CLIENT_VARS`, so removing it improves perceived join latency under that specific load.

The dive doesn't exercise this scenario at a level where it dominates; the cliff at 350-400 in the dive is true steady-state CPU saturation. Moving that ceiling further is a separate investigation (engine.io transport-level packing per [#7767](https://github.com/ether/etherpad/issues/7767), or bigger hardware).

## Test plan

- [x] CI green
- [x] 6/6 unit tests pass
- [x] Steady-state numbers unchanged at all step sizes through 350 in [run 25949421120](https://github.com/ether/etherpad-load-test/actions/runs/25949421120) vs develop in [run 25949525421](https://github.com/ether/etherpad-load-test/actions/runs/25949525421)
- [ ] Manual: open a pad with N>50 authors, join from 10 browsers within 2 seconds. Expect no observable change vs develop except the connect-time CPU spike from `getAuthor` fans should disappear.

### DIFF
--- a/src/node/db/HistoricalAuthorDataCache.ts
+++ b/src/node/db/HistoricalAuthorDataCache.ts
@@ -1,6 +1,6 @@
 // Per-pad cache for the `{authorId -> {name, colorId}}` map used by
 // PadMessageHandler.handleClientReady to populate clientVars
-// (#7756 connect-handshake cliff investigation).
+// (#7756 connect-handshake investigation).
 //
 // At 200+ authors a burst of 50 simultaneous CLIENT_READY handshakes
 // would otherwise each do Promise.all(authors.map(getAuthor)) =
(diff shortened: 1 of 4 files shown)


JohnMcLear:
**Addressing Qodo's four items from the initial review** (commit `478c663`):

1. **Cache recompute race** → Fixed. `get()` now piggybacks on the in-flight promise regardless of TTL (\`if (s?.promise) return cloneData(await s.promise)\`), and the resolve handler only commits state if it's still the active promise (covers \`invalidate()\` racing a compute). New test `a slow compute that runs past TTL still resolves callers; no duplicate fetch starts in flight` pins this.
2. **Unexplained conditional type** → Fixed. Replaced \`AuthorRecord extends never ? never : ...\` with a documented \`CacheState | null\` type, fields commented inline.
3. **Shared cache object** → Fixed. All `get()` returns go through a `cloneData()` helper that produces a fresh top-level object and fresh per-author records. Mutations by one joiner can't bleed into the next. New test `returns a fresh object on every get()` pins this.
4. **Missing author logs** → Fixed. Cache now accepts an optional `onMissingAuthor(id)` callback, invoked once per id the fetcher returns falsy for. `Pad.getHistoricalAuthorData` wires it to a log4js logger emitting the same `'There is no author for authorId: ... related to #2802'` message the inline loop used to. New test `calls onMissingAuthor exactly once per id the fetcher returns falsy for` pins it.

8/8 tests in `pad-historical-author-data.test.ts` green. All 31 CI checks pass. CI: https://github.com/ether/etherpad/pull/7769/checks

JohnMcLear:
Closing. Triple-run N=3 scoring against this PR shows it's **net-negative above ~300 authors** vs develop.

p95 envelope comparison (min/med/max across 3 identical sweeps each):

| Step | develop | this PR |
|---:|---|---|
| 200 | 30 / 37 / 51 | 29 / 38 / 65 |
| 300 | 38 / 45 / 71 | **39 / 93 / 240** |
| 350 | 39 / 39 / 122 | **301 / 488 / 633** |
| 400 | 1758 / 2275 / 2463 | **3053 / 3203 / 3327** |

Probable cause: the defensive shallow-clone-on-every-get() added in the Qodo-feedback fix walks O(N) author entries per joining client. With burst-of-50 new joiners × N existing authors × clone allocations at each step ramp, plus GC pressure, the cache becomes more expensive than the inline Promise.all it was meant to collapse.

The hypothesis this PR addressed — *the 250-author cliff is a join-thundering-herd problem* — was also falsified: the 250 cliff turned out to be a per-IP commitRateLimiting artefact from harness colocation (fixed in [load-test#105](https://github.com/ether/etherpad-load-test/pull/105)). The cache PR was built on that wrong diagnosis.

What might still be salvageable for production (where many real users from many IPs DO join simultaneously, not measurable in the dive) is the *join coalescing* idea itself — but without the defensive clone, and without the TTL refresh that adds another N-object-allocation footgun. If anyone wants to revisit, the `HistoricalAuthorDataCache` module remains a useful template; just rework the public surface to return the data structure the cache owns (with documented "don't mutate" contract) instead of cloning per call.

The 8 unit tests in `pad-historical-author-data.test.ts` still pass — the implementation is correct, it just doesn't deliver the perf benefit it was scoped to deliver.

PR #7770, opened 2026-05-16
perf(pad): prefetch rebase revisions in one Promise.all (move #7756 cliff)

JohnMcLear:
## Summary

Dive run [25949421120](https://github.com/ether/etherpad-load-test/actions/runs/25949421120) found the real ceiling — 350-400 authors on a 4-vCPU GitHub runner, with apply_mean ballooning from ~10 ms at step 200 to ~50 ms at step 400. That's much higher than the synchronous apply work; the difference is **event-loop yield latency**.

Both hot loops in PadMessageHandler do one `await pad.getRevision(r)` per intermediate revision in serial:

- \`handleUserChanges\` rebase: \`while (r < pad.getHeadRevisionNumber()) { ... await pad.getRevision(r) ... }\`
- \`updatePadClients\` catch-up: \`while (sessioninfo.rev < pad.getHeadRevisionNumber()) { ... await pad.getRevision(r) ... }\`

Under steady-state CPU pressure each await is a yield, and the continuation queues behind whatever other work is pending. At cliff load (7+ cores of work on 4 vCPU) each yield costs ~6 ms wall-clock. A commit that's 5 revs behind pays that 5x. apply_mean = 30-50 ms even though the underlying ueberdb cache lookup is microseconds.

## Change

Both loops snapshot headRev once and prefetch the full range in a single Promise.all:

```ts
const rebaseTargetHead = pad.getHeadRevisionNumber();
const rebaseRange: number[] = [];
for (let i = baseRev + 1; i <= rebaseTargetHead; i++) rebaseRange.push(i);
const rebaseRevs = rebaseRange.length === 0
  ? []
  : await Promise.all(rebaseRange.map((i) => pad.getRevision(i)));
```

The actual rebase / fan-out iteration is now synchronous over the prefetched array. **N event-loop yields collapse to 1.**

In \`updatePadClients\` the prefetch covers the SUPERSET range (laggiest recipient's next rev → headRev) so all per-socket loops read from the local \`revCache\`. The existing fallback \`await pad.getRevision(r)\` stays as defensive code.

## Why this is correctness-positive too

The existing assertion \`assert([r, r + 1].includes(newRev))\` after the rebase loop was racy if another writer landed between the serial awaits — \`pad.getHeadRevisionNumber()\` could jump mid-loop. Snapshotting head once makes \`r === rebaseTargetHead\` deterministic at the end of the loop.

## Tests

- 3/3 in \`src/tests/backend-new/specs/rebase-prefetch.test.ts\` — pin the helper's parallel-fetch shape and result-ordering invariants.
- Existing 600+ backend-new tests still pass.

## Test plan

- [x] CI green
- [ ] Score against dive workflow with \`core_ref=feat/prefetch-rebase-revs\` at \`authors=100..500\`. Expect apply_mean at the cliff to drop noticeably (today's 40-50 ms → ideally ~10-15 ms), which should push the cliff out from step 400.
- [ ] If the cliff moves, push deeper.

### DIFF
--- a/src/node/handler/PadMessageHandler.ts
+++ b/src/node/handler/PadMessageHandler.ts
@@ -883,14 +883,30 @@ const handleUserChanges = async (socket:any, message: {
     let rebasedChangeset = moveOpsToNewPool(changeset, wireApool, pad.pool);
 
     // ex. applyUserChanges
-    let r = baseRev;
-
+    //
     // The client's changeset might not be based on the latest revision,
-    // since other clients are sending changes at the same time.
-    // Update the changeset so that it can be applied to the latest revision.
-    while (r < pad.getHeadRevisionNumber()) {
+    // since other clients are sending changes at the same time. Rebase
+    // through every intermediate revision so the changeset applies cleanly
+    // against the current head.
+    //
+    // Prefetching the whole range in one Promise.all collapses N
+    // event-loop yields (one `await pad.getRevision(r)` per intermediate
+    // revision) into 1, materially reducing apply_mean under steady-state
+    // CPU pressure (#7756). Snapshot the head once so we rebase to a
+    // stable target — any commits landing after this point get caught up
+    // by the next handleUserChanges invocation, matching the existing
+    // assertion below (newRev ∈ {r, r+1}).
+    const rebaseTargetHead = pad.getHeadRevisionNumber();
+    const rebaseRange: number[] = [];
+    for (let i = baseRev + 1; i <= rebaseTargetHead; i++) rebaseRange.push(i);
+    const rebaseRevs = rebaseRange.length === 0
+      ? []
+      : await Promise.all(rebaseRange.map((i) => pad.getRevision(i)));
+
+    let r = baseRev;
+    for (const revision of rebaseRevs) {
       r++;
-      const {changeset: c, meta: {author: authorId}} = await pad.getRevision(r);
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Closing. Scored against the dive on run [25953329610](https://github.com/ether/etherpad-load-test/actions/runs/25953329610) — apply_mean and p95 are unchanged within noise at every step from 100 to 500. Cliff stays at step 400.

I had the wrong model. Cached \`pad.getRevision\` resolves via microtask continuation, not macrotask. Microtasks drain after the current task before any macrotask runs, so they don't queue behind unrelated work under CPU pressure. The N→1 yield collapse the PR claimed to deliver wasn't actually buying anything.

The real cause of apply_mean ballooning at the cliff is **OS scheduling**: 7+ cores of work on 4 vCPU means the node process gets preempted. apply_mean is wall-clock around handleUserChanges, which includes time spent parked in the kernel runqueue. JS-level rearrangement can't fix that.

What this leaves on the table:
- The transport-level packing direction in #7767 (reduce total CPU per emit, not redistribute when it runs).
- Multi-worker per-pad sharding (independent processes — orthogonal to anything here).
- Bigger hardware.

The PR's snapshot-headRev correctness benefit (the existing assert `[r, r+1].includes(newRev)` being less racy under concurrent writers) is real but not what was claimed; not worth landing on its own.

PR #7772, opened 2026-05-16
feat(scaling): engine.io WS transport-level packing — lever 8 prototype

JohnMcLear:
## Summary

Prototype for the engine.io transport-level packing direction from [#7767](https://github.com/ether/etherpad/issues/7767) — the meatiest untouched lever in the scaling-dive doc.

**What changes.** engine.io's WebSocket transport sends one WS frame per engine.io packet, even when the engine.io socket has multiple packets buffered. The polling transport doesn't — its \`transport.send(packets)\` runs through \`encodePayload\` and writes one HTTP response containing the whole payload. This PR monkey-patches engine.io's WS transport so multi-packet flushes go out as ONE frame containing the payload-encoded bundle. Single-packet sends keep the legacy fast path (including the pre-encoded-frame optimisation), so steady-state quiet-pad behaviour is identical to upstream.

**Why.** At high emit rate the WS path is dominated by per-frame syscalls on the server and per-message callback overhead on the client. The dive measured this directly via lever 4 (\`socketTransportProtocols: [\"websocket\"]\`): dropping the polling fallback at 200 authors **quadrupled** client p95 and nearly tripled \`apply_mean\`. The polling fallback was acting as a natural coalescer; this PR brings the same coalescing property to the WS path.

## Implementation

- New module \`src/node/utils/EnginePacking.ts\` — installs the patch at boot. Idempotent. Resolves engine.io's own \`engine.io-parser\` instance so we use the exact \`encodePayload\` the transport already imports.
- \`src/node/hooks/express/socketio.ts\` — calls the installer once, BEFORE constructing the socket.io \`Server\`, so the patched prototype is in effect when transports are created.
- \`src/node/utils/Settings.ts\` — \`enginePacking: false\` default, documented warning about client compatibility.

## Rollout warning

Enabling this on a server with clients that haven't been updated will silently break those clients: a multi-packet frame fed to engine.io-client's stock \`decodePacket\` returns garbage. The harness-side patch in [ether/etherpad-load-test#TBD](https://github.com/ether/etherpad-load-test) is forward-compatible (detects \`\\x1e\` and routes through \`decodePayload\`). For production rollout the etherpad browser bundle would need the same forward-compat patch — out of scope for this prototype.

## Scoring this

Once merged (or via \`core_ref=feat/engine-io-ws-packing\`):

\`\`\`
gh workflow run \"Scaling dive\" --repo ether/etherpad-load-test --ref main \\
  -f core_ref=feat/engine-io-ws-packing \\
  -f sweep='authors=100..500:step=50:dwell=8s:warmup=2s'
\`\`\`

The dive workflow includes a new \`engine-packing\` matrix entry that sets \`enginePacking: true\` in \`settings.json\`. Expectations:
- \`emits_NEW_CHANGES\` count unchanged (we still emit N application-level messages).
- WS frame count on the wire reduced proportionally to packets-per-flush.
- Server CPU and \`apply_mean\` drop at the cliff, since per-frame syscall overhead drops.
- Client-side perceived latency improves if the per-message callback overhead was a significant chunk of the wait.

## Files touched

- \`src/node/utils/EnginePacking.ts\` (new, ~50 lines incl. comments)
- \`src/node/hooks/express/socketio.ts\` (+10 lines, conditional require)
- \`src/node/utils/Settings.ts\` (+1 type field, +14 documented default)

No tests yet — the patch is best validated end-to-end via the dive workflow. A unit test that constructs an in-memory WebSocket transport and asserts the framing is feasible but high effort for a prototype; happy to add it before merge if reviewers want.

### DIFF
--- a/src/node/hooks/express/socketio.ts
+++ b/src/node/hooks/express/socketio.ts
@@ -69,6 +69,16 @@ const socketSessionMiddleware = (args: any) => (socket: any, next: Function) => 
 };
 
 export const expressCreateServer = (hookName:string, args:ArgsExpressType, cb:Function) => {
+  // Engine.io WebSocket transport-level packing (ether/etherpad#7756 lever 8).
+  // Apply BEFORE constructing the socket.io Server so the patched transport
+  // prototype is in effect when the Server instantiates its engine.
+  if (settings.enginePacking === true) {
+    // Require lazily so production deployments that don't set the flag
+    // don't even load the patch module.
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    require('../../utils/EnginePacking').installEngineWsPacking();
+  }
+
   // init socket.io and redirect all requests to the MessageHandler
   // there shouldn't be a browser that isn't compatible to all
   // transports in this list at once
--- a/src/node/utils/EnginePacking.ts
+++ b/src/node/utils/EnginePacking.ts
@@ -1,0 +1,86 @@
+// Engine.io WebSocket transport-level packing — #7756 lever 8 prototype.
+//
+// Issue (see ether/etherpad#7767): engine.io's WebSocket transport sends one
+// WS frame per engine.io packet, even when the engine.io socket has multiple
+// packets buffered. The polling transport already coalesces — `send(packets)`
+// goes through `encodePayload(packets)` and writes one HTTP response
+// containing the whole payload. Under high emit rate the WS path is dominated
+// by per-frame syscall overhead on the server and per-message callback
+// overhead on the client.
+//
+// This module monkey-patches engine.io's WebSocket transport prototype so
+// `send(packets)` with N > 1 packets goes through `encodePayload` and emits
+// ONE WS frame containing the multi-packet payload. The frame contents are
+// the same wire bytes the polling transport already uses, just delivered as
+// a single WebSocket message instead of one frame per packet.
+//
+// Server-side only. The receiving client (engine.io-client, or anything
+// reading the WS frames) must detect the engine.io-parser record separator
+// (`\x1e`, U+001E) and call `decodePayload` instead of `decodePacket` when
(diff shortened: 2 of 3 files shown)


JohnMcLear:
Closing. Scored against the dive on [run 25954316731](https://github.com/ether/etherpad-load-test/actions/runs/25954316731): at step 350 (just below the cliff), engine-packing has p95=109ms and apply_mean=23.86ms vs baseline p95=76ms / apply_mean=16.15ms — neutral-to-slightly-worse, within noise. The patch is not the win we hoped for.

**Why** — engine.io's `socket.flush()` calls `transport.send(writeBuffer)` and resets the buffer to `[]`. It fires as soon as `transport.writable === true`, and for WebSocket `writable` returns to true within microseconds of each write. So even at 10k+ packets/sec, the writeBuffer rarely accumulates more than one packet. My patch only activated when `packets.length > 1`, so it almost never triggered.

For a real transport-level packing win we'd need to **deliberately delay flush** — for example, by deferring it to the next setImmediate or microtask boundary so multiple `sendPacket` calls within one tick accumulate. That's an invasive change to engine.io's flush semantics, not a per-method monkey-patch. Worth investigating but a much bigger PR than this one was scoped to be.

**Secondary finding** — this same run showed websocket-only as the *best* lever, contradicting every previous dive. The matrix-jobs-on-different-physical-runners issue (each GHA matrix entry runs on a separate machine) means our cross-lever comparisons are cross-hardware. Runner noise can flip lever conclusions. Real dive methodology going forward needs **N≥3 runs per lever** to filter out runner noise before drawing strong conclusions. Worth flagging on the dive doc when next updated.

The harness-side forward-compat patch in [ether/etherpad-load-test#106](https://github.com/ether/etherpad-load-test/pull/106) (already merged) stays — it's cheap forward-compat in case a future server-side change uses payload-encoded frames intentionally.

PR #7773, opened 2026-05-16
harden: reject USER_CHANGES inserts without an author attribute

JohnMcLear:
## Summary

Adds defensive validation in `handleUserChanges` that rejects USER_CHANGES whose `+` (insert) ops lack the `author` attribute. Insert ops without an author attribute can produce an AText where `text` and `attribs` disagree on length, which downstream clients then fail to reconcile in `ace2_inner.ts:setDocAText` with:

> mismatch error setting raw text in setDocAText

Once a pad reaches that state, every subsequent browser load throws on init and the pad becomes effectively unloadable.

The check is a one-liner inside the existing per-op validation loop and reuses the `wireApool` that's already constructed for the author-match validation a few lines above — no extra parsing or new traversal.

## Why now

The standard JS web client always tags inserts with the author attribute (via `Changeset.builder(...).insert(text, [['author', authorId]], pool)`), so legitimate clients are unaffected. Less-careful clients (older code paths, third-party integrations, naive ports) can today send unattributed inserts that the server silently accepts — producing the broken AText state described above. This commit closes that gap server-side.

## Changes

1. `src/node/handler/PadMessageHandler.ts` — single new check at the end of the existing per-op validation loop. Reuses `opAuthorId` already computed there. ~6 lines.
2. `src/tests/backend/specs/messages.ts` — existing USER_CHANGES test fixtures updated to send proper author-attributed inserts (`*0+5` instead of `+5`, plus the matching apool entry). Mirrors what the JS client does in production. A new regression test `'insert without author attribute is rejected'` locks in the new server behaviour.

## Test plan

- [x] Existing USER_CHANGES tests still pass with the updated fixtures
- [x] New regression test asserts `disconnect: badChangeset` on an insert with empty attribs
- [x] Other test files that use unattributed inserts (`ImportEtherpad.ts`, `importexportGetPost.ts`) flow through different code paths (import API, not USER_CHANGES via socket.io) and are unaffected

### DIFF
--- a/src/node/db/Pad.ts
+++ b/src/node/db/Pad.ts
@@ -93,6 +93,17 @@ exports.cleanText = (txt:string): string => txt.replace(/\r\n/g, '\n')
     .replace(/\t/g, '        ');
 
 class Pad {
+  /**
+   * Stable author id used to attribute inserts coming from internal callers
+   * (HTTP API setText/appendText with no authorId, plugins like ep_post_data,
+   * server-side import flows). Without ANY author attribute, pad.atext.text
+   * and pad.atext.attribs drift out of sync — clients then fail
+   * setDocAText reconciliation in ace2_inner.ts when loading the pad. Using
+   * a fixed system author keeps the AText well-formed without requiring
+   * every plugin to allocate its own author up-front.
+   */
+  static readonly SYSTEM_AUTHOR_ID = 'a.etherpad-system';
+
   private db: Database;
   private atext: AText;
   private pool: AttributePool;
@@ -415,9 +426,19 @@ class Pad {
         (!ins && start > 0 && orig[start - 1] === '\n');
     if (!willEndWithNewline) ins += '\n';
     if (ndel === 0 && ins.length === 0) return;
-    const attribs = authorId ? [['author', authorId] as [string, string]] : undefined;
+    // An unattributed insert (empty authorId + non-empty ins) would produce
+    // an AText where `text` and `attribs` disagree on length — clients fail
+    // setDocAText reconciliation on load. Backward-compat fix: if the caller
+    // didn't provide an authorId, attribute the insert to a stable system
+    // author. ep_post_data and other plugins that want named attribution
+    // should still pass an explicit authorId.
+    const effectiveAuthorId =
+        (ins.length > 0 && !authorId) ? Pad.SYSTEM_AUTHOR_ID : authorId;
+    const attribs = effectiveAuthorId
+        ? [['author', effectiveAuthorId] as [string, string]]
+        : undefined;
     const changeset = makeSplice(orig, start, ndel, ins, attribs, this.pool);
-    await this.appendRevision(changeset, authorId);
+    await this.appendRevision(changeset, effectiveAuthorId);
   }
 
   /**
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -453,8 +453,8 @@ importers:
         specifier: ^4.0.5
         version: 4.0.5(eslint@10.4.0)(typescript@6.0.3)
       etherpad-cli-client:
-        specifier: ^4.0.2
-        version: 4.0.2
+        specifier: ^4.0.3
+        version: 4.0.3
       mocha:
         specifier: ^11.7.5
         version: 11.7.5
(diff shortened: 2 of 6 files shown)


JohnMcLear:
Pushed a second commit (631d8c5) that closes the parallel hole through `Pad.spliceText` — which is the path used by `API.setText` / `API.appendText`, the import flow, and plugins like `ep_post_data`.

The first commit only caught socket.io `USER_CHANGES`. But callers reaching `Pad.spliceText` with an empty `authorId` produce the same malformed AText: `attribs` field on `+` ops is `undefined`, so the resulting AText's `text` and `attribs` drift out of sync.

I went with **auto-substitute a stable system author** rather than throw, because rejecting would break:
- ~10 existing tests that call `pad.setText(...)` / `pad.appendText(...)` without an authorId
- `ep_post_data` and any other plugins that call `API.appendText(padId, content)` (the second arg defaults to `''`)

`Pad.SYSTEM_AUTHOR_ID = 'a.etherpad-system'` is now used when internal callers don't name an author. The AText stays well-formed. Plugins/callers that want named attribution should still pass an explicit `authorId` (allocate via `authorManager.createAuthor`).

New test `'spliceText with empty authorId attributes to the system author'` verifies the pool gets the binding.

Happy to split this into a separate PR if you'd prefer to land the socket.io check first.

JohnMcLear:
I need to also check for 
```
Uncaught (in promise) Error: doRepApplyChangeset length mismatch: 53/65
```

JohnMcLear:
Pushed `805766c` extending the PR with a second hardening pass, motivated by a related browser-disconnect vector surfaced by a non-JS client (etherpad-pad terminal editor):

**Problem (related but distinct from the author-attrib vector this PR opens with):**
Etherpad's pad text always ends with `\n`. When a client sends USER_CHANGES whose application strands the trailing `\n` mid-document, `_handleUserChanges` used to silently appended a second `nlChangeset` correction revision to fix the stored pad. But the FIRST NEW_CHANGES broadcast (the malformed user revision itself) reaches browsers BEFORE the correction does, and `applyToAttribution`'s `MergingOpAssembler` asserts `line assembler not finished` on a non-`\n`-terminated doc — the watching browser session drops the changeset and any later edits silently no-op until reload.

Repro from a non-JS client (this is what surfaced it):
- Wire `Z:1h>6|g=1h*0+6$hello ` from a 53-char doc ending in `\n` — Keep covers all 53 chars, Insert lands at position 53, past the existing `\n`.
- Server applies, sees no trailing `\n`, appends correction.
- Browser receives the bad rev first and crashes.

**Fix in 805766c:**
- Replace the silent auto-correction with an explicit reject. Compute `applyToText(rebasedChangeset, prevText)` before `appendRevision`; if the result doesn't end with `\n`, throw → `badChangeset` disconnect. Forces buggy clients to surface their issues in their own logs instead of stranding `\n` in pad history.
- Side-fix for a latent retransmission-detection bug uncovered by this PR's author-attrib changes: `moveOpsToNewPool` renumbers `*N` references to whatever slot the pad pool assigns, which can differ from the wire form's slot. Comparing raw client `changeset === c` then misses retransmissions and duplicates the edit. Snapshot the post-pool-mapping form and compare against `c` instead.
- Backend test `'changeset that would strand the trailing \\n is rejected'` locks in the new behaviour with wire `Z:6>1|1=6*0+1$X` against `hello\n`.
- Updated `handleMessageSecurity` test to capture `roSocket`'s own `authorId` for the apool — the PR's prior commit made `*0` referencing the wrong author a hard reject, which the test was tripping over.

All 1130 backend tests pass locally on Node 25.

Happy to split this into a separate PR if reviewers prefer that — but it's the same shape of bug (malformed-but-passes-checkRep USER_CHANGES → server accepts → browser chokes mid-stream) and the same defensive posture, so leaving it stacked here unless asked to separate.

src/node/db/Pad.ts
```diff
@@ -93,6 +93,17 @@ exports.cleanText = (txt:string): string => txt.replace(/\r\n/g, '\n')
...
+  /**
+   * Stable author id used to attribute inserts coming from internal callers
+   * (HTTP API setText/appendText with no authorId, plugins like ep_post_data,
+   * server-side import flows). Without ANY author attribute, pad.atext.text
+   * and pad.atext.attribs drift out of sync — clients then fail
+   * setDocAText reconciliation in ace2_inner.ts when loading the pad. Using
+   * a fixed system author keeps the AText well-formed without requiring
+   * every plugin to allocate its own author up-front.
+   */
+  static readonly SYSTEM_AUTHOR_ID = 'a.etherpad-system';
```
SamTV12345, line 105: Can that be abused? If I name myself a.etherpad-system as author. Can this have consequences?

SamTV12345 (approved):

src/node/db/Pad.ts
JohnMcLear: patch incoming

PR #7774, opened 2026-05-16
feat(scaling): engine.io socket flush deferral — modest tail-latency reduction at mid-load

JohnMcLear:
## Summary

Follow-up to the closed engine.io WS packing prototype ([#7772](https://github.com/ether/etherpad/pull/7772)). That patch only modified \`transport.send(packets[])\` and almost never fired because engine.io's \`Socket.sendPacket()\` calls \`flush()\` synchronously after each \`writeBuffer.push()\`, and \`flush()\` drains immediately when \`transport.writable === true\` — microseconds on WebSocket. The writeBuffer almost never contained more than one packet.

This patch flips that. \`Socket.prototype.sendPacket\` is re-implemented to push to writeBuffer and then schedule a single coalesced flush via \`queueMicrotask\`. Multiple \`sendPacket()\` calls in the same task all accumulate; the queued microtask drains the whole batch. The transport's \`send([packets])\` call then sees N > 1 packets — that's where engine.io's existing batched-send fast paths (per-packet vs payload encoding) have something to coalesce.

**Microtask deferral adds zero meaningful wall-clock latency** — microtasks drain before the next macrotask, so anything waiting on the next I/O callback or timer still sees the flush completed first. Wire bytes are unchanged.

## Measured impact (N=3, develop baseline vs setting-on)

| Step | develop baseline (min/med/max) | flush-defer (min/med/max) | takeaway |
|---:|---|---|---|
| 100 | 28/38/38 | 37/37/37 | within noise |
| 200 | 30/37/51 | 21/44/49 | within noise |
| 300 | 38/45/71 | 50/53/58 | **tighter max** (71→58) |
| 350 | 39/39/122 | 61/84/110 | **tighter max** (122→110) |
| 400 | 1758/2275/2463 | 1501/2157/2887 | overlapping at cliff |

**Not a cliff-mover.** Both develop and flush-defer cliff at step 400 — that's CPU saturation on the 4-vCPU runner.

**At step 300-350 the tail latency consistently shrinks.** The median may go up or down within noise, but the MAX excursion across 3 runs is reliably smaller with flush-defer on. The mechanism is exactly the predicted one: deferred flush gives more packets per WS frame → fewer syscalls/parser calls → smoother delivery, fewer p95-spiking incidents.

## Caveats

- N=3 is the minimum honest sample size. The noise envelope at this concurrency is large enough that median deltas are inside noise; the tail-tightening signal is what the data actually supports. N=5+ would give tighter conclusions.
- The cliff at step 400 is OS preemption (~7 cores of work on 4 vCPU). No application-level change in this PR's class can move it.
- The patch re-implements engine.io 6.6.5's \`sendPacket\` inline rather than wrapping it, so future engine.io upgrades need re-vetting if they change packet-shape semantics.

## Files

- \`src/node/utils/EngineFlushDeferral.ts\` (new, ~100 lines incl. comments) — installer; idempotent; gated behind \`settings.engineFlushDefer\`.
- \`src/node/hooks/express/socketio.ts\` — conditional \`require()\` before \`new Server(...)\`.
- \`src/node/utils/Settings.ts\` — \`engineFlushDefer: false\` default.

No tests yet; the patch is end-to-end measured via the dive workflow. A unit test that constructs an in-memory engine.io socket and asserts coalesced flush is feasible but high effort for a prototype.

## Recommendation

Merge as a modest mid-load improvement and a useful primitive for future engine.io-side investigations. Production deployments running on 4-vCPU equivalents and hitting 300+ concurrent authors per pad would see modestly fewer tail-latency excursions.


JohnMcLear:
**Addressing Qodo's 2 items** (commit `f0661a9`):

1. **Install guard blocks retries** → Fixed. `installed = true` now runs *after* require + shape validation. Transient require errors leave `installed=false` so a later boot path can retry. New comments call out the intent.

2. **engineFlushDefer undocumented** → Fixed. Added to `settings.json.template` next to `loadTest` with the same prose as `Settings.ts` (including the "wire bytes unchanged" / "no meaningful wall-clock latency" notes so operators see the safety case).

All 33 CI checks pass. The Windows-without-plugins flake from earlier was a post-job cache-save step, unrelated to this PR.

Issue #7778, opened 2026-05-16
a11y: Provide an AT-only read view of pad content (line-by-line + link navigation)


JohnMcLear:
## Background

Follow-up to [#7255](https://github.com/ether/etherpad/issues/7255). Murphy's
2026-05-16 re-test:

> The biggest problem is that you still can't cycle through the text properly
> line by line to press links and such. You can only really get to links if
> you already know where they are and navigate directly to them.

PR [#7777](https://github.com/ether/etherpad/pull/7777) addresses the targeted
toolbar / measurement-node / online-count AT regressions, but this complaint is
structural and needs a separate design effort.

## Root cause

Two structural exposures keep AT from navigating pad content the way it
navigates a regular web page:

1. **Each line renders as `<div class=\"ace-line\">`, not `<p>`** (`src/static/js/domline.ts:64`).
   AT users rely on the `<p>` element for paragraph / line-by-line navigation
   (P key in NVDA/JAWS rotor). With `<div>`, the whole pad is one
   undifferentiated block.

2. **`innerdocbody` carries `role=\"textbox\" aria-multiline=\"true\"`**
   (`src/static/js/ace.ts:299`). This tells AT the editor is a flat editable
   surface — links inside (`<a href>`) often don't surface in the rotor /
   elements list because the screen reader treats the editor as a text input,
   not a document.

Headings already work (ep_headings2 emits `<h1>`–`<h4>` via the
`aceDomLineProcessLineAttributes` hook).

## Why we can't just swap `<div>` → `<p>`

The naïve fix — change `document.createElement('div')` to `'p'` in
`domline.ts` — was scoped and rejected:

- Plugins extensively target line `<div>`s with selectors like
  `iframe.contents().find('div.ace-line')`, `.innerdocbody div`,
  `div[data-author=…]`. Silently switching the tag would break a long tail of
  plugins with no compile-time signal — bugs would trickle in over weeks.
- `<p>` is phrasing content per HTML5 and cannot contain block-level children
  (`<ul>`, `<ol>`, etc.). Existing list lines wrap themselves with
  `<ul><li>…</li></ul>` inside the line node — browser auto-fixup would close
  the `<p>` before the `<ul>`, breaking rendering.
- Even with the per-line list pattern, AT would see N single-item lists
  (one `<ul>` per line) instead of one multi-item list — also wrong.

A flag-gated, opt-in switch with a deprecation cycle is theoretically possible
but is a multi-release rollout, not an a11y bug fix.

[...]

PR #7782, opened 2026-05-16
fix(a11y): drop role=textbox / aria-multiline from innerdocbody (#7778)

JohnMcLear:
## Summary

- Removes `role="textbox"` and `aria-multiline="true"` from `innerdocbody`
- Keeps `aria-label="Pad content"` and `aria-describedby="editor-keyboard-hint"`
- Adds Playwright regression test asserting the attrs stay absent

Lighter alternative to the AT-only read mirror originally sketched in #7778 — pure ARIA strip, no DOM restructuring, no plugin impact.

## Why

Murphy's 2026-05-16 re-test of #7255 said "you still can't cycle through the text properly line by line to press links and such". The narrower fixes in #7777 don't address that — it's the textbox role doing the damage:

- `role="textbox"` + `aria-multiline="true"` pin NVDA/JAWS into **focus mode** for the entire pad.
- In focus mode: arrow keys move the caret one character at a time, P/H/K rotor shortcuts are suppressed, links don't appear in the rotor's links list.
- That matches the reported symptoms exactly.

`contenteditable="true"` alone is enough to tell AT this is editable. Without the textbox role, NVDA/JAWS browse the content as document-mode HTML, restoring line-by-line arrow nav and rotor access to headings + links.

## Acceptance criteria from #7778 — partial

This is the cheap experiment first. If the ARIA strip alone unblocks line / link / heading navigation in real AT, the heavier mirror approach in #7778 isn't needed. If headings still feel flat, we layer `role="paragraph"` onto each `<div class="ace-line">` next.

## Test plan

- [ ] CI: new `innerdocbody does not advertise role=textbox / aria-multiline (#7778)` Playwright test passes
- [ ] CI: existing #7255 regression tests (skip-to-content, line-number sidediv hidden, keyboard-hint text) still pass
- [ ] Local manual: arrow-down through a multi-line pad in NVDA/JAWS/VoiceOver and confirm caret moves line-by-line in browse mode
- [ ] Local manual: open the rotor / elements list and confirm headings (h1-h4 from ep_headings2) + links surface
- [ ] **@StrangeGirlMurph** real-screen-reader retest of #7255 once this is on a build

Refs #7255 #7777

### DIFF
--- a/src/static/js/ace.ts
+++ b/src/static/js/ace.ts
@@ -301,8 +301,12 @@ const Ace2Editor = function () {
     // <body> tag
     innerDocument.body.id = 'innerdocbody';
     innerDocument.body.classList.add('innerdocbody');
-    innerDocument.body.setAttribute('role', 'textbox');
-    innerDocument.body.setAttribute('aria-multiline', 'true');
+    // Deliberately no role="textbox" / aria-multiline: those put NVDA/JAWS
+    // into focus mode (the whole pad becomes one flat edit field), which
+    // hides links and headings from the rotor and suppresses arrow-key
+    // line navigation. contenteditable=true already tells AT this is
+    // editable; without textbox semantics AT can browse the content as a
+    // document. See #7778 / #7255.
     innerDocument.body.setAttribute('aria-label', 'Pad content');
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Waiting user testing.

PR #7784, opened 2026-05-17
harden: assorted server-side tightening for 3.0.2

JohnMcLear:
A bundle of defence-in-depth changes across the server entry points. Each change is small on its own; landing them together keeps the diff cohesive for review and the release notes simple.

## Areas touched

- HTTP API auth path (`APIHandler.ts`) — tightened JWT validation + constant-time API key comparison.
- Import / export tmp file handling — switched to CSPRNG-derived path tokens.
- `/tokenTransfer` flow — added TTL, single-use enforcement, and removed an unnecessary field from the response body.
- `x-proxy-path` header handling — extracted a shared sanitiser used by both the admin static serving and the legacy timeslider redirect; added `Vary` and `Cache-Control: private, no-store` on the admin responses that vary by that header.
- `Pad.appendRevision` — centralised an insert-op invariant the wire handler already enforces so non-wire callers get the same check.

## Tests

- 5 new mocha specs under `tests/backend/specs/`
- 1 new vitest spec under `tests/backend-new/specs/`
- 1 helper added to `tests/backend/common.ts`

20 new test cases, all passing.

## Regression posture

Ran the touched-surface tests (16 spec files) against this branch and unmodified `develop`: identical pass/fail counts. The five failures both branches share are pre-existing on develop — issues filed separately.

## Notes for reviewers

- Detailed per-fix breakdown will accompany the 3.0.2 release notes.
- Behaviour change worth flagging: pads created via internal flows that did not previously attribute the initial revision (e.g. `getPad` with no author, `setHTML` with no `authorId`) now attribute the initial write to the stable `a.etherpad-system` author — same substitution `setText`/`spliceText` already use. Public-facing API surfaces are unchanged.

### DIFF
--- a/src/node/db/API.ts
+++ b/src/node/db/API.ts
@@ -19,10 +19,15 @@
  * limitations under the License.
  */
 
+import AttributeMap from '../../static/js/AttributeMap';
 import {deserializeOps} from '../../static/js/Changeset';
 import ChatMessage from '../../static/js/ChatMessage';
 import {Builder} from "../../static/js/Builder";
 import {Attribute} from "../../static/js/types/Attribute";
+
+// Mirror of `Pad.SYSTEM_AUTHOR_ID`. Inlined to avoid a circular load
+// (API <-> Pad) at module init time.
+const SYSTEM_AUTHOR_ID = 'a.etherpad-system';
 import settings from '../utils/Settings';
 const CustomError = require('../utils/customError');
 const padManager = require('./PadManager');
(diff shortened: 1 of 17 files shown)


JohnMcLear:
Addressed the 4 Qodo findings — force-pushed an amend to the single commit.

- **Bug 1 (legacy restore/copy):** `API.restoreRevision` and `Pad.copyPadWithoutHistory` now merge in an author marker via `AttributeMap` before handing the changeset to `appendRevision`. Caller-supplied `authorId` wins; otherwise falls back to the system author. Added a regression test that copies a pad whose source atext has unattributed inserts (the symmetric `restoreRevision` test is skipped as a TODO — the historical revs need on-disk surgery rather than in-memory poison to exercise that path, and the merge logic is identical to the copyPad case).
- **Bug 2 (ImportHtml loses authorId):** `setPadHTML` now passes `effectiveAuthorId` to both `pad.setText('\n', ...)` and `pad.appendRevision(...)` so `meta.author`, the `padCreate`/`padUpdate` hook context, and the op attribs all reference the same identity.
- **Bug 3 (proxy-path may be relative):** `sanitizeProxyPath` now prepends `/` when the cleaned value is non-empty and doesn't already start with one. Docstring updated. New vitest cases assert `'pad/etherpad' → '/pad/etherpad'` and that double-prefix is avoided.
- **Bug 4 (createdAt type mismatch):** `TokenTransferRequest.createdAt` is now `createdAt?: number` to match the runtime treatment of legacy records.

Regression sweep across 16 backend specs: 207 passing, 13 pending, 5 failing — same 5 pre-existing failures present on unmodified `develop` (filed separately as #7785–#7788).

PR #7798, opened 2026-05-17
fix(admin/pads): apply filter chip server-side, before pagination

JohnMcLear:
## Summary
- thm reported on a 3.1.0 deployment: clicking the **empty pads** chip on the admin pads page only filtered the *current 12-row page slice*, not the full pad universe. Same bug applied to `active` / `recent` / `stale`.
- Cause: `PadPage.tsx:81–88` ran `filteredResults = pads.results.filter(...)` on the already-paginated payload from `padLoad`. Filter chip → pagination order was inverted.
- Fix: push the filter into `PadSearchQuery` and apply it server-side in the `/settings` socket's `padLoad` handler, before `offset/limit`. `total` then reflects the filtered universe and the pagination footer makes sense.

## Server-side refactor
The old handler had a 4-way `if/else if (query.sortBy === …)` that duplicated the hydrate-then-sort loop per sort key. Folded into one pipeline:
1. Pattern filter on names
2. Hydrate metadata for the candidate universe **only if** a non-`all` filter or a non-`padName` sort is requested (preserves the existing fast path for the default browse)
3. Apply filter chip on hydrated rows
4. Sort + slice → `total` becomes the filtered total

## Client-side
- Drop `filteredResults` client-side filter
- Chip click writes `filter` into `searchParams` (so it round-trips through the debounced refetch) and resets `currentPage` to 0
- Older clients that don't send `filter` keep working — server defaults to `all`

## Compatibility
- `PadSearchQuery.filter` is optional. Older clients (e.g. cached admin SPA) emit without it; server treats missing/unknown as `all`. No protocol break.

## Known limitation (out of scope)
The stats cards (`totalUsers` / `activeCount` / `emptyCount`) sum from the current page slice — a pre-existing limitation; making them reflect the global universe needs a separate aggregate endpoint.

## Test plan
- [x] `tsc --noEmit` clean (server + admin)
- [x] New backend spec `tests/backend/specs/admin/padLoadFilter.ts` — 5 passing locally on Node 25:
  - `filter:"empty"` returns only revisionNumber===0 from the full set
  - `filter:"empty"` with `limit=2` still reports `total=5` (the regression thm hit)
  - `filter` omitted → falls back to `all`
  - `filter:"all"` equals no-filter behaviour
  - `filter:"active"` excludes pads with `userCount === 0`
- [ ] CI: full backend + admin frontend playwright
- [ ] Manual: chip click on a >12-pad deployment, verify pagination spans the filtered set

### DIFF
--- a/admin/README.md
+++ b/admin/README.md
@@ -64,3 +64,25 @@ const SettingsPanel = () => {
 The admin endpoints are not yet present in the OpenAPI spec — this client is
 in place to support upcoming work (see issue #7638 follow-up). For now, it is
 exercised only by the smoke test.
+
+## Socket.io: `padLoad` query shape
+
+The admin `/settings` namespace's `padLoad` event accepts a `PadSearchQuery`
+defined in `src/node/types/PadSearchQuery.ts`:
+
+| field        | type                                                              | required | notes |
+| ------------ | ----------------------------------------------------------------- | -------- | ----- |
+| `pattern`    | `string`                                                          | yes      | Substring match on pad name. |
+| `offset`     | `number`                                                          | yes      | Pagination start, in items. Clamped server-side. |
+| `limit`      | `number`                                                          | yes      | Page size. Capped at 12. |
+| `ascending`  | `boolean`                                                         | yes      | Sort direction. |
+| `sortBy`     | `"padName" \| "lastEdited" \| "userCount" \| "revisionNumber"`    | yes      | Column to sort by. |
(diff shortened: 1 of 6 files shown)


JohnMcLear:
Qodo's remaining `Rule violation #2: padLoad filter lacks docs` is stale — the docs section (`Socket.io: padLoad query shape` table with the new `filter` field) was added to `admin/README.md` at L67–88 in commit 5646ea3f. Qodo seems to be re-evaluating the original `[1-66]` line range and not picking up the appended section. Bugs #1 and #3 are correctly marked ✓ Resolved.

Windows-with-Plugins (24) failure is the pre-existing Node 24 + Windows post-suite hard-kill flake — process exits 255 after the last completed test (`cookie identity: same-author second socket kicks the first`, pre-existing) with no test assertion failure. Same flake hit #7797 earlier and passed on rerun. Re-running now.

PR #7800, opened 2026-05-17
test(settings): default updates.tier to "off" so version-badge can't block chat-icon clicks

JohnMcLear:
## Why
v2.7.x added a "severely outdated" version-badge that paints absolutely-positioned at `bottom: 8px; right: 8px; z-index: 9999` when the running core is at least one major behind the latest GitHub release. `#chaticon` sits in the same corner (`bottom: 0; right: 25-30px; z-index: 400`). The banner intercepts pointer events on the chat icon — every Playwright spec that clicks it times out.

This bit ep_comments_page on 2026-05-17 the moment v3.0.0/v3.1.0 were tagged: their Node-22 matrix pins core to v2.7.3 (engineStrict compatibility) and v2.7.3 lets `updates.tier` default to `"notify"`, so the updater polls GitHub, finds 3.x, paints the banner, every chat-related test → 1.5min timeout × 3 retries × N tests. Fixed in [ether/ep_comments_page#421](https://github.com/ether/ep_comments_page/pull/421) with a per-workflow injection.

## What
Set `updates.tier: "off"` in `src/tests/settings.json`. With tier=off, `expressCreateServer` in `node/hooks/express/updateStatus.ts:65` short-circuits and never registers `/api/version-status` — the client fetch 404s and the badge stays hidden.

## Side effects
- Plugins that do the standard `cp src/tests/settings.json settings.json` against any future tag built from this commit get the fix automatically.
- Plugins pinned to v2.7.3 or earlier still need the per-workflow injection pattern (their checked-out test settings predate this commit).
- Existing `pad-version-badge.spec.ts` mocks the endpoint browser-side via `page.route()` — independent of the server-side setting, keeps passing.

## Test plan
- [ ] CI green (backend + frontend + admin)
- [ ] `pad-version-badge.spec.ts` continues to pass (specs mock the endpoint browser-side)


JohnMcLear:
Closing this in favor of another approach https://github.com/ether/etherpad/issues/7799

PR #7804, opened 2026-05-18
fix(pad): redesign outdated-version notice (#7799)

### LINKED ISSUE #7799: Inappropriate update message
A message "Etherpad on this server is severely outdated. Tell your admin." is being displayed on the user front end for a server that is only 1 version behind current.

Aside from being inaccurate, it's inappropriate to inform users of a site about maintenance tasks that they don't understand or have context to resolve. It wasting users time by having them try and contact me, and its wasting my time by having to respond. As a server admin it is my job to maintain the server, its not the job of site users.

This alert message should be disabled by default.

JohnMcLear:
## Summary

Closes #7799.

Replaces the persistent, all-visitor "Etherpad on this server is severely outdated" banner with a dismissable gritter toast, shown only to a pad's first author (pool position 0 in the attribute pool), only when the running server is at least one minor version behind the latest published release.

### Behaviour changes
- The persistent `#version-badge` bottom-right banner is gone. A gritter (auto-fades after 8s, X-dismissable) takes its place.
- The notice fires only on minor-or-more deltas. Patch-only deltas (e.g. `3.0.1 → 3.0.2`) no longer fire.
- Non-creators of a pad never see the notice.
- `updates.tier = 'off'` remains the full kill-switch and is unchanged.

### API changes
- `GET /api/version-status` now takes an optional `?padId=<id>` query parameter and returns `{outdated: 'minor' | null, isFirstAuthor: boolean}`.
- The `severe` and `vulnerable` enum values, the `vulnerable-below` directive scraping, the `vulnerableBelow` state field, and the `vulnerable`/`vulnerable-new-release` email kinds have all been removed.
- `isSevere` (the Notifier signal that triggers the "outdated" admin email) now fires on minor-or-more-behind, matching the new gritter signal. The email copy is updated to match.

### Implementation notes
- The route resolves the requester's author via the express session cookie (`resolveRequestAuthor`) and only computes the per-pad first-author match when both `state.latest` exists AND the current version is minor-or-more behind. The `padId` is validated via `padManager.isValidPadId` before any DB call.
- Results are cached per `(padId, authorId)` for 60s in a 1000-entry LRU with single-flight on misses.

Design spec + implementation plan committed under `docs/superpowers/specs` / `docs/superpowers/plans`.

## Test plan

- [x] Backend vitest suite green (702 / 703; the one pre-existing failure is the unrelated `backend-tests-glob.test.ts` which expects `../node_modules/ep_*/static/tests/backend/specs` to match — a fresh-clone harness limitation, present on develop)
- [x] New backend coverage: `tests/backend-new/specs/updater/versionCompare.test.ts` (semver helper, 18 cases), `tests/backend-new/specs/hooks/express/firstAuthorOf.test.ts` (6 cases), `tests/backend-new/specs/hooks/express/updateStatus.test.ts` (9 cases including LRU eviction, single-flight, first-author gating)
- [x] New Playwright coverage: `src/tests/frontend-new/specs/outdated_notice.spec.ts` (6 cases incl. auto-fade, X-dismiss, server-500, client-side `isFirstAuthor=false` guard)
- [x] `tsc --noEmit` clean
- [ ] Manual: dev server with `var/update.state.json` pinned to a higher `latest.version` — verify gritter appears once for the pad's first author, absent for second visitor


JohnMcLear:
Qodo follow-up — addressed in #7805:

- 🐞 **Bug: author lookup always null** — confirmed real bug. `resolveRequestAuthor` read `req.session.user.author`, which Etherpad never populates for pad visitors. Fixed in #7805 by switching to the same token-cookie path Etherpad's own socket.io handshake uses (`authorManager.getAuthorId(token, user)`).
- 📘 **Public route in admin spec** — clarified the admin spec's `info.description` to note pad-side endpoints are included for completeness. Larger structural split deferred.
- ⚠️ **Rule violation: vulnerable-below removed without deprecation WARN** — intentional. Per #7799 the maintainer asked to "drop it"; the directive was an internal release-notes scrape (no plugin or external system consumed it), so the deprecation rule doesn't usefully apply here.

PR #7826, opened 2026-05-20
feat(admin): explain env-var substitution in /settings, surface auth errors (#7819)

JohnMcLear:
## Summary

Three small admin /settings UX improvements driven by [#7819](https://github.com/ether/etherpad/issues/7819), where a Docker operator saved a plugin block in the raw view and reported it "disappeared." After diagnosing the issue, root cause turned out to be a mix of environmental factors (no volume mount on settings.json, env-var substitution misunderstanding) — but the Etherpad UI was confusing enough that *somebody* will hit this again. These changes close the gaps without touching the UX of non-container installs.

- **Banner** above the editor explaining the template / env-var substitution model. Only rendered when the loaded file contains a `\${VAR}` placeholder, so non-container installs see no change.
- **Effective tab** in the mode toggle — read-only view of the in-memory settings the server is actually using (with secrets redacted). The backend was already emitting `resolved` alongside every `load`; the SPA now exposes it. Same gate: only appears when `\${VAR}` is in the file.
- **\`admin_auth_error\` event** from the /settings socket. The connection handler previously silently returned when the session wasn't admin — making misrouted Traefik+SSO auth look like "save did nothing" with no error path. Now emits a dedicated event the SPA surfaces as a toast before dropping the socket.

No behaviour change for users who don't use env-var placeholders in their settings.json.

## Screenshots / behaviour

With `\${VAR}` in the file (Docker default):
- Top of /settings: blue note banner: "This file is a template, not the live config. Placeholders like \${VAR:default} are substituted into memory at startup; they are never written back to this file…"
- Mode toggle: Form | Raw | Effective
- Effective tab: read-only textarea showing the live in-memory config with passwords/secrets as \`[REDACTED]\`.

Without any \`\${VAR}\` placeholders:
- No banner, no Effective tab. Identical to today's UI.

## Test plan

- [x] \`pnpm run build\` (admin SPA) passes
- [x] \`pnpm run ts-check\` (src) passes
- [x] \`adminSettingsAuthError.ts\` — new spec proves the auth_error/disconnect contract fires for non-admin sockets
- [x] \`adminSettingsSave.ts\` — existing save tests still pass (no save-path change)
- [x] \`adminSettingsResolved.ts\` — existing resolved tests still pass (no load-path change)
- [x] \`adminsettings.spec.ts\` — new Playwright test asserts banner + Effective tab only appear after \`\${VAR}\` is added, and Effective view is readonly + shows [REDACTED]

Backend mocha output:
\`\`\`
18 passing (2s)
\`\`\`


JohnMcLear:
Thanks Qodo — first one (#1) is intentional, second (#2) is already addressed; quick notes on each:

**#1 — Effective tab / banner without dedicated feature flag**

These two additions are read-only informational UX that don't change Etherpad's runtime behaviour at all — no save semantics changed, no load semantics changed, no API surface added that wasn't already there. The backend already emitted `resolved` alongside every `load` (since #7670); this PR just exposes that existing data in a non-editable view.

The activation heuristic — render only when `${VAR}` placeholders are present in `settings.json` — is itself a form of soft gating: operators who don't use env-var substitution (the overwhelming majority of non-container installs) never see the banner or the Effective tab, so for them the UX is byte-identical to today.

That said, happy to add a `settings.adminUI.envVarHints` flag if @JohnMcLear prefers an explicit toggle — defaulting `true` (so the docker audience this PR exists for actually sees the UX) with the option to disable. Pinging you to weigh in.

**#2 — Auth error reconnect loop**

Already fixed in 5c6b164 (commit pushed before this review landed) — the SPA now tracks an `authErrored` flag that is reset on each successful connect and suppresses the auto-reconnect once an `admin_auth_error` has fired. Qodo correctly auto-marked this resolved in the same review.

JohnMcLear:
cc @SamTV12345

SamTV12345 (approved):

SamTV12345:
So far I checked the code and couldn't find any issues. Nice :)

Issue #981, opened 2012-09-05
Require script could not be embedded on windows [develop]


marcelklehr:
Git revision: 3cbd59c (develop)

Steps to reproduce:
1. Start etherpad lite
2. open `127.0.0.1:9001/` with you a browser of your choice
3. open any pad

In Firefox:

```
ErrorId: znrlBH7Tfzvy3RwDM6AG
UserAgent: Mozilla/5.0 (Windows NT 6.1; rv:15.0) Gecko/20100101 Firefox/15.0
Error: Require script could not be embedded. in http://127.0.0.1:9001/javascripts/lib/ep_etherpad-lite/static/js/pad.js?callback=require.define at line 2823
```

In IE9:

```
SCRIPT5009: "require" is undefined
testpad, Line 1 Column 1
```

In Opera:

```
ErrorId: 7OV8BLtKCWf8jUXU1XVh
UserAgent: Opera/9.80 (Windows NT 6.1; U; en) Presto/2.10.289 Version/12.02
Uncaught exception: Error: Require script could not be embedded. in http://127.0.0.1:9001/javascripts/lib/ep_etherpad-lite/static/js/pad.js?callback=require.define at line 2823
```

PR #7831, opened 2026-05-22
fix: page sessionstorage cleanup to avoid OOM (#7830)

### LINKED ISSUE #7830: Excessive Memory Use by Session Cleanup Leading to OOM
**Describe the bug**
After upgrading to to Etherpad 2.7.3 we noticed our etherpad instance would consistently hit `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory` and crash after about 15 minutes. After digging through some heap snapshots I'm reasonably confident that the source of the problem is session cleanup which begins on startup and was introduced in https://github.com/ether/etherpad/commit/da9f5ac4eed750af2448c42d7973b9b0b243fd54. I have since set sessionCleanup to false and memory usage is consistently low and the process has managed to live for over 15 minutes after this config change.

**To Reproduce**
1. Run etherpad for many years (I think our DB is over a decade old) to accumulate many `sessionstorage:.*` database records
2. Upgrade to 2.7.3 and set sessionCleanup to true (the default)
3. Watch memory usage climb until node and etherpad crash due to reaching the heap limit

I realize that run etherpad for a decade isn't a great reproducer, but I'm not super familiar with the database layout or how to artificially create session records. In theory this is possible and should reproduce things though.

We run etherpad without authentication (not sure if this impacts session handling in the DB). I suspect that we're affected by https://github.com/ether/etherpad/issues/5010 but then the new cleanup routine attempts to load all of these many records into memory at once and then we run out of memory.

**Expected behavior**
Session cleanup should use a reasonable amount of memory. Possibly by paging through the session records rather than loading them all at once. The new cleanup routine does `const keys = await DB.findKeys('sessionstorage:*', null);` maybe we can set limits on the number of records returned at one time and work our way through a reasonable batch size before proceeding to the next batch?

 - Etherpad version: 2.7.3
 - OS: Debian Trixie
 - Node.js version (`node --version`): nodejs_version_info{version="v24.15.0",major="24",minor="15",patch="0"}
 - pnpm version (`pnpm --version`): 11.0.6 (note the bug template asks for npm version but we're using pnpm to match the upstream docker container)
 - Is the server free of plugins: No we have ep_headings2 installed
 - Are you using any abstraction IE docker? Yes, we build our own images using node:24-trixie-slim as a base but try to follow the approach used by the upstream Dockerfile.

**Additional context**
Our database is a mariadb 10.11 database. Not sure if that makes a difference in how the internal memory for query results is structured.

Also, the way I tracked this down was to run etherpad with `NODE_OPTIONS: "--heapsnapshot-signal=SIGUSR2"` then I grabbed a heap snapshot shortly after startup then again about 2 minutes later. Viewing these snapshots in Chrome's developer tools I was able to see that this chain of objects was retaining a significant amount of memory compared to the initial startup: `_pool in mysql_db_default -> _allConnections in Pool -> _list in denque -> [$INDEX] in Array -> _command in PoolConnection -> _rows in Query -> [$INDEX] in Array -> [$INDEX] in Array -> key in {key}`. Those key values appear to be `sessionstorage:.*` records. Each of them is listed as 0.1kB in size, but in aggregate having many of them is significant enough to run out of heap space.

We are currently running with sessionCleanup set to false. I figure that is no worse than we were before the upgrade, but it doesn't OOM.

JohnMcLear:
## Summary
- Switch `SessionStore._cleanup()` from `findKeys('sessionstorage:*', null)` (loads everything) to `findKeysPaged` with a 500-key page size
- Yield to the event loop between pages so the DB driver can release the previous page's buffered rows and request handlers can interleave
- Bump `ueberdb2` to `^6.1.0` for the new method (see ether/ueberDB#981)
- Add regression test that seeds 50 sessionstorage rows, forces a 4-key page, and asserts cleanup walks the full keyspace

## Why
Per the reporter on #7830, after 2.7.3 their decade-old MariaDB install OOMed within ~15 minutes. Heap snapshots pinned retention on `_pool → _allConnections → … → _command._rows → keys` — the unbounded `findKeys` result tied up `mysql2` PoolConnection row buffers (and dominated JS heap with millions of session keys). Paging at the DB layer turns that into bounded N-row reads.

Cleanup termination is now driven by an empty page rather than `page.length < CLEANUP_PAGE_SIZE`, so the loop still iterates a stubbed/throttled paged source correctly (caught by the new test on the first run).

## Blocked on
- ether/ueberDB#981 — must publish ueberdb2 6.1.0 to npm before this PR's lockfile resolves.

## Test plan
- [x] `pnpm --filter ep_etherpad-lite run ts-check` clean
- [x] `mocha tests/backend/specs/SessionStore.ts` — 29 passing, including the new \"pages across a large sessionstorage keyspace\" case (\"removed 25 expired/stale sessions out of 50\" confirms cross-page iteration)
- [ ] CI green once ueberdb2 6.1.0 publishes

## Notes for the maintainer
- `findKeysPaged` falls back to JS-side slicing for backends that don't natively support it (sqlite/rusty/redis/mongo/...). Correctness preserved; OOM-resistance for those backends is a follow-up.
- The same unbounded `findKeys` pattern exists in `PadManager.ts:77` and `AuthorManager.ts:{361,369,484,495}`. Out of scope for this fix but worth tracking — happy to file as a follow-up issue.

Closes #7830


JohnMcLear:
Addressed Qodo's four findings in 63c2493b:

1. **Missing DB method hidden** — `DB.init()` now throws fast on any missing wrapper method (no more silent skip).
2. **Unbounded cleanup** — single `_cleanup()` run capped at 10 minutes; warning logged on exhaustion, next scheduled run picks up.
3. **Silent cursor-stall break** — defensive `page[0] <= after` guard now logs an error with both `after` and `page[0]`.
4. **Test cleanup symmetry** — finally block now removes both expiredSids and validSids.

29 SessionStore backend tests still green; ts-check clean. (Wider CI still blocked on the ueberdb2 6.1.0 npm publish.)

JohnMcLear:
Permanent fix path is now ether/ueberDB#983 — migrates the ueberDB publish workflow from a stored `NPM_TOKEN` (the credential that just expired and broke #981's auto-publish) to npm OIDC trusted publishing. No expiring secret on our side, signed provenance attached. One-time setup on https://www.npmjs.com/package/ueberdb2/access required after merge — full recipe in the #983 description and at the top of the publish-npm job.

Once #983 is merged and the npm UI step is done, the next push to ueberDB main publishes `ueberdb2@6.1.2` via OIDC. I'll then regenerate this PR's lockfile against `^6.1.0` and CI goes green for good. (#982 closed — that was the git-ref stopgap and is no longer needed.)

PR #7837, opened 2026-05-25
fix(oidc): fix OIDCAdapter broken flows

enlightened88:
Fixed critical bugs in MemoryAdapter that broke grant revocation, device flow, and token consumption, missing grantId/userCode indexes in upsert, unsafe null …Fixed critical bugs in MemoryAdapter that broke grant revocation, device flow, and token consumption, missing grantId/userCode indexes in upsert, unsafe null assertions in destroy and consume, and a stale userCode mapping leak.

### DIFF
--- a/src/node/security/OIDCAdapter.ts
+++ b/src/node/security/OIDCAdapter.ts
@@ -1,58 +1,52 @@
 import {LRUCache} from 'lru-cache';
 import type {Adapter, AdapterPayload} from "oidc-provider";
 
-
 const options = {
   max: 500,
-  sizeCalculation: (item:any, key:any) => {
-    return 1
-  },
-  // for use with tracking overall storage size
+  sizeCalculation: (item: any, key: any) => 1,
   maxSize: 5000,
-
-  // how long to live in ms
   ttl: 1000 * 60 * 5,
-
-  // return stale items before removing from cache?
   allowStale: false,
-
   updateAgeOnGet: false,
   updateAgeOnHas: false,
 }
 
 const epochTime = (date = Date.now()) => Math.floor(date / 1000);
-
-const storage = new LRUCache<string,AdapterPayload|string[]|string>(options);
+const storage = new LRUCache<string, AdapterPayload | string[]>(options);
 
 function grantKeyFor(id: string) {
   return `grant:${id}`;
 }
-
-function userCodeKeyFor(userCode:string) {
+function userCodeKeyFor(userCode: string) {
   return `userCode:${userCode}`;
 }
 
-class MemoryAdapter implements Adapter{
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Thanks for picking this up — the core fix is correct and overdue. The pre-fix `revokeByGrantId` was effectively a no-op (no grant index was ever written), meaning any "log out everywhere" / consent revocation flowing through this adapter silently did nothing. Device flow was similarly broken (`findByUserCode` had nothing to look up).

### Strong points
- Writing the `grant:<grantId>` and `userCode:<userCode>` indexes in `upsert` is the load-bearing fix.
- Widening the `upsert` payload type to `AdapterPayload` is correct — the prior hand-written subset excluded `grantId`, `userCode`, `consumed`, etc.
- `destroy` now cleans the userCode mapping and tolerates a missing grant list.
- No more `!` non-null assertions throwing on missing entries.

### Requested changes before merge

**1. `consume` resets the TTL to the cache default (5 min).**
```ts
storage.set(key, payload);   // no ttl arg → falls back to constructor ttl
```
This silently overrides the `expiresIn * 1000` TTL set by `upsert`. Either drop the `storage.set` entirely (LRUCache holds by reference, so the in-place `payload.consumed = epochTime()` is already persisted), or preserve the original expiry, e.g. `storage.set(key, payload, {ttl: storage.getRemainingTTL(key)})`.

**2. Add regression tests.**
The four bugs being fixed are exactly the kind that silently regress. Even a small spec covering:
- `upsert` → `revokeByGrantId` actually clears tokens
- `upsert` → `findByUserCode` returns the payload
- `destroy` clears the userCode mapping
- `consume` on a missing id doesn't throw

…would lock these in. Given this is OIDC, I'd really like to see this before merge.

### Notes (not blockers)
- The grant list and userCode index are written without an explicit TTL, so they fall back to the 5-min default while the underlying token entry uses `expiresIn * 1000`. Matches the upstream `panva/node-oidc-provider` reference adapter, but worth a one-line comment noting the trade-off — a late `revokeByGrantId` can miss tokens whose grant index has already expired.
- `findByUid` is still an O(n) scan over the whole cache — not in scope here, but a `uid:<uid>` secondary index (mirroring what this PR just added for `grantId`) would be a worthwhile follow-up.
- Cosmetic whitespace / type-annotation cleanup is bundled with the fix. Fine, just noting it.

Nice catch on a real security-relevant bug. Approving in spirit pending the `consume` TTL nit and at least minimal test coverage.

SamTV12345:
Thanks for the mr. That really fixed an issue as described by John ^^

enlightened88:
> Thanks for the mr. That really fixed an issue as described by John ^^

I am glad that I was useful to you

PR #7841, opened 2026-05-25
feat(settings): default settings.enablePluginPadOptions to true

### LINKED ISSUE #422: "pad-wide settings unavailable" message on console
Console message

```
[ep_plugin_helpers.padToggle ep_comments_page] pad-wide settings unavailable — settings.enablePluginPadOptions is false — set to true in settings.json to enable pad-wide options. Per-user cookie toggle still works.
```

As discussed on Discord, separate issues for each affected plugin.

JohnMcLear:
## Summary
- Flip `settings.enablePluginPadOptions` default from `false` → `true`. Matches the original intent of shipping the `ep_*` passthrough in 3.0.0 (PR #7698): let plugins like `ep_plugin_helpers`' `padToggle` / `padSelect` ride the existing `padoptions` broadcast/persist rail out of the box.
- Updates `Settings.ts`, `settings.json.template` (env-var default), `PluginCapabilities.ts` comment, and `doc/plugins.md` (now describes the flag as opt-out instead of opt-in).
- Backend test describe-blocks relabeled — `true` is now `(default)`, `false` is now `(operator opt-out)`. Same matrix; both branches still pass.

## Why now
- ether/ep_comments_page#422 (and sibling per-plugin reports discussed on Discord) — stock 3.x deployments `console.warn` on every pad load because `ep_plugin_helpers` detects `clientVars.enablePluginPadOptions === false` and tells the admin to flip it. With the flag default-true, the warning stops firing on fresh installs while still surfacing for operators who have explicitly opted out.
- Plugins on `ep_plugin_helpers >= 0.6` already expect pad-wide options to work; default-false silently no-op'd `pad.changePadOption('ep_*', …)` and made the helper UI inert.

## Compat
- Deployments with an explicit `"enablePluginPadOptions": false` in `settings.json` keep that value — no migration.
- `clientVars` shape is unchanged; older clients only read `enablePluginPadOptions` and keep working.

## Follow-up
- ether/ep_plugin_helpers: README and warning comments still describe the flag as opt-in. Separate PR queued.

Closes ether/ep_comments_page#422 (warning suppression for stock deployments).

## Test plan
- [x] `mocha tests/backend/specs/Pad.ts` — 31 passing, both `(default)` and `(operator opt-out)` branches green
- [ ] Full backend suite via CI
- [ ] Smoke a fresh `settings.json` (env-var path) to confirm `clientVars.enablePluginPadOptions === true` on default deploy

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -267,15 +267,16 @@ below.
 ### Runtime flag
 
 The passthrough is gated by `settings.enablePluginPadOptions`, default
-`false`. Operators must opt in via `settings.json`:
+`true`. Operators who want to lock plugins out of pad-wide state can flip
+it in `settings.json`:
 
 ```json
 {
-  "enablePluginPadOptions": true
+  "enablePluginPadOptions": false
 }
 ```
 
-When enabled, the server reflects the value to every client via
+When enabled (the default), the server reflects the value to every client via
 `clientVars.enablePluginPadOptions` so plugins can detect both *capable*
 (static) and *active* (per-pad request) at the same point.
 
--- a/settings.json.template
+++ b/settings.json.template
@@ -853,11 +853,11 @@
     * accepting pad-wide values under plugin-namespaced keys matching
     * /^ep_[a-z0-9_]+$/ (e.g. ep_table_of_contents). Values are validated
     * (JSON-safe, 64 KB per key, 256 KB total) and broadcast to every
-    * connected client just like native pad-wide toggles. Disabled by
-    * default; flip to true once your plugins (e.g. ep_plugin_helpers'
-    * padToggle) require it. See doc/plugins.md.
+    * connected client just like native pad-wide toggles. Enabled by
+    * default; set to false to lock plugins out of pad-wide state. See
+    * doc/plugins.md.
     **/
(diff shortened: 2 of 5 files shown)


JohnMcLear:
**CI note**: the two Windows backend test failures (\`Windows with Plugins (24)\` and \`Windows without plugins (24)\`) are the pre-existing silent-ELIFECYCLE flake on Windows runners — each run dies at a different test file (this one died inside \`messages.ts USER_CHANGES\` and \`socketio.ts Pad-wide settings creator gate\`, neither of which touches \`enablePluginPadOptions\`). The develop tip itself (sha \`d9dabe35\`) fails \`Windows without plugins (24)\` the same way 30 min before this PR was opened.

Every Linux variant (with + without plugins, plus 24-only matrix), Playwright Chrome ×2, Playwright Chrome with plugins, Playwright Firefox, Wrapper unit, perform type check, build, build-test, build-test-local-plugin, test, Analyze, CodeQL, Build snap (destructive-mode), and dependency-review all pass.

Heartbeat instrumentation to surface the silent Windows failure root cause is queued in #7838.

SamTV12345 (approved):

PR #7843, opened 2026-05-25
fix(pad): URL view-option params lost to padeditor.init race (#7840)

### LINKED ISSUE #7840: iframe showLineNumbers=false ignored
I am running etherpad 3.2 in docker and embedding the pad using an iframe. The URL for the iframe is `https://<domain>/p/<pad_name>?showChat=false&showLineNumbers=false&showControls=false&userName=<username>`. Etherpad is running behind a nginx reverse proxy.

No matter what I set `showLineNumbers` to, I still get line numbering in the pad. I have also set `PAD_OPTIONS_SHOW_LINE_NUMBERS: "false"` in my docker-compose.yml.

Clearing Cache and using other browsers didn't change anything.

Any ideas would be appreciated.

JohnMcLear:
## Summary

Closes #7840. `?showLineNumbers=false` (and `?useMonospaceFont=true`) on an iframe-embedded pad get re-applied to the server default a few hundred ms after pad boot. Same race that affected `?rtl=false` before #7464 — generalises that fix to the neighbouring URL view-toggles.

## Why the iframe reporter sees it and direct users don't

The race always fires, but it only becomes *observable* when the value `padeditor.init`'s deferred `setViewOptions(initialViewOptions)` call re-applies disagrees with the URL param. `initialViewOptions` is built from `cookie pref ?? server default`:

- Direct-browser users typically have a `prefs` cookie from a prior in-pad toggle that matches what they're asking for via URL → no observable race.
- Cross-context iframe embeds with no cookie → server default `showLineNumbers:true` fights the URL `false` → URL preference visibly lost.

## What the race looks like

1. `_afterHandshake` → `getParams()` sets `settings.LineNumbersDisabled = true`.
2. `_afterHandshake` calls `padeditor.init(view).then(postAceInit)` — async; ace iframes still loading.
3. Sync tail of `_afterHandshake` hits the URL-param block: `changeViewOption('showLineNumbers', false)` → queues `setProperty('showslinenumbers', false)` in `Ace2Editor.actionsPendingInit` (loaded=false).
4. `ace.init` resolves → `loaded=true` → queue flushes → URL-driven `false` applied.
5. `padeditor.init` resumes past its own `await` and runs `self.setViewOptions(initialViewOptions)` — `initialViewOptions.showLineNumbers === true` (server default) → `setProperty('showslinenumbers', true)` runs against `loaded=true` and immediately re-shows the gutter.

#7464 spotted this for RTL and moved the override into `postAceInit`. The neighbouring blocks for `showLineNumbers` / `noColors` / `useMonospaceFontGlobal` at the same synchronous-tail site never got the same treatment. This PR generalises that fix.

## Test plan

- [x] New spec \`src/tests/frontend-new/specs/url_view_options.spec.ts\`:
  - \`?showLineNumbers=false\` → outer body has \`.line-numbers-hidden\` on first paint
  - \`?showLineNumbers=true\` no-regression
  - \`?useMonospaceFont=true\` → \`#viewfontmenu\` value is \`RobotoMono\` on first paint
- [x] Verified the three race cases fail on \`develop\` before the fix.
- [x] Verified \`rtl_url_param.spec.ts\`, \`timeslider_line_numbers.spec.ts\`, \`pad_settings.spec.ts\`, \`embed_value.spec.ts\`, \`hide_menu_right.spec.ts\` still pass (26/26 green together).

## Not in scope

- The \`?noColors=true\` URL path turns out to be unaffected for the inner-body \`authorColors\` class (the override at \`pad_editor.ts:268-270\` always wins). The colorscheck checkbox UI doesn't reflect \`settings.noColors\` in either the racy or fixed state — that's a pre-existing cosmetic mismatch independent of this race and worth its own follow-up.

### DIFF
--- a/src/static/js/pad.ts
+++ b/src/static/js/pad.ts
@@ -720,9 +720,26 @@ const pad = {
       pad.refreshPadSettingsControls();
       pad.applyOptionsChange();
       pad.refreshMyViewControls();
+      // The following view overrides MUST run inside postAceInit (after
+      // padeditor.init resolves), not in the synchronous tail of
+      // _afterHandshake. Running them sync queues a setProperty in the
+      // Ace2Editor pending-init queue; the queue is flushed when ace
+      // finishes loading, but padeditor.init's own
+      // setViewOptions(initialViewOptions) call runs immediately *after*
+      // that flush and clobbers the URL-driven value. See #7464 for the
+      // RTL incarnation of this race (now generalised here for #7840).
       if (settings.rtlIsExplicit) {
         // URL or server config explicitly set RTL — takes priority over cookie
         pad.changeViewOption('rtlIsTrue', settings.rtlIsTrue === true);
+      }
+      if (settings.LineNumbersDisabled === true) {
+        pad.changeViewOption('showLineNumbers', false);
+      }
+      if (settings.noColors === true) {
+        pad.changeViewOption('noColors', true);
+      }
+      if (settings.useMonospaceFontGlobal === true) {
+        pad.changeViewOption('padFontFamily', 'RobotoMono');
       }
 
       // Prevent sticky chat or chat and users to be checked for mobiles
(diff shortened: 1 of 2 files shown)


SamTV12345 (approved):

SamTV12345:
Makes sense that it is now firing correctly.

PR #7852, opened 2026-05-25
fix(test): enable HTTP keep-alive on the global agent for backend tests

JohnMcLear:
## Summary

One-line behaviour change in `src/tests/backend/common.ts`: set `http.globalAgent.keepAlive = true` at module load. Targeted at the Windows backend-test silent-ELIFECYCLE flake we've been chasing across #7838, #7842, #7846.

## Smoking gun from PR #7846

The OS-level netstat sidecar in #7846 captured `run 26419467467 Win-no-plugins`. localhost server-side TIME_WAIT count on the Etherpad listening port `[::1]:50398`:

```
21:02:18.337       56
21:02:20.765      103
21:02:23.204      136
21:02:25.662      169
21:02:28.146      201
21:02:30.638      228   ← silent kill 37 ms later
```

~14 server-active-close TIME_WAITs per second, growing linearly. ESTABLISHED count is ~zero because each supertest request opens-and-closes a fresh TCP. The SERVER is the active closer (sends FIN after response), so the SERVER side enters TIME_WAIT — for ~120 s on Windows default `TcpTimedWaitDelay`.

228 half-dead sockets isn't enough to hit the Windows free-TCB cap (~2000), but it forces libuv's IOCP socket-table scan to walk a much larger working set on every TCP completion. The kill cluster is concentrated on tests that do rapid sequential HTTP roundtrips — `importexportGetPost.ts > Import authorization checks`, `pad.ts > Tests`, `import.ts > DOCX export -> import round-trip`, all of which match this profile.

## The fix

`http.globalAgent.keepAlive = true` makes supertest's underlying `http.request` reuse a single TCP connection for sequential requests to the same origin. Each test no longer leaves a TIME_WAIT socket behind. Churn collapses from ~14/s to ~0.

Linux is unaffected by the flake because its TCP recycling and kernel half-dead socket handling are far more permissive. This change is a strict improvement there too (less socket churn = less work overall).

## What this does NOT do

- Doesn't touch `diagnostics.ts` — the heartbeat / boundary / mid-test reports stay exactly as merged in #7838 + #7842. If this fix doesn't fully resolve the flake, we still capture death data.
- Doesn't touch the OS sidecar (#7846 is a separate PR, still in review).
- Doesn't change any production code path — tests-only.

## Test plan

- [ ] Existing backend matrix passes (Linux ± plugins, Windows ± plugins). No tests assert on per-request connect/disconnect cycles, but the change is observable.
- [ ] On the Win+plugins matrix, rerun the workflow 5-10x. Pre-fix flake rate was ~22%. Post-fix expectation: substantially lower (ideally near zero).
- [ ] If #7846 is also landed, netstat.log on a failure should show TIME_WAIT count flat at 0-5 instead of climbing to 200+.

### DIFF
--- a/src/tests/backend/common.ts
+++ b/src/tests/backend/common.ts
@@ -16,7 +16,32 @@ import TestAgent from "supertest/lib/agent";
 import {Http2Server} from "node:http2";
 import {SignJWT} from "jose";
 import {privateKeyExported} from "../../node/security/OAuth2Provider";
+import * as http from 'node:http';
 const webaccess = require('../../node/hooks/express/webaccess');
+
+// Enable HTTP keep-alive on the global agent for the test process. Without
+// this, every supertest request opens a fresh TCP connection and the server
+// closes it on response — the server side then enters TIME_WAIT for the
+// default Windows TcpTimedWaitDelay (~120 s) before the ephemeral port is
+// freed.
+//
+// The Windows backend-test job's OS-level netstat sidecar (PR #7846)
+// captured the smoking gun for the silent-ELIFECYCLE flake in run
+// 26419467467: localhost server-side TIME_WAIT counts on the Etherpad
+// listening port climbed linearly at ~14/s, reaching 228 active TIME_WAIT
+// entries on `[::1]:50398` 37 ms before the kill — all server-active-close
+// half-dead sockets, all from rapid sequential supertest requests with no
+// connection reuse. The kill cluster on Windows + Node 24 + plugins
+// correlates tightly with this TIME_WAIT accumulation: it gives libuv a
+// large pool of half-dead handles to walk on every IOCP completion.
+//
+// Setting keepAlive=true on http.globalAgent makes supertest's underlying
+// http.request reuse a single TCP connection for sequential requests to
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Closing — the TIME_WAIT smoking gun captured in #7846 (228 server-side TIME_WAITs accumulating at ~14/s by the kill moment) led me to claim this would fix the flake. It didn't. The follow-up run on this PR (job 77776265356, Win-no-plugins) still hit a silent ELIFECYCLE on `pad.ts > Tests > deletes a Pad`, exit 255, same fingerprint as the prior 11 captures. The node-report TCP handle count IS now 1 across every snapshot (was 2-3 pre-fix), so the keep-alive change took effect — supertest is reusing connections and TIME_WAIT churn is collapsed. The flake just isn't caused by TIME_WAIT.

The connection-reuse change might still be a small improvement in isolation, but I'm not going to ship a "fix" that isn't one. Open to reopening as `chore(test): reuse HTTP connections in backend tests` if the test-infra value alone is worth it — that's a separate decision from the flake investigation.

PR #7853, opened 2026-05-26
perf: room-broadcast NEW_CHANGES fan-out for steady-state clients behind feature flag (#7780)

### LINKED ISSUE #7780: Performance: [#7756 follow-up] Room-broadcast NEW_CHANGES fan-out to drop ~5-7% per-recipient packet construction
## Background

CPU profiling against develop (and against #7775 + #7776 stacked, which removes the upstream log4js cost) attributes ~10% of total process CPU at the dive cliff to socket.io's per-recipient packet construction chain inside `PadMessageHandler.updatePadClients`:

```
3.36% emit                [socket.io/dist/socket.js]
3.56% packet              [socket.io/dist/socket.js]
3.31% _packet             [socket.io/dist/client.js]
2.49% writeToEngine       [socket.io/dist/client.js]
2.23% sendPacket          [engine.io/build/socket.js]
```

The fan-out loop today does `socket.emit('message', msg)` once per recipient per missed revision:

```ts
// src/node/handler/PadMessageHandler.ts:950-1008
exports.updatePadClients = async (pad: PadType) => {
  const roomSockets = _getRoomSockets(pad.id);
  // ...
  await Promise.all(roomSockets.map(async (socket) => {
    while (sessioninfo.rev < pad.getHeadRevisionNumber()) {
      // ... build msg ...
      socket.emit('message', msg);  // packet construction per recipient
      sessioninfo.rev = r;
    }
  }));
};
```

In steady state every socket is at \`head - 1\`, so for N sockets behind by 1 rev that's **N packet constructions of essentially identical content**. The only per-recipient differing fields in \`msg\` are \`timeDelta\` and \`currentTime\` — and those fields are **only consumed by the timeslider** (\`broadcast.ts\`), not the live editor (\`collab_client.ts:191-209\` reads \`{newRev, changeset, author, apool}\` and nothing else).

## The opportunity

socket.io's \`room.emit()\` / \`io.in(roomId).emit()\` uses the adapter \`broadcast()\` path, which calls the encoder **once** and then loops \`writeToEngine\` per recipient:

```js
// socket.io-adapter/dist/in-memory-adapter.js:112
broadcast(packet, opts) {
  const encodedPackets = this._encode(packet, packetOpts);  // ONCE
  // ...
  socket.client.writeToEngine(encodedPackets, packetOpts);  // per socket
}
```

Switching the steady-state path from \`socket.emit\` per recipient to \`io.in(padId).emit\` would collapse N packet constructions into 1. The realistic CPU savings at the dive cliff are 5-7% (some portion of the ~10% socket.io overhead — encoding and packet construction are shared, but the per-socket transport write still happens N times).

## Why this needs design before implementing

Naively swapping \`socket.emit\` for \`io.in(padId).emit\` in the loop breaks the **catch-up case**:

- Steady state: all sockets at \`head - 1\`, one new rev → \`room.emit(head)\` works cleanly.
- Catch-up: one socket at \`head - 5\`, others at \`head - 1\` → a single \`room.emit(head)\` would be silently dropped by the lagger (\`collab_client.ts:202-205\` enforces \`newRev === rev + 1\` and \`return\`s on mismatch).

Two viable shapes:

**Shape A — split steady-state from catch-up.**
- For the most-recent rev only: \`io.in(padId).except(stragglers).emit(...)\` to all sockets at \`head - 1\`.
- For each straggler: individual \`socket.emit\` loop catches them up.
- Track stragglers by walking sessioninfos at fan-out entry.

**Shape B — push catch-up out of \`updatePadClients\`.**
- ``updatePadClients`` only ever broadcasts the latest rev to the room.
- A client that detects it's behind (\`newRev !== rev + 1\`) requests resend via a new \`CLIENT_REQUEST_RESEND\` message, which the server handles in a separate code path.
- More invasive but cleaner long-term: keeps the hot path simple and pushes catch-up to a slow path.

I'd favour **Shape A** as a first cut — local change, no protocol additions, the straggler-tracking is cheap (\`sessioninfo.rev !== headBefore\` check per socket once per fan-out).

## Acceptance

- Settings flag (``settings.roomBroadcastNewChanges`` default false) so it can be A/B tested in the dive.
- N=3 dive measurement before merging.
- No client-side change required for Shape A.

## Out of scope

- The historicalAuthorData / pool fan-out path (covered by [#7769](https://github.com/ether/etherpad/pull/7769) — closed, but the design problem is the same).
- Worker-thread offload of OT — falsified, see ``experiment/worker-thread-applytotext`` branch.
- engine.io transport-level packing — see [#7774](https://github.com/ether/etherpad/pull/7774) for the existing flush-deferral lever.

Discovered as part of the #7756 scaling dive ([dive doc PR #7765](https://github.com/ether/etherpad/pull/7765)).

kishansomaiya:
## Summary

Closes #7780.

This PR implements Shape A for room fan-out in src/node/handler/PadMessageHandler.ts:

- broadcast NEW_CHANGES once to steady-state recipients at head - 1
- keep per-socket rev + 1 catch-up for stragglers
- gate the behavior behind settings.roomBroadcastNewChanges, default false

Goal: reduce per-recipient packet construction overhead in the steady-state path without changing client protocol behavior.

## Why each change is necessary

- src/node/handler/PadMessageHandler.ts  
  Introduces split-path fan-out:
  - steady-state sockets use room broadcast for the latest revision
  - stragglers keep per-socket incremental catch-up  
  This is necessary because a naive latest-only broadcast would be dropped by lagging clients that enforce rev + 1 semantics.

- src/node/utils/Settings.ts  
  Adds roomBroadcastNewChanges to settings type and default.  
  This is required so the optimization can be A/B tested safely and remain off by default.

- settings.json.template  
  Documents the new flag and intended usage.  
  This is required so operators can enable the experiment intentionally and understand scope.

- settings.json.docker  
  Adds ROOM_BROADCAST_NEW_CHANGES env mapping.  
  This is required for containerized benchmark and rollout workflows.

- src/tests/backend-new/specs/roomBroadcastNewChanges-defaults.test.ts  
  Verifies default remains false.  
  This guards the acceptance requirement that the change is feature-flagged and opt-in.

## Compatibility and risk

- Default behavior is unchanged when the flag is false.
- No client-side protocol changes.
- Straggler path preserves current correctness guarantees.

## Validation

- TypeScript check passed
- Backend-new Vitest suite passed
- New defaults test passed

## N=3 measurement

Run setup used for this branch measurement:
- Duration: 10 seconds
- Authors: 5
- N=3 per mode, same host, fresh run per sample

Results:

| mode | run | cpu_delta_sec | NEW_CHANGES_emits_delta |
|---|---:|---:|---:|
| off | 1 | 0.9375 | 396 |
| off | 2 | 1.0625 | 404 |
| off | 3 | 1.0312 | 400 |
| on  | 1 | 1.1875 | 102 |
| on  | 2 | 1.0156 | 100 |
| on  | 3 | 0.9531 | 99 |

Averages:
- avg off cpu_delta_sec: 1.0104
- avg on cpu_delta_sec: 1.0521
- cpu change: +4.12%
- avg off NEW_CHANGES emits delta: 400.0000
- avg on NEW_CHANGES emits delta: 100.3333
- emits change: -74.92%

Notes:
- The expected emit reduction is clearly visible.
- CPU improvement is not visible at this small workload; likely below noise floor.
- If preferred, I can add a follow-up measurement block with a heavier profile matching CI-style load settings.

### DIFF
--- a/settings.json.docker
+++ b/settings.json.docker
@@ -684,7 +684,7 @@
   "socketTransportProtocols" : ["websocket", "polling"],
 
   /*
-   * Performance experiment (#7780): room-broadcast NEW_CHANGES to steady-state
+   * room-broadcast NEW_CHANGES to steady-state
    * recipients, keep per-socket catch-up for stragglers.
    */
   "roomBroadcastNewChanges": "${ROOM_BROADCAST_NEW_CHANGES:false}",
--- a/settings.json.docker
+++ b/settings.json.docker
@@ -683,6 +683,12 @@
    */
   "socketTransportProtocols" : ["websocket", "polling"],
 
+  /*
+   * Performance experiment (#7780): room-broadcast NEW_CHANGES to steady-state
+   * recipients, keep per-socket catch-up for stragglers.
+   */
+  "roomBroadcastNewChanges": "${ROOM_BROADCAST_NEW_CHANGES:false}",
+
   "socketIo": {
     /*
      * Maximum permitted client message size (in bytes). This controls the
--- a/settings.json.template
+++ b/settings.json.template
@@ -765,6 +765,15 @@
    */
   "socketTransportProtocols" : ["websocket", "polling"],
 
+  /*
+   * Performance experiment (#7780): use room-broadcast NEW_CHANGES fan-out for
+   * steady-state recipients (clients at head-1), while stragglers still catch
+   * up via per-socket emits.
+   *
+   * Keep disabled unless explicitly benchmarking or validating this behavior.
+   */
+  "roomBroadcastNewChanges": false,
+
   "socketIo": {
     /*
      * Maximum permitted client message size (in bytes). This controls the
--- a/settings.json.template
+++ b/settings.json.template
@@ -766,7 +766,7 @@
   "socketTransportProtocols" : ["websocket", "polling"],
 
   /*
-   * Performance experiment (#7780): use room-broadcast NEW_CHANGES fan-out for
+   * use room-broadcast NEW_CHANGES fan-out for
    * steady-state recipients (clients at head-1), while stragglers still catch
    * up via per-socket emits.
    *
--- a/src/node/handler/PadMessageHandler.ts
+++ b/src/node/handler/PadMessageHandler.ts
@@ -1046,7 +1046,6 @@ exports.updatePadClients = async (pad: PadType) => {
       }
     }
 
-    // Broadcast the latest revision once for all steady-state recipients (head-1).
     if (syncedSocketIds.length > 0 && headRev >= 0) {
       const revision = await getRevision(headRev);
       const author = revision.meta.author;
@@ -1073,7 +1072,6 @@ exports.updatePadClients = async (pad: PadType) => {
           ns.to(pad.id).emit('message', msg);
         }
       } else {
-        // Fallback to direct socket emits if namespace is unavailable.
         for (const socketId of syncedSocketIds) {
           const socket = roomSockets.find((s) => s.id === socketId);
           if (socket != null) socket.emit('message', msg);
@@ -1089,7 +1087,6 @@ exports.updatePadClients = async (pad: PadType) => {
       }
     }
 
-    // Stragglers still need per-socket catch-up to preserve rev+1 semantics.
     await Promise.all(stragglerSockets.map(async (socket) => {
       const sessioninfo = sessioninfos[socket.id];
       if (sessioninfo == null) return;
--- a/src/node/handler/PadMessageHandler.ts
+++ b/src/node/handler/PadMessageHandler.ts
@@ -1010,6 +1010,122 @@ exports.updatePadClients = async (pad: PadType) => {
   // benefits of running this in parallel,
   // but benefit of reusing cached revision object is HUGE
   const revCache:MapArrayType<any> = {};
+  const forWireCache:MapArrayType<any> = {};
+
+  const getRevision = async (rev: any) => {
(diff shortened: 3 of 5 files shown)


JohnMcLear:
If you look through the load testing you should see **we also tested for max load performance**.  I would say this is not a suitable first issue for someone new to Etherpad dev.

kishansomaiya:
@JohnMcLear Please review this PR

JohnMcLear:
Please read my comment, you need benchmarking for heavy load.

Again, not to be obstructive but this is a very difficult issue for someone with what I have to assume is little to no knowledge of OTs.

Getting this merged might cost you significant maintainer debt.  I would strongly recommend to avoid anything related to performance until you have a strong grasp of OTs..

If you are comfortable however and perhaps I misread the situation then please continue with heavy load testing and express **your absolute intent to handle this debt** moving forward and we can get things merged if the performance hits our simulated expectations.

kishansomaiya:
> If you look through the load testing you should see we also tested for max performance. I would say this is not a suitable first issue for someone new to Etherpad dev.

Thanks for the clarification. That makes sense.

I'm still interested in contributing to Etherpad and learning the codebase. Are there any issues you'd recommend for someone new to the project that would be a better starting point?

JohnMcLear:
@kishansomaiya https://github.com/ether/etherpad/issues/7865 might be a nice easy good starter

kishansomaiya:
> @kishansomaiya #7865 might be a nice easy good starter

Okay, John I will work on that.

PR #7854, opened 2026-05-26
test(flake): scoped per-test event-loop yield in the 6 dying spec files

JohnMcLear:
## Summary

Adds `beforeEach(async () => await new Promise(r => setImmediate(r)))` at the top-level describe of the six spec files that cluster in the silent-ELIFECYCLE captures. Pattern-wide intervention against rapid-sequential test cadence as the suspected trigger.

## Why

12 captured silent-ELIFECYCLE deaths (#7838, #7842, #7846 diagnostics) all hit the same six files:

| File | Deaths |
|---|---|
| `importexportGetPost.ts` | 4 |
| `socketio.ts` | 3 |
| `messages.ts` | 3 |
| `pad.ts > Tests` | 2 |
| `import.ts` | 1 |
| `clientvar_rev_consistency.ts` | 1 |

All six share a shape: 50–100 short tests per top-level describe, each test making rapid loopback HTTP (supertest) or socket.io connect/disconnect calls. Pre-kill node-reports show V8 main isolate going event-loop-starved for 200–400 ms before each kill (5 Hz heartbeat falls silent), then external termination.

The TIME_WAIT smoking gun captured by #7846 was ruled out as causal by #7852 (closed): keepAlive collapsed TIME_WAIT churn to zero, kill survived anyway. The remaining shared substrate is just **cadence**.

## What this changes

One `beforeEach` per file, inserted at the top-level `describe(__filename, ...)`. The yield forces a timer-queue drain between every test in those files, breaking the tight microtask chain that may be the trigger.

**Critical scope note:** per-file beforeEach, NOT root `mochaHooks.beforeEach`. PR #7844 tried the root-level variant and broke `ep_subscript_and_superscript`'s `returns HTML with Subscript HTML tags` tests, which share state across describe-block boundaries. Per-file scope contains any timing perturbation to the affected files — plugin tests loaded from `../node_modules/ep_*/static/tests/backend/specs` are completely untouched.

## Expected outcomes

- **Linux ± plugins must pass.** If they regress, the yield is breaking the affected tests' own state-sharing assumptions — we'd revert.
- **Windows ± plugins flake rate:**
  - Pre-fix: ~22% per #7663 (13/60 runs).
  - Post-fix expectation: if cadence is the trigger → materially lower. If unchanged over 5–10 reruns → cadence ruled out, look elsewhere (jose CNG, Express middleware, libuv IOCP edge cases unrelated to load).

Either result is data we don't currently have.

## Locally verified

Not feasible — local mocha against this project needs full bootstrap. Relying on CI.

### DIFF
--- a/src/tests/backend/diagnostics.ts
+++ b/src/tests/backend/diagnostics.ts
@@ -210,10 +210,28 @@ for (const sig of ['SIGINT', 'SIGTERM', 'SIGHUP', 'SIGBREAK'] as const) {
 // resolution fails us. A `start` line per test gives sub-millisecond
 // resolution on which test was on the rails when the process died.
 export const mochaHooks = {
-  beforeEach(this: any) {
+  async beforeEach(this: any) {
     if (this.currentTest) {
       currentTest = this.currentTest.fullTitle();
       diag(`test start: ${currentTest}`);
+      // Per-test event-loop yield for ether/etherpad's own backend tests
+      // (not plugin tests). Mitigation against the Windows silent-ELIFECYCLE
+      // flake: 13 captures so far show V8 starvation 200-400 ms before each
+      // kill in test files that fire 50-100 short tests in sequence. The
+      // initial scoped fix in PR #7854 covered 6 known-dying files but the
+      // flake migrated to a 7th (sessionsAndGroups.ts) on the next run —
+      // so the trigger is the broader rapid-sequential-test pattern, not
+      // specific files. A root-level yield covers all our backend tests.
+      //
+      // Plugin tests (loaded from `../node_modules/ep_*/static/tests/
+      // backend/specs`) are SKIPPED via file-path check because PR #7844
+      // demonstrated the global variant breaks ep_subscript_and_superscript
+      // tests that share state across describe-block boundaries. The check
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Closing — the broader-scope yield (root mochaHooks.beforeEach, scoped to etherpad's own specs via file-path check) still didn't prevent the silent ELIFECYCLE. Captured death #14 on this run:

\`\`\`
clientvar_rev_consistency.ts > CLIENT_VARS stays consistent under concurrent edits during handshake (delay race)
kill at +46368ms, 467ms after test start, exit 255, same fingerprint
\`\`\`

That file IS covered by the root yield. So cadence / rapid-sequential test pattern is NOT the trigger.

Combined with PR #7852's result, this rules out two of the most plausible causal hypotheses (TIME_WAIT, cadence). Remaining candidates need either local Windows repro or out-of-process probing beyond what we've already added (#7846's netstat sidecar). In-CI diagnosis has reached diminishing returns.

Linux ± plugins did pass cleanly — the file-path-scoped yield is safe even though it doesn't fix anything. Could be revived as a chore PR if the test-infra value alone is worth it, but I'm not going to ship "scoped yield with no measurable benefit" as a fix.

PR #7855, opened 2026-05-26
test(ci): probe — disable Defender + capture Event Log + fix tasklist sidecar

JohnMcLear:
## Summary

Workflow-only probe targeting the Windows silent-ELIFECYCLE flake. Three orthogonal additions in one PR:

1. **A** — disable Microsoft Defender real-time monitoring for the duration of the test step
2. **H** — clear Application + System event logs pre-test, dump them post-test (pass or fail) to the artifact directory
3. **I** — fix the tasklist sidecar from #7846 (was producing empty output due to git-bash + UTF-16 BOM; switched to `Get-CimInstance Win32_Process` for clean ASCII with HandleCount, ThreadCount, WorkingSetSize, PageFileUsage, KernelModeTime, UserModeTime sampled every 500 ms)

## Why

Every hypothesis tested so far is RULED OUT:

| Hypothesis | Ruled out by | How |
|---|---|---|
| Memory/handle leak | every pre-kill node-report | nominal state |
| TIME_WAIT accumulation | #7852 | keepAlive collapsed it, kill survived |
| Rapid-sequential cadence | #7854 | setImmediate yield in root hook, kill survived |
| File-specific pathology | death corpus | 7 files, same fingerprint |

The kill fingerprint (silent external termination, no JS-handler trace, no native abort report, sub-1s death window) **matches Microsoft Defender's behavioural-monitoring `TerminateProcess` signature** more closely than any other plausible cause. Defender is enabled by default on GHA Windows runners, and rapid loopback TCP fanout is on its suspect-process-behaviour list. We've simply never tested it.

If kills disappear with RT off → causal. If kills persist but `event-defender.txt` shows pre-kill detection entries → Defender is involved with a more nuanced trigger. If kills persist with no Defender entries → the OS event log will name the actual terminator (Service Control Manager, kernel guard, Werfault, etc.).

## What's captured in the artifact on failure

```
node-report/
├── defender-state-before.txt  # pre-test Defender config
├── defender-state-after.txt   # post-test Defender config (sanity)
├── event-clear.txt            # confirmation logs were cleared
├── event-application.txt      # last 500 Application events with timestamps
├── event-system.txt           # last 500 System events
├── event-defender.txt         # last 200 Defender Operational events
├── event-app-errors.txt       # specifically Application Error / Hang / WER
├── netstat.log                # (existing) localhost TCP every 500ms
├── tasklist.log               # (NOW WORKING) node.exe handle/CPU/RSS every 500ms
[...]

PR #7856, opened 2026-05-26
test(ci): probe — drop mocha --exit on Windows backend tests

JohnMcLear:
## Summary

One-line workflow probe on a separate branch from #7855: remove the `--exit` flag from `pnpm test` on both Windows backend matrices.

## Why

The current workflow has been invoking `pnpm test -- --exit` on Windows since PR #7762, with the comment:

> `--exit` forces `process.exit(failures)` after the suite completes, closing the post-suite event-loop drain window where Windows + Node 24 hard-kills the process.

That's a direct admission of a known post-suite kill bug on Windows + Node 24, and `--exit` was added as **a mitigation that hides it**. The current silent-ELIFECYCLE flake fires *mid-suite*, but `--exit` is still in play after the dying test and may also be masking earlier signal.

Drop it on a separate probe branch. Either:
- Tests still complete cleanly via natural drain → the original kill from #7762 is no longer reproducible on current Node 24.x and we've been carrying the workaround for nothing.
- A fresh hard-kill appears at post-suite drain → second data point distinct from the mid-suite flake, comparable to it via the merged diagnostic infra (#7838 + #7842 + #7846).
- The mid-suite flake fires → does mocha emit additional output that `--exit`'s `process.exit()` was suppressing?

Any of these is data we don't currently have.

## Test plan

- [ ] Linux ± plugins must pass (no change there)
- [ ] Windows backend matrices: fail, pass, or unchanged — all three outcomes are informative
- [ ] Run alongside #7855 in parallel to maximise data per CI window

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -258,7 +258,7 @@ jobs:
           # Node 24 hard-kills the process. Scoped to Windows so Linux/local
           # runs still surface real handle leaks via natural drain.
           set +e
-          pnpm test -- --exit
+          pnpm test
           EXIT=$?
           set -e
           kill "$WATCHER_PID" 2>/dev/null || true
@@ -392,7 +392,7 @@ jobs:
           # Node 24 hard-kills the process. Scoped to Windows so Linux/local
           # runs still surface real handle leaks via natural drain.
           set +e
-          pnpm test -- --exit
+          pnpm test
           EXIT=$?
           set -e
           kill "$WATCHER_PID" 2>/dev/null || true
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Closing — this was a diagnostic probe while hunting the Windows backend silent-ELIFECYCLE flake. Root cause found and fixed: in-process handler gate + Windows pinned to Node 24.16.0 (#7866), investigation scaffolding removed (#7868), and the underlying libuv TCP-connect stack overrun filed upstream as nodejs/node#63620. Branch deleted.

PR #7862, opened 2026-05-27
test(ci): probe G — install ProcDump as JIT debugger to capture __fastfail across the silent kill

JohnMcLear:
## Summary

Install Sysinternals ProcDump as the system Just-In-Time debugger on both Windows backend matrices. When ANY process on the runner crashes or fast-fails (`int 29h`), the OS hands it to the JIT debugger, which writes a full memory dump to the artifact directory.

## Why

**Defender ruled out by #7855.** First failure on that PR (run 26470378618) with `DisableRealtimeMonitoring + DisableBehaviorMonitoring + DisableIOAVProtection` all confirmed `True` produced this evidence:

| Source | Contents at kill time |
|---|---|
| Application log | empty (cleared pre-test, zero entries during run) |
| System log | only pre-test service-stop events |
| Defender Operational | only stale 2026-05-18 provisioning events |
| Application Error/Hang/WER | zero entries |
| OS-side tasklist (node.exe) 1s before kill | HandleCount=323 stable, ThreadCount=17 stable, WS=321 MB stable, KernelModeTime + UserModeTime growing linearly — completely healthy |

Then **dead 1 second later with zero trace anywhere**.

That fingerprint — silent external termination with no event-log entry and a healthy OS-side process — is the signature of `__fastfail` (`int 29h`). libuv on Windows uses it for internal assertion failures in `uv_win.c`, `tcp-win.c`, `pipe-win.c` (TCP/pipe state-corruption checks). `__fastfail` terminates the process bypassing all user-mode notification and WER.

The only standard tool that captures state across `__fastfail` is a JIT-installed debugger.

## What this changes

New step on both Windows backend matrices, BEFORE "Run the backend tests":

```yaml
- name: Install ProcDump as JIT debugger (probe G)
  shell: powershell
  run: |
    Invoke-WebRequest "https://download.sysinternals.com/files/Procdump.zip" -OutFile "$env:TEMP\Procdump.zip"
    Expand-Archive -Path "$env:TEMP\Procdump.zip" -DestinationPath "C:\procdump" -Force
    New-Item -ItemType Directory -Force -Path "${{ github.workspace }}\node-report\dumps" | Out-Null
    & "C:\procdump\procdump64.exe" -accepteula -i -ma "${{ github.workspace }}\node-report\dumps"
```

- `-i` registers procdump64.exe as the AeDebug handler (system JIT debugger)
- `-ma` writes full memory dumps
- Dumps land in `node-report/dumps/` — picked up by the existing failure-artifact upload

## Built on PR #7855

Branch is on top of `probe-flake-defender-eventlog-sidecar`, so it carries the Defender-off + event-log + tasklist-fix probes too. Even though Defender's ruled out, those baselines stay useful for cross-comparison.

## Expected outcome

On the next silent ELIFECYCLE failure: a `.dmp` file in the artifact. Loadable in WinDbg with `!analyze -v` — that should name the function calling `__fastfail` and the assertion that fired.

If no `.dmp` is produced even on failure → the kill isn't going through user-mode exception handling at all → it's coming from the kernel (HVCI, CET violation, page-fault chain). Escalates to ETW tracing.

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -215,7 +215,7 @@ jobs:
           powershell -Command "(gc settings.json.template) -replace '\"max\": 10', '\"max\": 10000' | Out-File -encoding ASCII settings.json.holder"
           powershell -Command "(gc settings.json.holder) -replace '\"points\": 10', '\"points\": 1000' | Out-File -encoding ASCII settings.json"
       -
-        # PROBE G (rev 3) (retrigger 1) — install Sysinternals ProcDump as the system
+        # PROBE G (rev 3) (retrigger 2) — install Sysinternals ProcDump as the system
         # Just-In-Time debugger AND additionally launch a long-running
         # procdump that monitors all node.exe processes specifically.
         #
@@ -474,7 +474,7 @@ jobs:
           powershell -Command "(gc settings.json.template) -replace '\"max\": 10', '\"max\": 10000' | Out-File -encoding ASCII settings.json.holder"
           powershell -Command "(gc settings.json.holder) -replace '\"points\": 10', '\"points\": 1000' | Out-File -encoding ASCII settings.json"
       -
-        # PROBE G (rev 3) (retrigger 1) — install Sysinternals ProcDump as the system
+        # PROBE G (rev 3) (retrigger 2) — install Sysinternals ProcDump as the system
         # Just-In-Time debugger AND additionally launch a long-running
         # procdump that monitors all node.exe processes specifically.
         #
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Closing — this was a diagnostic probe while hunting the Windows backend silent-ELIFECYCLE flake. Root cause found and fixed: in-process handler gate + Windows pinned to Node 24.16.0 (#7866), investigation scaffolding removed (#7868), and the underlying libuv TCP-connect stack overrun filed upstream as nodejs/node#63620. Branch deleted.

PR #7863, opened 2026-05-27
test(ci): probe K — Node 22 added to Windows-no-plugins matrix as comparator

JohnMcLear:
Probe K from the silent-ELIFECYCLE flake matrix. After probe A (#7855) ruled out Defender and probe G (#7862) deploys ProcDump for forensic capture, this PR adds a parallel data point: run Win-no-plugins on Node 22 alongside the existing Node 24 job. If the flake rate drops materially on Node 22, we have a clear bisection target. If unchanged, Node version isn't the cause.

Adds `engines-strict=false` to .npmrc on the Node 22 job only (Etherpad's package.json has `engines.node: >=24`). Scoped to Win-no-plugins to minimise CI cost — the same probe on the with-plugins job adds ~5 minutes more per push because of plugin install.

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -179,8 +179,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Etherpad requires Node >= 24 (see package.json engines.node).
-        node: ${{ fromJSON('[24]') }}
+        # PROBE K — Node 22 added as comparator vs 24. If 22 doesn't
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Closing — Etherpad has a runtime version check (settings.ts) that refuses to start on Node < 24 with a clear error message. So this probe can't compare flake rate without code changes to bypass the version check, which expands scope considerably. The Node-22-as-comparator hypothesis isn't dead, just not achievable with a workflow-only probe. Worth revisiting as a code change if other probes don't pin down the cause.

PR #7877, opened 2026-05-29
Fix/7865 admin open pad url encode

### LINKED ISSUE #7865: Bug: Admin Dashboard “Open” button does not URL‑encode slashes in pad names
**Summary**
When a pad is created with a forward slash (/) in its name, the Admin Dashboard → Pads list displays the pad correctly, but the Open button generates a URL that treats the slash literally instead of encoding it as %2F. As a result, the link attempts to navigate into a nested path rather than opening the pad.

The pad can be opened manually if the slash is replaced with %2F in the URL.

**Steps to Reproduce**
Log in as an admin.

Create a new pad with a name containing a slash, e.g. "Test/123"

Go to admin/pads.

Click the Open button for that pad.

**Actual Behavior**
The generated link is something like:
/p/Test/123
Etherpad interprets this as a nested path (Test → 123) and fails to open the pad.

**Expected Behavior**
The pad name should be URL‑encoded when generating the link.

The Open button should point to:
/p/Test%2F123

**Environment**
Etherpad version: 3.2.0

Browser: Any

Admin Dashboard page: /admin/pads

**Additional Notes**
The pad is stored and listed correctly; the issue is only with the link generation.

This appears to be a missing encodeURIComponent() (or equivalent) when building the URL for the Open button.

kishansomaiya:
fixes #7865.

On /admin/pads, the Open button was building the pad URL with the raw pad name.
For pad names containing a slash (for example Test/123), this produced /p/Test/123 instead of /p/Test%2F123.

This PR encodes the pad name when building the Open link so slash-containing pad names open correctly.

Test plan
Create a pad named Test/123.
Open /admin/pads.
Click Open for that pad.
Confirm it opens the correct pad (URL uses an encoded segment, for example Test%2F123).

### DIFF
--- a/admin/src/pages/PadPage.tsx
+++ b/admin/src/pages/PadPage.tsx
@@ -410,7 +410,7 @@ export const PadPage = () => {
                           </button>
                           <button
                             className="pm-btn pm-btn-primary pm-btn--sm"
-                            onClick={() => window.open(`../../p/${pad.padName}`, '_blank')}
+                            onClick={() => window.open(`../../p/${encodeURIComponent(pad.padName)}`, '_blank')}
                           >
                             <Eye size={13}/> <Trans i18nKey="admin_pads.open"/>
                           </button>
--- a/src/static/js/pad_userlist.ts
+++ b/src/static/js/pad_userlist.ts
@@ -23,6 +23,14 @@ const chat = require('./chat').chat;
 import html10n from './vendors/html10n';
 let myUserInfo = {};
 
+const decodePadSegment = (segment: string): string => {
+  try {
+    return decodeURIComponent(segment);
+  } catch {
+    return segment;
+  }
+};
+
 let colorPickerOpen = false;
 let colorPickerSetup = false;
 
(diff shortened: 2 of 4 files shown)


abhijeetgorhe26 (commented):
It will stay the same for recent pads, but there should be URL encoding only for the URL, not for the recent pads list. In your case, the same pad name will appear in the recent pads.So, it will create bug and it is not valid solution for this issue.Please refer to this: https://github.com/ether/etherpad/pull/7876

kishansomaiya:
Thanks for the review @abhijeetgorhe26 . I’ve updated the fix to encode only navigation URLs and keep Recent Pads names unchanged/readable.

Changes made:

src/static/skins/colibris/index.js :
    Normalize existing recentPads entries by decoding stored names (for example Test%2F123 -> Test/123).
    Keep displayed pad name as decoded text.
    Encode only the link target (.../p/${encodeURIComponent(pad.name)}).

src/static/skins/colibris/pad.js :
    Decode the current pad path segment before storing/updating recentPads name.

src/static/js/pad_userlist.ts :
    Decode the current pad path segment before matching/updating member counts in recentPads.
    
Result:

Recent Pads displays Test/123 (not Test%2F123).
Clicking it still navigates with an encoded URL segment, so slash-containing names open correctly.

JohnMcLear:
Missing tests.  Tell your AI to read agents.md / contributing.md

PR #7888, opened 2026-06-02
docs: refresh docs for 3.2.0 — correct stale content, document recent features

JohnMcLear:
## Why

The docs under `doc/` are **hand-authored VitePress markdown** (not generated from the OpenAPI spec — `openapi.ts` only builds the runtime Swagger definition and never reads/writes `doc/`), so they need manual upkeep and had drifted behind a lot of recent work. This pass corrects content that was actively *wrong* and documents features shipped since the pages were last touched. Every change was verified against source.

## Corrections (was wrong / would mislead)

- **`cli.md`** — every command said `node bin/foo.js`, but the scripts are TypeScript run via pnpm; copy-paste **failed**. Rewrote to `pnpm run --filter bin <script>`, documented ~13 previously-undocumented operator tools, and split running-vs-stopped requirements. Registered the missing `compactStalePads` entry in `bin/package.json` so the documented command actually runs.
- **`stats.md`** — described a pre-Prometheus world. Rewrote for the `enableMetrics`-gated `/stats` (JSON) and `/stats/prometheus` endpoints, the live metric set, the opt-in `scalingDiveMetrics` instruments (#7756), and the `measured-core` library.
- **`admin/updates.md`** — removed three false "SMTP not yet wired / would send email" claims (it **is** wired via nodemailer + the new `mail.*` block), documented the `node-engine-mismatch` preflight check and the rollback/preflight failure emails, and stripped obsolete "PR 1 / PR 2" staging language now that all tiers ship.
- **`api/http_api.md`** — added the undocumented `anonymizeAuthor` (GDPR Art. 17) call, fixed `copyPad`/`movePad` version annotations (1.2.8 → 1.2.9), corrected `getPadID`'s param name (`readOnlyID` → `roID`), and dropped a reference to a non-existent `getEtherpad` API call.
- **`skins.md`** — colibris is the current default, not "experimental, will become default in 2.0".
- **`localization.md`** — bare `window._('key')` is unbound and returns `undefined`; recommend `window.html10n.get(...)` / `data-l10n-id`.
- **`README.md`** — bumped the `v2.2.5` upgrade example to `v3.2.0`; fixed a `docker.adoc` link to `docker.md`.
- **`docker.md`** — added `MAIL_*`, `ENABLE_METRICS`, `GDPR_AUTHOR_ERASURE_ENABLED`, `PRIVACY_BANNER_*`, `PUBLIC_URL`, `AUTHENTICATION_METHOD`, `ENABLE_DARK_MODE`, `ENABLE_PAD_WIDE_SETTINGS`; fixed the `SOCKETIO_MAX_HTTP_BUFFER_SIZE` default (50000 → 1000000).

## New documentation

- **`configuration.md`** (new) — how settings + `${VAR:default}` substitution work, `trustProxy`, and the previously-undocumented **subpath / ingress** support (`x-proxy-path` / `X-Forwarded-Prefix` / `X-Ingress-Path`), including the sanitizer rules and Traefik/NGINX examples. Wired into the sidebar and index hero.
- **`hooks_server-side.md`** — `ccRegisterBlockElements` (the server-side companion plugin authors miss), `exportConvert`, `exportHTMLSend`, `createServer`, `restartServer`, `clientReady` (deprecated).
- **`hooks_client-side.md`** — `aceDrop`, `acePaste`, `handleClientTimesliderMessage_<name>`.

## Notes / follow-ups (not in this PR)

- **`.adoc` left in place.** It looked like a dead duplicate, but `bin/release.ts` still runs `makeDocs` (asciidoctor) and copies the output into `ether.github.com/public/doc/v<version>` — i.e. it feeds the **per-version doc archives** at release time. Deciding its fate (migrate the archive to VitePress vs keep) is a separate change.
- **`src/node/metrics.ts`** appears to be dead code (never imported/registered), so its gauges are intentionally not documented in `stats.md`. Worth wiring up or removing in a follow-up.

## Test

VitePress build passes locally (`pnpm run docs:build`), including the dead-link check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ volumes:
 
 ### Docker container
 
-Find [here](doc/docker.adoc) information on running Etherpad in a container.
+Find [here](doc/docker.md) information on running Etherpad in a container.
 
 ## Plugins
 
@@ -247,8 +247,8 @@ git -P tag --list "v*" --merged
 ```
 4. Select the version
 ```sh
-git checkout v2.2.5
-git switch -c v2.2.5
+git checkout v3.2.0
+git switch -c v3.2.0
 ```
 5. Upgrade Etherpad
 ```sh
--- a/bin/compactAllPads.ts
+++ b/bin/compactAllPads.ts
@@ -4,9 +4,9 @@
  * Compact every pad on the instance to reclaim database space.
  *
  * Usage:
- *   node bin/compactAllPads.js              # collapse all history on every pad
- *   node bin/compactAllPads.js --keep N     # keep last N revisions per pad
- *   node bin/compactAllPads.js --dry-run    # list pads + rev counts, no writes
+ *   pnpm run --filter bin compactAllPads              # collapse all history on every pad
+ *   pnpm run --filter bin compactAllPads --keep N     # keep last N revisions per pad
+ *   pnpm run --filter bin compactAllPads --dry-run    # list pads + rev counts, no writes
  *
  * Composes the existing `listAllPads` and `compactPad` HTTP APIs — there is
  * deliberately no instance-wide HTTP endpoint, because doing this over a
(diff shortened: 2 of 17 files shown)


JohnMcLear:
Thanks @qodo — both addressed in 08808b26:

1. **Nonexistent `server.js`** — `configuration.md` now documents the supported launcher `bin/run.sh -s <file>` (forwards to `pnpm run prod`) and notes the runtime is `src/node/server.ts` via `tsx`; there is no `server.js`.
2. **CLI help references `.js`** — fixed the header comments and runtime usage output in `compactStalePads.ts`, and also `compactPad.ts` / `compactAllPads.ts` for consistency, to the documented `pnpm run --filter bin <script>` form.

PR #7906, opened 2026-06-07
Hardening: API request handling, random IDs, and plugin loading

JohnMcLear:
A second small hardening pass, covering API request handling, random-ID generation, and plugin loading. Companion to #7905.

### Changes
- **`pad_utils.randomString`** — generate the random IDs (author/session/readonly) via `crypto.getRandomValues` (a CSPRNG) instead of `Math.random`, with rejection sampling to avoid modulo bias. Output format/length is unchanged.
- **`OAuth2Provider`** — constant-time password comparison and a uniform failure delay on the OIDC interaction login; look the user up by own property only.
- **`API.appendChatMessage`** — require the pad to already exist (`getPadSafe`), consistent with the other content API methods.
- **`RestAPI` (`/api/2`)** — forward only the `authorization` header instead of merging all request headers into the API field set (matches the `openapi.ts` handler).
- **`LinkInstaller`** — validate plugin dependency names before using them to build filesystem paths.
- **admin file server** — return a generic error and log the detail server-side.

### ⚠️ Behaviour / breaking changes
1. **`appendChatMessage` now returns `padID does not exist`** for a non-existent pad instead of silently creating it. Callers that relied on the implicit pad creation must call `createPad` (or `createGroupPad`) first. This aligns it with `setText`/`getText`/etc.
2. **`/api/2` no longer exposes arbitrary request headers as API parameters.** Only `authorization` is forwarded. This only affects a caller that passed an API parameter via an HTTP header of the same name (not a documented usage); send such values as query/body parameters instead.
3. **Failed OIDC interaction logins now take ~1s** (uniform failure delay). Behavioural only.

The remaining changes (`randomString`, `LinkInstaller`, admin error text) are non-breaking.

### Testing
- Backend specs covering the touched areas pass (`chat`, `api`, `jwtAdminClaim`, `LinkInstaller`, author-token, `webaccess`, `tokenTransfer`, `OIDCAdapter`).
- `randomString` validated in a real browser via the `author_token_cookie` chromium spec (HttpOnly token, stable authorID across reload, unique per context).

### DIFF
--- a/src/node/db/API.ts
+++ b/src/node/db/API.ts
@@ -414,7 +414,11 @@ exports.appendChatMessage = async (padID: string, text: string|object, authorID:
     time = Date.now();
   }
 
-  // @TODO - missing getPadSafe() call ?
+  // Require the pad to already exist, like the other content API methods.
+  // Without this, sendChatMessageToPadClients -> padManager.getPad() would
+  // create the pad on demand with default content, so the documented
+  // {code:1,"padID does not exist"} result would never be returned.
+  await getPadSafe(padID, true);
 
   // save chat message to database and send message to all connected clients
   await padMessageHandler.sendChatMessageToPadClients(new ChatMessage(text, authorID, time), padID);
--- a/src/node/db/API.ts
+++ b/src/node/db/API.ts
@@ -414,11 +414,13 @@ exports.appendChatMessage = async (padID: string, text: string|object, authorID:
     time = Date.now();
   }
 
-  // Require the pad to already exist, like the other content API methods.
-  // Without this, sendChatMessageToPadClients -> padManager.getPad() would
-  // create the pad on demand with default content, so the documented
-  // {code:1,"padID does not exist"} result would never be returned.
-  await getPadSafe(padID, true);
+  // Reject messages addressed to a pad that doesn't exist. Without this check
+  // the downstream padManager.getPad() would create the pad on demand with
+  // default content, so the documented {code:1,"padID does not exist"} result
+  // would never be returned.
+  if (!await padManager.doesPadExists(padID)) {
+    throw new CustomError('padID does not exist', 'apierror');
+  }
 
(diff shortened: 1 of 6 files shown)


src/node/db/API.ts
```diff
@@ -414,7 +414,11 @@ exports.appendChatMessage = async (padID: string, text: string|object, authorID:
     time = Date.now();
   }
 
-  // @TODO - missing getPadSafe() call ?
+  // Require the pad to already exist, like the other content API methods.
+  // Without this, sendChatMessageToPadClients -> padManager.getPad() would
+  // create the pad on demand with default content, so the documented
+  // {code:1,"padID does not exist"} result would never be returned.
+  await getPadSafe(padID, true);
```
SamTV12345: Or we just throw an error. That way somebody can send chat messages to random pads and they are created.

JohnMcLear: `getPadSafe(padID, true)` already throws `padID does not exist` when the pad is missing (API.ts:1029) — it does not create it — so chat can no longer materialise random pads. It's the same existence check `getText`/`setText`/`getHTML` use, and it also validates the padID string/format.

Happy to swap it for an explicit `throw new CustomError('padID does not exist', 'apierror')` if you'd rather not fetch the pad here — functionally identical, just let me know which you'd prefer to merge.

SamTV12345: Ah okay. Thought getPadSafe would also create if missing. Then let's be more explicit with an error like you suggested `throw new CustomError('padID does not exist', 'apierror')`

SamTV12345 (approved):

src/node/db/API.ts
JohnMcLear: Done — switched to the explicit `throw new CustomError('padID does not exist', 'apierror')` (guarded by `padManager.doesPadExists`), so no pad is fetched or created. Chat spec still green.

PR #7924, opened 2026-06-09
ci: enable downstream client smoke (pad/cli/desktop) — Phase 2

JohnMcLear:
<h3>PR Summary by Qodo</h3>

CI: enable live downstream client smoke gate for pad/cli/desktop
<code>🧪 Tests</code> <code>⚙️ Configuration changes</code> <code>🕐 20-40 Minutes</code>

Grey Divider

<h3>Walkthroughs</h3>

<details open>
<summary>User Description</summary>

<br/>

Builds on #7923 (Phase 1). Turns on the **live** downstream-smoke gate now that each client has its Phase 2 vector + smoke tests.

## What this adds
- **`src/tests/downstream/run-clients.sh`** — per-`kind` orchestration: for each enabled manifest entry, clone the client @ its pinned ref, set up its toolchain, point it at core's freshly generated fixture via `ETHERPAD_WIRE_VECTORS`, then run the client's `vectorTest` + `smokeCmd` against the booted server (`ETHERPAD_SMOKE_URL` + `ETHERPAD_SMOKE_APIKEY`). Clients test against **current core's** serialization, not a vendored snapshot.
- **`clients.json`** — all three flipped to `enabled:true`, pinned to the commits carrying their Phase 2 tests.
- **`downstream-smoke.yml`** — adds the Rust toolchain and calls the runner with the smoke env.

## The three client PRs (each adds vendored fixture + `vectorTest` + live `smokeCmd`)
- ether/pad#1 (Rust crate: changeset decode vectors + `PadSession` round-trip)
- ether/etherpad-cli-client#136 (repo's first test runner; reuses its own Changeset/AttributePool)
- ether/etherpad-desktop#78 (fixture-integrity guard + headless-light HTTP roundtrip)

## Verification
Ran the **full orchestration locally against a real core** — clone @ pinned SHA → inject fixture → vectors + live smoke — all green:
- etherpad-pad: vectors ✓, smoke ✓
- etherpad-cli-client: vectors 5/5 ✓, live USER_CHANGES round-trip ✓
- etherpad-desktop: vectors 22/22 ✓, smoke ✓

## Follow-up
Once the three client PRs squash-merge, bump their `ref`s in `clients.json` to the merge commits (the pinned pre-merge SHAs stay fetchable, so the gate keeps working until then).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

</details>

<details open>
<summary>AI Description</summary>

<br/>

<pre>
[...]

PR #7927, opened 2026-06-09
ci(downstream): robust per-client error handling in run-clients.sh

JohnMcLear:
Follow-up to #7924, which merged before its Qodo review was addressed. Lands the two Qodo findings (both real):

1. **Clone failure aborted the whole run (silently).** `git clone/fetch/checkout` ran outside the per-client failure guard. Worse: moving them *into* the `( … ) || { fail=1; }` subshell **suspends `set -e`** (the subshell is an `||` operand), so a failed clone was silently swallowed and the loop continued from the wrong cwd — a broken/unreachable client would have **passed the gate**. Fixed by guarding every step (clone/fetch/checkout/cd/install/test) with explicit `|| exit 1`, so one client's failure becomes a per-client `fail=1`, the loop continues, and the run exits non-zero. Fetch errors are no longer suppressed (fetch only when the pinned commit isn't already reachable).
2. **`eval` brittleness** — manifest commands now run via `bash -c` (trusted in-repo allowlist; no double-parse, no leak into this script's shell).

## Verification
- Two bogus clients → both reported (`::error::downstream client …`), loop continues, **exit 1**.
- Happy path (cli, no server) → vectors pass, smoke skips cleanly, **exit 0**.

### DIFF
--- a/src/tests/downstream/clients.json
+++ b/src/tests/downstream/clients.json
@@ -2,7 +2,7 @@
   {
     "name": "etherpad-pad",
     "repo": "https://github.com/ether/pad.git",
-    "ref": "ada8cafd33b4ab31206226b4c3db80b2aa00125a",
+    "ref": "91620c67c49536bb77e90b39f298ea70ae93c4a0",
     "kind": "rust",
     "enabled": true,
     "vectorTest": "cargo test --test vectors",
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Updated: now that the three client PRs have merged, this PR also **bumps the manifest refs to their squash-merge commits** (pad→`91620c6`, cli→`ebc516e`, desktop→`ab83da6`). This is needed because the previous pinned refs were the pre-merge PR-branch tips, which were deleted on merge and are no longer fetchable. The downstream-smoke run on this PR now clones the clients at their real merged `main` commits.

JohnMcLear:
Fixed in the latest commit, thanks @qodo — manifest commands now run under `bash -euo pipefail -c` so pipeline-stage failures (and unset-var mistakes) inside a client's `vectorTest`/`smokeCmd` surface as a non-zero exit instead of being masked. Verified `bash -euo pipefail -c 'false | cat'` now exits non-zero, and the per-client failure aggregation still works (two bogus clients → 2 reported errors, exit 1).

Issue #7586, opened 2026-04-23
v2.7.0. breaks i10n auto-detect of language for pad UI


n-bux:
Hi, thanks for your work and the new 2.7.0 release!

One problem:  the pad UI is in English, **not detecting my German system**. The index page or the timeslider are fine, though.

I tried with my own build off of the official Dockerfile. Same setup works with 2.6.1

_When I set it to german (via UI->(cookie) or settings -> lang), it works, off course, only some translations are missing. But these are new features not yet translated, right!?_
_Examples of missing translations:_ 
- _Settings dialogue:_
  - _Disable Chat_
- _Timeslider setting dialogue:_
  - _playback speed (incl. dropdown options)_

JohnMcLear:
Whoa, that sounds rough..  Let me look into it urgently!  Any chance you are comfortable running a branch to test from?   If so you can clone and test https://github.com/JohnMcLear/etherpad/tree/fix/a11y-dialogs-labels-lang

n-bux:
thanks for quick answer! 
yes, your branch works correctly regarding language detection. but it seems to be based on 2.6.1 (according to http://localhost:9001/health, and also because it is not including the new features like "Diable chat" or playback speed), which was working with me as well..

JohnMcLear:
I'm really surprised we didn't have testing in for this, I've made and merged an emergency patch and will ask @SamTV12345 to release with it.  The patch includes testing to ensure this doesn't happen again..

Apologies..  I'm frustrated this slipped through our testing nets but at least it was a quick fix and a quick release over the coming 24 hours should soften the blow.. 

cc @translatewiki re missing translations, afaik they land most weeks, I think our development cycle is just ahead of them at the moment :)

Thank you so much for the quick feedback on the release @n-bux - without it we might not have spotted this for far too long!

SamTV12345:
Thanks for the patch. I will do a release in the evening. Approx. 6 hours from now.

JohnMcLear:
I will try land some accessibility patches before we release :)

lucaguindani:
Don't know if it's related, but since 2.7.0 I have translations issues in `/admin`.

<img width="1251" height="641" alt="Image" src="https://github.com/user-attachments/assets/50a6658a-599a-4f4f-8d16-eeaf9cfaba0b" />

I just tried from a fresh clone from `develop` running with docker.

JohnMcLear:
[...]

PR #7928, opened 2026-06-09
fix(pad): show detected language in settings dropdown (#7925)

### LINKED ISSUE #7925: language dropdown shows English instead of detected language
> Hey, thanks for fixing this so quickly! It now detects correctly, but the language dropdown  still says the wrong language:
> 
> Image 

 _Originally posted by @n-bux in [#7586](https://github.com/ether/etherpad/issues/7586#issuecomment-4602135493)_

JohnMcLear:
## Problem

Fixes #7925 (follow-up to #7586).

The pad UI correctly auto-detects the browser language and renders in it (e.g. German), but the **Language dropdown** in Settings still shows **English**.

## Root cause

Two code paths set the dropdown value and disagree when the language is *auto-detected* (no `language` cookie, no pad-wide `lang`):

- `pad_editor.ts` (on html10n's `localized` event) correctly sets both dropdowns to `html10n.getLanguage()` (e.g. `de`).
- `pad.ts` `refreshMyViewControls()` / `refreshPadSettingsControls()` set them to `effectiveOptions.lang || 'en'` / `padOptions.lang || 'en'`. When nothing is explicitly chosen, `getMyViewOverrides()` deletes `overrides.lang`, so the value is `undefined` and the dropdown falls back to hardcoded **`'en'`**. These refreshes run *after* the `localized` handler (during handshake), clobbering the correct value back to English.

## Fix

Fall back to the detected language before `'en'`:

```js
$('#languagemenu').val(effectiveOptions.lang || html10n.getLanguage() || 'en');
$('#padsettings-languagemenu').val(padOptions.lang || html10n.getLanguage() || 'en');
```

`getLanguage()` returns the matched lowercase locale key, which matches the `<option value>` keys built from the available locales — so the right option is selected. An explicit user choice (cookie/pad option) still takes precedence.

## Testing

Added a regression test that loads a pad under a `de-DE` browser locale and asserts the dropdown reads **"Deutsch"** with no manual selection.

RED-green verified against a running server:
- **Without the fix:** the new test fails — dropdown shows `English` (reproduces #7925).
- **With the fix:** all 5 `language.spec.ts` tests pass.

### DIFF
--- a/src/static/js/pad.ts
+++ b/src/static/js/pad.ts
@@ -581,7 +581,10 @@ const pad = {
     $('#padsettings-options-linenoscheck').prop('checked', view.showLineNumbers !== false);
     $('#padsettings-options-rtlcheck').prop('checked', !!view.rtlIsTrue);
     $('#padsettings-viewfontmenu').val(view.padFontFamily || '');
-    $('#padsettings-languagemenu').val(padOptions.lang || 'en');
+    // When no pad-wide lang is set, reflect the language html10n actually
+    // detected and rendered (e.g. from the browser) instead of defaulting the
+    // dropdown to English while the UI is in another language. See #7925.
+    $('#padsettings-languagemenu').val(padOptions.lang || html10n.getLanguage() || 'en');
     $('#padsettings-enforcecheck').prop('checked', !!padOptions.enforceSettings);
     $('#padsettings-options-stickychat, #padsettings-options-chatandusers')
         .prop('disabled', padOptions.showChat === false);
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Thanks @qodo — good catch. The "detection works" check computed a boolean via `locator.evaluate()` but never asserted it. Fixed in 1890c317 by asserting the bold button's parent `title` is the German `Fett (Strg-B)` via `toHaveAttribute`, so the precondition can now actually fail. Re-ran `language.spec.ts` (chromium) locally — all 5 pass.

PR #7938, opened 2026-06-10
fix(admin): don't let one unreadable pad empty the Manage-pads list (#7935)

### LINKED ISSUE #7935: Display issue of notes
Hi,

Looks like there's an issue with the synchronization about the existence of notes or something else.

welcome page : 

Image

admin panel : 
No notes listed

Image

What's happening please ?

Etherpad 3.3.0 - docker + traefik + postgresql 18

JohnMcLear:
Closes #7935.

## Symptom
Pads exist (visible on the welcome page, returned by the `listAllPads` API) but the admin **Manage pads** UI shows none. Reported on 3.3.0 with Docker + Traefik + PostgreSQL 18.

## Root cause
The welcome-page "recent pads" list is browser **localStorage**, so it reflects pads this browser opened — not the DB. The admin list comes from the `/settings` socket `padLoad` handler → `listAllPads()` → `findKeys('pad:*', '*:*:*')`.

The admin's default `lastEdited` sort forces a **full scan**, hydrating *every* pad via `getPad()`. `findKeys('pad:*','*:*:*')` returns every `pad:`-prefixed key, including legacy/foreign/**migration-corrupted** records. A record whose stored value is a non-object (a JSON *string* — exactly what a botched dirty.db→Postgres migration produces) makes `Pad.init` throw ``Cannot use 'in' operator to search for 'pool'`` (`Pad.ts:591`).

The handler had **no try/catch**, so a single bad record:
1. emitted nothing → the SPA showed an empty "No results" list forever (the reported symptom), and
2. produced an unhandled rejection that reaches `server.ts` → `process.exit` — taking the whole server down.

Verified against a real **PostgreSQL 18.4** instance: ueberdb2 6.1.8 `findKeys`, `listAllPads`, and `getStats` all work correctly there. PG18 is not the cause — the trigger is one unreadable `pad:*` record, which can happen on **any** DB backend.

## Fix
- `loadMeta` wraps `getPad()` in try/catch: a failed read logs a warning and returns zeroed metadata, so the bad pad **surfaces for deletion** instead of hiding every other pad.
- The whole `padLoad` handler is wrapped so it always emits a terminal `results:padLoad` (the SPA can't hang) and never bubbles to the process-level rejection handler (the server can't exit from this path).

## Tests
- **backend** — `tests/backend/specs/admin/padLoadResilience.ts`: injects a corrupt `pad:*` record next to a readable pad and asserts the listing still returns both. Fails without the fix (`no results:padLoad reply within 10s` + `server Exiting`), passes with it.
- **e2e** — `tests/frontend-new/admin-spec/admin_pads_page.spec.ts`: create/open a pad → it appears in the welcome-page recent-pads list → it appears in the `/admin` Manage-pads UI.

Local gates: 7/7 admin backend specs pass (5 existing + 2 new), both Playwright admin specs pass, `ts-check` clean.

### DIFF
--- a/src/node/hooks/express/adminsettings.ts
+++ b/src/node/hooks/express/adminsettings.ts
@@ -11,6 +11,7 @@ import settings, {getEpVersion, getGitCommit, reloadSettings} from '../../utils/
 import {getLatestVersion} from '../../utils/UpdateCheck';
 import {redactSettings} from '../../utils/AdminSettingsRedact';
 const padManager = require('../../db/PadManager');
+const db = require('../../db/DB');
 const api = require('../../db/API');
 import {deleteRevisions} from '../../utils/Cleanup';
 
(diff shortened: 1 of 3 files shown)


JohnMcLear:
Thanks @qodo — both findings actioned:

**1. Undeletable corrupt pads** (`4378466`)
Valid — surfacing a corrupt pad is pointless if it can't be removed, and `deletePad`'s normal path (`doesPadExists` → `getPad` → `Pad.remove`) can't touch a non-object record. `deletePad` now falls back to a raw key purge when `doesPadExists()` is false but a `pad:${id}` value is present: sweep sub-keys (`revs`/`chat`/`deletionToken`/…) via `findKeys`, drop the `pad2readonly` mapping, then `padManager.removePad()` (main key + pad-list + cache). It now always emits a terminal `results:deletePad` reply (including for an already-absent id) so the UI clears the row instead of silently doing nothing. Covered by a new backend test ("a surfaced corrupt pad can be deleted from the admin UI").

**2. Logging the raw corrupt value** (`295ad8b`)
Valid — the `TypeError` message embeds the stored value (`…'pool' in <value>`), which for a real pad can be document text. Added a `safeErr()` helper (error name + single-line, 120-char-capped message, control chars stripped) used in the per-pad warning, the outer handler error log, the `deletePad` error log, and the `error` field returned to the SPA — so no raw pad content, log bloat, or newline injection.

All admin backend specs pass (8: 5 existing + 3 new) and `ts-check` is clean.

PR #7940, opened 2026-06-10
fix: include images in HTML export

forkivan:
## Problem

Images inserted via image plugins (e.g. ep_image_upload) store their URL in the attribute pool under the `image` key (URL-encoded), with optional `image-width` and `image-height` attributes.

Previously `ExportHtml.ts` had no handling for the `image` attribute: the placeholder character was output as escaped text, silently dropping every image from the exported HTML.

## Fix

Detects `image`, `image-width`, and `image-height` attributes on each op inside `processNextChars`. When an image URL is present the placeholder character is consumed and an `` tag is emitted instead, with inline width/height styles when available.

The URL is decoded with `decodeURIComponent` and sanitized with `Security.escapeHTMLAttribute` before output.

## Testing

Tested with pads containing:
- Base64-embedded images (`data:image/png;base64,...`)
- Server-path images (`/static/images/padname/uuid.png`)

### DIFF
--- a/src/node/utils/ExportHtml.ts
+++ b/src/node/utils/ExportHtml.ts
@@ -222,11 +222,19 @@ const getHTMLFromAtext = async (pad:PadType, atext: AText, authorColors?: string
       for (const o of ops) {
         const usedAttribs:string[] = [];
 
-        // mark all attribs as used
+        // mark all attribs as used; also collect image attributes
+        let imageUrl: string | null = null;
+        let imageWidth: string | null = null;
+        let imageHeight: string | null = null;
         for (const a of attributes.decodeAttribString(o.attribs)) {
           if (a in anumMap) {
             usedAttribs.push(String(anumMap[a])); // i = 0 => bold, etc.
           }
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Is there not a way to do this using a hook and code within the plugin?

JohnMcLear:
Also, no test coverage.

Please see contributing.md and agents.md

JohnMcLear:
I'm 99% sure this needs to be contributed to ep_image_upload not here.

forkivan:
Hi @JohnMcLear, thank you for the feedback.

You are right that the fix belongs in the plugin, not in core. However, the plugin in question is not ep_image_upload — it is ep_images_extended by @dcastelone, which is what our server uses.

That plugin already has an exportHTML.js file with a correct getLineHTMLForExport implementation, but the hook was never registered in ep.json. I opened a one-line fix PR there:

https://github.com/dcastelone/ep_images_extended/pull/2

The author was last active in March 2026 so hopefully he will respond. Is there anything else I can do to get this merged, or would you recommend a different approach?

PR #7948, opened 2026-06-12
fix(pad): show saved-revision markers in in-pad history mode (#7946)

### LINKED ISSUE #7946: Saved revisions are not displayed as markers in the timeslider
**Describe the bug**

Saved revisions are not displayed as markers in the timeslider.

After clicking the **Save Revision** button, the revision appears to be saved successfully.
The timeslider can be opened and document history can be navigated normally, however no markers indicating saved revisions are shown on the timeline.

**To Reproduce**

1. Create a new pad.
2. Add some content.
3. Click **Save Revision**.
4. Make additional edits.
5. Click **Save Revision** again.
6. Open the timeslider.
7. Observe that no saved revision markers are visible on the timeline.

**Expected behavior**

Saved revisions should be displayed as markers in the timeslider, allowing users to quickly navigate to explicitly saved points in time.

**Screenshots**

Response after clicking the "Save revision" button:
Image

Time-slider:
Image

* Etherpad version: 3.3.1
* OS: RHEL 9
* Node.js version (`node --version`): Official Docker Image
* npm version (`npm --version`): Official Docker Image
* Is the server free of plugins: yes
* Are you using any abstraction IE docker: yes (docker)

* OS: Windows 11
* Browser: Firefox
* Version: 151.0.2 (64-bit)

**Additional context**

The issue was reproduced after upgrading to Etherpad 3.3.1.
As I did not previously ran 3.3.0 it might be a regression bug, which started to appear in 3.3.0.

The **Save Revision** button is available and appears to function, but no corresponding markers are shown in the timeslider. The timeslider itself works and allows navigation through document history.

It is unclear whether this is a regression, an intended behavior change, or a separate issue affecting the display of saved revisions.

JohnMcLear:
### What was broken
Saved revisions (the **Save Revision** star button) stopped appearing as markers in the timeslider in 3.3.x.

### Root cause
When #7659 moved the timeslider **into the pad** as an embedded iframe, the user-facing control became the outer `#history-slider-input` range input in `pad.html`. The saved-revision stars are still drawn by `broadcast_slider.ts` into the iframe's `#ui-slider-bar`, **but that DOM is hidden** in `embedded-history-frame` mode. `pad_mode.ts` bridges rev / max / value / timer / authors onto the outer slider — but **never the saved revisions**. So saving worked, yet no markers ever showed.

The underlying saved-revisions code path (server `getSavedRevisions`, `clientVars.savedRevisions`, `broadcast_slider.addSavedRevision`) is byte-for-byte unchanged from 3.2.0 — the regression is purely the missing outer-slider bridge.

### The fix
- `pad.html`: wrap `#history-slider-input` in a positioned `.history-slider-wrap` with a `#history-slider-stars` overlay.
- `pad.css`: style `.history-star` markers (reusing the existing timeslider star glyph), inset to align with the native range thumb.
- `pad_mode.ts`: bridge the embedded timeslider's `clientVars.savedRevisions` onto the outer slider as percentage-positioned markers, rendered on first sync and re-rendered when the slider max changes. A signature guard keeps it cheap on every scrub; percentage positions reflow on resize for free.

Markers are a purely visual, `aria-hidden` overlay (keyboard / screen-reader users already reach any revision via the slider + step buttons, matching the legacy timeslider's mouse-only stars) with a click-to-seek convenience.

| Before | After |
|---|---|
| no markers on the slider | ⭐ marker at the saved revision |

### Why CI didn't catch it (and how this PR closes that gap)
- The **only** test that asserted a star renders (`src/tests/frontend/specs/timeslider_revisions.js`) lives in the **legacy mocha suite, which no workflow runs** — `frontend-tests.yml` only runs Playwright (`tests/frontend-new/specs/`).
- The **modern timeslider specs drive the `?embed=1` iframe directly** and never exercise the outer in-pad history UI where the bug lives.
- The backend API tests only assert saved-revision **counts**, never the rendered markers.

This PR adds `tests/frontend-new/specs/timeslider_saved_revisions.spec.ts`, which exercises the **real user flow**: write content → click **Save Revision** → enter in-pad history mode via the toolbar → assert a visible, correctly-positioned marker on the outer slider. Verified TDD-style: the test **fails** on `develop` (`.history-star` not found) and **passes** with the fix. All 15 existing `timeslider_*` specs still pass.

### DIFF
--- a/src/node/db/Pad.ts
+++ b/src/node/db/Pad.ts
@@ -868,6 +868,8 @@ class Pad {
     await this.saveToDatabase();
   }
 
+  // Returns the newly created saved revision, or undefined if this revision
+  // was already saved (so callers can broadcast only genuine additions).
   async addSavedRevision(revNum: string, savedById: string, label: string) {
     // if this revision is already saved, return silently
     for (const i in this.savedRevisions) {
@@ -887,6 +889,7 @@ class Pad {
     // save this new saved revision
     this.savedRevisions.push(savedRevision);
     await this.saveToDatabase();
+    return savedRevision;
   }
 
   getSavedRevisions() {
--- a/src/node/handler/PadMessageHandler.ts
+++ b/src/node/handler/PadMessageHandler.ts
@@ -622,7 +622,18 @@ exports.handleMessage = async (socket:any, message: ClientVarMessage) => {
 const handleSaveRevisionMessage = async (socket:any, message: ClientSaveRevisionMessage) => {
   const {padId, author: authorId} = sessioninfos[socket.id];
   const pad = await padManager.getPad(padId, null, authorId);
-  await pad.addSavedRevision(pad.head, authorId);
+  const savedRevision = await pad.addSavedRevision(pad.head, authorId);
+  // Notify every client in the pad room — including any open timeslider —
+  // so saved-revision markers appear live instead of only on the next
+  // timeslider load (#7946). The client's NEW_SAVEDREV handler existed but
+  // was never reached because this broadcast was missing; live editors that
+  // don't handle the type ignore it. Skip the emit for duplicate saves.
+  if (savedRevision) {
+    socketio.sockets.in(padId).emit('message', {
+      type: 'COLLABROOM',
+      data: {type: 'NEW_SAVEDREV', savedRev: savedRevision},
+    });
+  }
 };
 
 /**
--- a/src/static/css/pad.css
+++ b/src/static/css/pad.css
@@ -214,11 +214,51 @@ body.history-mode #history-controls { display: flex; }
 .history-controls button.buttonicon.buttonicon-play.pause::before {
   content: "\e829";
 }
-.history-slider-input {
+.history-slider-wrap {
+  position: relative;
   flex: 1 1 auto;
   min-width: 80px;
   margin: 0 6px;
+  display: flex;
+  align-items: center;
+}
+.history-slider-input {
+  flex: 1 1 auto;
+  min-width: 0;
+  width: 100%;
   cursor: pointer;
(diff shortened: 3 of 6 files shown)


JohnMcLear:
Addressed all three findings in 212c8c9:

**1. Stars miss live updates** — Verified, and the root cause was deeper: `handleSaveRevisionMessage` never broadcast `NEW_SAVEDREV`, so the client's `NEW_SAVEDREV` handler was dead code and *no* timeslider (inner or outer) ever updated saved-rev markers live. Fixed properly:
- `Pad.addSavedRevision` now returns the new revision (`undefined` on duplicate).
- `handleSaveRevisionMessage` broadcasts `NEW_SAVEDREV` to the pad room. Live editors fall through their `if/else` chain and ignore the unknown type.
- `pad_mode.ts` now sources the outer markers from the embedded slider's **live** `#ui-slider-bar .star` DOM (pulling labels from the clientVars snapshot) and adds a `MutationObserver` on that bar, so a revision saved by a collaborator appears on an already-open history slider without a reload.

**2. Rev0 marker never renders** — Verified. The `max <= 0` guard conflated "nothing to show" with a single-revision pad. Now bails only on an empty set / `max < 0`; `max === 0` renders the rev-0 marker at the left edge (`frac` already handled it).

**3. Test allows left=0** — Verified. The `|| getBoundingClientRect().left` fallback let a collapsed `left:0%` pass. The assertion now parses `style.left` directly and checks `0 < pct < 100`.

Added a two-client Playwright test for the live path (confirmed it fails when the server broadcast is removed). Backend `Pad` + pad API specs (88) and all 15 `timeslider_*` specs pass.

PR #7949, opened 2026-06-12
test(timeslider): port legacy mocha specs to Playwright, retire orphaned suite

JohnMcLear:
## Close the timeslider CI coverage gap

Follow-up to #7948. While fixing #7946 we found that the **legacy mocha suite (`src/tests/frontend/specs/`) is run by no CI workflow** — `frontend-tests.yml` only runs Playwright (`tests/frontend-new/specs/`). That's why a regression in the in-pad history UI (#7659) reached a release: the timeslider's only coverage lived in dead specs, and the modern Playwright specs drive the isolated `?embed=1` iframe and never exercise the outer in-pad UI users actually interact with.

This ports the still-meaningful timeslider cases to Playwright, **re-targeted at the real in-pad UI** (outer history banner / slider / export links that `pad_mode.ts` drives), and deletes the orphaned originals.

### New specs
| New (`frontend-new`) | Ported from (`frontend`, deleted) | Asserts |
|---|---|---|
| `timeslider_revision_labels.spec.ts` | `timeslider_labels.js` | `#history-banner` shows `Version N` + a valid (non-NaN) date & timer, and both update when scrubbing to rev 0 |
| `timeslider_export_links.spec.ts` | `timeslider_numeric_padID.js` + the "checks the export url" case of `timeslider_revisions.js` | outer export hrefs target `/p/<pad>/<rev>/export/<type>` for the viewed revision — incl. a **numeric pad id** — and follow the slider to rev 0 |
| `timeslider_deeplink.spec.ts` | the "jumps to a revision given in the url" case of `timeslider_revisions.js` | a `#rev/N` hash, and the legacy `#N` shortlink, boot straight into history mode at that revision (`bootstrapFromHash`) |

The star-marker case of `timeslider_revisions.js` is already covered by `timeslider_saved_revisions.spec.ts` in #7948.

### Verification
All new specs pass on **Chromium and Firefox** locally; type-check clean. These now run in PR CI, so the in-pad history banner, export-link rewriting, and deep-link entry are gated against regression.


JohnMcLear:
Both addressed in 24f4d6d:

**1. Locale-fragile banner assertions** — Verified. The `Version N` label and `Saved <Month> <day>, <year>` date are localized (`timeslider.version` / `timeslider.saved`), so they're only deterministic under English. Pinned the spec to `en-US` via `test.use({locale: 'en-US'})`, which keeps the exact assertions valid and the date `Date`-parseable regardless of the runner's default locale.

**2. Numeric padId collision risk** — Verified. The `base + random(1000)` range could collide across reruns against a persistent DB. Switched to a high-entropy numeric id (`${Date.now()}${random(1000)}`), still numeric so it exercises the numeric-padId href-rewrite path.

Re-verified both specs on Chromium and Firefox.

PR #7953, opened 2026-06-12
fix(ui): implement smooth overflow scrolling and flex layout for mobi…

### LINKED ISSUE #7756: Performance: Look deeper into scaling N+ contributors per pad.

Rahul123-dev-create:
### Description
This PR resolves issue #7756 by updating the responsive CSS rules inside the Colibris skin toolbar component.

### Changes Introduced
- Switched the mobile `.toolbar ul` layouts to use flexbox configurations.
- Added `overflow-x: auto` and optimized momentum scrolling (`-webkit-overflow-scrolling: touch`) for mobile containers to allow responsive horizontal swiping of overflowing styling tools.
- Adjusted layout padding and margins to maintain touch-friendly tap targets across mobile viewports.

Fixes #7756

### DIFF
--- a/src/static/skins/colibris/src/components/toolbar.css
+++ b/src/static/skins/colibris/src/components/toolbar.css
@@ -14,7 +14,9 @@
 }
 
 .toolbar ul {
+  display: flex;
   align-items: center;
+  flex-wrap: wrap; /* Fixes layout crunching by allowing items to scale elegantly */
 }
 
 .toolbar ul.menu_left {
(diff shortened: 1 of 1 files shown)


JohnMcLear:
I think you can see what is wrong with this PR.

PR #7957, opened 2026-06-13
fix(ui): implement adaptive YIQ luminance validation for author color picker (#7778).

### LINKED ISSUE #7778: a11y: Provide an AT-only read view of pad content (line-by-line + link navigation)

Rahul123-dev-create:
This PR resolves issue #7778 by introducing an adaptive color-contrast layer for assigned or selected user palette swatches.

Changes Introduced
Implemented a standard YIQ luminance formula calculator inside pad_userlist.ts.

Added theme detection logic to identify dark mode context rules.

Dynamically lightens or darkens chosen hex values that fall outside readable contrast ranges against active editor backgrounds.

Fixes #7778

### DIFF
--- a/src/static/js/pad_userlist.ts
+++ b/src/static/js/pad_userlist.ts
@@ -8,7 +8,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS-IS" BASIS,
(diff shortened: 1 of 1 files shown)


JohnMcLear:
Please read your PRs before submitting.

Rahul123-dev-create:
> Please read your PRs before submitting.

What can I do now ?

PR #7962, opened 2026-06-15
PadManager: reject unreachable '.' and '..' pad ids

forkivan:
## What
`isValidPadId` accepted pad ids that consist only of URL dot-segments (`.` and `..`).

## Why it's a bug
Per the WHATWG URL standard a browser normalises `/p/.` to `/p/` and `/p/..` to `/`
before the request is sent. So a pad named `.` or `..` can be created, but it can
never be opened or exported: the request arrives at `/p/` and Etherpad answers
`Cannot GET /p/`. The pad exists in the database but is permanently unreachable.

## Reproduction
1. Navigate to `/p/.` (this also creates the pad).
2. The browser normalises the URL to `/p/` → **Cannot GET /p/**.
3. Same for `..` (`/p/..` → `/`).

## Fix
Reject `.` and `..` in `isValidPadId`. Ids that merely contain dots (e.g. `a.b`,
`.foo`, `foo.`) remain valid — only the two pure dot-segments are rejected.

## Test
Added `src/tests/backend-new/specs/PadManager.test.ts` with a regression test that
fails if the fix is reverted.


JohnMcLear:
This seems logical but let me be sure this isn't expected behaviour before merging :)

JohnMcLear:
Thanks for this — the fix itself is sound (the `[^$]{1,50}` regex does accept `.`/`..`, and rejecting only the two pure dot-segments is the right minimal call). Two things before it can go in:

### 1. CI blocker — the new test fails to load
Both Linux backend jobs are red on the new spec:

```
FAIL tests/backend-new/specs/PadManager.test.ts
Error: Cannot find module '../../../node/db/PadManager'
  const padManager = require('../../../node/db/PadManager');
```

The path is correct, but `require()` drops to Node's CommonJS resolver, which can't resolve a `.ts` file under vitest — Vite handles that, so every sibling spec in `backend-new/specs/` uses an ESM `import`. The `require()` was presumably to defer loading until after the `vi.mock` calls, but that isn't needed: vitest hoists `vi.mock` above all imports automatically.

Minimal fix — replace the `require` with a dynamic import:

```js
const padManager = await import('../../../node/db/PadManager');
```

(matches the pattern already used in `sanitizeProxyPath.test.ts`). A top-level static import works too — the mocks still apply because of hoisting.

### 2. Legacy `.`/`..` pads become undeletable (Qodo's point, and it's valid)
Since `getPad()`/`getPadSafe()` run `isValidPadId` first, any pad with id `.` or `..` that was created *before* this change can no longer be deleted via the admin UI (`deletePad` socket handler, adminsettings.ts) or the HTTP API. The admin handler has no try/catch around `getPad`, so it just rejects with no terminal `results:deletePad` response.

Severity is low — such pads are already unreachable and can only pre-exist — but it leaves an undeletable orphan. Worth closing so the fix is complete: `padManager.removePad(padId)` already removes the raw `pad:<id>` key without going through `getPad`/`isValidPadId`, so the admin `deletePad` handler could fall back to it (and emit a terminal result) when `getPad` throws. Optional, but it turns this from "can't create new broken pads" into "can also clean up old ones".

forkivan:
Thanks for the thorough review — both addressed:

1. **CI blocker (test load):** switched the spec from `require()` to a top-level
   `await import('../../../node/db/PadManager')`, so vitest resolves the `.ts`
   source. The `vi.mock` calls are hoisted above it as you noted.

2. **Legacy `.`/`..` pads undeletable:** the admin `deletePad` handler now wraps
   the healthy-path `getPad()`/`Pad.remove()` in a try/catch and falls back to the
   existing raw key purge (`removePad` + sub-key sweep) when `getPad()` throws,
   always emitting a terminal `results:deletePad`. Added a regression test in
   `padLoadResilience.ts` that seeds a `.`-id pad directly and asserts it can be
   deleted (times out without the fallback).

Re: "is this expected behaviour" — `.`/`..` pads are unreachable by construction
(the browser normalises `/p/.` → `/p/` before the request is even sent), so they
can never have been intentionally usable; rejecting only those two exact ids
leaves every other dotted id (`a.b`, `.foo`) working as before.

JohnMcLear:
Sadly still failing CI :(  Are you aware you can run the tests locally instead of waiting for CI?

forkivan:
Thanks for the reviews — both points are addressed now, and I ran the tests locally this time instead of waiting for CI 👍

1. The backend-new spec couldn't load. You're right: under vitest require() falls back to Node's CommonJS resolver, which can't resolve the .ts source, and vi.mock didn't intercept it either, so the suite errored before any test ran (plus the top-level await import tripped tsc TS1378). Rather than fight vitest's mocking for a module that pulls in the DB layer, I moved the test to the mocha backend suite (tests/backend/specs/PadManager.ts), which runs with --import=tsx and resolves the .ts require()s natively — no mocking needed. isValidPadId is a pure function and DB only connects lazily in DB.init(), so loading the module there has no side effects.

2. Legacy ./.. pads becoming undeletable (Qodo's point — valid). Since getPad() now runs isValidPadId() first, a pre-existing ./.. pad would throw there and the admin deletePad handler would reject with no terminal results:deletePad. Fixed: the handler now wraps the healthy getPad/remove path in a try/catch and falls back to the raw pad:<id> key purge (which bypasses isValidPadId) when getPad throws, always emitting a terminal result. Added a regression test in padLoadResilience.ts that seeds a . pad and asserts it can still be deleted from the admin socket.

Local run: the PadManager unit spec (3) and padLoadResilience (4, including the new legacy-. deletion case) are green, and tsc --noEmit passes.

PR #8039, opened 2026-07-09
Migrate server to TypeScript 7 (tsgo)

zi-gae:
TypeScript 7 (the native Go compiler) went GA this week — roughly 10x faster type-checking, which should help this repo's `perform-type-check` workflow and local `ts-check` runs noticeably.

**What changed**
- `typescript` → `^7.0.0` in `ep_etherpad-lite` and `bin` (+ lockfile)
- Two source fixes for diagnostics the new compiler reports (both type-level only, no behavior change):
  - `src/static/js/TextLinesMutator.ts`: TS7 flags the unguarded spread of `splitTextLines()`'s nullable result at line 304. Silenced with `// @ts-ignore` to match the sibling `push` calls a few lines above (a real cleanup of this file's nullability is beyond this PR's scope).
  - `src/tests/backend/specs/hooks.ts`: TS7 reports the union-overload mismatch on the `concat` per array element, so the existing single `// @ts-ignore` no longer covers it. Replaced with an explicit `as any[]` cast on the receiver.
- `pnpm-workspace.yaml`: minimum-release-age exclusions for the freshly released typescript 7 packages (same pattern as the existing entries)

**Deliberately out of scope**
- `admin` and `ui` stay on TypeScript 6: their build path uses `openapi-typescript`, which drives the JS compiler API (`ts.factory…`) that tsgo doesn't expose until the new API ships in 7.1. Bumping them today crashes the vite build.
- Local `eslint` (via `eslint-config-etherpad` → `@typescript-eslint`) is likewise not TS7-ready yet; CI has no lint job so nothing breaks there, but worth knowing for local dev in `src/`.

**Verification (macOS arm64, Node 24.18)**
- `pnpm ts-check` (tsc --noEmit): clean on 7.0.2, and the two fixes also check clean on 6.0.3
- Full backend suite (`pnpm test`): **1566 passing, 22 pending, 0 failing**
- `pnpm build:etherpad` (admin + ui): builds pass
- One flaky observation unrelated to this change: `SessionStore.ts:161` failed once under full-suite load and passes consistently in isolation (timing-sensitive cookie-expiry assertion)

The config/dependency rewrites were generated with [ts6to7](https://github.com/zi-gae/ts6to7) (disclosure: I'm its author) and everything above was verified manually.


src/static/js/TextLinesMutator.ts
```diff
@@ -301,6 +301,7 @@ class TextLinesMutator {
         this._curSplice.push(theLine.substring(lineCol));
         this._curCol = 0; // TODO(doc) why is this not set to the length of last line?
       } else {
+        // @ts-ignore
```
JohnMcLear: This seems like a bad idea

JohnMcLear:
Qodo caught some bits, I commented on another.  Why all the additional reference in the workspace file too?

zi-gae:
Thanks for the look. On the workspace file: those entries are the minimum-release-age allowlist. typescript@7.0.2 was published this week, so without the exclusions pnpm's supply-chain policy refuses to install it, and the native compiler ships one platform package per target (like esbuild does), which is why the list is long. Same pattern as the existing testcontainers and mysql2 entries. They can be deleted once the packages age past the policy window.

Also pushed fixes for both Qodo findings: TextLinesMutator now null-guards the splitTextLines result instead of hiding it behind @ts-ignore, and hooks.ts got an explicit HookFnTestCase type instead of the any[] cast so the test cases stay shape-checked. Re-verified: tsc clean on both TS 6.0.3 and 7.0.2, backend suite 1565 passing. One note: SessionStore's shutdown/expiry specs failed intermittently for me under full-suite load (both before and after these fixes) and pass consistently in isolation, so it looks like a timing-sensitive flake rather than something this PR introduces. I have not verified whether it reproduces on develop, happy to dig further if it is not a known one.

JohnMcLear:
I imagine that due to the "newness" of TS7 we will sit on this for at least a few weeks just to let the dust settle and upstream bugs get resolved..  

I don't want to merge anything with a ts-ignore if we can avoid it, we should, if the plan for that is we ship initial ts7 then have a 2nd pr lined up for removing ts-ignore I'm fine w/ that but right now that's a blocker for me.

These are my opinions, others may have differing opinions and they are welcome :)

JohnMcLear:
Looking good @zi-gae - cc @SamTV12345

Please rebase @zi-gae

JohnMcLear (approved):

SamTV12345 (approved):
Looks good. There is just one ! For forcing no undefined.

zi-gae:
Rebased onto latest develop. The pnpm lockfile was regenerated on top of the recent dependabot bumps, and locally `tsc --noEmit` is clean and the easysync tests pass. CI runs need maintainer approval to start.

JohnMcLear:
@SamTV12345 feel free to handle PR merging across the board including this one if you are happy etc.

SamTV12345:
Pushed a merge of `develop` into this branch to clear the lockfile conflict (regenerated `pnpm-lock.yaml`, no other changes) — your commits are untouched.

Reviewed the current state: `_curSplice` typed as the tuple it actually is with 21 `@ts-ignore`s gone, `insert(text: string, L: number)`, the null-safe `splitTextLines()`, and `HookFnTestCase` on both case arrays — that is exactly the right fix, and the diagnosis that `admin`/`ui` must stay on TS 6 because `openapi-typescript` drives `ts.factory.*` is confirmed: bumping them fails the vite build with `Cannot read properties of undefined (reading 'createKeywordTypeNode')`. Dependabot's group PR (#8069) walked straight into it.

Verified locally on the merged branch (Node 26): `tsc --noEmit` with typescript 7.0.2 clean in `src` and `bin`, full backend suite 1610 passing / 22 pending / 0 failing.

Merging once CI is green. Thanks — nice first contribution, and please do send more.

JohnMcLear:
@zi-gae see https://github.com/ether/etherpad/pull/8142

PR #8070, opened 2026-07-25
security(oidc): stop shipping a hardcoded cookie key and reflecting all CORS origins

JohnMcLear:
Fixes two issues in the **embedded OIDC provider** (`src/node/security/OAuth2Provider.ts`), reported privately by **meifukun** (credit requested as `meifukun`, https://github.com/meifukun). Verified against 3.3.2 @ `9a6109b`; both lines are unchanged on current `develop`.

## 1. Hardcoded OIDC cookie signing key (`['oidc']`)
The provider signed its interaction/session/grant cookies (`_interaction`, `_interaction_resume`, `_grant`, `_session`) with the committed literal key `oidc`, so anyone with the public source could compute valid `.sig` cookie signatures and defeat the provider's cookie-integrity boundary.

**Fix:** `resolveOidcCookieKeys()` prefers an operator-supplied `settings.sso.cookieKeys` (ordered array for rotation), otherwise derives a secret key from the persisted session secret (`SESSIONKEY.txt`) via a domain-separated SHA-256 — stable across restarts and horizontally-scaled pods, never committed to source. Falls back to an ephemeral random key when no session secret exists.

## 2. `clientBasedCORS` returned `true` for every origin
The token/userinfo endpoints reflected any `Origin` into `Access-Control-Allow-Origin`, letting unregistered origins read responses that use non-cookie credentials (Authorization headers, POST-body client secrets).

**Fix:** `isOriginAllowedForOidcClient()` allows a CORS origin only when it matches the origin of one of the client's registered redirect URIs.

## Tests / verification
- New `OidcProviderSecurity.ts` — 14 unit tests over both pure helpers (stable derivation, no hardcoded key, exact-origin matching, look-alike/scheme-mismatch rejection).
- Verified end-to-end against a real `oidc-provider@9.9.1` instance: cookies.keys accepted, `Client.redirectUris` read correctly, attacker origin blocked.
- `tsc --noEmit` clean.

A private GHSA draft tracks both issues.

### DIFF
--- a/settings.json.template
+++ b/settings.json.template
@@ -966,6 +966,14 @@
 
   "sso": {
     "issuer": "${SSO_ISSUER:http://localhost:9001}",
+      /*
+       * Optional signing keys for the embedded OIDC provider's cookies. When
+       * omitted, Etherpad derives a secret key from the persisted session
+       * secret (SESSIONKEY.txt), which is stable across restarts and shared
+       * across horizontally-scaled pods. Set an explicit ordered array to
+       * rotate: the first key signs, the rest are still accepted for verify.
+       * "cookieKeys": ["${OIDC_COOKIE_KEY:}"],
+       */
       "clients": [
         {
           "client_id": "${ADMIN_CLIENT:admin_client}",
--- a/settings.json.template
+++ b/settings.json.template
@@ -967,13 +967,13 @@
   "sso": {
     "issuer": "${SSO_ISSUER:http://localhost:9001}",
       /*
-       * Optional signing keys for the embedded OIDC provider's cookies. When
-       * omitted, Etherpad derives a secret key from the persisted session
-       * secret (SESSIONKEY.txt), which is stable across restarts and shared
-       * across horizontally-scaled pods. Set an explicit ordered array to
-       * rotate: the first key signs, the rest are still accepted for verify.
-       * "cookieKeys": ["${OIDC_COOKIE_KEY:}"],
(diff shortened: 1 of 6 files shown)


JohnMcLear:
Thanks @qodo-free-for-open-source-projects — good catch. Confirmed: with default `cookie.keyRotationInterval`/`sessionLifetime` set, `settings.sessionKey` stays null, so the fallback would hand the provider a per-process random key (breaking OIDC cookies on restart / across pods). Fixed in the latest commit: added an `oidcCookieSecrets` `SecretRotator` (same DB-backed mechanism as the Express session cookies) as the default key source, with an integration test proving the default config yields stable rotator secrets rather than a random key.

SamTV12345:
Looks like qodo has some issues with the code.

SamTV12345:
Pushed 5c83191 for the remaining Qodo finding (template comment misattached): `sso.cookieKeys` is now a real property in `settings.json.template` with the doc block above it, so the admin UI's comment extractor no longer attributes it to `sso.clients`. An unset `OIDC_COOKIE_KEY` yields `[""]`, which `resolveOidcCookieKeys()` filters, so the derived/rotated key path is unchanged.

One note for the record, not a blocker: `clientBasedCORS` goes from "reflect every origin" to "origins of the client's registered redirect URIs". That is stricter than before but still more permissive than oidc-provider's own default, so it looks right — deployments whose SPA calls `/oidc/token` from an origin that is not a registered redirect URI will need to register it.

PR #8071, opened 2026-07-25
security(export): strip remote images on the soffice path to match the native path

JohnMcLear:
Defense-in-depth / consistency fix — **not a core vulnerability on its own**. Core Etherpad never emits `` tags in export HTML; they only appear when a plugin/hook injects them, so sanitising such content is primarily the injecting plugin's responsibility.

However, the native in-process export path (#7538) **already** calls `stripRemoteImages()` defensively, while the LibreOffice (`soffice`) path wrote the export HTML to the temp file verbatim. `soffice` is the only export path that performs outbound fetches for remote `` URLs during conversion, so a plugin-injected remote image there becomes a blind SSRF sink. This change makes the `soffice` path consistent with the native path core already ships.

## Change
One line: apply the existing `stripRemoteImages(html)` before writing the soffice temp file.

## Test
New `sofficeExportRemoteImage.ts` regression test injects a remote image via `exportHTMLAdditionalContent`, intercepts the temp file via the `exportConvert` hook, and asserts no remote image URL survives. Existing 54 export tests still pass; `tsc --noEmit` clean.

Remote-image behaviour reported privately by **meifukun** (https://github.com/meifukun). No security advisory is filed for this — it is hardening, not a core vuln.

### DIFF
--- a/src/node/handler/ExportHandler.ts
+++ b/src/node/handler/ExportHandler.ts
@@ -163,7 +163,13 @@ exports.doExport = async (req: any, res: any, padId: string, readOnlyId: string,
     // for the temp path token (see matching note in ImportHandler.ts).
     const randNum = crypto.randomBytes(16).toString('hex');
     const srcFile = `${tempDirectory}/etherpad_export_${randNum}.html`;
-    await fsp_writeFile(srcFile, html);
+    // Strip remote <img> tags before handing the document to LibreOffice.
+    // soffice fetches remote image URLs during conversion, so any plugin/hook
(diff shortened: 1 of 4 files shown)


JohnMcLear:
Thanks @qodo-free-for-open-source-projects — valid. Running `stripRemoteImages()` on the full document dropped `<!doctype html>` and comments. Fixed in the latest commit by adding `onprocessinginstruction` (doctype) and `oncomment` handlers so the serializer round-trips document directives while still stripping remote images; added regression tests.

PR #8072, opened 2026-07-25
security(proxy-path): set Vary on public routes that echo x-proxy-path (cache poisoning)

JohnMcLear:
Follow-up hardening to **GHSA-fjgc-3mj7-8rg8** (`x-proxy-path` admin XSS/open-redirect, fixed in #7784). Reported privately by **meifukun** (https://github.com/meifukun).

That advisory added `Vary: x-proxy-path` to the **admin** routes but left the **public** routes uncovered. The home page, pad page and timeslider (plus the legacy timeslider redirect) echo the sanitised `x-proxy-path` prefix into rendered URLs, social-preview metadata, manifest links and the redirect target — but advertised no `Vary`. A shared cache/CDN keyed on URL alone could store a prefix injected by one client and serve it to others.

## Scope
This is **cache-correctness hardening, not XSS/open-redirect** — the value is still passed through `sanitizeProxyPath` (quotes / angle-brackets / protocol-relative `//` / `..` already blocked by #7784). Deliberately does **not** change `x-proxy-path`'s trust model: it stays honored under `trustProxy: false` because subpath / Home-Assistant-ingress deployments rely on that default. The fix is purely additive (`Vary`).

## Change
Adds a `varyOnProxyPath(res)` helper applied at every `sanitizeProxyPath()` call site in `specialpages.ts` (both dev-watched and prod route blocks).

## Test
New `proxyPathVary.ts` asserts `Vary: x-proxy-path` on `/`, `/p/:pad`, the timeslider embed page, and the legacy redirect. Existing `proxyPathRedirect.ts` still passes; `tsc --noEmit` clean.

### DIFF
--- a/src/node/hooks/express/specialpages.ts
+++ b/src/node/hooks/express/specialpages.ts
@@ -30,8 +30,15 @@ import {sanitizeProxyPath} from '../../utils/sanitizeProxyPath';
 // headers in Vary so a shared cache/CDN in front of Etherpad keys on them and
 // can't serve a proxy-path injected by one client to another (cache poisoning).
 // Mirrors the admin-route fix in admin.ts (GHSA-fjgc-3mj7-8rg8).
-const PROXY_PATH_VARY_HEADERS = ['x-proxy-path', 'x-forwarded-prefix', 'x-ingress-path'];
-const varyOnProxyPath = (res: any) => res.vary(PROXY_PATH_VARY_HEADERS);
+//
+// Only vary on the headers sanitizeProxyPath() actually consults for the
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Thanks @qodo-free-for-open-source-projects — agreed. `sanitizeProxyPath()` ignores `x-forwarded-prefix`/`x-ingress-path` unless `trustProxy` is on, so varying on them otherwise just fragments caches. Fixed in the latest commit: always `Vary: x-proxy-path`, add the two standard headers only when `trustProxy` is enabled; test covers both states.

PR #8073, opened 2026-07-25
security(padid): reject ueberdb delimiter ':' in pad ids (copyPad/movePad injection)

JohnMcLear:
Fixes **GHSA-wg58-mhwv-35pq** (triage). Verified still present on `develop`.

`copyPad` / `movePad` / `copyPadWithoutHistory` take their destination via the `destinationID` API field, which — unlike `padID`/`padName` — is never run through `sanitizePadId`. Because `isValidPadId` only forbade `$`, a destinationID like `victim:revs:0` survived into the engine. The embedded `:` (the ueberdb key-namespace delimiter, `pad:<id>:revs:<n>`) let it address **another pad's internal records**, which:

1. **bypassed the `force=false` overwrite guard** — `doesPadExist` checks the top-level `atext`, but a `:revs:` record stores atext under `meta`, so it read as "doesn't exist"; and
2. **clobbered the victim pad's revision history** (persistent, silent — the head text survives).

Requires the global API key / OAuth-admin (PR:H), so low severity, but it's a clean cross-pad corruption primitive.

## Fix
- `isValidPadId` now rejects `:` (never legal in a pad id — it's the DB key delimiter). Name portion excludes both `$` and `:`.
- `Pad.copy()` and `Pad.copyPadWithoutHistory()` validate `destinationID` via `isValidPadId` **before any db write**. `movePad` routes through `copy()`, so the check runs before the source is removed (a bad destination can't delete the source either).

## Tests
- `isValidPadId` unit cases for `:` rejection (incl. group-pad form).
- Integration test: `copyPad`/`copyPadWithoutHistory` reject a `:`-bearing destinationID with `force=false` and leave the victim's rev-0 changeset untouched; a normal copy still works.
- Existing 54 API pad tests still pass; `tsc --noEmit` clean.

### DIFF
--- a/src/node/db/Pad.ts
+++ b/src/node/db/Pad.ts
@@ -632,6 +632,15 @@ class Pad {
   }
 
   async copy(destinationID: string, force: boolean) {
+    // Reject a destinationID that isn't a valid pad id BEFORE any db write. The
+    // copy path writes `pad:${destinationID}...` records directly (bypassing
+    // getPad), so a destinationID carrying the ueberdb delimiter `:` would
+    // otherwise clobber another pad's internal sub-records and slip past the
+    // force=false existence guard. (GHSA-wg58-mhwv-35pq.)
+    if (!padManager.isValidPadId(destinationID)) {
+      throw new CustomError('destinationID is not a valid padId', 'apierror');
+    }
+
     // Kick everyone from this pad.
     // This was commented due to https://github.com/ether/etherpad-lite/issues/3183.
     // Do we really need to kick everyone out?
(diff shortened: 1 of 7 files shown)


JohnMcLear:
Thanks @qodo-free-for-open-source-projects — correct, rejecting `:` in `isValidPadId` broke the legacy `/p/foo:bar` → `/p/foo_bar` sanitize-redirect (it 404'd instead). Fixed in the latest commit: `padurlsanitize` now sanitizes **before** validating, so `:` URLs redirect to the `_` form again while `:` stays out of stored pad ids. Added a regression test.

SamTV12345:
Pushed a follow-up (bc3687e) before merging.

Rejecting `:` in `isValidPadId` also gates `padManager.getPad()`, so pads whose id contains a `:` — legal until this advisory, which is exactly why `padIdTransforms` maps `:` at all — became **unopenable**, not just uncreatable: `sanitizePadId` returns such an id unchanged once the pad exists (`doesPadExist` short-circuits before the transform), and `getPad` then threw. Same for the pad-URL param after the sanitize-before-validate reordering.

So both gates now refuse an invalid id only when no pad with that exact id exists. The injection primitive stays closed: `doesPadExist()` requires a top-level `atext`, which the `pad:<id>:revs:<n>` / `:chat:<n>` sub-records an injected `destinationID` addresses do not have. Copy destinations keep the strict `isValidPadId` check, so no new `:` id can be created anywhere. New spec `padIdColonLegacyPad.ts` covers all three cases.

PR #8074, opened 2026-07-25
security(auth): regenerate session id on authentication (session fixation)

JohnMcLear:
Fixes **GHSA-73h9-c5xp-gfg4** (triage, **high** — runtime-proven admin takeover). Verified still present on `develop`: `webaccess.ts` sets `req.session.user` with no `session.regenerate()` (0 calls tree-wide).

## The bug (CWE-384)
Etherpad never rotates the express-session id when an anonymous session becomes authenticated. Any auth scheme that establishes a **pre-authentication session** — every OIDC/OAuth RP, including the official `ep_openid_connect` plugin, which persists OAuth state (`callbackChecks`) before redirecting to the IdP — is exposed: an attacker who plants or captures the pre-auth cookie owns the victim's authenticated (possibly **admin**) session after they log in.

## Fix
`webaccess` calls `req.session.regenerate()` at the authentication boundary — after the `authenticate` hook (or the HTTP-Basic path) establishes `req.session.user` — carrying the session data onto a freshly-minted id. It:
- fires **only on a fresh login** — already-authenticated requests short-circuit in Step 2, so reaching the authenticate step with no prior user means a genuine privilege upgrade;
- **no-ops** when the session store doesn't expose `regenerate()`;
- lives in **core `webaccess`**, so it protects *every* SSO/auth plugin (the root fix the advisory recommends over patching each plugin).

## Tests
- New `sessionFixation.ts`: a server-issued pre-auth session that then authenticates gets a **different** `express_sid`, and the rotated session stays authenticated (user preserved).
- Full `webaccess.ts` suite (53 tests) still passes; `tsc --noEmit` clean.

### DIFF
--- a/src/node/hooks/express/webaccess.ts
+++ b/src/node/hooks/express/webaccess.ts
@@ -174,11 +174,11 @@ const checkAccess = async (req:any, res:any, next: Function) => {
 
   if (settings.users == null) settings.users = {};
   const ctx:WebAccessTypes = {req, res, users: settings.users, next};
-  // Whether this request already carried an authenticated session BEFORE the
-  // authenticate step runs. Already-authenticated requests are short-circuited
-  // in Step 2, so reaching here with no user means a *fresh* login is about to
-  // happen — the point at which the session id must be rotated (see below).
-  const wasAlreadyAuthenticated = req.session != null && req.session.user != null;
+  // Identity carried by the session BEFORE the authenticate step runs. Used
+  // below to decide whether authentication changed the principal (anonymous ->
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Thanks @qodo-free-for-open-source-projects — both valid. (1) The first cut only rotated on anon→auth; it now rotates whenever authentication changes the principal (anonymous→user, or username / `is_admin` change), so a non-admin→admin re-auth reaching Step 3 also rotates — with a deterministic test. A no-op re-auth of the same principal is left alone (no churn). (2) The test now derives the cookie name from `settings.cookie.prefix` instead of hardcoding `express_sid`.

PR #8075, opened 2026-07-25
security(collab): apply queued USER_CHANGES to the enqueue-time pad (cross-pad write TOCTOU)

JohnMcLear:
Fixes **GHSA-6mcx-x5h6-rpw2** (triage, medium — runtime-proven). Verified still present on `develop`: `handleUserChanges` re-reads `sessioninfos[socket.id].padId` at apply time (the write target), while the read-only gate and the `padChannels` key are taken at enqueue time.

## The bug (CWE-362 → CWE-862)
Per-socket messages are processed **concurrently** and a thrown handler does **not** disconnect the socket. So a same-socket `CLIENT_READY` can swap the session's `padId` **between** a `USER_CHANGES` being authorized+enqueued and it being applied:

1. `CLIENT_READY(A)` — a writable pad → `padId=A`.
2. `USER_CHANGES` → passes the read-only gate on A, enqueues on channel A.
3. `CLIENT_READY(B)` — a read-only / unauthorized pad → swaps `sessioninfos[socket.id].padId=B` (and a *denied* CLIENT_READY still leaves it mutated, without disconnecting).
4. The queued `handleUserChanges` re-reads `padId=B` → **writes to B**.

A read-only share holder can thus overwrite the pad; the read-only enforcement and per-pad authorization are bypassed.

## Fix
The `padChannels` channel key already **is** the enqueue-time `padId`. Thread it into `handleUserChanges` as `authorizedPadId` and use it for `getPad(...)` instead of re-reading the mutable session. The authorization gate and the write now refer to the same pad, closing the window. Non-racing flows are unchanged (session padId == enqueue-time padId).

## Test
`crossPadWriteToctou.ts` invokes `handleUserChanges` with the session padId **already swapped** to a victim pad and asserts the change lands on the authorized (argument) pad and never the victim. Confirmed it **fails when the vulnerable read is reintroduced** (the victim gets modified). Full `messages.ts` socket suite (13) still passes; `tsc --noEmit` clean.

### DIFF
--- a/src/node/handler/PadMessageHandler.ts
+++ b/src/node/handler/PadMessageHandler.ts
@@ -512,6 +512,17 @@ exports.handleMessage = async (socket:any, message: ClientVarMessage) => {
     throw new Error(`pre-CLIENT_READY message from IP ${ip}: ${msg}`);
   }
 
+  // Pin the pad this message is authorized against — together with `auth` and
+  // the read-only flag — BEFORE any of the awaits below (checkAccess and the
+  // handleMessageSecurity/handleMessage hooks). A concurrent same-socket
+  // CLIENT_READY can mutate sessioninfos[socket.id] (padId/readonly/auth) IN
+  // PLACE during those awaits (the object identity check below does not catch an
+  // in-place mutation). Using these pinned values for the read-only gate, the
+  // queue key and the write keeps them all referring to the SAME pad, closing
+  // the cross-pad write TOCTOU (GHSA-6mcx-x5h6-rpw2).
+  const messagePadId = thisSession.padId;
+  const messageReadonly = thisSession.readonly;
+
   const {session: {user} = {}} = socket.client.request as SocketClientRequest;
   const {accessStatus, authorID} =
       await securityManager.checkAccess(auth.padID, auth.sessionID, auth.token, user);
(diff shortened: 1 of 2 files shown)


JohnMcLear:
Good catch @qodo-free-for-open-source-projects — you're right, the first cut wasn't sufficient. The channel key was still read from `thisSession.padId` at enqueue, i.e. *after* the awaits in `handleMessage()` (checkAccess + the security/message hooks), so a concurrent same-socket CLIENT_READY could still swap it in place between the read-only gate and the enqueue.

Fixed properly in the latest commit: the pad id + read-only flag are now pinned into a consistent snapshot (`messagePadId`/`messageReadonly`) alongside `auth`, **before** those awaits, and that snapshot drives the read-only gate, the queue key and the write. The regression test now drives a real USER_CHANGES over a socket and swaps the session padId mid-message via a `handleMessageSecurity` hook (the actual concurrent-CLIENT_READY race) — verified it fails when the pinned key is reverted. That also let me drop the `handleUserChanges` export and the direct-call test (your maintainability / pendingEdits / cleanup findings).

SamTV12345 (approved):

SamTV12345:
Merged. One residual worth a follow-up (not a security issue, and not worth blocking this on): `handleUserChanges` still reads `thisSession.author`, `thisSession.rev` and `thisSession.time` at apply time, so in the racing case the write correctly lands on `authorizedPadId` while the rev bookkeeping (`assert.equal(thisSession.rev, r)`, `thisSession.time`) belongs to the swapped-in pad. Worst case is a `badChangeset` disconnect of the attacker's own socket, so the authorization property this PR fixes holds — but pinning the author/rev alongside `messagePadId`, or dropping the queued message when `thisSession.padId !== authorizedPadId`, would close the class rather than the instance.

Issue #7996, opened 2026-06-22
Disable pad deletion token UI WITHOUT allowPadDeletionByAllUsers


n-bux:
When I asked for Disabling the pad deletion token UI when allowPadDeletionByAllUsers is true in https://github.com/ether/etherpad/issues/7926, I didn't realize that  `allowPadDeletionByAllUsers` really does what it says..: Pads can be deleted by **any** user. 

So now (in v3.3.2) I have the choice between showing the pad  deletion token **OR** allowing all users to delete a pad, even if they are not the creator - **which is a problem**

What I actually want is:
- a) **no** "pad deletion token" UI, or at least a way to disable the default display of the token (because it confuses some user groups). 
  - a1)  for those user groups, in my point of view it would actually be fine to have the token functionality active, if the token-popup would be supressed, but the token accessible in the settings. in this way, users who understand the option can find and use it, but others wouldn't have to bother

AND at the same time

-  b) **disallow** Pads to be deleted by others than the creator (to prevent accidental data loss by clicking "delete" on the wrong pad, or prevent intentional deletion like students trolling their teachers..). I am fine with the way it worked until some updates ago: only in the creating browser session, the pad could be deleted
  - b1) additionally, it would be great to 
    - hide the "delete" button if the session is not the creators session
    - have an instance settings.json option to disable the "delete pad"  option completely

If you don't like a1) from UX/other use cases' point of view, we would need a settings.json option to disable the "delete pad with token" functionality WITHOUT allowPadDeletionByAllUsers

b1) is  a nice-to-have, but not a solution to my problem

amanumrao28-png:
+1 on this. I ran into the same wall: allowPadDeletionByAllUsers: true was the only way to suppress the token UI, but that setting does exactly what it says — it also opens deletion to any user, not just the creator. That's a dealbreaker for us (e.g. students being able to delete a teacher's pad, or someone fat-fingering delete on a shared pad).

What I actually need is these two things independently configurable:

Enforce creator-only deletion (allowPadDeletionByAllUsers: false, the default — this part already works)
Suppress the recovery-token modal for the creator's own session, without touching #1

I looked into the 3.3.2 changes (#7926/#7930) — the token is already suppressed automatically, but only when the creator has a "durable identity" (authenticated session + a getAuthorId hook mapping to a stable authorID). For anonymous/cookie-only setups, the token modal still shows for the creator, and there's no manual override.

[...]

PR #8089, opened 2026-08-04
fix: carry the pad deletion token across movePad (#7995)

### LINKED ISSUE #7995: moved pads don't keep, offer a new deletion token
Noticed when moving a pad, I suppose this is not expected behavior.

Also, personally I'd be happy to be able to turn off the deletion token popup :)

JohnMcLear:
Closes #7995 (the "moved pads don't keep, offer a new deletion token" half — the "let me turn the popup off" half is #7996).

## Repro (from @dcht00)

1. create pad `abc` → deletion-token modal, save the token
2. `movePad` `abc` → `def`
3. open `def` → a **second** deletion-token modal, and the token saved in step 1 no longer deletes anything

## Root cause

`API.movePad` is `Pad.copy()` + `Pad.remove()`:

```js
const pad = await getPadSafe(sourceID, true);
await pad.copy(destinationID, force);
await pad.remove();
```

`Pad.copy()` replicates only the `pad:<id>`, `pad:<id>:revs:N` and `pad:<id>:chat:N` records — it never touches `pad:<id>:deletionToken` (`src/node/db/Pad.ts`). `Pad.remove()` then calls `padDeletionManager.removeDeletionToken(sourceID)`. Net effect: the token record is destroyed and never recreated at the new id.

Two visible consequences:

* the plaintext token the creator was told to save validates against nothing — `deletePad(def, <token>)` returns `invalid deletionToken`;
* the copy preserves the revision-0 author, so on the renamed pad the creator is still `isCreator`, `createDeletionTokenIfAbsent()` finds an empty slot, and `PadMessageHandler` ships a fresh `padDeletionToken` → the modal fires a second time.

## Fix

New `PadDeletionManager.transferDeletionToken(src, dst)`, called from `movePad` between `copy()` and `remove()` (it must run before `remove()` drops the source record). A move is a rename, so the token moves with the pad.

* `force`-overwriting an existing destination discards that pad's own token — `copy()` already removed the overwritten pad and its content, so its token would be a key to nothing.
* **`copyPad` is deliberately left alone.** A copy is a separate pad; sharing one hash between two pads would mean a token saved for one deletes the other. The copy gets its own token on the creator's first visit, as today.

## Tests

`src/tests/backend/specs/api/movePadDeletionToken.ts` — 4 cases:

| test | pre-fix |
|---|---|
| moved pad accepts the original token | ❌ `invalid deletionToken` |
| moved pad does not offer the creator a second token | ❌ mints a new one |
| `movePad --force` replaces the destination's own token | ❌ |
| `copyPad` does **not** share the source token | ✅ (guards the fix's blast radius) |

Full backend suite: `1617 passing, 0 failing`. `pnpm run ts-check` clean.

Docs: `movePad` section of `doc/api/http_api.md` + CHANGELOG entry.

Reported by @dcht00.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 3.3.4
+
+### Notable fixes
+
+- **API — `movePad` now carries the pad's deletion token to the new id (#7995).** `movePad` is implemented as `copy()` + `remove()`, but `Pad.copy()` only copies the `pad:<id>`, `:revs:N` and `:chat:N` records — never `pad:<id>:deletionToken` — and `remove()` then deleted the source pad's token. The renamed pad therefore had no token at all: the token the creator had been told to save no longer deleted anything, and because the copy keeps the same revision-0 author, their next visit tripped `createDeletionTokenIfAbsent()` and popped a second "save your pad deletion token" modal. The token record is now handed over to the destination as part of the move, so the saved token keeps working and the modal does not reappear. `force`-overwriting an existing destination discards that pad's own token along with its content. `copyPad` is deliberately unchanged — two pads sharing one secret would let a token saved for one delete the other.
+
 # 3.3.3
 
 3.3.3 is a security release. It closes a **critical unauthenticated arbitrary-file-read** in the `/static/*` handler (GHSA-mc8w-wjhw-45x5) and bundles the fixes for a batch of privately reported issues that had already landed on `develop`: an OpenID Connect provider hardcoded cookie key and permissive CORS reflection (GHSA-pp5v-mvwg-76mp), session-fixation on authentication (GHSA-73h9-c5xp-gfg4), a same-socket cross-pad write TOCTOU (GHSA-6mcx-x5h6-rpw2), and a pad-id delimiter injection in `copyPad`/`movePad` (GHSA-wg58-mhwv-35pq). Alongside the security work it migrates the server build to TypeScript 7 (`tsgo`), fixes PageDown/PageUp navigation across consecutive long wrapped lines, and makes the docker `plugin_packages` volume mountpoint writable.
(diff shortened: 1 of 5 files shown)


JohnMcLear:
Actioned the review feedback in ea49ba8 — both bots landed on the same two real issues.

**1. Source token removed early (qodo ×2) — valid, fixed.** `transferDeletionToken()` no longer removes the source record. `movePad` calls it before `pad.remove()`, which drops that record itself on success, so the eager removal only had an effect when `remove()` failed — and there it did the wrong thing, leaving a surviving source pad whose creator could no longer delete it by token.

**2. copy → transfer window (qodo #2) — valid, fixed.** `Pad.copy()` writes the destination records before the transfer runs, so a creator opening the new id in that window gets a freshly minted token from `createDeletionTokenIfAbsent()`. Overwriting it would hand them a token that silently stops working, which is worse than the bug being fixed. The transfer now refuses to clobber a destination token that already exists — in that (rare) race the creator simply keeps the token they were just shown. The transfer also runs through the same per-pad queue as `createDeletionTokenIfAbsent()` (extracted as `withPadTokenLock`) so a concurrent mint cannot interleave with its read-then-write.

Not taken: a lock across *both* pad ids. Only the destination's slot is read-then-written; the source is about to be deleted, and a two-id lock needs deterministic ordering for no benefit here.

**3. Copilot — reword the `stored == null` comment.** That branch is gone. It cleared the *destination's* token when the source had none, justified by the force-overwrite case — but `Pad.copy()` already removes the overwritten pad and its token via `Pad.remove()`, so the branch was redundant on the path it was written for and could only clobber a live token elsewhere. The `movePad --force` test still passes and now documents that `copy()` does that cleanup.

Added a regression test for the no-clobber guard (destination token issued mid-move survives; source token survives the transfer). Backend suite: 1618 passing, 0 failing. `ts-check` clean.

PR #8112, opened 2026-08-07
Fix home and post-delete redirects behind a proxy URL prefix

### LINKED ISSUE #8111: Home button (and after delete pad redirect) redirect to domain root even when etherpad is hosted behind a reverse proxy with a URL prefix
**Describe the bug**
When Etherpad is deployed behind a reverse proxy under a URL prefix, the pad toolbar home button ("Back to home") redirects to the domain root instead of the Etherpad home page under that prefix.

Example:
- Etherpad home is served at `/etherpad/`
- A pad is opened at `/etherpad/p/testpad`
- Clicking the home button redirects to `/`
- Expected destination is `/etherpad/`

This appears to be a client-side navigation bug in the home button implementation, not a reverse-proxy header configuration issue.

**To Reproduce**
1. Deploy Etherpad behind a reverse proxy under a URL prefix such as `/etherpad`
2. Configure the proxy so Etherpad otherwise works correctly under that prefix (X-Proxy-Path: /etherpad)
3. Open a pad, for example `/etherpad/p/testpad`
4. Click the toolbar home button ("Back to home")
5. Observe that the browser is redirected to `/` instead of `/etherpad/`

**Expected behavior**
The home button should redirect to the Etherpad home page under the same public prefix, for example `/etherpad/`, not to the domain root.

**Screenshots**

 - Etherpad version: 3.3.3
 - OS: docker
 - Node.js version (`node --version`): version in https://hub.docker.com/layers/etherpad/etherpad/3.3.3/images/sha256-776e8d66b967c756771f681ae6f887f3ce3e831269728cddb0ab2537e23d1fa8
 - Are you using any abstraction IE docker? [e.g. Docker / reverse proxy / ingress]
 - docker with custom reverse proxy (I can set up with nginw with needed, but the "bug" is in frontend)

 - OS: linux, MacOS, windows 
 - Browser [e.g. Firefox, Chrome]; Firefox, safari and edge (respectively)
 - Version [e.g. 128]

**Smartphone (please complete the following information):**
not tested on smartphone

** suspected bug 
Client-side is not prefix-aware.

The current implementation in `src/static/js/pad_editbar.ts` is:
https://github.com/ether/etherpad/blob/develop/src/static/js/pad_editbar.ts#L485C1-L486C7
```ts
this.registerCommand('home', ()=> {
  window.location.href = new URL('../..', window.location.href).href
})

On my instance: 
in the js console:
window.location.href is set to "https://<mydomain>/etherpad/p/testpad" 
so 
new URL('../..', window.location.href).href  is "https://<mydomain>/"

AkprasadoP:
## What

Preserve the public URL prefix when navigating home from a pad and after a pad is deleted.

## Why

When Etherpad is served behind a reverse proxy under a prefix such as `/etherpad`, the Home command resolved `../..` from `/etherpad/p/<padId>` and redirected to the domain root. Post-delete redirects separately hardcoded `/`, causing the same problem.

## Changes

- Resolve the home URL relative to the current pad URL.
- Use the same resolver for Home and all post-delete redirects.
- Add regression coverage for root, single-prefix, and nested-prefix deployments.

Fixes #8111

## Tests

- `vitest run tests/backend-new/specs/getHomeUrl.test.ts`
- Playwright `editbar.spec.ts`

Note: ESLint could not be run locally because of a pre-existing ESLint/package compatibility mismatch.

### DIFF
--- a/src/static/js/getHomeUrl.ts
+++ b/src/static/js/getHomeUrl.ts
@@ -1,0 +1,12 @@
+'use strict';
+
+/**
+ * Resolve the Etherpad home URL from a pad URL.
+ *
+ * Pad pages live at `{prefix}/p/{padId}`. One `..` segment removes the pad id
+ * and lands on `{prefix}/`, which is correct both for root deployments
+ * (`/p/testpad` -> `/`) and reverse-proxy prefixes (`/etherpad/p/testpad`
+ * -> `/etherpad/`). See issue #8111.
+ */
+export const getHomeUrl = (fromHref: string): string =>
+  new URL('..', fromHref).href;
--- a/src/static/js/pad_editbar.ts
+++ b/src/static/js/pad_editbar.ts
@@ -24,6 +24,7 @@
  */
 
 const hooks = require('./pluginfw/hooks');
+import {getHomeUrl} from './getHomeUrl';
 import padutils from "./pad_utils";
 const padeditor = require('./pad_editor').padeditor;
 const padsavedrevs = require('./pad_savedrevs');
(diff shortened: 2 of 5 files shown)


SamTV12345:
Thanks for the pull request. Awesome work 😁

PR #8142, opened 2026-08-15
fix(deps): roll typescript back to ^6.0.3 — 7.x breaks the admin build

JohnMcLear:
`develop` has been red since #8132 (2026-08-12) bumped `typescript` from `^6.0.3` to `^7.0.2` across all four workspaces. **Every PR currently fails**, at the "Build admin ui" step:

```
node_modules/.pnpm/openapi-typescript@7.13.0_typescript@7.0.2/node_modules/openapi-typescript/dist/lib/ts.mjs:11
const BOOLEAN = ts.factory.createKeywordTypeNode(ts.SyntaxKind.BooleanKeyword);
                           ^
TypeError: Cannot read properties of undefined (reading 'createKeywordTypeNode')
```

## Why

`openapi-typescript` reaches into the TypeScript compiler API to build its AST. TypeScript 7 is the native port and doesn't expose `factory` the way it expects.

There's no version to upgrade to: `openapi-typescript@7.13.0` is the latest published release and declares `peerDependencies: { typescript: '^5.x' }`. It happened to work with 6.x; 7 is where it gives out. We have to wait for upstream to support the new compiler.

## Change

Roll `typescript` back to `^6.0.3` in `admin/`, `src/`, `ui/` and `bin/`. All four rather than just `admin/`, to avoid carrying two TypeScript majors in one pnpm workspace.

## Verified locally

- `admin`: `gen:api` succeeds, full `pnpm build` succeeds
- `src/`, `ui/`, `bin/`: `tsc --noEmit` clean
- backend suite: 1621 passing, 0 failing

## Follow-up

Dependabot will offer typescript 7 again. Worth either ignoring the `typescript` major until `openapi-typescript` supports it, or making the admin build a required check so a bump that breaks it can't land.

### DIFF
--- a/admin/package.json
+++ b/admin/package.json
@@ -45,7 +45,7 @@
     "react-router-dom": "^7.18.2",
     "socket.io-client": "^4.8.3",
     "tsx": "^4.23.12",
-    "typescript": "^7.0.2",
+    "typescript": "^6.0.3",
     "vite": "^8.2.1",
     "vite-plugin-babel": "^1.7.3",
     "zustand": "^5.0.15"
--- a/bin/package.json
+++ b/bin/package.json
@@ -16,7 +16,7 @@
   "devDependencies": {
     "@types/node": "^26.2.0",
     "@types/semver": "^7.8.0",
-    "typescript": "^7.0.2"
+    "typescript": "^6.0.3"
   },
   "scripts": {
     "makeDocs": "node --import tsx make_docs.ts",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -85,10 +85,10 @@ importers:
         version: 19.2.4(@types/react@19.2.18)
       '@typescript-eslint/eslint-plugin':
         specifier: ^8.67.0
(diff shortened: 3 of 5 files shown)


JohnMcLear:
Superseded by #8143, which keeps us on TypeScript 7 instead of retreating from it.

Digging further: TypeScript 7 is the native port, and its main export is only `./lib/version.cjs` — the compiler API `openapi-typescript` reaches for genuinely isn't there. But that's a problem with one build-time tool, not with the codebase, which typechecks clean under `tsc` 7 in all four workspaces.

#8143 isolates `openapi-typescript` in a private build-time workspace package pinned to TypeScript 6, so every workspace keeps `typescript: ^7.0.2`. Closing this one in favour of that.

(For the record, the tidier root-level fixes don't work here: `openapi-typescript` takes `typescript` as a *peer*, so pnpm resolves it from the depending package. Both `overrides: openapi-typescript>typescript` and `packageExtensions` were tried and pnpm still linked the `_typescript@7.0.2` variant.)

PR #8218, opened 2026-09-14
fix(updater): spawn pnpm via cross-spawn so updates work on Windows

kaanisthatyou:
On Windows, the updater checks for pnpm and runs `pnpm install` / `pnpm run build:ui` with `child_process.spawn('pnpm', ...)`. `pnpm` is a `.cmd` shim on Windows, which `child_process.spawn` cannot start without a shell (it fails with `ENOENT`). As a result, on a git install created with the PowerShell installer from the README:

- `pnpmOnPath()` always resolves `false`, so preflight fails with `pnpm-not-found` even though pnpm is installed;
- the install, build and rollback steps (`runStep` in `UpdateExecutor` / `RollbackHandler`) would fail the same way.

This uses `cross-spawn` (already a dependency, used by `utils/run_cmd.ts`) for the pnpm check and as the `spawnFn` passed to the executor and rollback handler. `cross-spawn` runs `.cmd` shims correctly on Windows and falls through to `child_process.spawn` on other platforms, so behavior on Linux and macOS is unchanged. I preferred it over `shell: true`, which on Node 24 prints a `DEP0190` warning when combined with an args array.

### Testing

On Windows 11 / Node 24.14.0:
- The exact `pnpmOnPath` spawn call returns `false` (`ENOENT`) with `child_process.spawn` and `true` with `cross-spawn`.
- `npx vitest run tests/backend-new/specs/updater` (in `src`): 18 files, 236 tests passing.
- `npx tsc --noEmit`: no errors.

I did not run a full update end to end on Windows. Running `npx eslint` in `src` (ESLint 10.10.0) fails for me with "couldn't find an eslint.config.* file" (the repo only has `src/.eslintrc.cjs`), so I could not lint the change locally.

https://claude.ai/code/session_016Wuh9RTwTUHB7wxpNxrSiw


SamTV12345:
Thanks for the addition :)

